PR #49, opened 2016-11-28
Reduce kernel size by 2MB

zaolin:
Hey,

I have reworked the linux kernel config via make tinyconfig and updated the sources.


osresearch:
I'm not sure that will boot on real hardware -- `CONFIG_BINFMT_ELF` is not set, so will any executables work?

If we can find 2MB in the kernel to strip out, that would be great.  Using uclibc would also help, although that would require a full toolchain and building everything (which we'll want to do anyway for reproducibility).

zaolin:
I would recommend to use alpine linux with musl libc to build all sources. It's really small and build with hardened flags. Yeah need to fix some things let me check that.

osresearch:
The kernel `.config` needs analysis for areas where we might have extraneous features enabled, but the `tinyconfig` throws out everything that makes it work.

Issue #37, opened 2016-10-04
Coreboot config should support multiple targets


osresearch:
The makefile target for coreboot config only supports the x230; it should support building the qemu target or the (soon to be create) chell chromebook target without having to copy files around.

zaolin:
There is the abuild tool under util/abuild. It can build mutilple devices directly also we are planing to bring up a service for building coreboot images (GPG signed/encrypted) -> https://github.com/zaolin/coreboot-builder

osresearch:
Supported as of https://github.com/osresearch/heads/commit/4fbd6ca58bff576c5ba149de472a18615e82a665

Run `make TARGET=qemu` or `make TARGET=x230`  to generate the different ROM images.  Chell support still needs to be merged (minimal changes, but needs to be back tested)

PR #50, opened 2016-11-28
WIP: Add initial support for plymouth splash screen

zaolin:
Signed-off-by: Philipp Deppenwiese <zaolin@das-labor.org>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -143,10 +143,12 @@ endef
 
 $(foreach _, $(call bins,kexec), $(eval $(call initrd_bin_add,$_)))
 $(foreach _, $(call bins,tpmtotp), $(eval $(call initrd_bin_add,$_)))
+$(foreach _, $(call bins,plymouth), $(eval $(call initrd_bin_add,$_)))
 
 $(foreach _, $(call libs,tpmtotp), $(eval $(call initrd_lib_add,$_)))
 $(foreach _, $(call libs,mbedtls), $(eval $(call initrd_lib_add,$_)))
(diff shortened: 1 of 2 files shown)


zaolin:
Thanks good point. Can we remove libgcrypt or is it needed somehow. Normally it's optional for cryptsetup. It would safe 1.2MB

osresearch:
I've reworked the makefile to make it easier to build the qemu ROM image (issue #37) and fixed issue #53 that prevented qemu from starting.  This makes it easier to debug, since qemu doesn't have the 4MB size issue of the real x230 ROM.  With plymouth included it prints an error when I try to run it:

```
plymouth: ply-event-loop.c:497: ply_event_loop_new: Assertion `loop->epoll_fd >= 0` failed.
```

Haven't had time to debug it further yet.

zaolin:
Should be fixed by adding epoll to kernel config as syscall. I updated my #49 changeset to include some more syscalls..

osresearch:
How do I test plymouth and the theme?  qemu (and chell) support 12 MB or more of payload, so we can make it there without having to squeeze into the x230's 4MB ROM.

zaolin:
You can test the plymouth theme on your local machine or in qemu itself. Currently I am totally busy with orga stuff. Maybe I can fix the remaining issues this weekend and get it running for the 33c3

osresearch:
I'd like to revisit this now that we have a more streamlined boot process and the coreboot to Heads to dracut to Qubes handoffs are working smoothly.  Can you rebuild the plymouth start screen and see about adding the totp or other displays to it?

PR #58, opened 2016-11-30
WIP: Replace gnupg with signify

zaolin:
Signed-off-by: Philipp Deppenwiese <zaolin@das-labor.org>


osresearch:
gpgv is big, but it does have a fairly mature ecosystem around it.  One of the reasons for using GPG is that it integrates well with hardware tokens like Yubikeys, allowing the signing key to be separated from the systems. A potential threat is if the signing key lives on the machine then an attacker might be able to use code execution on the machine to be able to update and sign the kernel or root hashes.

zaolin:
Okay but if someone has access to my system everything is exposed. So I can understand that GPG is a good decision if it comes to security ( I am also a yubikey user) but in my opinion this isn't needed for boot chain security. Also the signify key is stored encrypted on the operating system.

zaolin:
The xen module is more a compability layer for qubes os. Maybe we can use signify for dm-verity stuff and offer an optional gpg support for qubes os or other modules

jonathanfemideer:
@zaolin:
> Okay but if someone has access to my system everything is exposed.

Yes, and that would be bad enough. But if that happens, do you also want the attacker to be able to update and ***sign*** the kernel or root hashes?!

By letting them sign those hashes, you eliminate your last line of defense against an attacker who modifies your kernel/etc: the chance to detect those changes when you next boot. You negate the primary purpose of Heads.

> Also the signify key is stored encrypted on the operating system.

Sure, but a patient attacker will just wait until the user decrypts the key to use it, and will then grab it from RAM. Offloading the key to a smart card should prevent this, but that is not possible with Signify at the moment, only with GnuPG.

I am sure that your pull request was well-intentioned, but I hope that @osresearch will reject it for the reasons above.

zaolin:
Yeah for me that's no real security if the system is compromised everything is exposed (the code quality of gnupg is worse than everything else i have seen..). I made this pull request because of the payload size for coreboot. So I don't mind to close it..

jonathanfemideer:
> Yeah for me that's no real security if the system is compromised everything is exposed

Think of it as having an [IDS](https://en.wikipedia.org/wiki/Intrusion_detection_system). No-one wants a breach, but if one happens then it is far better to have *some* chance of knowing that it has happened.

PR #59, opened 2016-11-30
Fix coreboot buildgcc detection

zaolin:


osresearch:
How does this improve the detection of coreboot gcc?  There is only a dependency target for i386-elf-gcc; we should have a separate one for arm if we need it (which chromebook EC does).

zaolin:
For example x86_64 is also a target :) So in my opinion we should allow every target.

osresearch:
The makefile seems to be detecting coreboot's gcc fine at this point.  Reopen this if it is still an issue.

PR #78, opened 2017-01-04
Merge musl-libc branch

osresearch:
Appears to work and build cleanly now.


patches/kexec-2.0.12.patch
```diff
@@ -0,0 +1,67 @@
...
+-		unsigned long long start, end;
++		uintptr_t start, end;
+ 		char *str;
+ 		int type;
+ 		int consumed;
+ 		int count;
+ 		if (memory_ranges >= MAX_MEMORY_RANGES)
+ 			break;
+-		count = sscanf(line, "%Lx-%Lx : %n",
++		count = sscanf(line, "%"SCNxPTR"-%"SCNxPTR" : %n",
```
rofl0r: nope, it's not a musl bug, it's a conformance bug in GLIBC and a bug for apps that use it.
according to POSIX (man 3p fscanf)
```
     L    Specifies  that  a  following  a,  A, e, E, f, F, g, or G conversion
          specifier applies to an argument with type pointer to long double.
```
the author wanted to actually use "%llx" but he did it the GLIBC-only-crash-everywhere-else way.
i've fixed this in various linux-specific opensource projects, the most prominent being [alsa-lib](http://git.alsa-project.org/?p=alsa-lib.git;a=commitdiff;h=1d3f7975f920f47e6a8a324f547da2180e64171a)

richfelker: Strictly speaking it's a conformance bug in applications; on the glibc side it's just doing something unhelpful in the case of UB and masking the bug by guessing what the programmer wanted and doing it rather than reporting the error. The fix looks correct though.

modules/musl
```diff
@@ -8,9 +8,17 @@ musl_hash := 97e447c7ee2a7f613186ec54a93054fe15469fe34d7d323080f7ef38f5ecb0fa
 
 musl_output := $(INSTALL)/bin/musl-gcc
 
+#
+# Note that for syslibdir to be /lib the install will likely fail.
+# this is unfortunate, but can be worked around by making sure that
+# a /lib/ld-musl-x86_64.so.1 symlink exists and points to libc.so
```
rofl0r: maybe you should opt for static linked binaries, musl is quite good at that (unlike GLIBC, where a static linked hello world is almost as big as musl's libc.so).

PR #81, opened 2017-01-16
modules/coreboot: remove 'time' for dash compat

iseeareddoor:
the 'time' builtin is a bashism which is not supported in Debian's standard sh ('dash'), which is used implicitly here when building on Debian systems.


iseeareddoor:
alternatively the bash dependency could be made explicit, or /usr/bin/time (https://tracker.debian.org/pkg/time) could be marked as explicit build-dependency for dash-as-sh systems like Debian.

I guess most people will just run the whole "make" invocation wrapped with whatever 'time' variant is appropriate if they are interested in this kind of information, or have their shell configured to automatically output "time"-like information for long running commands.

osresearch:
Thanks for the suggestion. Signed and merged into master: https://github.com/osresearch/heads/commit/85dcbf668732388911b9f13ede757d675191b57d

PR #112, opened 2017-02-26
Find Qubes root with /dev/mapper instead of label

jonathanfemideer:
Avoids having to assign labels during Qubes installation and avoids
having to deal with UUIDs.

See https://github.com/osresearch/heads/issues/110 for discussion.

### DIFF
--- a/initrd/bin/start-xen
+++ b/initrd/bin/start-xen
@@ -21,7 +21,7 @@ gpgv "${KERNEL}.asc" "${KERNEL}" || die "Kernel signature failed"
 
 kexec \
 	-l \
-	--module "${KERNEL} root=LABEL=root rhgb" \
+	--module "${KERNEL} root=/dev/mapper/qubes_dom0-root rhgb" \
 	--module "${INITRD}" \
 	--command-line "no-real-mode reboot=no console=vga dom0_mem=min:1024M dom0_mem=max:4096M" \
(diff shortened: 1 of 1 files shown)


osresearch:
I like the simplicity of this; much easier than finding the UUIDs and requiring the script to be updated.

Do you have a GPG key for signing commits?

osresearch:
Although one of my test machines doesn't have `/dev/mapper/qubes_dom0-root`. They all show up as `/dev/mapper/luks-XXXX`. I'll need to check on other ones.

mfc:
as [mentioned](https://github.com/osresearch/heads/issues/110#issuecomment-282922021) in the other thread and you've found I don't think this is a general solution / this doesn't work for encrypted hard drives.

jonathanfemideer:
> Although one of my test machines doesn't have /dev/mapper/qubes_dom0-root. They all show up as /dev/mapper/luks-XXXX. I'll need to check on other ones.

Odd. What is the output of `blkid` in your `dracut` shell?

mfc:
I take back what I said, it works for me now! thanks for helping me troubleshoot @jonathanfemideer.

osresearch:
I think is no longer necessary, with the updates in `boot.sh` that use LVM to find the various volume groups and LUKS devices.

PR #158, opened 2017-04-04
Automatically load network kernel modules and request DHCP

geedo0:
This loads the kernel modules for the Thinkpad NICs and then starts up udhcp inside init. I wasn't sure whether to use -b or -n with udhcp, but I did find that having neither hangs up the system if you don't have an ethernet cable plugged in.

### DIFF
--- a/initrd/init
+++ b/initrd/init
@@ -28,6 +28,11 @@ if ! unsealtotp.sh ; then
 fi
 echo
 
+# Load the network modules and get an IP address
+insmod /lib/modules/e1000.ko
+insmod /lib/modules/e1000e.ko
+udhcpc -n
+
 # Start an interactive shell
 export PATH=/sbin:/usr/sbin:/bin:/usr/bin
 exec /bin/ash


initrd/init
```diff
@@ -28,6 +28,11 @@ if ! unsealtotp.sh ; then
 fi
 echo
 
+# Load the network modules and get an IP address
```
osresearch: This should go into a separate MOC startup script that is specified in `CONFIG_BOOTSCRIPT`-- `initrd/init` is only for things that are common to all platforms.  See https://github.com/osresearch/heads/blob/master/config/x230-qubes.config#L21 and  `initrd/bin/qubes-init` for an example

osresearch:
This should go into a separate MOC startup script that is specified in `CONFIG_BOOTSCRIPT`-- `initrd/init` is only for things that are common to all platforms.  See https://github.com/osresearch/heads/blob/master/config/x230-qubes.config#L21 and  `initrd/bin/qubes-init` for an example

geedo0:
I noticed you pushed up some changes for us on the master branch that does this inside /etc/keylime-init. I'll pull down that branch and continue our work from there in a new PR.

Issue #160, opened 2017-04-05
USB mass storage should be a module


osresearch:
Otherwise it is not possible to initialize the Yubikey without also bringing in USB storage devices.

osresearch:
Works on the x230. "Ship it!"

osresearch:
borked out and forgot to commit the patch to `config/linux.config`

PR #189, opened 2017-04-16
load usb-storage module in x230-flash.init

flammit:


osresearch:
Thanks! When I split the `usb-storage.ko` out from the main USB drivers for security reasons (issue #160 ), I forgot that the flashing script had a helper to load them.

flammit:
No problem!  Just didn't want someone else to get stuck in the same place =)

PR #190, opened 2017-04-21
Add support for the lenovo X220

jgrip:
This pull request adds support for the lenovo X220. I've added a top level folder called blobs to hold blobs that can't be distributed easily.

Flashrom is enabled for the X220 and the chip I have in mine.

Firmware built from this is running on my X220 with qubes installed and all working.


config/x220-qubes.config
```diff
@@ -0,0 +1,25 @@
+# Configuration for a x230 running Qubes OS
```
jedahan: s/x230/x220 ?

osresearch:
The layout file is something that we might want to install into the intird.  There is [/etc/x230-layout.txt](https://github.com/osresearch/heads/blob/master/initrd/etc/x230-layout.txt) for instance.

On the x230 I've relocated the TPM measurement to be earlier in the boot process so that we can measure the ROM stage before bringing up memory; this change hasn't been pushed to mainline coreboot and probably needs further discussion. https://github.com/osresearch/coreboot/commit/716a7328e2cfd84ace3ef901da6fe3e71dbf89ed

The diff-of-patches for flashrom is hard to read, but that's not your fault...

osresearch:
Merged.  Thanks for adding x220 support!

PR #191, opened 2017-04-22
Add Tails boot option

flammit:
Took a quick pass at adding an option to boot Tails in Heads.  It follows the same pattern as the Qubes boot for signing and checking the hashes of the relevant boot files.

Not sure I did this all correctly, but it seems to works locally.  Would appreciate any feedback.

### DIFF
--- a/initrd/init
+++ b/initrd/init
@@ -32,14 +32,14 @@ hwclock -l -s
 if [ ! -x "$CONFIG_BOOTSCRIPT" ]; then
 	recovery 'Boot script missing?  Entering recovery shell'
 	# just in case...
-	tpm extend -ix 4 recovery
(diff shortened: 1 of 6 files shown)


flammit:
Closing this as it was poorly thought out with respect to the state it leaves behind.

initrd/init
```diff
@@ -32,14 +32,14 @@ hwclock -l -s
 if [ ! -x "$CONFIG_BOOTSCRIPT" ]; then
 	recovery 'Boot script missing?  Entering recovery shell'
 	# just in case...
-	tpm extend -ix 4 recovery
+	tpm extend -ix 4 -ic recovery
```
osresearch: Oops. This is a serious PCR handling bug. #198 to track it.

Issue #196, opened 2017-04-23
Boot menu and grub.cfg parser


osresearch:
Right now the path to the boot scripts are hard coded and very specialized to booting Qubes (as mentioned in #192).  For non-Qubes users, for installing a new OS from a USB device or for situations when you want to boot something like Tails (cancelled pull request #191), it would be very useful to have a way to parse the existing `/boot/grub/grub.cfg` file to build a boot menu.

The syntax is very much like `/bin/sh`, so it might be possible to build enough scaffolding to just execute it. The PCRs would have already been extended, so an attacker who controlled the file would not be able to retrieve the disk encryption keys (and the system will very soon be executing code from that device anyway).

flammit:
#200 seems to cover a lot of the functionality required for non-Qubes boot now.  In fact, this works for Qubes boot as well, but you're required to enter the disk password on start.

FYI - to get to feature parity on the qubes-update/qubes-init scheme, the generic mechanism still needs the rollback protection (TPM counters are not used in hashes yet) and TPM secret key management.

PR #197, opened 2017-04-23
add boot support for Tails

### LINKED ISSUE #195: add boot support for Tails
Not sure if there's a desire to add more boot options before there's a more complex boot screen, but I've taken a shot at doing this for Tails.

There are a couple of boot options that are supported: 

1) a signed Tails ISO in the root of the drive (`/tails-i386-2.12.iso` and `/tails-i386-2.12.iso.sig`) 

2) the standard USB key made by Tails Installer.  

The signed ISO option requires that the `/tails-signing.key` be present as the signing key is quite large (450k).  The init script will expect the file be user-signed (like the `boot.hashes.asc`) or that it matches a hash in the script.

With the USB drive made from the Tails Installer, there's no good way to determine if the USB key has been tampered with, so there's a warning/confirmation on boot to proceed.

Will post a PR shortly.

flammit:
Closes #195


flammit:
Would appreciate comments / feedback on this PR.  After many different attempts, I couldn't get the GRUB parsed kernel command line injection to kexec to work and left it commented in case someone else wanted to give it a shot.

flammit:
Apologies for the commit noise - I messed up the rebase after adding support to verify the TOTP.

flammit:
Closing this in favor of a PR that supports both #195 and #196

PR #199, opened 2017-04-23
Updates keylime-init

geedo0:
Changes the host for stage 2 keylime.tar.gz payload and adds failover for keylime_node.

### DIFF
--- a/initrd/etc/keylime-init
+++ b/initrd/etc/keylime-init
@@ -1,5 +1,7 @@
 #!/bin/sh
-# Bring up the x230's NIC, get a DHCP address and invoke keylime
+# Bring up the x230's NIC, get a DHCP address, and invoke keylime
+
+. /etc/functions
 
 insmod /lib/modules/e1000e.ko
 udhcpc -n
--- a/initrd/etc/keylime-init
+++ b/initrd/etc/keylime-init
@@ -1,17 +1,25 @@
 #!/bin/sh
 # Bring up the x230's NIC, get a DHCP address and invoke keylime
 
+recovery_on_error() {
+  echo "!!!! $1"
+  tpm extend -ix 4 -ic "recovery"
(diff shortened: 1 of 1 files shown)


initrd/etc/keylime-init
```diff
@@ -1,17 +1,25 @@
 #!/bin/sh
 # Bring up the x230's NIC, get a DHCP address and invoke keylime
 
+recovery_on_error() {
```
osresearch: This function is now provided in `/etc/functions` in the initrd.

initrd/etc/keylime-init
```diff
@@ -1,17 +1,25 @@
...
+}
+
 insmod /lib/modules/e1000e.ko
 udhcpc -n
 
 cd /
-wget-measure.sh 6 http://192.168.1.5/keylime.tar.gz
-tar xf keylime.tar.gz
+wget-measure.sh 6 http://keylime-server/keylime.tar.gz
+tar -xvzf keylime.tar.gz -C /
```
osresearch: the `-C /` shouldn't be necessary if you've also done the `cd /` a few lines above.

initrd/etc/keylime-init
geedo0: Neat! I'll make sure to reuse that.

osresearch:
Do you have any further updates after this weekend's hacking? This can be merged if it is ready.

geedo0:
The hack we did to handle the library conflict with Keylime is not pretty so we'll have to hold off on this. Would you be opposed to either invoking ld-musl explicitly or manipulating LD_LIBRARY_PATH in this script to launch Keylime?

osresearch:
What's the status on the PR?  Is it still relevant?

geedo0:
I'll close this PR and resubmit when we have something end to end ready.

PR #200, opened 2017-04-29
adds a USB boot option with basic parsing to kexec

### LINKED ISSUE #195: add boot support for Tails

flammit:
Supports booting from USB media using either the root device or
a signed ISO as the boot device.  Boot options are parsed with
quick/dirty shell scripts to infer kexec params.

Closes #195 and begins to address #196


initrd/etc/functions
```diff
@@ -28,3 +28,38 @@ recovery() {
 pcrs() {
 	head -7 /sys/class/tpm/tpm0/pcrs
 }
+
+confirm_totp()
```
osresearch: This makes lots of sense to move it into a function.  Thanks for extracting it.

initrd/bin/usb-parse-boot
```diff
@@ -0,0 +1,171 @@
...
+				syslinux_multiboot_append
+			else
+				append="$val"
+			fi
+			;;
+	esac
+}
+
+state="search"
+while read line
```
osresearch: Is this the best way that we have to parse grub config files? If we can come up with a general way to do this it would also simplify the normal Qubes setup, rather than having very specialized tools that the user has to run.

flammit: Definitely not the best we can do, especially for fixed installations like the normal Qubes installation.  My goal was to get some external options working first (including cases where grub might not exist), so this simple parsing was just quick and dirty.

I've looked into replicating the grub environment/commands necessary to make at least the qubes install/update/boot process a little more generic and more extensible later (support loopback command for iso etc.).  It should be straight forward to pull out the boot devices and kexec params (if we're focusing Qubes), but I think there will still be special handling for the secret management and the xen binary replacement.

Did you want me to update this PR for this or start another one to try to rework the fixed media / normal Qubes procedure to require less user input?  If there are other fixed OS setups that you're targeting to handle, let me know and I can take a look at the boot routines too.

flammit:
Any thoughts on this PR? I'm happy to change it or close it out as not necessary.

flammit:
Changed the usb-boot to always use custom Heads xen whenever a multiboot config is detected (not great but not sure what would be a better filter).  

This allows Qubes to be installed from USB without the use of the custom qubes-install script.  Other xen-based distros seem to work; successfully tested alpine-xen from USB.

flammit:
Just added a new `x230-generic.config` which should allow for arbitrary installations of OSs that are supported by the kexec loading, albeit in a less secure manner by default.

Will post an example/docs for customizing the persistent configuration in a bit.

flammit:
Successfully tested out `x230-generic.config` on a clean setup with default Ubuntu 16.04 and Fedora 25 installations (installation from USB then subsequent normal boots from HD).

A user has the option to make persistent modifications to the non-Qubes boot process by creating one or more of the following files:

| file | description |
| ---- | ---- | 
| kexec_menu.txt | contains multiple options for parameters to the kexec command | 
| kexec_hashes.txt | a sha256sum file from within the respective boot directory |
| kexec_iso_add.txt | a sh variable to override the standard ISO kernel argument additions |
| kexec_iso_remove.txt | a sh variable to override the standard ISO kernel argument removals |

These can be placed in any of the following locations:

| location | description | 
| ---- | ---- | 
| /boot/ | used during x230-generic.config HD boot | 
| /media/ | used during standard USB boot |  
| /media/kexec_iso/$ISO_FILENAME/ | used during USB boot from a particular ISO file |

These files are only used if there is an appropriate signature for them in `kexec.sig` covering all `kexec*.txt` in that location.  This can be generated by running `kexec-sign-config /boot/`, etc.

If there is no persistent `kexec_menu.txt`, the boot directory will be searched for grub/syslinux-like configurations and it will be generated on-the-fly (for any of the HD/USB/USB-ISO locations).  Creating a persistent `kexec_menu.txt` can be useful to limit the options displayed or to make custom persistent alterations to xen or kernel params.

`kexec_menu.txt` has a simple layout with a single line per boot option:

```
description 1|elf|kernel ... |initrd ... |append ...
description 2|multiboot|kernel ... |module ... |module ... 
```

This is a sample `kexec_menu.txt` covering the expected options (derived from grub.cfg):

```
Ubuntu|elf|kernel /vmlinuz-4.8.0-58-generic|initrd /initrd.img-4.8.0-58-generic|append root=/dev/mapper/ubuntu--vg-root ro quiet splash crashkernel=384M-:128M crashkernel=384M-:128M
Memory test (memtest86+, serial console 115200)|elf|kernel /memtest86+.bin|append console=ttyS0,115200n8
Qubes, with Xen hypervisor|multiboot|kernel /xen-4.6.5.gz placeholder |module /vmlinuz-4.4.67-13.pvops.qubes.x86_64 placeholder root=/dev/mapper/luks-UUID ro rd.qubes.hide_all_usb|module /initramfs-4.4.67-13.pvops.qubes.x86_64.img
```

If there is a persistent `kexec_hashes.txt`, the boot will fail when the file hashes don't match the expected values.  By default, no such checks are made.

When booting from an ISO file on a USB drive, it must be signed by a valid key in the Heads ROM and the boot process will fail if invalid. The `kexec_iso_add.txt` and `kexec_iso_remove.txt` are useful to inject the appropriate kernel arguments to allow it to load properly. ISOs for Ubuntu, Fedora Live, Qubes Install require that `kexec_iso_add.txt` contains to load properly:

```
iso-scan/filename=/${ISO_PATH}
```

Note that currently, any multiboot entry is interpreted as a Xen-variant and `kexec-boot` uses the internal Heads ROM `/bin/xen.gz` with `no-real-mode reboot=no`.  A slightly better solution would be to parse unknown grub/syslinux files should be use a third type called `xen` for which this occurs, and leave `multiboot` entries alone to manually support non-Xen multiboot kernels.

flammit:
FYI - please hold off on review of this PR.  I have a commit I'm still testing which enables a generic form of TPM secret management that'll work with Qubes as well as Ubuntu, etc.

flammit:
As of this last commit, you should be able to do everything that the qubes-* scripts supported when using the generic config along with the ability to do so with non-Qubes systems (tested successfully w/ Qubes and Fedora 25, but couldn't figure out how to make Ubuntu's initramfs cryptroot use the `/secret.key`).  I've updated some preliminary doc additions here: https://gist.github.com/flammit/fc77718651d59d720547dda6c544a04b.

PTAL when you get a chance.  I'm going to hold off on any further changes as there is already quite a lot here.  Sorry if the file naming scheme is inappropriate - I didn't spend much time on it and figure these can be fixed later.  There is a bit of older redundant code just to allow the coexistence of the qubes-init and generic-init, but it should be safe to remove in the future.

flammit:
Updated doc to include instructions for using `kexec-save-key` to use the TPM secret in Ubuntu 16.04 and Debian 9.

osresearch:
When the `kexec` command is invoked (https://github.com/osresearch/heads/pull/200/files#diff-c41bd3f22c6c36bddeb4ccf8fed6ba2bR133), I think it should be measured.  Otherwise it might be possible to pass arguments that will lead to insecure operation.

How long does the `initrd.cpio` rebuild take? (https://github.com/osresearch/heads/pull/200/files#diff-4ea64964f165faa83eed5691943775e1R61) I tried a similar approach but found that it took several seconds, which is why I wanted to instead use the concat method to add just the few files as a separate initrd.

I plan to try installing this on my x230 test machine tomorrow.

osresearch:
Ah, I missed that you had fixed the initrd padding size:
```
dd if="$INITRD" of="$SECRET_CPIO" bs=512 conv=sync \
|| die "Failed to copy initrd to /tmp"
```

flammit:
I agree with the kexec comment and I wrestled a bit with how to "protect" the process (both the file contents as well as all the command line parameters).  Definitely happy to take suggestions on how to make it better.

Right now, `kexec-boot` is like `qubes-boot` where it is assumed that the inputs are trusted.  Establishing some trust in these parameters happens earlier in `kexec-select-boot` (which is the only non-recovery path to `kexec-boot`).

If `CONFIG_BOOT_REQ_HASH=n`, we assume that the user doesn't require any signed file hashes and will allow any boot option to proceed (make like most other BIOSes).

If `CONFIG_BOOT_REQ_HASH=y`, then a `kexec-boot` call will only be attempted if a valid signature has covered all the custom boot parameters and some form of boot file hashes has been verified.  The valid signature can be satisfied either in the form of a detached signature (`kexec.sig`) of the hashes of `kexec*.txt` in the associated parameter directory via `kexec-sign-config`, or as a valid ISO signature (since the boot content and config are immutable).

For convenience, saving a default boot option via the menu or `kexec-save-default` does this `kexec-sign-config` call in addition to locking down all file hashes and kexec-boot parameters. A valid default boot implies that the expected parameters are always used and files not tampered with.

There are scenarios though where a user may manually sign a situation where boot parameters are not trusted (using only a `kexec_hashes.txt` and not locking down a custom `kexec_menu.txt`), but not sure where to draw the flexibility line here.

Good luck with your testing!

osresearch:
It boots!

Some initial thoughts: the TOTP confirmation could also be used to select the next boot mode.  There is currently a fast timeout for recovery versus USB, followed by an unbounded wait for TOTP `y`.

The first time through I forgot that I had to have my yubikey to sign the hashes, but it fails in the right way (back to the boot menu).

It isn't clear when the LUKS key is not retrieved from the TPM.  This should probably be a more verbose error message (with an ack from the user) since it indicates that something has seriously gone wrong
or that they have not completed the setup with running `seal-key`.  They could be prompted to do that at this time.

I'm not being prompted for my TPM unlock password, so Qubes is dropping me into its recovery shell since it can't unlock the disks.  (The does make me worry about a possible local security failure -- if a local attacker can convince the Qubes initrd to drop to a recovery shell, `/secret.key` will be present, along with USB drivers. They would have needed the TPM unlock password to get there, but maybe they were able to shoulder surf it)

osresearch:
There appears to be a minor race condition in `check_config` in `/etc/functions` between when it checks the signatures on `/boot/kexec*.txt` and when it copies them into `/tmp/kexec`.

It also took me a try to figure out that the filenames in the signature file are `/boot//kexec.whatever.txt`, with two `/` since there is a trailing /.

Easiest fix might be to sign a file that has no leading path indicator, which would also make them portable.  Are there issues with doing this?

osresearch:
Is it possible to interrupt the default boot to select an alternate kernel from the grub.cfg?  And should this prevent TPM LUKS key access? (I think the answer is yes, it should, since a "recovery mode" option in grub.cfg might allow policy circumvention)

flammit:
Good call on the TOTP confirm to drive the initial boot direction.  I didn't want to mess too much with anything that could interfere with init/qubes-init, but if you're okay with it, I'll refactor that a bit.

As for exposure of the `/secret.key` in initrd, for Qubes/dracut you can set `rd.shell=0 rd.emergency=reboot` in the kernel command line to prevent shell access.  Maybe adding `panic=N` can also help (see http://hmarco.org/bugs/CVE-2016-4484/CVE-2016-4484_cryptsetup_initrd_shell.html).  But point taken, some time should be spent to figure out the risk here.

Not sure I follow the `check_config` race condition.  Are you saying we should copy, then check sigs and then delete from tmp if bad?

Sorry I knew about that trailing `/` problem, there's no problem with fixing it.  I'll update that ASAP.

There's a few ways to that default boot can fail but leave you available to select an alternate kernel AND have the key accessible.  The primary use case is when there's a kernel upgrade and you just want to reset the default and boot right away.  Maybe it's safer thought to be in recovery mode for the resign and then require reboot for TPM key unseal.

osresearch:
I'm going to do a pull and merge on this branch and then we'll patch up a few of these minor issues.

osresearch:
Hold off on changing your branch; I'm about to pull it into master.  I fixed the trailing / and cleaned up the remaining qubes-* files that are no longer relevant.

Issue #202, opened 2017-06-06
LVM thin-provisioned volumes


jpouellet:
It is my understanding that Qubes plans to moves to LVM thin provisioning by default in R4.0 (https://github.com/QubesOS/qubes-issues/issues/2412), and this does not appear to be supported by Heads at the moment:
![image8](https://user-images.githubusercontent.com/1231207/26848186-914e21b2-4acd-11e7-91ed-5d27d839e0e9.jpeg)

osresearch:
Can you try turning it on in the `config/linux.config` file and seeing if it works?
```
diff --git a/config/linux.config b/config/linux.config
index 7ad8bab..14cc7f8 100644
--- a/config/linux.config
+++ b/config/linux.config
@@ -946,9 +946,11 @@ CONFIG_BLK_DEV_DM=y
 # CONFIG_DM_DEBUG is not set
 CONFIG_DM_BUFIO=y
 # CONFIG_DM_DEBUG_BLOCK_STACK_TRACING is not set
+CONFIG_DM_BIO_PRISON=y
+CONFIG_DM_PERSISTENT_DATA=y
 CONFIG_DM_CRYPT=y
 # CONFIG_DM_SNAPSHOT is not set
-# CONFIG_DM_THIN_PROVISIONING is not set
+CONFIG_DM_THIN_PROVISIONING=y
 # CONFIG_DM_CACHE is not set
 # CONFIG_DM_ERA is not set
 # CONFIG_DM_MIRROR is not set
@@ -2496,7 +2498,7 @@ CONFIG_CRC32_SLICEBY8=y
 # CONFIG_CRC32_SARWATE is not set
 # CONFIG_CRC32_BIT is not set
 CONFIG_CRC7=m
-CONFIG_LIBCRC32C=m
+CONFIG_LIBCRC32C=y
 CONFIG_CRC8=m
 # CONFIG_AUDIT_ARCH_COMPAT_GENERIC is not set
 # CONFIG_RANDOM32_SELFTEST is not set
```

osresearch:
(unrelated to the LVM topic -- are you running Xen 4.6.1?)

jpouellet:
> Can you try turning it on in the config/linux.config file and seeing if it works?

After a while of rebuilding & reflashing Heads, and reinstalling Qubes with thin partitioning, same error. None of the usual ways to read config of the running kernel seem to work, and I didn't make clean. Rebuilding again to see if that might have been it...

[...]

PR #210, opened 2017-07-09
Started update of xen to 4.8.1-4 for Qubes 4

flammit:
No reproducibility tests yet, but compilation, installation and boot seem to work (though VMs are boot seem unstable).

Had to alter the `head.S` Xen patch due to a change in a register.


flammit:
Does it make sense to make a separate branch for Qubes 4 for testing purposes?

osresearch:
I haven't tested Qubes 4.0 yet; is there a working ISO for experimentation?

LVM thin provisioning #202 is probably a blocker.

osresearch:
Removing the `vga.c` hacks from Xen (issue #227) seems to have caused a conflict, but hopefully makes for an easier overall merge.

flammit:
There are a couple posts in qubes-devel with preliminary isos. Here's the latest:

https://groups.google.com/forum/m/#!topic/qubes-devel/jeAsPQjImlU

When I tried the first one on x230, I couldn't get sys-USB to start (USB pci issue), but it seemed to install and boot okay. LVM-thin is an option but not the default yet.

flammit:
Rebased patch to current master (w/o the vga tweaks).

Tested the 20170718 ISO on an x230.  Everything seems to work (usb/network/firewall) but with the caveats described in the qubes-devel post.  Be sure to use LVM instead of LVM thin provisioning.

flammit:
Looks like there are additional qubes xen command line parameters from grub that may need to be supported.

flammit:
FYI - https://www.qubes-os.org/news/2017/07/31/qubes-40-rc1/

flammit:
FYI there's a commit in qubes-vmm-xen that might be necessary if we update the Heads iommu/igfx behavior - https://github.com/QubesOS/qubes-vmm-xen/commit/f6fdebef909d8a8f8ffd735de5514d00402c289d 

I believe I had the iommu enabled for intel graphics (in the non-i915 mode Heads mode) and it was fine, but will test again this weekend and the new RC1 cut.

flammit:
Please hold off on this merge. I'll update PR so that the xen version is configurable.

flammit:
PTAL - the xen version can be controlled by `CONFIG_XEN_VERSION=4.6` or `CONFIG_XEN_VERSION=4.8` in the board config files: `x230-generic.config` etc.

osresearch:
Thanks for tracking the xen updates.  I'll merge this in.

Issue #223, opened 2017-07-17
Ensure that kernel updates force signing of image


osresearch:
https://github.com/osresearch/heads/pull/200#issuecomment-315801765

> There's a few ways to that default boot can fail but leave you available to select an alternate kernel AND have the key accessible. The primary use case is when there's a kernel upgrade and you just want to reset the default and boot right away. Maybe it's safer thought to be in recovery mode for the resign and then require reboot for TPM key unseal.

A kernel update should definitely require a recovery mode resign/reseal.  That might indicate a successful attack that has managed to modify the `/boot` filesystem to gain persistence.

flammit:
Absolutely.  Right now it does require a resign of the updated kernel parameters, but not a reseal of the disk key (which is the old behavior in qubes-init - just needed to do `qubes-update` to update the `boot.hashes`).

I wanted to try to allow the user to resign and immediately boot (including unsealing the TPM key) all without touching the recovery shell.  That doesn't actually work right now because TPM 5 is extended in the signing process and the subsequent TPM unseal operation will fail.

As you said, it's safer in this case to die and go to recovery to manually run `kexec-save-default` or some other means to assert the altered boot config is good.  Maybe in the future if we can trigger a reboot, the user can resign in the menu (after TPM 4 is extended) and reboot to continue in normal mode.

Issue #221, opened 2017-07-17
Combine boot mode selection with tpmtotp confirmation


osresearch:
The generic boot path has multiple possible choices for the user:

* Drop into recovery mode
* Boot the default kernel and command line (after validating signatures)
* Try to boot from a USB device
* Select a different kernel/command line from `grub.cfg`

Additionally, they need to confirm the tpmtotp output and input their TPM pin to retrieve the LUKS disk key.

`/init` should have a way to drop into recovery mode that hopefully doesn't add to the boot time.  `/bin/generic-init` can display the tpmtotp code and the user can select confirm it to boot the defaults, drop into recovery to adjust the defaults, or try to boot form USB.

osresearch:
Having a very short wait in `/init` doesn't work since busybox's ash doesn't implement read like bash: timeout 0 == infinite wait and no support for fractional seconds.  Bug filed: https://bugs.busybox.net/show_bug.cgi?id=10101

There is almost all the code in `shell_builtin.c` to do it; a minimal patch appears to work.

PR #230, opened 2017-07-22
generic boot cleanup

flammit:
Starting a PR to clean up a bunch of issues with the `generic-init` boot.  So far addresses the #223 and fixing the extra / when tab-completing and manually running the scripts.  I assumed the `show_menu` parameter to `kexec-select-boot` was created to force showing the menu and avoid using the default parameters, so that was added as well.


flammit:
After an OS install, setting the default boot entry also asks if you wish to add the TPM LUKS key, so this can now all be done from boot menu (w/o having to go to recovery shell).

It also reboots automatically after resign/reseal using the `/proc/sysrq-trigger` method.

flammit:
PTAL - hopefully these address your remaining open boot issues (looks like you already handled issue #221).

osresearch:
`reboot` and `poweroff` are definitely missing!  On qemu the busybox ones don't actually reboot and the sysrq technique seems to hang.  I need to test it on real hardware today -- I've been waiting on a fresh chipclip to be able to reflash my bricked x230.

flammit:
I should add that this branch includes the updates to Qubes 3.2/Xen 4.6.6-29 to cover the latest Qubes Security Bulletin (issue #238)

PR #285, opened 2017-11-12
Changed to coreboot patch not to call prog_segment_loaded in smm.

persmule:
If USE_OPTION_TABLE is not set when building coreboot, cbfs_boot_map_with_leak() within src/lib/cbfs.c is not called by smm codes, and will be stripped out at link time, otherwise, other modules may call cbfs_boot_map_with_leak() within smm codes, which breaks its build for prog_segment_loaded() is not built for smm codes (tested when building for Thinkpad x230 with USE_OPTION_TABLE enabled).
I believe that prog_segment_loaded() is not intended to be called within smm codes, so I create this patch, not to call prog_segment_loaded in smm, which enables us to [set USE_OPTION_TABLE and STATIC_OPTION_TABLE to make tuning cmos options easier while keeping security](https://github.com/osresearch/heads/issues/274).

It works well atop [flammit's branch for coreboot-4.6](https://github.com/flammit/heads/pull/3#issuecomment-343711911)

 [This file](https://github.com/flammit/heads/files/1440412/coreboot-x230-optable.config.txt) may be used as the config file of the coreboot part.

### DIFF
--- a/patches/gpg-1.4.21.patch
+++ b/patches/gpg-1.4.21.patch
@@ -1,6 +1,15 @@
-diff -u --recursive ../clean/gnupg-1.4.21/configure gnupg-1.4.21/configure
---- ../clean/gnupg-1.4.21/configure	2016-08-17 09:20:25.000000000 -0400
-+++ gnupg-1.4.21/configure	2017-04-05 17:59:31.785284325 -0400
+diff -u --recursive /home/hudson/build/clean/gnupg-1.4.21/configure gnupg-1.4.21/configure
+--- /home/hudson/build/clean/gnupg-1.4.21/configure	2016-08-17 09:20:25.000000000 -0400
++++ gnupg-1.4.21/configure	2018-01-20 16:55:14.502067084 -0500
+@@ -572,7 +572,7 @@
+ ac_clean_files=
+ ac_config_libobj_dir=.
+ LIBOBJS=
+-cross_compiling=no
++cross_compiling=yes
+ subdirs=
+ MFLAGS=
+ MAKEFLAGS=
 @@ -17100,7 +17100,7 @@
          _libusb_try_libs=`$LIBS $_usb_config --libs`
          _libusb_try_cflags=`$LIBS $_usb_config --cflags`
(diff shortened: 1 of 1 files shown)


flammit:
@persmule's patch will work to enable proper compilation with the USE_OPTION_TABLE settings.

I'm wondering though if it's just better to drop all the cbfs / cmos option dependencies from the smm by removing `cbfs.c` and dropping most of `src/drivers/pc80/rtc/mc146818rtc.c` during smm compilation. This way the smm can't load cbfs data without measurement (even though `cmos_layout.bin` is currently properly measured once now when writing the coreboot tables outside of smm).

It looks like the only use it at `src/southbridge/intel/bd82x6x/smihandler.c:415`:

```
static void southbridge_smi_sleep(void)
{
...
  u8 s5pwr = CONFIG_MAINBOARD_POWER_ON_AFTER_POWER_FAIL;

  // save and recover RTC port values
  u8 tmp70, tmp72;
  tmp70 = inb(0x70);
  tmp72 = inb(0x72);
  get_option(&s5pwr, "power_on_after_fail");
  outb(tmp70, 0x70);
...
```

@osresearch thoughts on which approach is better?

persmule:
New commit added: ebc6dbf14100c82bc9ca92a602fa2e9c5e060e0b

flammit:
@persmule here's a gist with the alternative changes I referenced above: https://gist.github.com/flammit/d00210803849c9dee66a0cc931152242

persmule:
for your gist, I suggest you not make patch for x230 only, but for every board with TPM available or installable (e.g. gigabyte ga-b75m-d3h) coreboot CAN support, so if you touched drivers/pc80/rtc/mc146818rtc.c, please check codes for other RTC chips as well, otherwise you had better make patches for coreboot independent to any specific board. That is why I tend to make least changes to existing codes to achieve my goal.

persmule:
The primary part of Heads is a payload based on linux kernel and initrd which in theory can run atop any coreboot build, just like [what I wrote in this article](https://github.com/hardenedlinux/Debian-GNU-Linux-Profiles/blob/master/docs/hardened_boot/heads-atop-coreboot.md).

PR #293, opened 2017-12-06
Make TPM dependency optional and controlled by flag CONFIG_TPM

persmule:
if "CONFIG_TPM=n" is set in the config file, functionalities needing TPM
could be disabled, while leaving other functionalities intact.

This will make Heads a more general-usage bootloader payload atop coreboot.


osresearch:
Looks like a pretty good addition for configurability.  The convention for `CONFIG_FOO` variables is that `y` is the only defined value; anything else is false, so the tests of `if [ "$CONFIG_TPM" != "n" ];` should be testing for equality `if [ "$CONFIG_TPM" = "y" ];` instead

persmule:
Currently CONFIG_TPM is an existing option, and in most config files it is NOT SET, so if positive condition checking is used, "CONFIG_TPM=y" should be explicitly stated.

persmule:
Condition already changed.

osresearch:
merged!

PR #294, opened 2017-12-06
Not to generate hashes for initrd or modules if module.sig_enforce=1 …

persmule:
…is present

Digital signatures could be embedded into linux kernel modules, and a kernel
with a certificate embedded in can load them only if their signature is valid,
when booted with parameter "module.sig_enforce=1" present.

In such situation, for bootloader (e.g. Heads) to verify the hashes of initrd
or modules may become unnecessary, and leaving them to the kernel may ease the
updating of initrd from GNU/Linux OSes.

Option CONFIG_BOOT_RESPECT_MOD_SIG_ENFORCE is used to enable this feature.

Test passed on my x230 atop flammit's coreboot-4.6 branch.


osresearch:
The kernel option prevents `kexec` from invoking unsigned kernels and `insmod` from loading unsigned modules, right?

How would you see this working in practice? What are the tools like for manipulating those signatures?

persmule:
The option `module.sig_enforce=1` does not prevent "unsigned" kernels being loaded, but prevents `insmod` from loading modules not signed with the private key corresponding to a certificate embedded in the kernel. It is unrelated to signing the vmlinuz itself.

Debian has a [sub-project](https://packages.debian.org/source/experimental/linux-signed) to release its kernel image in this form. The project however seems to be experimental currently, though it once released in Jessie.

PR #295, opened 2017-12-25
Use HTTPS URL for flashrom

flammit:


paulmenzel:
I wouldn’t use *fix* but *Use HTTPS URL for flashrom*. Could you please update the descriptions?

flammit:
Done

osresearch:
Merged!

We should also think about the 1.0.0 release which was just announced by Nico on the coreboot list: https://mail.coreboot.org/pipermail/coreboot/2018-January/085890.html

flammit:
Thanks!  And agreed - I'll test flashrom 1.0.0 and also start a branch targeting the upcoming coreboot 4.7 release.

PR #301, opened 2018-01-21
Add OHCI and UHCI drivers to initrd.

persmule:
USB smart card readers are most full speed devices, and there is no
"rate-matching hubs" beneath the root hub on older (e.g. GM45) plat-
forms, which has companion OHCI or UHCI controllers and needs cor-
responding drivers to communicate with card readers directly plugged
into the motherboard, otherwise a discrete USB hub should be inserted
between the motherboard and the reader.

This time I make inserting linux modules for OHCI and UHCI controllable
with option CONFIG_LINUX_USB_COMPANION_CONTROLLER.

Tested on my x200s and elitebook revolve 810g1.


osresearch:
Looks reasonable to me. I wonder if we'll need a better way to handle Linux modules as we start to build for different systems.  Perhaps a `Kconfig` based menu would be useful in the near future.

persmule:
By the way, linux-librem13v2.config is identical to linux-x230.config, so it may be possible to maintain linux-*.config for chipset traits instead of concrete boards.

e.g. companion controller is the trait of chipsets with discrete southbridge, while rate-matching hub is the trait of chipsets with pch.

PR #322, opened 2018-02-20
Add graphical init menu that uses whiptail

kylerankin:
This is a modified version of the generic-init script that uses whiptail
to generate a graphical menu. I changed two of the options so that the
user can refresh the menu to get an updated TOTP code if needed.


osresearch:
Merged with minor tweaks to formatting in `kexec-select-boot` and making the boards/ files export their config variables.

Issue #227, opened 2017-07-18
Remove Xen vga driver hacks


osresearch:
Xen supports `vga=current` as a command line option, which appears to work in the absence of any hacks on the `xen/drivers/video/vga.c` file.  This would simplify our porting to just the `no-real-mode` patches.

tlaurion:
@osresearch : It is unclear to me if `no-real-mode` patch was included upstream into xen/QubesOS? If not, why? 

> no-real-mode (x86)
> 
>     = <boolean>
> 
> Do not execute real-mode bootstrap code when booting Xen. This option should not be used except for debugging. It will effectively disable the vga option, which relies on real mode to set the video mode.
[SRC](http://xenbits.xen.org/docs/unstable/misc/xen-command-line.html)

That would provide the `no-real-mode` switch at hypervisor boot, since "vga=current" already fixes vga issues that needed xen patches from Heads.

That would greatly reduce the need from users to update Heads firmware as often as they need now, following Qubes security patches.  

@flammit's new UX would see a boot config change, could provide boot menu(m) as a result and prompt the user with a choice to set new default or boot actual config (y) since the stored and signed boot config is different then the one genrated on the fly from what was found on disk.  The user would only need to resign config changes with it's OpenGnuPG card and set new config defaults, which would link to newly patched xen, kernel and initrd provided by Qubes recent security upgrade. 

That, combined with #182 would suppress most actual needs of upgrading Heads, confining those upgrades to when Heads fixes it's own security bugs or add functionality. 

@osresearch : Any link to discussions that justified upstream rejection of `no-real-mode`?

Thanks.

osresearch:
`no-real-mode` was already part of Xen, although it still had a legacy dependency on the EBDA (Extended BIOS Data Area).  I haven't tried `vga=current` to see if that eliminates the need for the patch.

It would be ideal to be able to boot stock Qubes' Xen, so this is worth tracking down.

Here is the xen-devel thread about the patch: https://lists.xenproject.org/archives/html/xen-devel/2016-08/msg01195.html

osresearch:
> > Since this is the only mention I've seen of this setup since 2008,
> I understand that merging this patch into the Xen kernel might not be
> the highest priority.  But for time travellers looking to boot
> Coreboot + Linux payload + kexec + Xen + Qubes on their Thinkpads,
> these patches might prove to be useful.

Going forwards, there are a number of usecases for booting Xen in
non-legacy environments which have been proposed, and all would benefit
[...]

PR #327, opened 2018-02-25
qubes 4.0rc4 and init refactoring

flammit:


flammit:
Split out the KGPE-D16 support to another branch

osresearch:
The qubes part of this PR is no longer relevant after fixing #227, right?

I'm hesitant on the split of `boards/$(BOARD).config` and `.mk` since it makes for more files that have to be tracked and possibly maintained in sync.  What do you think about a way to generate the `/etc/config` file from all of the `$(CONFIG_.*)` variables?

`export | grep '^CONFIG_'` almost works, except that most of these configs are not exported.

flammit:
Yup the qubes isn't needed now and the export way to get `/etc/config` SGTM.  Let me know if you need me to fix.

osresearch:
I've pushed a fix that generates `/etc/config` from the environment: https://github.com/osresearch/heads/commit/4f5432bb461ad63995afb45aba34534e1d8c3d23

What else needs to be merged from this PR?  I think the TPM fixes went in as part of merging #293

flammit:
Just the `initrd/init` changes, but as long as it looks like #331 in the end, it should be fine to close / ignore this PR.

initrd/init
```diff
@@ -16,50 +16,36 @@ mkdir /proc /sys /dev /tmp /boot /media 2>&- 1>&-
 mount /dev 2>/dev/ttyprintk
 mount /proc 2>/dev/ttyprintk
 mount /sys 2>/dev/ttyprintk
-mount /sys/firmware/efi/efivars
-
-# Setup the pty psudeo filesystem
-mkdir /dev/pts
```
osresearch: What's wrong with always mounting `/dev/pts`?

initrd/init
```diff
@@ -69,35 +55,76 @@ if [ ! -z "$CONFIG_USB_BOOT_DEV" ]; then
 	echo >> /etc/fstab "$CONFIG_USB_BOOT_DEV /media auto defaults,ro 0 0"
 fi
 
-if [ ! -x "$CONFIG_BOOTSCRIPT" ]; then
-	recovery 'Boot script missing?  Entering recovery shell'
-	# just in case...
-	tpm extend -ix 4 -ic recovery
-	exec /bin/ash
+# Now it is safe to print a banner
+if [ ! -z "$CONFIG_LINUXBOOT" ]; then
```
osresearch: In my recent `export CONFIG_*` commit I didn't export the linuxboot config. Do you think it is important to be able to distinguish them?

Also, `[ ! -z "$CONFIG_LINUXBOOT" ]` will go down this path if `CONFIG_LINUXBOOT=n`.  It is better to test against `y`.

initrd/init
```diff
@@ -69,35 +55,76 @@ if [ ! -z "$CONFIG_USB_BOOT_DEV" ]; then
...
-	boot_option
-echo
-
-if [ "$boot_option" = "r" ]; then
-	# Start an interactive shell
-	recovery 'User requested recovery shell'
-	# just in case...
-	tpm extend -ix 4 -ic recovery
-	exec /bin/ash
+# Setup remote attestation interface
```
osresearch: The network setup should move to a separate setup file, rather than living here in `/init`.

initrd/init
flammit: Mmm... `CONFIG_UNIX98_PTYS=y` isn't always built in the kernel so mounting devpts can fail but I guess that's not awful (added `CONFIG_LINUX_PTY` in board config in PR #331).

initrd/init
flammit: Fine by me - I just wanted minimal equal functionality changes to what you had done for linuxboot.

flammit:
If it makes sense, I can just rebase/refactor #327, #330 and #331 to current master and incorporate all your comments.

PR #329, opened 2018-02-27
refactor coreboot-4.7.patch for coreboot's HEAD

persmule:
coreboot has heavily refactored its TPM framework, which makes the
original patch designed for coreboot-4.7 unable to apply to its HEAD,
so I provide a refactored patch against coreboot's current head
(5fd1d5ad1258407ade7fe8f72672c878bdfd8f05), which may be useful for
you during future developments or to integrate heads atop your customed
coreboot builds.

### DIFF
--- a/patches/coreboot-HEAD.patch
+++ b/patches/coreboot-HEAD.patch
@@ -1,0 +1,476 @@
+diff --git ./src/Kconfig ./src/Kconfig
+index 02abed2dd4..7f5155a8c5 100644
+--- ./src/Kconfig
++++ ./src/Kconfig
+@@ -253,6 +253,21 @@ config BOOTSPLASH_FILE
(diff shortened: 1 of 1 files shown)


flammit:
Maybe add as `patches/coreboot-4.8` or `patches/coreboot-HEAD` in case someone wants to use it with an altered `module/coreboot` file?

persmule:
@flammit

osresearch:
Hopefully the recent commit for PR #348 will make this easier to do.  I've split the coreboot patch into a measured boot, sandybridge and kgpe-16 patches, and the purism team has a series of skylake patches for their board.

osresearch:
I think the plan is to move to coreboot 4.8.1, which is in process by the Librem team and @flammit's WIP #420

PR #331, opened 2018-02-28
Additional KGPE-D16 work

flammit:


flammit:
This is dependent on #330 and #327.

PR #332, opened 2018-03-01
exclude gnupg keyring files from revision control

persmule:
Since gnupg and ssh keyring files are highly user-specific, they should not be
managed with revision control system, otherwise they can cause trouble
when rebasing.


flammit:
Maybe include /.ssh/authorized_keys too?

persmule:
@flammit 
Just amended.

osresearch:
Those files should move into the currently unimplemented and unspecified non-volatile user partition.  Which we should figure out how to make it work for coreboot and linuxboot, on both real hardware and qemu.

osresearch:
This was merged as part of #357, right?

PR #336, opened 2018-03-01
Walkaround envvars passed through make(1) when generating etc/config

persmule:
"export" statements included or declared in a Makefile proves literally
(with no escape) passed to the shell, which may result in shell envvars
containing literal double quote if SHELL is set as bash, and they further
becomes statements containing `\"` when printed with command export.

This behavior could be observed by the makefile inlined at the end.

This commit adds a regexp to sed to remove those `\"`.

export QUOTE="QUOTE"
SHELL := /bin/bash
.SHELLFLAGS := -o pipefail -c
export-quote:
        export|grep QUOTE

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -426,7 +426,8 @@ $(build)/$(BOARD)/tools.cpio: \
 		mkdir -p "$(initrd_dir)/etc" ; \
 		export \
 			| grep ' CONFIG_' \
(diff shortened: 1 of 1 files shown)


osresearch:
FYI: This causes a build failure if there are no `$CONFIG_*` variables in the environment, which is likely an error in the user's configuration.  Not sure if we should hack around this case.

Issue #338, opened 2018-03-01
Allow local package directory


osresearch:
For CI and other applications it would be useful to have an environment variable that points to an external directory for packages so that they don't have to be downloaded everytime.

osresearch:
This works for the Heads packages, but musl-cross sitll downloads from the network.

Perhaps we should re-implement musl-cross so that it is part of our build rather than depending on this external tree.

@vejmarie - what do you think?

osresearch:
This breaks for coreboot and musl-cross.  We should address those so that we can have an offline build environment.

PR #341, opened 2018-03-04
Add PKGS_PROXY as a Makefile parameter. The variable is providing the…

vejmarie:
… base url of a Proxy Pkgs server.

The directory structure on it shall be preconfigured with PKGS_PROXY/packages/ (or http://10.0.0.1/packages)
In the case a tarball is not found within the proxy, the standard url is used as a backup option.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,7 @@ config		:= $(pwd)/config
 INSTALL		:= $(pwd)/install
 log_dir		:= $(build)/log
 
+PKGS_PROXY	?= ""
 BOARD		?= qemu-coreboot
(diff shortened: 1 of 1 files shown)


osresearch:
Related to issue #338

osresearch:
This might be easier done if the `$(packages)` directory could be over-ridden on the command line or config file.  That would remove the need to special case fetching from a proxy or package directory.

osresearch:
That works: https://github.com/osresearch/heads/commit/ab460645a38fbb9e1c34b81c9a1caedf7ec6bfb2

PR #344, opened 2018-03-07
kexec-boot patch so that xen can be booted from bootdir and referred …

tlaurion:
…xen.gz and not internal flash xen.gz


osresearch:
The commit wasn't signed, so I've merged and amended the commit to sign it.

osresearch:
(this was merged, but github can't tell since I re-signed it)

PR #348, opened 2018-03-09
Add ability to specify series of patches

kakaroto:
Coreboot patches are becoming hard to maintain, and resolving conflicts in the patch if two people decide to update the patch file makes it even harder. 
This adds the ability to use a .series file which contains filenames for patches to apply. 
Also add various patches to coreboot 4.7 to add measured boot support to skylake devices, as well as IOMMU, and TPM support to librem 13v2.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -190,6 +190,14 @@ define define_module =
 		( cd $(build)/$($1_dir) ; patch -p1 ) \
 			< patches/$1.patch; \
 	fi
+	if [ -d patches/$1 ] && \
+	   [ -r patches/$1 ] ; then \
+		for patch in patches/$1/*.patch ; do \
+			echo "Applying patch file : $$$$patch " ;  \
+			( cd $(build)/$($1_dir) ; patch -p1 ) \
+				< $$$$patch ; \
+		done ; \
+	fi
 	@touch "$$@"
   else
     # Fetch and verify the source tar file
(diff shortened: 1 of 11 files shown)


osresearch:
I like this idea.  Do you think it is necessary to have a file that lists the patches, or should it just patch every file in the directory?

Pros: one less place to have to list things

Cons: ordering is dependent on naming convention.

kakaroto:
I just made the coreboot-4.7 directory out of convenience to avoid having too many files in patches/ but I hadn't thought of just getting all files in the subdir with the same name. That would work.
I don't really have a preference one way or the other actually, so whichever way you decide, just let me know and I can adapt the patch if needed.

Additional pros in using a series file is that
*  it's easier to disable patches if needed or while testing without actually deleting the file
* Could apply the same patch file to multiple modules (or multiple versions of the same module) instead of duplicating files.. though I don't think this is useful for heads since every board uses the same module versions, so there's no sharing required.

osresearch:
The directory idea makes sense -- I was wondering just patching everything in `patches/$(module)/*.patch`, which would expand based on filename (instead of listing them in the `$(module).series`).

Being able to turn patches on/off with the series file makes some sense, although you can also just move the patch in or out of the directory.  I lean towards the existence approach since it a) allows us to do things like "detect if any applied patches are newer than the .canary file in the directory" and b) eliminates the class of error "I commited the series file but forgot the patch" or "I added the patch but it isn't being applied".

osresearch:
Iterating over the patches might also be better than cating them together -- that way we can give a better error message as to which one failed.

kakaroto:
Sounds good to me. I'll change it to look for the files instead and ignore the series file. Also yeay signed commits!

kakaroto:
PR updated. I left the patches/coreboot-4.7.patch as is instead of moving it into the directory, it gets applied first, then the other ones get applied on top of it in sequence.

PR #350, opened 2018-03-09
Librem13v2 blobs and config

kakaroto:
Add a script to get the binary blobs needed for the librem13v2 and update the corbeoot config to use the correct blobs path (and remove bootsplash/bootorder/vgabios/vbt). Update linux config to add i915 support.

This however probably won't work to build a correct image until PR #348 is merged.


osresearch:
I've merged this one, although should the naming convention be the same as the board: `blobs/librem13v2`? Or are there multiple skylake machines that will use these blobs?

kakaroto:
Yeah, it's just because we don't have a librem15v3 board config in heads for now, but it will use the same blobs. Coreboot itself is using mainboard/purism/librem_skl for the mainboard dir, that's why I named it librem_skl for the blobs as well.

PR #355, opened 2018-03-11
Init changes

flammit:
This PR moves the network initialization logic to a separate bootscript as well as introduces an optional recovery console in situations where multiple are available.

Tested on qemu-linuxboot, qemu-coreboot, kgpe-d16.


flammit:
Tested on x230 now too - slight weirdness with double Heads motd but Qubes 4rc5 USB install is good.

config/busybox.config
```diff
@@ -470,9 +470,9 @@ CONFIG_FEATURE_XARGS_SUPPORT_ARGS_FILE=y
 # CONFIG_BOOTCHARTD is not set
 # CONFIG_FEATURE_BOOTCHARTD_BLOATED_HEADER is not set
 # CONFIG_FEATURE_BOOTCHARTD_CONFIG_FILE is not set
-CONFIG_HALT=y
-CONFIG_POWEROFF=y
-CONFIG_REBOOT=y
+# CONFIG_HALT is not set
```
osresearch: Why were these removed?

boards/x230.config
```diff
@@ -4,6 +4,7 @@ CONFIG_COREBOOT_CONFIG=config/coreboot-x230.config
 CONFIG_LINUX_CONFIG=config/linux-x230.config
 
 CONFIG_CRYPTSETUP=y
+CONFIG_FLASHROM=y
```
osresearch: I really should modify the x230-flash script to use flashtool. It is so much simpler than flashrom and takes up a fraction of the space.

config/busybox.config
flammit: These interfere with the sysrq-based scripts in bin on x230. I think those work always (as long as the kernel has it enabled).

initrd/init
```diff
@@ -67,11 +51,12 @@ if [ ! -z "$CONFIG_USB_BOOT_DEV" ]; then
...
 
-if [ ! -x "$CONFIG_BOOTSCRIPT" ]; then
-	recovery 'Boot script missing?  Entering recovery shell'
-	# just in case...
-	tpm extend -ix 4 -ic recovery
-	exec /bin/ash
+# Setup recovery serial shell
+if [ ! -z "$CONFIG_BOOT_RECOVERY_SERIAL" ]; then
+	stty -F "$CONFIG_BOOT_RECOVERY_SERIAL" 115200
+	pause_recovery 'Console recovery shell' \
```
osresearch: Does this require user intervension on every boot with `CONFIG_BOOT_RECOVERY_SERIAL` set?

flammit: Nope - you can boot normally via generic-init through the internal serial connected to the BMC. This is just here to enable recovery shell access on the external serial port on the back of the motherboard (in cases when the BMC is not active or down). This allows you to flash openbmc when physically local to the machine. You can also do this remotely if you configure network-init-recovery as the network bootscript (same drop to recovery shell on ssh into the machine)

PR #357, opened 2018-03-12
Read and measure CBFS files into initrd

### LINKED ISSUE #182: GPG key and config should live in separate cbfs file
Right now the user's GPG key has to be baked into the `initrd.cpio` file along with all of the system utilities.  It would be easier to update if it could live in a separate CBFS file in the ROM so that users do not need to rebuild all of Heads.

flammit:
Closes #182 and depends on PR #355.  Allows the user to add files such as keys to the reproducibly-build coreboot.rom via cbfstoo as long as there CBFS name is prefixed as `initrd/*`:

```
./build/coreboot-4.7/x230/cbfstool ./build/x230/coreboot.rom add -t raw -f ~/.ssh/id_rsa.pub -n "initrd/.ssh/authorized_keys"
./build/coreboot-4.7/x230/cbfstool ./build/x230/coreboot.rom add -t raw -f ~/.gnupg/pubring.gpg -n "initrd/.gnupg/pubring.gpg"
./build/coreboot-4.7/x230/cbfstool ./build/x230/coreboot.rom add -t raw -f ~/.gnupg/secring.gpg -n "initrd/.gnupg/secring.gpg"
./build/coreboot-4.7/x230/cbfstool ./build/x230/coreboot.rom add -t raw -f ~/.gnupg/trustdb.gpg -n "initrd/.gnupg/trustdb.gpg"
```

Tested on x230.


flammit:
Also successfully tested measurement by altering CBFS file name and data with flashtool on x230 after TOTP seal

initrd/bin/network-init-recovery
```diff
@@ -0,0 +1,29 @@
...
+
+if [ -e /sys/class/net/eth0 ]; then
+	# Setup static IP
+	if [ ! -z "$CONFIG_BOOT_STATIC_IP" ]; then
+		ifconfig eth0 $CONFIG_BOOT_STATIC_IP
+	fi
+	# TODO: setup DHCP if available
+	ifconfig eth0 > /dev/ttyprintk
+
+	# Setup the ssh server, allow root logins and log to stderr
```
paulmenzel: The verb is spelled with a space *Set up*.

initrd/bin/network-init-recovery
```diff
@@ -0,0 +1,29 @@
...
+if [ -f /lib/modules/e1000e.ko ]; then
+	insmod /lib/modules/e1000e.ko
+fi
+
+if [ -e /sys/class/net/eth0 ]; then
+	# Setup static IP
+	if [ ! -z "$CONFIG_BOOT_STATIC_IP" ]; then
+		ifconfig eth0 $CONFIG_BOOT_STATIC_IP
+	fi
+	# TODO: setup DHCP if available
```
paulmenzel: The verb is spelled with a space *Set up*.

initrd/bin/network-init-recovery
```diff
@@ -0,0 +1,29 @@
...
+if [ -f /lib/modules/e1000.ko ]; then
+	insmod /lib/modules/e1000.ko
+fi
+
+if [ -f /lib/modules/e1000e.ko ]; then
+	insmod /lib/modules/e1000e.ko
+fi
+
+if [ -e /sys/class/net/eth0 ]; then
+	# Setup static IP
```
paulmenzel: The verb is spelled with a space *Set up*.

initrd/bin/network-init-recovery
```diff
@@ -0,0 +1,29 @@
...
+# bring up the ethernet; maybe should do DHCP?
+ifconfig lo 127.0.0.1
+
+if [ -f /lib/modules/e1000.ko ]; then
+	insmod /lib/modules/e1000.ko
+fi
+
+if [ -f /lib/modules/e1000e.ko ]; then
+	insmod /lib/modules/e1000e.ko
+fi
```
paulmenzel: The modules are device specific, right? Should that depend on some configuration?

flammit: They are device specific.  I could use the variables in `module/linux` that control the inclusion of the kernel modules (`CONFIG_LINUX_E1000`, `CONFIG_LINUX_E1000E`, `CONFIG_LINUX_IGB`, `CONFIG_LINUX_SFC`, `CONFIG_LINUX_MLX4`) and export them in all the board configs, but that's equivalent to testing for existence in the initrd.  @osresearch what do you think?

osresearch:
I'm so excited to merge this so that we can move the user config and volatile things out of the initrd.

The cbfs reader patch should be sent as a patch to https://github.com/osresearch/flashtools so that we don't have the maintain it as a patch.  I hear the owner of that tree accepts PRs.

In QEMU when I last tried to read the ROM from `/dev/mem` it fails due to something not being mapped. `/proc/iomem` does not contain physmap-flash, which is likely why it doesn't show up. As a total hack, if I comment out this check in `read_mem()` in `linux-4.9.80/drivers/char/mem.c`, qemu is able to dump the contents of the ROM via `/dev/mem`:

```if (!valid_phys_addr_range(p, count))```

So it is something in the way that the memory map is being communicated to qemu that prevents us from being able to see the ROM.

We could patch the Heads kernel to allow this, although it would be good to know why it is not communicated.

**Comment by flammit** [comment]
Confirmed, in QEMU the `/dev/mem` reads fail in the current `cbfs` command, but using the busybox `devmem` tool and your `peek` work:

```
~# devmem 0xfffffffc
0xFF900138
~# devmem 0xff900138
0x4342524F
```

I'll fix `cbfs.c` up to work in QEMU by using your `map_physical` before pushing the patch to https://github.com/osresearch/flashtools.

**Comment by kakaroto** [comment]
I really like this solution! Great work!
flashtools doesn't work for me (on skylake) but I already reviewed the skylake port of flashrom so I should still know what needs to be done for it to work and I'll look into it soon.

File: patches/flashtools.patch [hunk 5]
```diff
@@ -0,0 +1,302 @@
...
++
++	if (header.magic != CBFS_HEADER_MAGIC) {
++		fprintf(stderr, "Failed to find valid header\n");
++		return EXIT_FAILURE;
++	}
++
++	uint32_t align = header.align;
++	// loop through files
++	uint64_t off = end - ((uint64_t) header.romsize) + ((uint64_t) header.offset);
++	while (1) {
```
kakaroto: Instead of an infinite loop (which breaks when it's outside the bounds of the cbfs), it would be better to do : 
`   while (off < end) { `

patches/flashtools.patch
```diff
@@ -0,0 +1,302 @@
...
++			if (verbose) {
++				printf("Seeking to %lx\n-------- Start Data\n", file_off);
++			}
++			if (lseek(fd, file_off, SEEK_SET) == -1) {
++				fprintf(stderr, "Failed to seek to file data at %lx\n", file_off);
++				return EXIT_FAILURE;
++			}
++
++			char *file_data = malloc(file.len);
++			if (read(fd, file_data, file.len) < 0) {
```
kakaroto: Verify file_data was allocated correctly.. A corrupted image could cause you to try and allocate GBs of data which would fail.

patches/flashtools.patch
```diff
@@ -0,0 +1,302 @@
...
++			printf("File attributes_offset : %x\n", file.attributes_offset);
++			printf("File offset            : %x\n", file.offset);
++		}
++
++		if (strncmp((char *)file.magic, CBFS_FILE_MAGIC, 8) != 0) {
++			break;
++		}
++
++		size_t name_size = file.offset - sizeof(struct cbfs_file);
++		char *name = calloc(name_size, sizeof(char));
```
kakaroto: name is allocated but never freed.

patches/flashtools.patch
```diff
@@ -0,0 +1,302 @@
...
++			if (write(STDOUT_FILENO, file_data, file.len) == -1) {
++				fprintf(stderr, "Failed to write file to stdout: %s\n",
++					strerror(errno));
++				return EXIT_FAILURE;
++			}
++			free(file_data);
++			if (verbose) {
++				printf("\n-------- End Data\n");
++			}
++			do_read++;
```
kakaroto: Maybe break once the file is read ?

patches/flashtools.patch
```diff
@@ -0,0 +1,302 @@
...
++	int fd;
++	fd = open("/dev/mem", O_RDONLY);
++	if (fd < 0) {
++		fprintf(stderr, "Failed to open /dev/mem : %s\n", strerror(errno));
++		return EXIT_FAILURE;
++	}
++
++	uint64_t end = 0x100000000;
++	if (verbose) {
++		printf("Seeking to %lx\n", end-4);
```
kakaroto: Since you print the file to stdout, all the verbose printfs should probably go to sdterr.

flammit:
@kakaroto thanks for the comments!  I'll incorporate them into the upcoming https://github.com/osresearch/flashtools PR

flammit:
The `cbfs` command PR is now up at https://github.com/osresearch/flashtools/pull/3.  Please review there and I'll adjust this PR once that is merged.

flammit:
@osresearch PTAL should be good to review and merge.  I've tested on `qemu-coreboot`.

flammit:
Minor change to require that cbfs file names be prefixed as `heads/initrd/*`.  The idea is that any CBFS file with the prefix `heads/` will attempt to be preserved between ROM updates.

flammit:
By default, both `flashrom-x230.sh` and `flashrom-kgpe-d16.sh` now use `preserve_rom` in `/etc/function` to inject previously loaded, non-conflicting `heads/*` files into new update ROMs prior to flashing.

modules/linuxboot
```diff
@@ -74,4 +79,5 @@ linuxboot.run: $(build)/$(BOARD)/linuxboot.rom
 		BOARD:=$(linuxboot_board) \
 		KERNEL=$(build)/$(BOARD)/bzImage \
 		INITRD=$(build)/$(BOARD)/initrd.cpio.xz \
+		CUSTOM=$(CUSTOMPWD) \
```
osresearch: What is `$CUSTOMPWD`?

PR #358, opened 2018-03-12
Create per-board configuration directory, allow init scripts to be installed from there

vejmarie:
- Refactore the board directory as to be able to support more than a config file per board. The idea is to put there also the initial init scripts per board which could be used before handing over the shell prompt to pre-initialize hardware. (aka device drivers tree etc ...) We could also put there specific kernel patches required per board (like ahci support for winterfell, or upcoming patches for pci bifurcation)
- Add an autoremove feature for u-root.cpio if it does exist before Makefile run otherwise it is not regenerated
- Add a UROOT_CMDS Makefile option as to be able to select which cmds are integrated within the u-root initramfs. Default is ALL.


modules/u-root
```diff
@@ -7,20 +7,32 @@
...
 
 export GOPATH=$(build)/go
+src_cmds := $(foreach cmd,$(UROOT_CMDS),github.com/u-root/u-root/cmds/$(cmd))
 
 $(GOPATH):
 	$(call do,GO GET,$(u-root_url),\
 		go get $(u-root_url) \
 	)
 
 $(u-root_output): $(GOPATH)
```
osresearch: The rm shouldn't be required -- I think the problem is that the dep is on the GOPATH directory, which will have been created the first time, rather than a dep on something intermediate.

osresearch: Should the ifeq tests be done in the shell rather than the Makefile? I think you're trying to allow the board to provide its own set of u-root commands; perhaps rather than specializing just this command it would make sense to have a boards/$(BOARD)/u-root subdirectory that is copied into the gopath directory.

(Also, this doesn't make it possible to do multiple boards that use different files without potentially forcing a clean rebuild)

vejmarie: I will have a look to the rm stuff tomorrow, you are right about it I believe. I was so focused at recompiling my u-root stuff each time I kicked it off, that I implemented it using brute force

vejmarie: the UROOT_CMDS is just a list of commands which will be integrated into the u-root ramfs, something like "pxeboot boot cat mount rsdp" reducing the initramfs to the smallest size and with only the required commands.

vejmarie: Just made an update with a "clean" rule definition into the u-root.config module

kakaroto:
This would be interesting for having per-board coreboot patches for example.

osresearch:
The only problem with per-board patches is that all boards build from the same source trees.

kakaroto:
Humm.. that's a good point. What we need to do is maybe have the canary keep track of which patches were applied. If the patch file changed or the list of patches changes (when using a series of patches), or the currently being built board is having a different set of patches, then it should recreate the source directory, patch it and rebuild it. That would work, right ? 
I'm not too makefile-fluent though to know how to achieve this at the moment.

PR #361, opened 2018-03-14
Impose Access Restriction to the Recovery Shell

### LINKED ISSUE #356: Impose Access Restriction to the Recovery Shell
Currently, HEADS will prompt a recovery shell under any fault conditions, it may not be a concern in a perspective of security. However, unlike other bootloaders, such as GRUB which only has read-only access to the disk with minimum risks of data destruction, HEADS is a fully-functional Linux-based operating system with full R/W access to the hard drive and hardware.

It enables extremely easy vandalism for script kiddies with physical access to the machine via the recovery shell, i.e. a keyboard-only denial-of-service attack. With keyboard-only access to an unguarded machine (such as a public machine in a lab), the script kiddie is able to...

* Removing any data from the hard drive.
* Changing the system time and invalidate TOTP.
* Resetting the TPM.
* Destroying the boot firmware. 

within a minute or two.

It may be a serious issue depending on one's threat model. For example, public computers are especially vulnerable. Major bootloaders, such an GRUB, allows programming a hashes password to the configuration script for access restrictions to the boot entry and the shell prompt. With access control, vandalism is only possible by physically disassembling the machine, which can be much harder.

In order to implement access control, two types of changes are needed. 

1. First, the recovery shell is provided unconditionally under a fault condition, for example, in `initrd/init`.

    ```
    # We should never reach here, but just in case...
    recovery 'Boot script failure?  Entering recovery shell'
    # belts and suspenders, just in case...
    tpm extend -ix 4 -ic recovery
    exec /bin/ash
    ```

    In an unexpected fault conditions, `exec /bin/ash` should be replaced with `exit` to trigger a kernel panic immediately.

2. For expected fault conditions and system maintenance, such as a TOTP mismatch, `initrd/bin/generic-init` should include an authentication scheme, for example a hashed password checkpoint, or a digital signature verification routine, to ensure the recovery shell access is obtained only under proper authentication.

persmule:
With this commit, to use recovery shell, users should authencate
themselves by connecting their OpenPGP card to the machine, in order
to prove they hold the very same secret key used to sign the boot file
list used by Heads. If they fail to pass the authencation, the machine
will exit from init and go kernel panic.

This digital signature verification routine is grafted from another
commit of mine.

fix #356


osresearch:
I like the idea of having this as an option, but not a requirement for dropping to the recovery shell.

We're starting to have enough `CONFIG_FOO` things that we should consider making a Kconfig menu...

persmule:
This time, the feature is controlled with an exported option
"CONFIG_RECOVERY_NEED_AUTH" in board files.

initrd/etc/functions
```diff
@@ -3,26 +3,68 @@
...
-	# Remove any temporary secret files that might be hanging around
-	# but recreate the directory so that new tools can use it.
-	rm -rf /tmp/secret
-	mkdir -p /tmp/secret
-	if [ "$CONFIG_TPM" = y ]; then
-		tpm extend -ix 4 -ic recovery
+	if [ "$CONFIG_RECOVERY_NEED_AUTH" != y ] || gpg_auth; then
+		# Remove any temporary secret files that might be hanging around
+		# but recreate the directory so that new tools can use it.
+		rm -rf /tmp/secret
```
flammit: the `rm -rf /tmp/secret` and `mkdir -p /tmp/secret` should both be done always before the if block, right?

persmule: It is just changed as you pointed out.

tlaurion:
Todo: review die usage and probably create error and replace lots of them. Also review recovery usage by calling error instead, which should prompt user to hit enter to continue.

tlaurion:
@persmule the problem here is if the public key has expired, the user would be locked out of the recovery shell.
Discussion: https://github.com/osresearch/heads/issues/881#issuecomment-917095442

tlaurion:
@persmule interesting enough, this is a plus for non fsp enabled boards. All sandy/ivy bridge can take advantage of this. Thanks a lot and looking forward to push that in with coreboot 4.15 version bump.

PR #363, opened 2018-03-14
Create nested menus and add option to rehash/sign /boot

kylerankin:
The number of options we want in the menu is starting to get large
enough that it's worth slimming things down in the main menu and move
options to nested menus. Along with this nested menu change is the
option to re-sign and re-hash files in /boot directly from the menu.

### DIFF
--- a/initrd/bin/gui-init
+++ b/initrd/bin/gui-init
@@ -48,12 +48,33 @@ while true; do
       'y' ' Default boot' \
       'r' ' TOTP does not match, refresh code' \
       'n' ' TOTP does not match after refresh, troubleshoot' \
+      'o' ' Other Boot Options -->' \
+      'a' ' Advanced Settings -->' \
+      'x' ' Exit to recovery shell' \
+      2>/tmp/whiptail || recovery "GUI menu failed"
+
+    totp_confirm=$(cat /tmp/whiptail)
+  fi
+
+  if [ "$totp_confirm" = "o" ]; then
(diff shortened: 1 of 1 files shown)


initrd/bin/gui-init
```diff
@@ -134,6 +155,33 @@ while true; do
...
+      DEFAULT_FILES=$(cat /boot/kexec_default_hashes.txt | cut -f3 -d ' ')
+      echo $DEFAULT_FILES | xargs sha256sum > /boot/kexec_default_hashes.txt
+
+      # sign and auto-roll config counter
+      extparam=
+      if [ "$CONFIG_TPM" = "y" ]; then
+        extparam=-u
+      fi
+      kexec-sign-config -p /boot $extparam \
+      || die "Failed to sign default config"
```
osresearch: die does not drop to a recovery shell.
a) should it by default?
b) should this script?
b) it should probably also try to remount /boot ro just in case

kylerankin: Perhaps we should put the remount and other safety features we'd like to see in the die function itself?

kylerankin:
I have merged a change that addresses the signing failure case a bit more gracefully. Instead of dying I let them know about the error so they can return to the main menu and drop to a recovery shell from there if they want to.

kylerankin:
This is independent of that change. My changes so far have just been using the standard whiptail package. Fbwhiptail is to be a drop-in replacement for it but so far all of my testing has used the standard newt-based whiptail.

PR #364, opened 2018-03-14
Allow insecure boot mode to bypass kexec sig checks

kylerankin:
There was a bug in the "force" boot mode where it would still fail if signatures didn't match. This was because the check_config function validates the signatures for kexec files. I've added a few conditionals
here so that in the case of a forced boot mode, we can bypass those signature checks that would prevent boot and error out to a recovery console.

### DIFF
--- a/initrd/bin/kexec-select-boot
+++ b/initrd/bin/kexec-select-boot
@@ -294,7 +294,11 @@ do_boot()
 }
 
 while true; do
-	check_config $paramsdir
+	if [ "$force_boot" = "y" ]; then
+	  check_config $paramsdir force
+  else
+	  check_config $paramsdir
+  fi
 	TMP_DEFAULT_FILE=`find /tmp/kexec/kexec_default.*.txt 2>/dev/null | head -1` || true
 	TMP_MENU_FILE="/tmp/kexec/kexec_menu.txt"
 	TMP_HASH_FILE="/tmp/kexec/kexec_hashes.txt"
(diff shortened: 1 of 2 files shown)


osresearch:
We should figure out how we want to merge this sort of functonality with #361, which is trying to prevent unauthorized access to the recovery shell.

kylerankin:
Won't the user still be able to hit 'r' during /bin/init to drop to a recovery shell?

persmule:
@kylerankin 
If the feature introduced with the newest #361 is built as enabled, even hitting 'r' during /bin/init is designed to need an authentication to access the recovery shell.

kylerankin:
@persmule with #361 in place, what will you use instead of 'die' when a signature check fails? I can just adopt the same method in my change to initrd/etc/functions

persmule:
@kylerankin If you mean a substitute when no `die()` from /etc/function available (e.g. in `/init`, before /etc/function is included), I use `exit 1`.

kylerankin:
@persmule I was referring more to @osresearch questions about how to merge this PR with #361. Would you prefer I exit 1 so as to have better compatibility with the restricted recovery shell?

persmule:
@kylerankin `die()` just terminates the shell instance which call it, so if you call `kexec-select-boot` from the recovery shell and it `die()`, you just return to the recovery shell.

So if `die()` is available, just use `die()` to terminate the shell instance which call it, otherwise, use `exit 1`.

kylerankin:
In that case I should leave that section as is in this PR. From my perspective this would be ready to merge and we should make sure the die() function does the "right thing" in the case of a restricted recovery shell.

PR #369, opened 2018-03-28
Add Cairo/FBWhiptail to the build process

kakaroto:
This makes the boot menu much nicer on the eyes, and at least with bigger fonts.
The GUI design is definitely a work in progress and fbwhiptail is not done yet, but most of the work remaining will be done in fbwhiptail rather than head and heads uses the git repo, so I think this is ready to be merged.


kakaroto:
It's not that big actually, once compressed, it adds about 600KB to the initrd while whiptail/newt adds 300KB. Currently the images/buttons/background/etc.. are programmatically generated by fbwhiptail but of course, once you add png files to customize/brand it, that will also increase the size, but for those who have small ROMs, they can fall back on using the newt whiptail instead.

Issue #249, opened 2017-09-20
NERF: flashrom doesn't work


osresearch:
The flashrom in the heads tree doesn't find the chipset (no dmi, etc). How do we do upgrades?

osresearch:
Recompiling with all the drivers results in:
```
~ # flashrom --programmer internal -r /tmp/test.rom
flashrom v0.9.9-r1955 on Linux 4.9.38-heads (x86_64)
flashrom is free software, get the source code at https://flashrom.org

Calibrating delay loop... OK.
No DMI table found.
Found chipset "Intel C610/X99 (Wellsburg)".
This chipset is marked as untested. If you are using an up-to-date version
of flashrom *and* were (not) able to successfully update your firmware with it,
then please email a report to flashrom@flashrom.org including a verbose (-V) log.
Thank you!
Enabling flash write... FREG2: Warning: Management Engine region (0x00020000-0x0030ffff) is locked.
Not all flash regions are freely accessible by flashrom. This is most likely
due to an active ME. Please see https://flashrom.org/ME for details.
Writes have been disabled for safety reasons. You can enforce write
support with the ich_spi_force programmer option, but you will most likely
harm your hardware! If you force flashrom you will get no support if
something breaks. On a few mainboards it is possible to enable write
access by setting a jumper (see its documentation or the board itself).
OK.
Found Winbond flash chip "W25Q128.V" (16384 kB, SPI) mapped at physical address 0x00000000ff000000.
Reading flash... Transaction error!
Read operation failed!
FAILED.
```

need to unlock the IFD. #259

osresearch:
Writes work, too.

PR #370, opened 2018-03-28
Update flashrom to 1.0

kakaroto:
This adds skylake support in flashrom.

### DIFF
--- a/modules/flashrom
+++ b/modules/flashrom
@@ -5,11 +5,11 @@ flashrom_depends := pciutils $(musl_dep)
 #flashrom_version := git
 #flashrom_repo := https://github.com/osresearch/flashrom
 
-flashrom_version := 0.9.9
+flashrom_version := 1.0
 flashrom_dir := flashrom-$(flashrom_version)
 flashrom_tar := flashrom-$(flashrom_version).tar.bz2
 flashrom_url := https://download.flashrom.org/releases/$(flashrom_tar)
-flashrom_hash := cb3156b0f63eb192024b76c0814135930297aac41f80761a5d293de769783c45
+flashrom_hash := 3702fa215ba5fb5af8e54c852d239899cfa1389194c1e51cb2a170c4dc9dee64
 
 flashrom_target := \
 	$(MAKE_JOBS) \


flammit:
Does the 0.9.9 patch still work?  I recall trying to do this upgrade and running into problems there.

kakaroto:
Oh, there's a pretty big flashrom patch for 0.9.9 that I forgot about. It adds a bunch of stuff, some ast1100 device support, 4 byte addressing mode for spi, progress reporting, thinkpad x220 whitelisted, and a bunch of other small features I suppose. Would need to go through it and see what needs to be ported to 1.0 (and why they weren't upstreamed)

For flashrom 1.0, it works, I even tested it from the recovery shell, it dumped the flash without issues (I didn't try to write, but reading was enough of a test to confirm it works).

osresearch:
The 0.9.9 patch was all about reducing the size of the binary (nearly 1 MB with all the drivers?) and ensuring that it was possible to flash even if IFD was not unlocked (ignore read failures).

Moving to `flashtool` where supported will help reduce the dependency on the patch.

Or we can try to convince the flashroom devs to make it more modular in the build and/or allow partial reflashes without trying to read the entire ROM.

kakaroto:
Ok. but most of the drivers are disabled in the make arguments, the patch that removed some chipsets was reverted in commit f7de7d73 due to issue #249.
The binary of 0.9.9 was 193KB, the one for 1.0 is 490KB. Once compressed with 'xz', the size becomes 71KB for 0.9.9 and 106KB for 1.0, so a 35KB difference in the resulting initrd.

Commit 0f299fe4 however is what brought the most changes to the patch. If these need to go back in, they need to be made into separate patches in a series.

osresearch:
It's not possible* to unlock the IFD from software, so flashing the unlocked IFD and cleaned ME requires physical access.

@kakaroto will the IFD be unlocked on the LIbrem machines?  The issue with flashrom reading the entire chip even for a partial reflash can cause problems if the ME region is locked.



____
* Without discovering a new vulnerability in the PCH or ME handling of the flash descriptor write permission checks

kakaroto:
On the librems, the IFD gives full read/write access to the entire flash. 
By the way, for the non-librems out there, can the '--ifd -i bios' option to flashrom be used to make it read/write only the BIOS region and ignore the ifd/me regions if they are locked. This way, you won't need a patch to flashrom to make it ignore read errors.

By the way, I see in the 0.9.9 patch that there was some code to add a progress bar/report of some sort.
I wrote a simple parser of flashrom -V output that transforms it into a progress bar (the function flashrom_progress) : https://code.puri.sm/kakaroto/coreboot-files/src/master/build_coreboot.sh#L246
Maybe it can be copied/reused in heads (note, the total rom size is hardcoded to 16MB in that script).

osresearch:
The `--ifd -i bios` must be new in 1.0.0.  It 0.9.9 it would still try to read the entire ROM even if `-i bios` were used, which was one of the reasons that the patch was required.

kakaroto:
I never tried it, the `-i bios` might still make it try to read the entire rom but only write the bios region, I don't know. Either way, it might be a good idea to patch it to not read the me region if `-i bios` is given as argument. The `--ifd` just means 'read the layout from the ifd' instead of reading the layout from a `--layout` argument

kakaroto:
Cool, thanks for merging. Did you test the '--ifd -i BIOS' option in 8108e419fe4789e5671cde8f6aa7a38a2be7a382  (in the code I see the region named as 'bios' when loaded from ifd, I don't see the comparison as being case insensitive) ?

PR #371, opened 2018-03-28
Librem13v2: Enable NVME support in kernel config

kakaroto:
This enables NVMe support in the kernel so we can at least go into recovery and mount it manually, but it doesn't yet add proper NVMe support in heads (i.e: not hardcoding the boot device so it can boot /dev/sda1 or /dev/nvme0n1p1 depending on which one is available, or providing a boot menu option for "Default boot from NVMe" and "Default boot from HDD" if both are available).


flammit:
@kakaroto we can definitely switch to scan for fixed boot devices as well as usb boot devices.  If you want to open an issue with the desired behavior, I can take a shot at it.

osresearch:
Should be a module, rather than linked in?

in `config/linux-linuxboot.config` for servers we've made every external device a module so that there are fewer external attack vectors until we've decided that the boot process wants to interact with that device.

kakaroto:
@osresearch The SATA one was linked in so I made NVMe linked in too. Also, I figured that even if it's a module, you would always need to load it to check if an NVMe drive is connected before you could boot. If it's always going to be loaded, then there's no need to have it as a module.
One possibility would be if we can detect that there is an nvme drive without loading the module then only load it if a drive is present... It might be doable with an lspci and parse a PCIe device class or something...

@flammit ok, thanks, I'll open the issue now. I figured that @kylerankin would handle it when he gets to that point but we'd appreciate the help! :)

osresearch:
My potential (and totally hypothetical) concern is that a device driver loaded pre-attestation could attempt to do something to subvert it. However, if the device driver is *always* loaded post-attestation, then that same potential vulnerability could be used to take over the boot process.

osresearch:
I'll go ahead and merge this and we'll revisit the "should we build in the device modules" question later.

osresearch:
https://github.com/osresearch/heads/commit/fe3e6dd5ec007e4fbb4895f87a726de63569ab12

(Not sure why github didn't detect the merge)

PR #377, opened 2018-04-01
X230 - working fbwhiptail support

tlaurion:
Changes required for the x230 to boot fbwhiptail.


osresearch:
Does this change the startup time significantly? I recall testing with the x230 i915 driver and seeing five seconds before anything was displayed.

config/linux-x230.config
```diff
@@ -1505,7 +1505,7 @@ CONFIG_DRM_FBDEV_EMULATION=y
 # ACP (Audio CoProcessor) Configuration
 #
 # CONFIG_DRM_NOUVEAU is not set
-# CONFIG_DRM_I915 is not set
+CONFIG_DRM_I915=y
```
paulmenzel: Why is a module not enough? Because no initrd is used?

paulmenzel:
Nice, but could you please elaborate on the commit *Reduce required changes to minimum to x230 Linux config*. It says *Reduce*, but the configuration is extended.

Additionally, it’d be great if you could statements as the commit message summaries, that means, start with a verb in imperative mood. But that’s only my preference.

tlaurion:
> Nice, but could you please elaborate on the commit Reduce required changes to minimum to x230 Linux config. It says Reduce, but the configuration is extended.

Those additional cofig parameters are added when DRM is activated. This is what I meant by "reduced Linux  changes to minimum" for fbwhiptail to function, with glitches I do not know how to fix. @kakaroto told there was a fb module in fbwhiptail that could work?

As previously said, this adds a total of 15 seconds and garbage on the screen. If there is no way of passing the fb from linux payload to kexec'd one, i'm not sure this is better then going with the console based whiptail option (based on NEWT and SLANG options).

tlaurion:
Failing attempts to date: https://github.com/tlaurion/heads/tree/x230_fbwhiptail

kakaroto:
@tlaurion  How does it fail exactly ? Do you get a /dev/dri/card0 device in heads or not? if not, do you get a /dev/fb? What errors does it print when you run fbwhiptail ?

tlaurion:
@kakaroto Difficult to say, cause there is no screen at all :) 

My last attempt of today was to add FB_SIMPLE into linux.

> git diff osresearch/master config/linux-x230.config
```
diff --git a/config/linux-x230.config b/config/linux-x230.config
index 9a31dc12b3..84c9f01c9d 100644
--- a/config/linux-x230.config
+++ b/config/linux-x230.config
@@ -1591,7 +1591,7 @@ CONFIG_FB_VESA=y
 # CONFIG_FB_MB862XX is not set
 # CONFIG_FB_BROADSHEET is not set
 # CONFIG_FB_AUO_K190X is not set
-# CONFIG_FB_SIMPLE is not set
+CONFIG_FB_SIMPLE=y
 # CONFIG_FB_SM712 is not set
 CONFIG_BACKLIGHT_LCD_SUPPORT=y
 # CONFIG_LCD_CLASS_DEVICE is not set
```

While using high-resolution frame-buffer support with native VGA init of coreboot.

> git diff osresearch/master config/coreboot-x230.config
```
diff --git a/config/coreboot-x230.config b/config/coreboot-x230.config
index 7edc29996d..1c4ac7d0fa 100644
--- a/config/coreboot-x230.config
+++ b/config/coreboot-x230.config
@@ -12,7 +12,7 @@ CONFIG_CBFS_PREFIX="fallback"
 CONFIG_COMPILER_GCC=y
 # CONFIG_COMPILER_LLVM_CLANG is not set
 # CONFIG_ANY_TOOLCHAIN is not set
-# CONFIG_CCACHE is not set
+CONFIG_CCACHE=y
 # CONFIG_FMD_GENPARSER is not set
 # CONFIG_UTIL_GENPARSER is not set
 # CONFIG_USE_OPTION_TABLE is not set
@@ -447,8 +447,9 @@ CONFIG_MAINBOARD_DO_NATIVE_VGA_INIT=y
 #
 # Display
 #
-CONFIG_VGA_TEXT_FRAMEBUFFER=y
-# CONFIG_GENERIC_LINEAR_FRAMEBUFFER is not set
+# CONFIG_VGA_TEXT_FRAMEBUFFER is not set
+CONFIG_GENERIC_LINEAR_FRAMEBUFFER=y
+CONFIG_LINEAR_FRAMEBUFFER=y
 # CONFIG_SMBUS_HAS_AUX_CHANNELS is not set
 CONFIG_PCI=y
 CONFIG_MMCONF_SUPPORT=y
```

Any suggestion into linux FB/DRI configuration options I should try? Or any troubleshooting path?

kakaroto:
@tlaurion what do you mean there's no screen at all? You should still have a configured screen. Just don't use gui-init as your start script and use the old generic-init script instead and in the recovery shell you should be able to see if /dev/fb exists or /dev/dri/card0. Also, if you run fbwhiptail manually in the recovery shell, you should see what errors it prints. 
If you compile fbwhiptail with -DDRI_DEBUG and if you uncomment the prints in https://github.com/kakaroto/fbwhiptail/blob/master/cairo_linuxfb.c then you should have all the debug info you need so that I could advise you on how to proceed.

tlaurion:
@kakaroto 
> @tlaurion what do you mean there's no screen at all?
coreboot 4.7 initializing VGA FB with libgfx or native high res, without linux having i915 driver doesn't bring any console up. There is only screen trash. 

On slack:
`compare the librem13v2 linux kernel config with the x230 and pay particular attention to differences around intel graphics and fb? also compare the x230 overall board config and librem13v2 for kernel command-line args`

There is some advancement, though, while still having some screen corruption when kexec'ing.

tlaurion:
@osresearch @flammit : any idea why this screen corruption takes places? Can you guys test it?

tlaurion:
This is working, no flicker at all.
Questioning the need to have FB support at all in coreboot, though.

tlaurion:
Combined with https://github.com/osresearch/heads/pull/404, I think this is ready to be UX tested in training environments.

@osresearch Would you guys be ready to reach out? 
Guys from https://usable.tools, accessnow, QubesOS and Freedom of Press foundation are ready for this!

kakaroto:
@tlaurion yes, coreboot doesn't need to do any kind of graphics init. On the librem with heads, there was no vgabios and no vbt, and coreboot graphics initialization was set to "None" (I eventually added the VBT back into coreboot only because without it, the OS wasn't able to control the screen brightness).
The i915 driver is able to initialize the graphics entirely on its own, but with IOMMU support in coreboot, it does cause issues when the OS boots, so we had to add the iommu=pt to the kernel boot args (or intel_iommu=igfx_off). 
That issue doesn't happen when using the original AMI firmware, so we think it has something to do with the RMRR entries in the DMAR ACPI table. We have someone looking into this and trying to get it fixed so we can drop the iommu=pt argument.

tlaurion:
@osresearch @flammit : This is ready to merge.

tlaurion:
@osresearch 

> Also need to check the size of the resulting kernel against the x230 flasher image -- it shares the same config/linux-x230.config file.

This is why boards/x230-flash/x230-flash.config depends on config/linux-x230-flash.config so that i915 is not included since unnecessary.

In an ideal world, x230-flash would boot gui-init without TPMTOTP and other silly dependencies. The user could then add his keys through GUI in one single step, prior to flashing downloaded reproducible built rom, from a trusted environment.

tlaurion:
In an ideal world, x230-flash would boot gui-init without TPMTOTP and other silly dependencies. The user could then add his keys through GUI in one single step, prior to flashing downloaded reproducible built rom, from a trusted environment.

tlaurion:
Smaller change set under #465. Closing.

PR #386, opened 2018-04-20
modules: update linux to 4.9.95

merge:
I didn't test it! Just built with the various board configs.


osresearch:
Thanks for the patch, although we'll stick with 4.9.80 and skip ahead prior to the next tagged Heads release.

paulmenzel:
Wouldn’t it be good for testing to get intermediate release in?

PR #387, opened 2018-04-20
config: remove x200 linux config

merge:
It's not used anymore, I think. I might be wrong. Please review. thanks!


osresearch:
Merged, thanks!

tlaurion:
@osresearch @merge 
If I remember well, work have been done to have a lightweight linux config and board config that deactivates TPM support. That may need to be maintained a bit, but I do not think it should be deleted.

The goal of such config would be to have heads able to boot signed tail isos directly from a usb keys, Tails keys being bundled in Heads recently.

@merge: what is inteded by "not used anymore?"

merge:
Thanks for the feedback. "Not used" means that there's no board / make target that uses said config. In that case this was missing and thus not readable. Could you add a "tails" target and re-add the config if that's appropriate?

tlaurion:
@merge Kinda lost track of it.
But the goal of @persmule changes were to permit a NOTPM mode. There is no board configuration in Heads repository for it, but inclusion is welcome: 
https://github.com/hardenedlinux/Debian-GNU-Linux-Profiles/blob/master/scripts/heads/autoboot-notpm.config#L14

herbsmn:
Is there any chance the x200 linux config could be added back?

tlaurion:
@herbsmn absolutely. The problem here is that the device has no TPM, so https://github.com/osresearch/heads/pull/836 would be required for additional security, where inclusion of the HOTP toolchain + GPG2 will probably fit really tight.

Also note that not all x200 are equal. If I remember well, only the labeled as "PRO"  (AMT PRO on intel label?) on the tests i've done in the past had a 8mb(?) SPI chip.... Would need to find traces of that in the backlog...

herbsmn:
My larger hope is that the Montevina / G45 / Centrino 2 devices that can be run without Intel ME can be supported by heads. I noticed that @persmule was running heads on this x301 based on the text at the bottom of this page: https://doc.coreboot.org/mainboard/lenovo/x301.html

When I asked persmule about this they said that the heads x200 board configs in question here were used to build the x301 Rom.

In their original heads patch set that added the x200 configs it says they also tested the configs on an x200s device: https://github.com/osresearch/heads/commit/baa30a20266086d7886fcb92b521b7e220abb774

Regarding which x200 models and, more broadly, Montevina models have 4MiB vs 8MiB SPI chips, you can maybe check the ThinkPad RAW Image Versioning's Device IDs using this handy table here: https://notabug.org/libreboot/libreboot/issues/731

You are correct that the vPro variants have 8MiB SPI chips while all others have 4MiB SPI chips.

tlaurion:
@herbsmn have not tested but see and comment https://github.com/osresearch/heads/issues/878

PR #391, opened 2018-04-25
Colorize warning and error messages in fbwhiptail

kylerankin:
Since fbwhiptail allows us to customize the background colors, we should
colorize warnings and error messages to provide a user with an
additional subtle cue that there might be a problem. I have added two
additional configuration options:

CONFIG_WARNING_BG_COLOR
CONFIG_ERROR_BG_COLOR

and in the librem13v2.config file you can see an example for how to set
them to be yellow and red gradients, respectively. I've also updated the
main two scripts that use whiptail to include those background colors.

If you decide to use regular whiptail, just don't set these config
options and it should behave as expected.


rminnich:
colorizing is problematical however. It makes it harder for colorblind folks and it's hugely painful to grep because you inject all this escape goop into a simple character stream, which also makes building continuous integration and test systems a headache. I'm not a fan but ...

kylerankin:
@rminnich This isn't colorizing bash with escape codes--this is specifically changing the background color in the fbwhiptail framebuffer-based port of the whiptail program. If you take a look at the changes to the librem13v2.config file you can see an example of the command-line options it adds to fbwhiptail.

Currently the background color for this framebuffer program is a grey gradient, and the error dialogs list "ERROR" in the title. This is just to provide an additional cue, not the only cue. Presumably anyone who is colorblind would see a regular gray gradient like is currently in place.

rminnich:
sounds good.

PR #406, opened 2018-05-10
Add Flashrom script for Librem laptops

kylerankin:
This change adds a flashrom-librem.sh script based off of
flashrom-x230.sh and usb-scan to flash Librem laptops using their
particular flashrom options. I've also added a menu option to gui-init
and so it's compatibile with multiple boards I've added the
CONFIG_FLASHROM_SCRIPT environment variable that points to a particular
flashrom script to use.


flammit:
Yeah it's a bit hokey the way these files are now.  What about having a generic `flash.sh` and `flash-bmc.sh` that reads an exported CONFIG_BOARD parameter in `/etc/config` to alter its behavior so that the installation procedures on multiple boards are the same?  We can put things like board specific PCIEXBAR for flashtool and custom flashrom parameters but also maintain a single set of logic for handling persistent config injection across all boards (cbfs or uefi).

kylerankin:
That works for me. I just didn't want to forge ahead with any new approach to doing this by myself. Since there's some consensus that it's a good time to integrate it, let me close this PR and I'll reimplement this based on this feedback and submit a new PR.

PR #411, opened 2018-05-11
Add generic flash script

kylerankin:
Based on the conversation for PR #406, we decided to go with a more
generic script for general-purpose flashing instead of having individual
(and therefore very similar) flash scripts for each board type. This
script currently handles flashrom on Librem and X230 board types and
introduces a new CONFIG_BOARD option that sets specific flashrom
arguments based on the board.

It also adds support to gui-init to call this flash script.


flammit:
Cool - any chance we can split up the new UI workflow from the actual flash command? Basically just leave the `flash_rom()` function and the `FLASHROM_OPTION` in `flash.sh` and put the UI stuff into another `flash-gui.sh` or something?  That way we can centralize the flash logic even when the whiptail stuff isn't being used.

Also would you mind adding the KGPE-D16 board option as well and drop the `flashrom-x230.sh` and `flashrom-kgpe-d16.sh` (or I can do that in a later commit)?

boards/librem13v2/librem13v2.config
```diff
@@ -34,3 +34,4 @@ export CONFIG_BOOT_GUI_MENU_NAME="Purism Librem 13v2 Heads Boot Menu"
 export CONFIG_USB_BOOT_DEV="/dev/sdb1"
 export CONFIG_WARNING_BG_COLOR="--background-gradient 0 0 0 150 125 0"
 export CONFIG_ERROR_BG_COLOR="--background-gradient 0 0 0 150 0 0"
+export CONFIG_BOARD="librem"
```
flammit: Maybe the `CONFIG_BOARD` variable should be injected into `/etc/config` the same way `GIT_HASH` 
 and `GIT_STATUS` are in the main Makefile?

kylerankin: On one hand that would remove an area of redundancy (such as if it reused the BOARD environment variable). In the Librem case it would mean that we would need to define each board revision within the flash.sh script instead of having a more generic label as we have boards that require different coreboot configuration but would be able to use the same flashrom options. I suppose your suggestion would make it more consistent with how we have the board configurations organized thus far. I'm open to either approach.

flammit: I'd defer to @osresearch. I just figured it would be one less thing to worry about in board-specific config files having `BOARD` be injected automatically.  And to handle multiple versions with single logic, you could just change to use `librem*)` in the case statement in `flash.sh` (for as long as that holds true for all revisions of your boards).

kylerankin: The more I think about it, the more I like the idea of reusing a configuration option we already have in place. It makes things more consistent and it ties directly into the idea of options for a particular board. I will take that approach.

kylerankin: Makefile change incorporated.

kylerankin:
flash.sh and flash-gui.sh split, and other boards accounted for.

initrd/bin/flash.sh
```diff
@@ -0,0 +1,53 @@
...
+#
+set -e -o pipefail
+. /etc/functions
+. /etc/config
+
+case "$CONFIG_BOARD" in
+  librem* )
+    FLASHROM_OPTIONS='-p internal:laptop=force_I_want_a_brick,ich_spi_mode=hwseq' 
+  ;;
+  "x230" )
```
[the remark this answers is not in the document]
kylerankin: I somehow missed this comment, just pushed this change.

Makefile
```diff
@@ -389,6 +389,7 @@ bin_modules-$(CONFIG_FLASHTOOLS) += flashtools
 bin_modules-$(CONFIG_NEWT) += newt
 bin_modules-$(CONFIG_CAIRO) += cairo
 bin_modules-$(CONFIG_FBWHIPTAIL) += fbwhiptail
+bin_modules-$(CONFIG_NITROKEY) += nitrokey-hotp-verification
```
osresearch: Is there a separate PR coming for nitrokey support?

(This is a NOP since `CONFIG_NITROKEY` will not be yes, so it's ok for now)

kylerankin: Yes, that was unfortunately copied over by mistake. The config options et al will be a bit different with the PR related to Librem Key support.

PR #417, opened 2018-05-23
Add VBT file to Librem coreboot image.

kakaroto:
Adding the VBT file makes it available through some ACPI memory area
and apparently the VBT contains the information needed by the i915 driver
in order to figure out how to control the screen's backlight.
Without the VBT, we can't control the screen backlight with Fn-F5/Fn-F6
anymore.


kylerankin:
I've tested this on my test Librem 13v2 and can confirm that I can control the backlight now.

PR #420, opened 2018-05-24
Update to coreboot 4.8.1

### LINKED ISSUE #419: Update coreboot from 4.7 to 4.8.1
coreboot 4.8.1 was released and should be integrated. I could work on that.

flammit:
Closes #419


patches/coreboot-4.8.1/0000-measuredboot.patch
```diff
@@ -0,0 +1,474 @@
...
++uint32_t tlcl_measure(int pcr_num, const void * start, size_t len)
++{
++	VBDEBUG("TPM: pcr %d measure %p @ %zu: ", pcr_num, start, len);
++
++	struct sha1_ctx sha;
++	sha1_init(&sha);
++	sha1_update(&sha, start, len);
++
++	const uint8_t * hash = sha1_final(&sha);
++	for(unsigned i = 0 ; i < SHA1_DIGEST_SIZE ; i++)
```
kakaroto: I would remove the dumping of the hash from here, it is just making it easy for someone to see how the PCRs were extending by simply doing a 'cbmem'. So let's not put that in any 'production code' and comment out this loop.

patches/coreboot-4.8.1/0009-Add-heads-TPM-measurements-to-Skylake-Kabylake.patch
```diff
@@ -0,0 +1,181 @@
...
+ 
+ u8 acpi_checksum(u8 *table, u32 length)
+ {
+@@ -1020,6 +1021,10 @@ unsigned long write_acpi_tables(unsigned long start)
+ 		return current;
+ 	}
+ 
++	if (IS_ENABLED(CONFIG_MEASURED_BOOT)) {
++	  tlcl_measure(2, (const void*) dsdt_file, dsdt_size);
++	}
```
kakaroto: Why do this? the dsdt_file was just loaded using `cbfs_boot_map_with_leak()` which itself calls the prog_segment_loaded so it's already been measured.

patches/coreboot-4.8.1/0009-Add-heads-TPM-measurements-to-Skylake-Kabylake.patch
```diff
@@ -0,0 +1,181 @@
...
+ 
+ 	/*
+ 	 * Initialize the TPM, unless the TPM was already initialized
+-	 * in verstage and used to verify romstage.
++	 * in verstage and used to verify romstage, or for measured boot.
+ 	 */
+ 	if (IS_ENABLED(CONFIG_LPC_TPM) &&
+-	    !IS_ENABLED(CONFIG_VBOOT_STARTS_IN_BOOTBLOCK))
++	    !IS_ENABLED(CONFIG_VBOOT_STARTS_IN_BOOTBLOCK) &&
++			!IS_ENABLED(CONFIG_MEASURED_BOOT))
```
kakaroto: Indentation issue here.

patches/coreboot-4.8.1/0009-Add-heads-TPM-measurements-to-Skylake-Kabylake.patch
```diff
@@ -0,0 +1,181 @@
...
+ 		init_tpm(s3wake);
+ }
+ 
+@@ -483,8 +484,33 @@ void fsp_memory_init(bool s3wake)
+ 	if (status != CB_SUCCESS)
+ 		die("Loading FSPM failed!\n");
+ 
++	if (IS_ENABLED(CONFIG_MEASURED_BOOT) && IS_ENABLED(CONFIG_LPC_TPM)) {
```
kakaroto: Do we need to check for CONFIG_LPC_TPM ? The Kconfig option for CONFIG_MEASURED_BOOT already depends on MAINBOARD_HAS_LPC_TPM and it does a 'select TPM' so I assume that if MEASURED_BOOT is enabled then LPC_TPM will also always be enabled.

patches/coreboot-4.8.1/0030-sandybridge.patch
```diff
@@ -0,0 +1,58 @@
...
++#include <security/tpm/tss.h>
++#include <program_loading.h>
+ #include <northbridge/intel/sandybridge/chip.h>
+ #include "southbridge/intel/bd82x6x/pch.h"
+ #include <southbridge/intel/common/gpio.h>
+@@ -72,6 +74,19 @@ void mainboard_romstage_entry(unsigned long bist)
+ 	/* Initialize superio */
+ 	mainboard_config_superio();
+ 
++	if (IS_ENABLED(CONFIG_MEASURED_BOOT) && IS_ENABLED(CONFIG_LPC_TPM)) {
```
kakaroto: Do we need to check for CONFIG_LPC_TPM ? The Kconfig option for CONFIG_MEASURED_BOOT already depends on MAINBOARD_HAS_LPC_TPM and it does a 'select TPM' so I assume that if MEASURED_BOOT is enabled then LPC_TPM will also always be enabled.

patches/coreboot-4.8.1/0009-Add-heads-TPM-measurements-to-Skylake-Kabylake.patch
```diff
@@ -0,0 +1,181 @@
...
++		const void * const bootblock = (const void*) 0xFFFFF800;
++		const unsigned bootblock_size = 0x800;
++		tlcl_measure(0, bootblock, bootblock_size);
++
++		tlcl_measure(1, _romstage, _eromstage - _romstage);
++	}
++
++	if (IS_ENABLED(CONFIG_MEASURED_BOOT)) {
++	  tlcl_measure(1, (const void*) hdr.image_base, hdr.image_size);
++	}
```
kakaroto: Why are you doing this here ? First, it might as well go in the if() just above it if there is a justifitication of doing it this way.. but also, I don't see why it's measured separately.. because you're measuring the hdr.image_base into PCR 1 right here, and then on the next line, you add the SEG_NO_MEASURE so you don't measure hdr.image_base (but the platform_segment_loaded function defined just below is measuring into PCR 1, so it's not because you want it in PCR 1 instead of whatever the prog_segment_loaded measured it into). So what is the point of doing all that ?
I suggest you remove this and remove the SEG_NO_MEASURE from the next line.

patches/coreboot-4.8.1/0009-Add-heads-TPM-measurements-to-Skylake-Kabylake.patch
```diff
@@ -0,0 +1,181 @@
...
++#include <security/tpm/tss.h>
+ #include <stage_cache.h>
+ #include <string.h>
+ #include <timestamp.h>
+@@ -101,6 +102,10 @@ void fsps_load(bool s3wake)
+ 	if (rdev_readat(&rdev, dest, 0, size) < 0)
+ 		die("Failed to read FSPS!\n");
+ 
++	if (IS_ENABLED(CONFIG_MEASURED_BOOT)) {
++	  tlcl_measure(1, (const void*) dest, size);
```
kakaroto: Same thing here, why do you manually measure and then add the SEG_NO_MEASURE flag to prog_segment_loaded ?

patches/coreboot-4.8.1/0009-Add-heads-TPM-measurements-to-Skylake-Kabylake.patch
```diff
@@ -0,0 +1,181 @@
...
+ #include "intel_bios.h"
+ #include "opregion.h"
+ 
+@@ -57,6 +58,10 @@ void *locate_vbt(size_t *vbt_size)
+ 	if (vbt_size)
+ 		*vbt_size = file_size;
+ 
++	if (IS_ENABLED(CONFIG_MEASURED_BOOT)) {
++	  tlcl_measure(2, (const void*) vbt_data, file_size);
++	}
```
kakaroto: There is no need to do this here either because the vbt is loaded using cbfs_boot_load_file which itself calls cbfs_load_and_decompress which itself calls (via the measurements patch) prog_segment_loaded, which means you're causing this to be measured twice.

patches/coreboot-4.8.1/0009-Add-heads-TPM-measurements-to-Skylake-Kabylake.patch
flammit: Yup - my mistake. I willre move.

patches/coreboot-4.8.1/0009-Add-heads-TPM-measurements-to-Skylake-Kabylake.patch
flammit: fixed

patches/coreboot-4.8.1/0009-Add-heads-TPM-measurements-to-Skylake-Kabylake.patch
flammit: it's to mirror the normal call to `init_tpm`.  seems safer in case we change `CONFIG_MEASURED_BOOT` to allow for a different TPM where there is another initialization path

patches/coreboot-4.8.1/0009-Add-heads-TPM-measurements-to-Skylake-Kabylake.patch
flammit: Sorry - this was leftover from when I thought segments_loading here were for ramstage (I forgot about postcar on this platform) so I wanted it to specifically be in PCR1 (instead of PCR2). Reverting

patches/coreboot-4.8.1/0009-Add-heads-TPM-measurements-to-Skylake-Kabylake.patch
flammit: Ah - I'm wanted to measure the output before any potential changes or relocation happens (similar thing to what happens in rmodule).  I didn't want to assume much about what `fsp_component_relocate` does now or in the future.

patches/coreboot-4.8.1/0009-Add-heads-TPM-measurements-to-Skylake-Kabylake.patch
flammit: Yup - my mistake. Removing

patches/coreboot-4.8.1/0000-measuredboot.patch
```diff
@@ -0,0 +1,474 @@
...
+ }
+diff --git ./src/security/tpm/Makefile.inc ./src/security/tpm/Makefile.inc
+index 2385635..52d088c 100644
+--- ./src/security/tpm/Makefile.inc
++++ ./src/security/tpm/Makefile.inc
+@@ -4,8 +4,12 @@ verstage-$(CONFIG_TPM) += tss/tcg-1.2/tss.c
+ verstage-$(CONFIG_TPM2) += tss/tcg-2.0/tss_marshaling.c
+ verstage-$(CONFIG_TPM2) += tss/tcg-2.0/tss.c
+ 
+-ifeq ($(CONFIG_VBOOT_SEPARATE_VERSTAGE),y)
```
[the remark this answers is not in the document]
flammit: I just assumed that all Heads should have MEASURED_BOOT enabled.

patches/coreboot-4.8.1/0000-measuredboot.patch
flammit: @osresearch what do you think? I don't see any harm in enabling debug since the contents of what goes into the PCR is not sensitive.

patches/coreboot-4.8.1/0030-sandybridge.patch
flammit: Same comment as on the fsp2.0 side, seems safer in the future with the added guard.

patches/coreboot-4.8.1/0000-measuredboot.patch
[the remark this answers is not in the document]
kakaroto: Yeah, but I'm assuming these patches will need to be sent upstream at some point to avoid porting them everytime there's an update. In that case, this is a comment you'll get when pushing to gerrit.

patches/coreboot-4.8.1/0009-Add-heads-TPM-measurements-to-Skylake-Kabylake.patch
kakaroto: Ok ,sounds good to me to keep it in that case.

flammit:
@kakaroto updated the makefiles.  PTAL/approve if it looks good to you.

flammit:
@kakaroto ping - could you test librem and approve when you get a chance?

kakaroto:
@flammit sorry, I've been busy. Will test it next week!

kakaroto:
Confirmed working, I needed gcc8 patches to both linux and coreboot, I also added new patches to coreboot (fix for iommu, and ME cleaning capabilities).
I'll clean my commits up and send a proper PR next week.

osresearch:
I'm building and installing this on my x230 today for testing. How does it look for everyone else?

osresearch:
As part of #439 I'm converting the coreboot config files to "defconfig" format, which removes anything that is a default for that platform.  This should reduce the complexity of this sort of commit when we want to upgrade coreboot -- this commit removed 4.6k lines of config file entires: https://github.com/osresearch/heads/commit/037bbfebdffb4e2c1975bd0c5c3e32aed6f91801

If you want to streamline this merge, cherry pick this commit: https://github.com/osresearch/heads/commit/2bcbd0bfe0788d1447a5a66bca39bb4bf4b059fe
and run `make BOARD=x230 && make BOARD=x230 coreboot.saveconfig` for all of the platforms to convert the `config/coreboot-$(BOARD).config` files.

rexredux:
Hi, what is the current state of this pullrequest? What is keeping it from being merged?

kakaroto:
I think 2 things were keeping it from being merged : 
1 - the config files are in long format, but @osresearch said that heads switched to short format
2 - I confirmed it working but only after patching it to move to gcc8 because the crossgcc toolchain doesn't build for me with gcc8 installed locally. I unfortunately never had time to review/push my own changes for that.

flammit:
Okay - sorry I didn't realize the config format was a blocker to merge.  I'll do that then test build on a Fedora 28 machine w/ gcc8.

kakaroto:
By the way, here are my gcc8 patches if you want to look at them, review/test and integrate into your PR: https://github.com/kakaroto/heads/commits/gcc8
(note: this is my last week at Purism, and I may not get the time to actually review/clean up/send PR/answer questions, for those commits)

PR #424, opened 2018-06-04
x230: remove 4M and 8M split-images from the build

### LINKED ISSUE #302: Documentation on how to install on X230 needs to be updated
This documentation is fairly outdated:
http://osresearch.net/Installing-Heads

I followed the steps, flashed the Rom... and then... the instructions are pretty unclear as to what to do next. 

I followed some older instructions to get the TPM set up, that didn't work. (here: https://github.com/mfc/flashing-docs/blob/master/walkthrough%20for%20flashing%20heads%20on%20an%20x230.md)

I tried to run start-xen, that doesn't exist

There's instructions on how to get a patched Xen from USB, but not sure that is still needed (as per one closed issue)
There's instructions about using a Yubikey to set up a pgp key; but yubikeys are considered unsafe now so I'd like to use my own key. Also kind of confused as the key is needed in "initrd.cpio" file, but the yubikey generated result is then in initrd/.gnupg - not initrd.cpio. 

I did tpm-reset, which seemed to work.
I did seal-totp, which worked, but unseal failed with an error message; maybe because of the missing gpg key? 

I'd like to be able to run my existing Qubes install, or recover it somehow. Prior to flashing the ROM I had coreboot installed and was running Qubes 4 rc3. 

I'd be happy to document the process, but currently nothing is really working. It's late so I will try to recover my qubes install tomorrow.

Configuration
Hardware Thinkpad X230
ROM: coreboot, intel ME neutered with intel ME cleaner.
Software: Qubes 4 rc3
I cloned the github repo and did a build from that (latest as per yesterday). Build worked fine, I built x230.flash.rom and also the "make xen.intermediary" xen version.

### LINKED ISSUE #307: Provide IFD, Gbe and ME blobs
The 12 MB x230 ROM image isn't flashable since it doesn't include the IFD and ME sections.  This isn't normally a large problem for updates since `flashrom-x230` won't touch the ME region, but it does lead to problems for initial install.

----
tlaurion hijacking top post here to give actual status in a glimpse.

- @osresearch accepted that heads distribute extracted IFD from an original Lenovo's rom, lower down in this issue [here](https://github.com/osresearch/heads/issues/307#issuecomment-552961256). The goal of this is to not generate the IFD ourselves and just unlock the one being present in the repository, on which BIOS region has been extended resulting of freed ME region, where the ME region is reduced to [the minimal size possible](https://github.com/osresearch/heads/pull/703/files#diff-0493cf37597ff5795588f8343ab268b7R3-R4). This results in a BIOS region of 11.5mb instead of 7mb, mitigating #590 issue until help joins forces for boards with limited SPI flash space without the need of soldering...
- [3 collaborators decided to support the adventure](https://github.com/osresearch/heads/issues/307#issuecomment-613514816) of [not waiting for legal approval](https://github.com/osresearch/heads/issues/307#issuecomment-615403721), nor expecting [Intel reply](https://github.com/osresearch/heads/issues/307#issuecomment-583887020), and going forward and include the ME cleaned binary.  Some efforts are still needed so the blob is downloaded directly from Lenovo, on which innoextract needs to be applied to remove the blob itself from local blob directory and do the required magic in a script having downloaded the blob from Lenovo website directly. Meanwhile the cleaned blob is present in tree with manual reproducibility instructions in xx30's blobs's directory README. [Read the xx30 blobs README here for more details](https://github.com/osresearch/heads/pull/703/files#diff-0493cf37597ff5795588f8343ab268b7). Please show interest for being paid fixing this.
- [Along the adventure](https://github.com/osresearch/heads/issues/307#issuecomment-626882321), it was discovered and [documented upstream](https://github.com/osresearch/heads/issues/700#issuecomment-627208266) that the GBE region needs to be present to have a functional ethernet card. This GBE blob [could be totally generated](https://github.com/osresearch/heads/issues/307#issuecomment-646897081). But would need some low level people involved. I'm not good at [reading reference manual and implementing what is required](https://github.com/osresearch/heads/issues/700#issuecomment-647003701). Please show interest for being paid fixing this.

Mitigations from dropping those binaries would be to:
- [ ] [Claim right to repair laws](https://gizmodo.com/right-to-repair-is-less-complicated-and-more-important-1834672055)? 
- [ ] Claim [Fair use laws](https://djtechtools.com/2017/06/05/basics-dj-copyright-laws/)?
- [ ] Generate those blobs from source code (see below)

So the result of the ongoing #703 PR is a PoC of that, which requires Heads to fake MAC address on boot to not have a MAC collision on the same LAN network segment if many Heads xx30 empowered machines are present.

Todo:
- [ ] [Generate GBE blob ](https://github.com/osresearch/heads/issues/700#issuecomment-647003701) to remove the GBE blob from tree. Seperate issue: #796 
- [ ] [Extract ME from Lenovo website (innoextract) and run me_cleaner on top of it](https://github.com/osresearch/heads/pull/703/files#diff-0493cf37597ff5795588f8343ab268b7) to put resulting blob locally to remove it from tree. Seperate issue: #797

merge:
The bios regions of the 12M coreboot image is 7M: 4M and 3 of the 8M split
image. The rest of the 8M image _generated_ with fake data and not usable
on real systems! It's dangerous to create them and suggest flashing them
externally.

That's exactly why the x230-flash build target is there: To
have a self-contained 4M image and enable easy unlocking of the 8M image
using the _original_ data.

the heads-wiki project is updated accordingly.

Closes #307
Closes #302

### DIFF
--- a/boards/x230/x230.config
+++ b/boards/x230/x230.config
@@ -33,14 +33,6 @@ export CONFIG_USB_BOOT_DEV="/dev/sdb1"
 # the ME image and part of the coreboot image, and a 4 MB one that
 # has the rest of the coreboot and the reset vector.
 #
-# When flashing via an external programmer it is easiest to have
-# to separate files for these pieces.
-all: $(build)/$(BOARD)/$(BOARD)-8.rom
-$(build)/$(BOARD)/$(BOARD)-8.rom: $(build)/$(BOARD)/coreboot.rom
-	$(call do,DD 8MB,$@,dd of=$@ if=$< bs=65536 count=128 skip=0 status=none)
-	@sha256sum $@
-
-all: $(build)/$(BOARD)/$(BOARD)-4.rom
-$(build)/$(BOARD)/$(BOARD)-4.rom: $(build)/$(BOARD)/coreboot.rom
-	$(call do,DD 4MB,$@,dd of=$@ if=$< bs=65536 count=64 skip=128 status=none)
-	@sha256sum $@
+# Only flashing to the bios region is safe to do. The easiest is to
+# flash internally when the IFD is unlocked for writing, and x230-flash
+# is installed first.


osresearch:
This needs to be fixed as part of #307 - the problem with the `x230-flash` image is that it doesn't unlock the IFD or clean the ME, leading to issues with reflashing the entire ROM with software.

merge:
Right. That's why I did https://github.com/osresearch/heads-wiki/pull/17/commits/1e8eec5ac155014945bb1178ff724ea284349ed1 in order to resolve this issue. It's a documentation thing and all not that complicated: Use a 4M coreboot build for hardware flashing and unlock+flash back _your_ 8M. From that point on you're good to go. (actually because of this confusion I created the "[Skulls](https://github.com/merge/skulls/tree/master/x230)" project). Would such Skulls-style helper scripts fit anywhere inside of Heads too?

tlaurion:
@merge : What is the beneficial gain in merging skull to heads, as opposed to https://github.com/osresearch/heads/pull/377#issuecomment-408463520

merge:
We need a plan here :) we don't even know _what_ to do yet

* First decision to make: we should agree that generating the 4M and 8M images of the `x230` board is wrong in any case. We'll never be able to safely generate a flashable thing of 12M. And there's no point in having the splits other than flashing externally, which is always wrong.
Also, flashing externally is only done the first time. And that's what the `x230-flash` board is for, which makes the x230 4M/8M split even _more_ misleading. But then, `x230-flash` is not sufficient. At least touching the other chip's IFD is necessary.

* So, second decision to make: Do we want to support "bootstrapping" explicitely inside of Heads?
  * If so, let's add scripts that unlock the IFD on the one chip and flash `x230-flash` on the other, and guide users through _all_ of it.
  * If not, drop `x230-flash` completely and simply add "unlocked IFD" to the requirements and add a script or (even better?) live-USB-image that flashes `x230` from outside of Heads, which makes "bootstrapping" quite neat.

thoughts?

tlaurion:
@merge I completely agree with your last comment. Checked what you are doing with skulls, and it seems completely right. 

So basically, ifd is what is missing into x230-flash if I get it right?

merge:
@tlaurion ifdtool is what's missing, yes. but not "into x230-flash", but applied to the 8M chip, just like documented in the Skulls project. (read, run "unlock", write).

PR #425, opened 2018-06-19
blobs/librem_skl: Update URLs and add ME rar check

kakaroto:
Purism coreboot-files repo moves from code.puri.sm to source.puri.sm
and the ME rar file moved from a mediafire to a mega.nz file which
doesn't allow us to get a direct link to the file, so we need to
tell the user about it and ask them to download the file manually.

### DIFF
--- a/blobs/librem_skl/get_blobs.sh
+++ b/blobs/librem_skl/get_blobs.sh
@@ -20,16 +20,16 @@ SKL_FSP_SPLIT_SHA="f654f6363de68ad78b1baf8b8e573b53715c3bc76f7f3c23562641e49a703
 ME_CLEANER_URL="https://github.com/corna/me_cleaner/raw/9e1611fdf21426d66a29a5ea62b7e30d512859e6/me_cleaner.py"
 ME_CLEANER_SHA="412e95538c46d6d4d456987a8897b3d0ad1df118c51378a350540eef51c242d4"
 
-SKL_DESCRIPTOR_URL="https://code.puri.sm/kakaroto/coreboot-files/raw/master/descriptor-skl.bin"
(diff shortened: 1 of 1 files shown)


osresearch:
Is there a legit site for the ME firmware image?  Can purism host a non-free repo similar to the https://github.com/tianocore/edk2-non-osi or https://github.com/coreboot/blobs

kakaroto:
Not as far as I know. I think ME images are sent to OEMs for inclusion in their BIOS and aren't made publicly available officially. That repository is contributed to by people from their own machines. Purism can't host the files since Purism doesn't own the copyright to those files. It is possible to distribute the files but it would involve accepting Intel's license which would prevent us from reverse engineering it basically.

osresearch:
I've been asking Intel about fixing the licensing terms on the ME and FSP blobs so that it is possible for the owner of a system to build a complete firmware image.  Hopefully they will add a clause to make this possible.

Since the user will be flashing the firmware anyway, is it too burdensome to require that they use flashrom or a programmer to read out the existing firmware and extract the ME blob from there?

kakaroto:
No, it's not burdensome to do that, and actually that's exactly what our updater script does. The need for downloading the ME externally was actually because of the HAP bit. The original ME version that came with the original librems didn't support the HAP bit (if enabled, it would brick and not boot at all), so we had to upgrade the ME to a newer version before enabling HAP. I'm guessing most users now have the newer ME, so it should be fine. If the script finds an older ME, then it could download it.

Note: we've debianized the 'megadown' tool which can be used to download from the mega.nz URL and are working on updating the script to use it that instead of asking the user to download the file manually.

tlaurion:
@kylerankin @kakaroto : https://mega.nz/#!DNdDVQ7I!hronBMVN8m82JciiT6UQwtwh-LVlHXIo-NzTB0324rk is not valid anymore.

kylerankin:
Closing PR as this was fixed in PR #512

PR #431, opened 2018-07-24
Update the path within the readme.md from the x220

Akendo:
The path was incorrect for the configuration file of the x220. Leading to some
confusion. Adjusting to the newer path, where the file is located.


osresearch:
merged!  For future contributions, please create a gpg key and use `git commit -S` to sign your commits.

PR #433, opened 2018-07-27
X230-flash DROPBEAR commented out to integrate GPG

tlaurion:
Since there is not enough space in the 4MB flash chip.
@flammit : In what use case DROPBEAR support is really required?


osresearch:
I'm not sure why we would want dropbear on a flasher image, especially if we don't have a network device driver...

osresearch:
This looks like it also enables i915 support.  Is #377 a pre-req before merging?

osresearch:
Merged just the config change: https://github.com/osresearch/heads/commit/cc4976b3dadbdc131b1cc0d9aed281d776ea4159

Issue #296, opened 2018-01-04
Revise use of GnuPG 1.4.x


duncanguthrie:
It would be useful to have a newer version of GnuPG, or to include different options.
By using the latest stable series, we can make use of cv25519 keys, which have smaller key sizes for equivalent security as RSA-2048, and produce smaller signatures.

Some thoughts:
- We can reduce some space in the initramfs, although I'm not sure if this is really valid, because GnuPG 2.x may be more bloated than the savings in key size. It may depend on the configuration of GnuPG.
- We can sign and verify signatures more quickly, which may be useful for reducing boot-time, especially on lower-end devices, should we employ the set-up in such contexts.
- We can take advantage of ECC on certain experimental cryptographic tokens, such as devices implementing the Gnuk token (http://www.fsij.org/doc-gnuk/intro.html), and certain Java cards implementing the OpenPGP Card version 3 standard. (These may or may not be of the same standard of security as the Yubikey, it should be noted - I know that in the case of Gnuk token, they make use of a generic microcontroller. It's useful as a reference platform, I think, though.)

I understand there are problems in that there are more dependencies with GnuPG 2.x.
Other options might be: 
- The NeoPG fork (https://github.com/das-labor/neopg- this is an experimental fork of GnuPG that changes perceived problems in the program being spun out into multiple components. It may or may not be suitable.
- OpenBSD's signify program (https://www.openbsd.org/papers/bsdcan-signify.html) - this is purely for signing, and makes compact signatures, and the program is very small. I'm not sure how well it supports smartcards, though, if at all. However, for some setups that don't want to install the entire GnuPG program, this could be very desirable.

osresearch:
The dependencies for Gnupg 2.x were quite large, especially when we were trying to fit entirely in the 4MB upper flash of the x230.  Now that we've abandoned that goal and are using both chips (and on servers have upwards of 64 MB), it might be reasonable to revisit.

I'm hesitant to use signify since it is a different tool than the "normal" gpg (and didn't support yubikey when I looked at it).

Can you build the newer gnupg version and figure out how much space is required?

duncanguthrie:
I played with the build system somewhat, and was able to make it build. Unfortunately, libgcrypt is not cross-compiling correctly, and I don't know why.
GnuPG 2.2 depends on the following components:
- libgpg-error: 349KB
- libgcrypt: a massive 4.0MB!
- libassuan: 404KB
- libksba: 845KB
- libnpth: 42KB
It installs the following components:
[...]

PR #454, opened 2018-09-18
Add support for GnuPG2

### LINKED ISSUE #435: Upgrade GnuPG to 2.2.9
Attempting to add GnuPG dependencies to Heads to resolve https://github.com/osresearch/heads/issues/296#issuecomment-409080487

Not really good at Makefile config file generation, but i'll try. The goal being to be able to own Nitrokey pro/Yubikey Neo devices and generate user keys from within Heads prior to flashing downloaded reproducible rom.

```
configure:
***
*** You need libgpg-error to build this program.
**  This library is for example available at
***   https://gnupg.org/ftp/gcrypt/libgpg-error
*** (at least version 1.24 is required.)
***
configure:
***
*** You need libgcrypt to build this program.
**  This library is for example available at
***   https://gnupg.org/ftp/gcrypt/libgcrypt/
*** (at least version 1.7.0 (API 1) is required.)
***
configure:
***
*** You need libassuan to build this program.
*** This library is for example available at
***   https://gnupg.org/ftp/gcrypt/libassuan/
*** (at least version 2.5.0 (API 2) is required).
***
configure:
***
*** You need libksba to build this program.
*** This library is for example available at
***   https://gnupg.org/ftp/gcrypt/libksba/
*** (at least version 1.3.4 using API 1 is required).
***
configure:
***
*** It is now required to build with support for the
*** New Portable Threads Library (nPth). Please install this
*** library first.  The library is for example available at
***   https://gnupg.org/ftp/gcrypt/npth/
*** (at least version 1.2 (API 1) is required).
***
configure: error: 
***
*** Required libraries not found. Please consult the above messages
*** and install them before running configure again.

```

osresearch:
This is based on the patches by @duncanguthrie and @tlaurion to fix #435 and #296.

gpg2 and its various libraries adds 430KB to the initrd compared to gnupg1.4.21.  This might be a concern for some of the smaller flash systems.


modules/npth
```diff
@@ -0,0 +1,24 @@
...
+	--disable-nls \
+	--with-libgpg-error-prefix="$(INSTALL)" \
+	--disable-asm \
+
+npth_target := $(MAKE_JOBS) \
+	DESTDIR="$(INSTALL)" \
+	$(CROSS_TOOLS) \
+	install \
+
+npth_libraries := src/.libs/libnpth.so
```
tlaurion: needed:
`npth_libraries := src/.libs/libnpth.so.0 `

modules/gpg2
```diff
@@ -0,0 +1,59 @@
...
+	--disable-finger \
+	--disable-dns-srv \
+	--disable-dns-cert \
+	--disable-regex \
+	--disable-nls \
+	--disable-all-tests \
+	--disable-wks-server \
+	--disable-wks-tools \
+	--disable-gnutls \
+	--disable-dirmgnr \
```
tlaurion: typo:
` npth_libraries := src/.libs/libnpth.so `

modules/libksba
```diff
@@ -0,0 +1,24 @@
...
+	--disable-nls \
+	--with-libgpg-error-prefix="$(INSTALL)" \
+	--disable-asm \
+
+libksba_target := $(MAKE_JOBS) \
+	DESTDIR="$(INSTALL)" \
+	$(CROSS_TOOLS) \
+	install \
+
+libksba_libraries := src/.libs/libksba.so
```
tlaurion: didn't see any error about it yet, but following the logic, I guess it should be:
`libksba_libraries := src/.libs/libksba.so.0`

osresearch:
With those library changes and enabling UNIX Domain Sockets with `CONFIG_UNIX` in the linux config to avoid the socket creation error, I'm able to startup`gpg-agent` and `scdaemon`.  Generating keys with `gpg --gen-key` fails due to a missing `pinentry` program.

I'm trying `gpg --gen-key --pinentry-mode loopback`, but it is hanging forever while generating entropy, I think.

osresearch:
Building pinentry-1.1.0 in tty mode works, although I'm not sure if it is currently hung waiting for entropy (on qemu) or if there is something else wrong.

tlaurion:
@osresearch : pinentry complains about curses.h missing.

duncanguthrie:
For the record, you don't need pinentry to be built with ncurses support. It can run in the tty on its own.

modules/pinentry
```diff
@@ -0,0 +1,39 @@
...
+# and datarootdir on the configure line so that the Makefiles will
+# be generated with the correct paths, but then re-write them when
+# we use the install target so that they will be copied to the correct
+# location.
+pinentry_configure := ./configure \
+	$(CROSS_TOOLS) \
+	--host x86_64-linux-musl \
+	--prefix "/" \
+	--enable-pinentry-tty \
+	--disable-pinentry-curses \
```
tlaurion: --disable-pinentry-qt5 \
        --disable-pinentry-fltk \
        --disable-pinentry-emacs \
+       --disable-fallback-curses \
        --with-libgpg-error-prefix="$(INSTALL)" \
        --with-libassuan-prefix="$(INSTALL)" \

tlaurion:
> For the record, you don't need pinentry to be built with ncurses support. It can run in the tty on its own.

@duncanguthrie It was supposed to. It was just not respected :)
the` --disable-fallback-curses` wasn't either, until placed [at the right place](https://github.com/osresearch/heads/pull/454#discussion_r218914515).

@duncanguthrie : have insight on [this](https://github.com/osresearch/heads/issues/435#issuecomment-422918053)?

jandryuk:
You might need to start gpg-agent with option `--scdaemon-program /bin/scdaemon`

image

jandryuk:
gpg-agent looks in /libexec (${libexecdir}) for scdaemon, but Heads packages scdaemon into /bin.  A workaround is to pass --libexecdir=/bin to gpg2_configure.  The other option is to extend the Heads Makefile to put files into /libexec, but that is more involved.

I haven't gotten to actually testing this with my Nitrokey yet...

tlaurion:
@jandryuk : I've added a gpg-agent.conf that gives scdaemon path and `gpg --card-edit` works on my Nitrokey Pro V2 :)

tlaurion:
@osresearch  @flammit : pinentry is still not invoked the right way to be usable. Actual stopper [here](https://github.com/osresearch/heads/pull/462#issuecomment-428186021)

osresearch:
Everything builds, `pinetry` seems to run. I'm trying to generate a key in qemu (no nitrokey) right now and it is taking quite some time.  I'll let it run for a while and check back...

tlaurion:
> Everything builds, `pinetry` seems to run. I'm trying to generate a key in qemu (no nitrokey) right now and it is taking quite some time. I'll let it run for a while and check back...

as stated [here](https://github.com/osresearch/heads/pull/462#issuecomment-428186021) gpg --card-edit works. The problem arrises when attempting to sign configuration. Please review posted screen capture in the linked comment.

tlaurion:
@osresearch : basically, from precendent screenshot, you can see that pinentry should be launched to ask for my passphrase but doesn't.

osresearch:
Not sure what failed in the circleci build since I can't see the coreboot logs:
```
failed. Check 'build-i386-elf-GCC/build.log'.
make[2]: *** [Makefile:48: build-i386] Error 2
make[1]: Leaving directory '/root/project/build/coreboot-4.7'
make: *** Waiting for unfinished jobs....
```

`gpg --gen-key` never finished -- it stayed like this overnight (after correctly invoking `pinentry`):
```
~ # gpg --gen-key --debug-all
gpg: Note: no default option file '//.gnupg/gpg.conf'
gpg (GnuPG) 2.2.10; Copyright (C) 2018 Free Software Foundation, Inc.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

gpg: enabled debug flags: packet mpi crypto filter iobuf memory cache memstat trust hashing ipc clock lookup extprog
Note: Use "gpg --full-generate-key" for a full featured key generation dialog.

GnuPG needs to construct a user ID to identify your key.

Real name: Foo Bar
Email address: foo@bar
You selected this USER-ID:
    "Foo Bar <foo@bar>"

Change (N)ame, (E)mail, or (O)kay/(Q)uit? o
We need to generate a lot of random bytes. It is a good idea to perform
some other action (type on the keyboard, move the mouse, utilize the
disks) during the prime generation; this gives the random number
generator a better chance to gain enough entropy.
gpg: DBG: [not enabled in the source] start
gpg: DBG: chan_4 <- OK Pleased to meet you, process 4095
gpg: DBG: connection to agent established
gpg: DBG: chan_4 -> RESET
gpg: DBG: chan_4 <- OK
gpg: DBG: chan_4 -> OPTION ttyname=/dev/console
gpg: DBG: chan_4 <- OK
gpg: DBG: chan_4 -> OPTION ttytype=linux
gpg: DBG: chan_4 <- OK
gpg: DBG: chan_4 -> OPTION lc-ctype=C.UTF-8
gpg: DBG: chan_4 <- OK
gpg: DBG: chan_4 -> OPTION lc-messages=C
gpg: DBG: chan_4 <- OK
gpg: DBG: chan_4 -> GETINFO version
gpg: DBG: chan_4 <- D 2.2.10
gpg: DBG: chan_4 <- OK
gpg: DBG: chan_4 -> OPTION allow-pinentry-notify
gpg: DBG: chan_4 <- OK
gpg: DBG: chan_4 -> OPTION agent-awareness=2.1.0
gpg: DBG: chan_4 <- OK
gpg: DBG: chan_4 -> RESET
gpg: DBG: chan_4 <- OK
gpg: DBG: chan_4 -> GENKEY
gpg: DBG: chan_4 <- S INQUIRE_MAXLEN 1024
gpg: DBG: chan_4 <- INQUIRE KEYPARAM
gpg: DBG: chan_4 -> D (genkey(rsa(nbits 4:2048)))
gpg: DBG: chan_4 -> END
gpg: DBG: chan_4 <- INQUIRE PINENTRY_LAUNCHED 301 tty 1.1.0 /dev/console linux -
gpg: DBG: chan_4 -> END
Please enter the passphrase to
protect your new key
Passphrase: 
Repeat: 
```

Regarding your attempt to use it in a script, does the same command work if run interactively? If so, it might be a problem of controlling tty or something terminal related.

tlaurion:
@osresearch :

Here is the command manually executed:
gpg2_invalid_ipc

Same, here with `gpg --debug-all`:
gpg2_invalid_ipc_debug-all

tlaurion:
@osresearch : https://www.gnupg.org/documentation/manuals/gnupg-2.0/Invoking-GPG_002dAGENT.html

So basically, the problem sits at making GPG_TTY variable aware of the tty. Where should we put that?

tlaurion:
@osresearch @flammit : still not perfect.

Unfortunately, instructing gpg for no-tty through gpg.conf makes pinentry works for signing, but fails requesting information from console when invoking gpg --card-edit  or gpg --gen-key which requires the user to comment gpg.conf's no-tty line and try again.

Why is gpg not picking up GPG_TTY and still tries to open /dev/tty?!?

tlaurion:
@osresearch : update on the stopper to merge?

jandryuk:
Is `gpg --gen-key` supposed to be work?  It hangs for me after entering a passphrase.

tlaurion:
This branch is not atop of master and builds with coreboot 4.7.....

tlaurion:
[CONFIG_HW_RANDOM, CONFIG_HW_RANDOM_INTEL and CONFIG_HW_RANDOM_AMD are not enabled in linux configs. ](https://dev.gentoo.org/~swift/docs/security_benchmarks/kernel.html#item-gt-kernelconfig-general-random)

Enabling them add no extra code. Testing.

tlaurion:
> [CONFIG_HW_RANDOM, CONFIG_HW_RANDOM_INTEL and CONFIG_HW_RANDOM_AMD are not enabled in linux configs. ](https://dev.gentoo.org/~swift/docs/security_benchmarks/kernel.html#item-gt-kernelconfig-general-random)
> 
> Enabling them add no extra code. Testing. 

Gave up after 45 minutes trying to generate 4096 bits with gpg --gen-key...
@osresearch : reverted to osresearch/master and testing gpg1 gpg --gen-key. Having entropy errors saying that 239 more bytes are missing and it never gets it.... Have gpg --gen-key already worked for someone with default gpg1 provided by master, anyone?

tlaurion:
GPG1 from master:
@osresearch : reverted to osresearch/master and testing gpg1 gpg --gen-key. Having entropy errors saying that 239 more bytes are missing and it seems it will never get it cause the recovery shell doesn't do anything else.... 


**Have `gpg --gen-key` already worked for someone with default gpg1 provided by master, anyone?**
gpg2 branch never gives output at all, even not that entropy warning. Will investigate more, but I do not understand for the moment. Always generated keys from smartcard, never from the shell, even though I swear I tested it ang it was working....!?!

Will try to merge [CONFIG_HW_RANDOM, CONFIG_HW_RANDOM_INTEL and CONFIG_HW_RANDOM_AMD. ](https://dev.gentoo.org/~swift/docs/security_benchmarks/kernel.html#item-gt-kernelconfig-general-random) into master to see if I can get gpg1 work before testing gpg2?!

jandryuk:
@tlaurion You've tested gpg2 w/ nitrokey on an x230?

My concern is if `gpg --gen-key` doesn't work, then something is wrong with gpg2.  I'm only working from QEMU at this time.  I ran Heads under QEMU/KVM.  The CPU wasn't particularly active, so I'm not sure what gpg is stuck on.  I wish strace were available to see why it's hanging.

tlaurion:
@jandryuk : I've used gpg2 ang gpg1 only with `gpg --card-edit `to generate keys, and that works. 

Trying to figure out what is wrong with pseudo-random generators. I'm testing on x230, not qemu, but can see that neither works (osresearch/master gpg1 branch and osresearch/gpg2 branch, rebased on master)

tlaurion:
I confirm, by running `gpg --debug-all --gen-key` that entropy is missing and filled if typing random keys on keyboard. 

Since Heads relies on smartcards to sign configs, is `gpg --key-gen` missing entropy a false problem?
@osresearch @flammit ?

A [couple of solutions ](https://wiki.archlinux.org/index.php/Random_number_generation) could be taken into consideration if entropy generation in Heads is really a need ( I don't think so ).

tlaurion:
The only need I see would be to generate keys from Heads to keep an offline backup of master key, and insert subkeys in the smartcard for daily operations, like proposed, for example, in this [guide](https://we.riseup.net/debian/using-the-openpgp-card-with-subkeys).

@kyle @flammit, @osresearch: False need?

jandryuk:
@tlaurion Thanks for tracking down the problem!  I am okay with leaving it as-is since it's due to the lack of entropy.  I was more concerned there was still something broken with the gpg2 installation causing the hang.

initrd/init
```diff
@@ -5,6 +5,9 @@ echo "hello world" > /dev/ttyprintk
 # Setup our path
 export PATH=/sbin:/bin:/usr/sbin:/usr/bin:/usr/local/bin
 
+#export GPG_TTY STATICALLY (NEED TO FIX)
+export GPG_TTY=/dev/console
```
jandryuk:
```suggestion
export GPG_TTY=$(tty)
```
tty is installed, so it could be used here.  tty returns /dev/console, so that doesn't change.

tlaurion: @jandryuk : I've took that approach initially. Exiting to recovery shell consequently gives:
```
echo $GPG_TTY
not a tty

tty
/dev/console
```

init script doesn't have an access to a TTY.
This is why gpg command verifying signatures  is `gpg --no-tty`, because from a whiptail perspective, it doesn't have access to a shell until ash is spawned. The reason why gpg signing works is because pinentry-tty is called for which requires the tty to be set, else it also fails.

I haven't found a more elegant solution then to hardcode the tty, force the paths to binaries in config and force gpg-agent usage. If you have a better solution, please advise. :)

jandryuk: I'm pretty sure I used `$(tty)` for qemu's recovery shell and it worked. But that wasn't the whiptail stuff.  Can you run the whiptail stuff under qemu?

Anyway, I don't have a better suggestion.

jandryuk: Oh, your call to `tty` was before /dev/ was mounted.  Setting GPG_TTY later works for me in QEMU.  Check out this commit: https://github.com/jandryuk/heads/commit/ab8046b5cc51b8c9863c639ee3d614c144723f3d

flammit: @jandryuk if you want to test fbwhiptail/gui-init, see this branch: https://github.com/flammit/heads/tree/qemu-gui-init.  Feel free to drop by u-root.slack.com / #heads channel if you wanna chat live (invite available from slack.u-root.com)

tlaurion:
@osresearch : please merge #491 here.

initrd/init
jandryuk: @flammit Oh, I forgot to thank you for the gui branch.  Thanks!

itay-grudev:
Is this PR stable at the moment?

tlaurion:
@itay-grudev : yes. You have to merge #491 on top of this one and you're good to go.

itay-grudev:
I have the following problem when I attempt to update my signatures after switching to the gpg2 branch. 

```txt
gpg: gpg-agent is not available in this session
gpg: Sorry, no terminal at all requested - can't get input
/boot: Unable to sign kexec hashes
Failed to sign default config
```

It ends with a kernel panic.

itay-grudev:
I also had a problem adding my GPG key. I ended up flashing to a non `gpg2` ROM, adding it and then switching to `gpg2`.

itay-grudev:
```
gpg: Sorry, no terminal at all requested - can't get input
```

Fixed in #505.

itay-grudev:
After I switched to this branch and compiled this I noticed that gpg is still using `1.x`. And I am also getting a: 

```
gpg: BAD signature from key_id_here
```

Please advise.

itay-grudev:
Fixed in #505

itay-grudev:
With #505: Compiled successfully. GPG2 works fine. Signing works fine. Verification works fine. Tested on a Librem 13 v3 with a 4096 bit RSA key.

tlaurion:
Please merge #510 instead.

PR #467, opened 2018-10-15
Couple of tweaking for Winterfell linuxboot build

vejmarie:
- Higher granularity on USB driver selection
- Reduce initial kernel footprint

These optimizations are required to lower the size of the kernel and user space due to migration to elvish as default shell in u-root.


config/linux-linuxboot.config
```diff
@@ -104,7 +104,7 @@ CONFIG_DEVTMPFS_MOUNT=y
 CONFIG_BLK_DEV_LOOP=y
 CONFIG_BLK_DEV_RAM=y
 CONFIG_BLK_DEV_RAM_SIZE=65536
-CONFIG_BLK_DEV_NVME=m
+CONFIG_BLK_DEV_NVME=y
```
osresearch: Shouldn't NVME remain as a module since not all platforms need it?

config/linux-linuxboot.config
```diff
@@ -119,16 +119,16 @@ CONFIG_SCSI_SCAN_ASYNC=y
...
+# CONFIG_MEGARAID_MM is not set
+# CONFIG_MEGARAID_MAILBOX is not set
+# CONFIG_MEGARAID_SAS is not set
 CONFIG_SCSI_GDTH=m
 CONFIG_SCSI_ISCI=m
 CONFIG_SCSI_DH=y
 CONFIG_ATA=m
-CONFIG_SATA_AHCI=m
-CONFIG_SATA_AHCI_PLATFORM=m
+CONFIG_SATA_AHCI=y
```
osresearch: The SATA should stay a module - we don't want to spend the time to initialize it and probe the buses on platforms that don't require it (plus it adds to the potential attack surface to initialize them prior to attestation)

config/linux-linuxboot.config
```diff
@@ -163,7 +163,7 @@ CONFIG_E1000E=m
 CONFIG_IGB=m
 # CONFIG_NET_VENDOR_I825XX is not set
 # CONFIG_NET_VENDOR_MARVELL is not set
-CONFIG_MLX4_EN=m
+CONFIG_MLX4_EN=y
```
osresearch: MLX4 should stay a module as well; not every platform needs it (and it's huge).

boards/x220/x220.config
```diff
@@ -17,7 +17,9 @@ CONFIG_QRENCODE=y
 CONFIG_TPMTOTP=y
 CONFIG_DROPBEAR=y
 
-CONFIG_LINUX_USB=y
+CONFIG_LINUX_USB_EHCI=y
```
osresearch: Instead of changing every config to specifically bring in the different USB drivers, let them define `CONFIG_LINUX_USB` and then `modules/linux` can check to see if that variable is defined and create the various drivers.  That way only platforms that want to specialize need to tweak their `CONFIG_LINUX_USB_EHCI` etc definitions.

osresearch:
I added a few comments to the changes; in general having the ability to specialize the USB modules makes sense, but we can do it in a way that doesn't require most platforms to define all of the config variables.

I'm more concerned about bringing in device drivers into the kernel; in addition to the attack surface issues, some of them like MLX4 are huge (and slow to init), and ones like SATA takes a long time to scan the buses, which is not necessary on network booting platforms.

boards/x220/x220.config
vejmarie: ok I will try to do that

config/linux-linuxboot.config
vejmarie: I agree, one of the issue I face, is that in space constraint environment like winterfell, if we want support pxe boot with u-root, we do have to support mlx4 driver, and the blob is getting to fat to fit as a kernel module + elvish + pxeboot code in u-root. Is there a way to move such specific kernel configuration at the board level definition and perhaps concatenate these specific needs to build final kernel config ?

config/linux-linuxboot.config
vejmarie: agree, we shall have I believe a minimal kernel config, but when it is defined as a module it is automatically added to the user land space which makes the system fat. So I think it might be good to make it happens the same way than mlx4 ?

tlaurion:
@vejmarie?

tlaurion:
@vejmarie is there still interest for linuxboot/Heads on coreboot platforms at HPE?

tlaurion:
Abandoned

PR #469, opened 2018-10-18
fix the fbwhiptail source URL

merge:
The current source URL is not available anymore.

kakaroto changed his copy of heads to point to his own github account's fbwhiptail:
https://github.com/kakaroto/heads/commit/b13cc5e68d9d77ef9bfd6623b812f4205698dafb

But it seems that source.puri.sm/coreboot is a more accessible home for the
project.


kylerankin:
Confirming that source.puri.sm/coreboot is the proper repository to use for fbwhiptail.

PR #471, opened 2018-10-22
Fix usb-scan so it proposes found ISOs

tlaurion:
$TMP_MENU_FILE is an unknown variable. Replaced with /tmp/iso_menu.txt so usb-scan works.


tlaurion:
Partly linked to https://github.com/osresearch/heads/issues/470 to install Ubuntu

Issue #215, opened 2017-07-16
Video glitches when kexec into Ubuntu kernel


osresearch:
![Video glitches](https://pbs.twimg.com/media/DEzH5yiUQAACY1C.jpg)

At the coreboot session during http://hackon.nl, @kliment found that his x230 would cause VGA display glitches when `kexec` into a stock 4.4.0-83 Ubuntu kernel.

It looks like @flammit solution to getting Tails to boot https://github.com/osresearch/heads/issues/195#issuecomment-296489243 might fix it.  The problem is that the Heads kernel has turned on the IOMMU and Ubuntu is not expecting this, so it is necessary to add `intel_iommu=on` to the kexec kernel command line.

merge:
I see the same glitches. This is __not__ related to Ubuntu. I run a stock 4.9 Debian kernel.

The thing is: The glitches last for as long as runtime-Linux hasn't set the fb-mode. After that, the display manager starts fine. However we don't have Linux console logs and this is a problem: If you use dm-crypt disk encryption, you are stuck and can't boot since Linux will stop and ask for the passphrase before setting modes!

`intel_iommu=on` is set, as in the master branch for the X230. I have https://github.com/osresearch/heads/pull/465 applied, but this should actually be unrelated, shouldn't it?

Any idea what else would make sense to test here?

merge:
I've found that I need `intel_iommu=on`. Having __only__ `intel_iommu=igfx_off` makes things worse and I'm stuck with the glitches forever.

merge:
> Are we talking about [this](https://github.com/osresearch/heads/pull/377/files#diff-2d4ed2860b55bb308eacc61e6c18a86fR640)? or [this](https://github.com/osresearch/heads/pull/452/files#diff-2d4ed2860b55bb308eacc61e6c18a86fR19) ? Those are required, with i915, to properly boot without screen flickers.
> 
sorry, but what's the differnce? isn't this the same thing in different pull requests? thanks!

tlaurion:
@merge : no difference. Just not present in the change subset being applied in #465 :)

tlaurion:
@merge : which version of Ubuntu were you able to install/ you attempt to use?
Not able to install [18.10 FTM](https://github.com/osresearch/heads/issues/470#issuecomment-431917443).

merge:
I don't use Ubuntu, only Debian 9. I didn't yet install it when running Heads, but my report is about booting an installed stock Debian kernel.

[...]

PR #480, opened 2018-11-18
Fixing i915 FB corruption with other OSes FB

tlaurion:
Fixing i915 FB on x230 to boot ISOs and debian kernels, permitting to deal with graphical LUKS passphrase requests and the like. 

There is a lateral flip of a millisecond still, [just like when booting QubesOS](https://github.com/osresearch/heads/issues/219#issuecomment-318857229).


tlaurion:
Fixes #215

paulmenzel:
Could you please update the commit message to describe why your changes 
fix the issue?

flammit:
I've tested x230 on the master branch and it boots up all the distros that I would expect to work (Ubuntu, Debian, Fedora, Tails, Qubes).  I've also booted to Ubuntu, Fedora with the disk encryption and I get the graphical display prompt just fine.

@tlaurion If you're trying to enable i915 by default on x230, then yes you need to add the extra parameters to the boot kernel and the target kernel adjust the iommu settings (though I thought `iommu-igfx=off` was safer than `iommu=pt`).  However, maybe it's worth having a `x230-gui` board config that has these changes along with `gui-init` by default?

merge:
This indeed fixes #215 for me too. Instead of testing this PR directly, I steal the necessary parts off of this fix and updated #465 accordingly which now boots cleanly as expected.

tlaurion:
@flammit: why iommu=pt is considered less safe?
@merge @flammit:  I was merging flammit/i915 branch in my tests the past. Completely forgot about it recently. My bad. Merged those changes in my PR. Will close.

So here is what worked/still works:
[boards/x230-gui/x230.config](https://github.com/tlaurion/heads/blob/x230-FB-Tests/boards/x230/x230.config): 
`export CONFIG_BOOT_KERNEL_ADD="intel_iommu=on intel_iommu=igfx_off"`

[config/coreboot-x230.config](https://github.com/tlaurion/heads/blob/x230-FB-Tests/config/coreboot-x230.config):
```
CONFIG_NO_GFX_INIT=y
CONFIG_LINUX_COMMAND_LINE="intel_iommu=igfx_off quiet"
```

[config/linux-x230.config](https://github.com/tlaurion/heads/blob/x230-FB-Tests/config/linux-x230.config):
`CONFIG_DRM_I915=y`
(Getting rid of a lot of unneeded lines that were linked to generic framebuffer that I never got to work anyway.)

tlaurion:
@merge : As pointed out by flammit, #480 now contains his flammit/i915 branch which fixed the issue but was never merged.

merge:
@tlaurion thanks for your updates. I've shamelessly included them in #465 which I hope to be the minimal necessary change for switching to gui-init on the x230.

tlaurion:
CLosing in favor of #465

PR #483, opened 2018-11-21
fix install directory handling for git and builds

merge:
The install directly should basically behave like the "build" directory.
Since it's tracked by git, containing a gitignore file, we shouldn't
have it in the toplevel gitignore (just like the build directory).

But then, the toplevel Makefile's real.clean target removes the install
directory. This is changed so that only it's content is being removed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,7 +15,6 @@ config/*.old
 *.log
 *~
 crossgcc
-install
 clean
 *.map
 *.sec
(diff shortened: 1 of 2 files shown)


merge:
since `make real.clean` taints the tree, this change might even be necessary in order to build reproducibly using real.clean

osresearch:
Good point. I almost never use `clean`; I just do a new checkout...

One minor tweak - the `rm -rf *` should probably be `rm -rf -- *` just to avoid any accidental extra arguments

osresearch:
Merged in https://github.com/osresearch/heads/commit/371b65ff5827e1048511b91be1f9ce656040cb9d

PR #487, opened 2018-11-27
GPG trustdb export from .ash_history and import in key-init

tlaurion:
Add support for trustdb after upgraded command history to export it at key generation, so that the signing key is trusted and heads doesn't complain about it.

### DIFF
--- a/initrd/.ash_history
+++ b/initrd/.ash_history
@@ -1,14 +1,17 @@
-mount /dev/sda1 /boot
-mount -o remount,rw /boot
-rm /boot/kexec_*
-mount-usb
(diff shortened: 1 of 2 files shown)


flammit:
@tlaurion can you move this PR to be against the `osresearch/gpg2` branch so once gpg2 is ready to merge it can be done from that branch?  hard to tell what you've changed here.

flammit:
@tlaurion this is still against osresearch/master, you need to make a new PR and select the gpg2 branch instead of master (which is default).

tlaurion:
replaced by #492, being applied on top of GPG2 branch.

PR #490, opened 2018-11-30
Add empty keyring detection, clean up main menu

kylerankin:
To help with onboarding new users to Heads, this change will detect when
Heads does not have any keys in its keyring and will guide the user
through adding a key to the running BIOS. It's important that this
happen *before* guiding them through setting up an initial TOTP/HOTP
secret because adding a GPG key changes the BIOS, so the user would have
to generate TOTP/HOTP secrets 2x unless we handle the keyring case
first.

In addition to this change I've simplified the main menu so that the
majority of the options appear under an 'advanced' menu.


tlaurion:
@flammit: fonctionnaly, I do not see any problem with this and tested it extensively.
[I added to this menu the possibility of factory resetting the GPG card to it in my PR](https://github.com/osresearch/heads/pull/507/files#diff-9a66e05586b37d4779ee1eb88802e759R252)

PR #491, opened 2018-12-03
Set GPG_TTY before calling gpg in key-init

tlaurion:
Cherry-pick of @jandryuk to merge atop osresearch/gpg2 branch:

> 
> gpg2 needs GPG_TTY set to function properly.  We set it in /init so it
> is inherited by all children.  The call to $(tty) must be after /dev and
> (preferably) /dev/pts are mounted.
> 
> Signed-off-by: Jason Andryuk <jandryuk@gmail.com>


tlaurion:
With that #454 seems complete.

PR #493, opened 2018-12-06
Use the Librem Key as a TPM work-alike in the absence of a TPM

kylerankin:
On machines without a TPM, we'd still like some way for the BIOS to attest that it has not been modified. With a Librem Key, we can have the BIOS use its own ROM measurement converted to a SHA256sum and truncated so it fits within an HOTP secret. Like with a TPM, a malicious BIOS with access to the correct measurements can send pre-known good measurements to the Librem Key.
    
This approach provides one big drawback in that we have to truncate the SHA256sum to 20 characters so that it fits within the limitations of HOTP secrets. This means the possibility of collisions is much higher but again, an attacker could also capture and spoof an existing ROM's measurements if they have prior access to it, either with this approach or with a TPM.

### DIFF
--- a/initrd/bin/seal-totp
+++ b/initrd/bin/seal-totp
@@ -17,35 +17,36 @@ fi
 TOTP_SECRET="/tmp/secret/totp.key"
 TOTP_SEALED="/tmp/secret/totp.sealed"
 
-dd \
-	if=/dev/urandom \
-	of="$TOTP_SECRET" \
-	count=1 \
-	bs=20 \
-	2>/dev/null \
-|| die "Unable to generate 20 random bytes"
+if [ "$CONFIG_TPM" = "y" ]; then
+  dd \
+    if=/dev/urandom \
+    of="$TOTP_SECRET" \
+    count=1 \
+    bs=20 \
+    2>/dev/null \
+  || die "Unable to generate 20 random bytes"
 
-secret="`base32 < $TOTP_SECRET`"
+  secret="`base32 < $TOTP_SECRET`"
 
 # Use the current values of the PCRs, which will be read
 # from the TPM as part of the sealing ("X").
 # PCR4 == 0 means that we are still in the boot process and
 # not a recovery shell.
 # should this read the storage root key?
-if ! tpm sealfile2 \
-	-if "$TOTP_SECRET" \
-	-of "$TOTP_SEALED" \
-	-hk 40000000 \
-	-ix 0 X \
-	-ix 1 X \
-	-ix 2 X \
-	-ix 3 X \
-	-ix 4 0000000000000000000000000000000000000000 \
-	-ix 7 X \
-; then
-	rm -f "$TOTP_SECRET"
-	 die "Unable to seal secret"
-fi
+  if ! tpm sealfile2 \
+    -if "$TOTP_SECRET" \
+    -of "$TOTP_SEALED" \
+    -hk 40000000 \
+    -ix 0 X \
+    -ix 1 X \
+    -ix 2 X \
+    -ix 3 X \
+    -ix 4 0000000000000000000000000000000000000000 \
+    -ix 7 X \
+  ; then
+    rm -f "$TOTP_SECRET"
+     die "Unable to seal secret"
+  fi
 
 
 # to create an nvram space we need the TPM owner password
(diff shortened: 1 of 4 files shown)


tlaurion:
Tested [this](https://github.com/tlaurion/heads/commits/kgpe-d16-merging_good_order) on KGPE-D16 with unreliable results, since for some reason, the kgpe-d16 coreboot configuration with `CONFIG_RELOCATABLE_RAMSTAGE` breaks boot. Also note that CBMEM console output would need to be disabled in all coreboot configs.

My insight on this is that extracting all rom for integrity attestation may not be the right solution. For example, on the kgpe-d16, we have 16MB of rom, which takes around 60 seconds to extract. I haven't tested this [on the x200](https://github.com/tlaurion/heads/tree/x200_readd) yet, but will at the end of this week.

What I would propose would be:
- export in board config a list of blacklisted modules to skip measurements for (eg: kgpe-d16: s3nv since it changes at each boot)
- list rom files with `cbfs --list`
- loop in module list, extract unblacklisted modules files and sha256sum them on the fly, keeping new_value = Hash(old_value || new_measurement))
- truncate final hash and store it in HOTP.

@kylerankin @flammit : any thoughts?

tlaurion:
On KGPE-D16 s3nv, from Timothy Pearson on [coreboot ML](https://mail.coreboot.org/pipermail/coreboot/2018-December/087985.html)
```
The memory is not (and cannot be) retrained, the
previous settings are loaded from the s3nv region of Flash (this is also
why the settings change on each boot -- the last known good training
data is loaded into Flash to support resume from suspend).

My understanding is that because the current coreboot native AMD
codebase doesn't support relocatable ramstage (yet?) we're hitting a
slow path somewhere in resume.  I think there was some work being put
into adding the relocateable ramstage support but I don't know current
status.
```

tlaurion:
@kylerankin:

I'm playing around and exploring options. Testing the following:
```
diff --git a/initrd/bin/unseal-hotp b/initrd/bin/unseal-hotp
index c6baa78..1945267 100755
--- a/initrd/bin/unseal-hotp
+++ b/initrd/bin/unseal-hotp
@@ -35,11 +35,12 @@ else
 # without a TPM, use the first 20 characters of the ROM SHA256sum 
   echo "TPM not configured, measuring ROM directly" 1>&2
   # use a previously-copied image if it exists
-  if [ -f ${ROM_IMAGE} ]; then
-    sha256sum ${ROM_IMAGE} | cut -f1 -d ' ' | cut -c 1-20 | tr -d '\n' > $HOTP_SECRET
-  else
-    flash.sh -s ${ROM_IMAGE} | cut -c 1-20 | tr -d '\n' > $HOTP_SECRET
-  fi
+  #if [ -f ${ROM_IMAGE} ]; then
+  #  sha256sum ${ROM_IMAGE} | cut -f1 -d ' ' | cut -c 1-20 | tr -d '\n' > $HOTP_SECRET
+  #else
+    #flash.sh -s ${ROM_IMAGE} | cut -c 1-20 | tr -d '\n' > $HOTP_SECRET
+  cbfs --list | grep -E 'fallback|heads|microcode|bootblock' | while read CBFS_FILES; do cbfs -o $CBFS_FILES; done | sha256sum | cut -f1 -d ' ' | cut -c 1-20 | tr -d '\n' > $HOTP_SECRET
+  #fi
 fi

```

tlaurion:
@kylerankin :

I updated my small PoC to [extend measured modules](https://github.com/tlaurion/heads/commit/100751d2683342495a9ff269a6952d26f1b0153f)

Here is the modules present in x200 coreboot 4.9:

```
FMAP REGION: COREBOOT
cbfs master header             0x0        cbfs header        32 none
fallback/romstage              0x80       stage           53836 none
cpu_microcode_blob.bin         0xd340     microcode       49152 none
fallback/ramstage              0x193c0    stage           88270 none
cmos_layout.bin                0x2ed00    cmos_layout      1664 none
fallback/postcar               0x2f3c0    stage           14920 none
fallback/dsdt.aml              0x32e40    raw             14207 none
data_ccfl.vbt                  0x36640    raw              1501 LZMA (3854 decompressed)
data_led.vbt                   0x36c80    raw              1439 LZMA (3863 decompressed)
fallback/payload               0x37280    simple elf    6483537 none
(empty)                        0x666140   null           628824 none
bootblock                      0x6ff9c0   bootblock         976 none

Built lenovo/x200 (ThinkPad X200)

	** WARNING **
coreboot has been built without an Intel Firmware Descriptor.
Never write a complete coreboot.rom without an IFD to your
board's flash chip! You can use flashrom's IFD or layout
parameters to flash only to the BIOS region.
```

Here are the modules present in kgpe-d16 coreboot 4.9:
```
FMAP REGION: COREBOOT
cbfs master header             0x0        cbfs header        32 none
fallback/romstage              0x80       stage          173132 none
config                         0x2a540    raw               613 none
revision                       0x2a800    raw               561 none
cmos.default                   0x2aa80    cmos_default      256 none
cmos_layout.bin                0x2abc0    cmos_layout      3524 none
fallback/dsdt.aml              0x2b9c0    raw              9895 none
(empty)                        0x2e0c0    null             7384 none
s3nv                           0x2fdc0    raw             65536 none
fallback/ramstage              0x3fe00    stage           88189 none
fallback/payload               0x556c0    simple elf    6310996 none
microcode_amd.bin              0x65a380   microcode       12684 none
microcode_amd_fam15h.bin       0x65d580   microcode        7876 none
(empty)                        0x65f4c0   null         10092568 none
bootblock                      0xfff500   bootblock        2232 none

```

initrd/bin/seal-totp
```diff
@@ -17,69 +17,77 @@ fi
...
+      -if "$TOTP_SEALED" \
+    || die "Unable to write sealed secret to NVRAM"
+  fi
 
-rm -f "$TOTP_SEALED"
+  rm -f "$TOTP_SEALED"
+else
+# without a TPM, use the first 20 characters of the ROM SHA256sum 
+  echo "TPM not configured, measuring ROM directly"
+  flash.sh -s /tmp/coreboot.rom | cut -c 1-20 | tr -d '\n' > $TOTP_SECRET
```
jtracey: Why truncate the hash prematurely? Could you not get the full 120 bits of the HOTP with something like
```suggestion
  secret="`flash.sh -s /tmp/coreboot.rom  | xxd -r -p | base32 | cut -c 1-24 | tr -d '\n'`"
```
(i.e., decode the base16 string then re-encode as base32)?

tlaurion: @kylerankin thoughts?

tlaurion:
@kylerankin: did I dream or did the HOTP code got upgraded to permit bigger HOTP size?

tlaurion:
Confirmed: https://github.com/Nitrokey/nitrokey-storage-firmware/pull/85

tlaurion:
@kylerankin : Confirmed 40 bytes (320 bits) for NitroKey 0.8+ [here and coming to NitroKey Storage next week]( https://github.com/Nitrokey/nitrokey-pro-firmware/issues/59)

tlaurion:
@kylerankin: ping :)

tlaurion:
@MrChromebox :
https://github.com/osresearch/heads/pull/493#issuecomment-487250771
https://github.com/osresearch/heads/pull/493#issuecomment-487379379

MrChromebox:
@tlaurion think it will be easier if I just push a new PR for this rather than trying to take over Kyle's. This one can be closed once I do

kylerankin:
Closing this PR as we have a newer one that solves the same issue.

tlaurion:
@szszszsz and on HID issue?

PR #494, opened 2018-12-06
Add GUI to change /etc/config configuration options

kylerankin:
This change adds a config-gui.sh script that provides a GUI to change exported configuration options like you'd see in /etc/config. To protect against any future additions to /etc/config that this would otherwise overwrite, this change adds an additional /etc/config.user file that is intended to be stored in CBFS and at boot combines all /etc/config* files into /tmp/config. Existing scripts have been updated to source this new combined config file.

This script behaves much like on network hardware where the changes apply to the running instance only, and to make them persist you have to explicitly tell Heads to save them, whereby it uses cbfs to add /etc/config.user to the running BIOS and reflash.

### DIFF
--- a/initrd/bin/config-gui.sh
+++ b/initrd/bin/config-gui.sh
@@ -67,7 +67,7 @@ while true; do
       exit 0
     ;;
     "b" )
-      CURRENT_OPTION=`grep 'CONFIG_BOOT_DEV=' /tmp/config | cut -f2 -d '=' | tr -d '"'`
+      CURRENT_OPTION=`grep 'CONFIG_BOOT_DEV=' /tmp/config | tail -n1 | cut -f2 -d '=' | tr -d '"'`
       find /dev -name 'sd*' -o -name 'nvme*' > /tmp/filelist.txt
       file_selector "/tmp/filelist.txt" "Choose the default /boot device.\n\nCurrently set to $CURRENT_OPTION."
       if [ "$FILE" == "" ]; then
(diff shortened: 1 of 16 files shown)


merge:
Good work as far as I can tell! Maybe that's just me, but you could simplify this a lot for reviewers, and probably make merging more easy too:

* squash together all changes to the new config-gui.sh script. at least the first 4 commits + 6th + 11th.
* just make a _different_ PR for the 7th (grep on empty file)

so you'd save at least 7 commits and all-in-all have (semantically) "add the script" and "properly use it in Heads userspace", plus the 4 or 5 smaller changes you have below, instead.

And as long as the end-result is the same, I'd do a force-push for updating the PR.

kylerankin:
I'll try to have more simplified branches for future PRs.

tlaurion:
@flammit  @kylerankin : pass fonctional testing and the code is clean.
Approved :)

tlaurion:
@kylerankin : you should [add this](https://github.com/tlaurion/heads/commit/4ee95c2061072e37a52d9842719e746919c7160b#diff-10d5be21d425b1f3dd8bb331ef7d9ad6R22) to make sure pressing "r" early to trigger recovery console doesn't break UX of calling other scripts.

kylerankin:
@tlaurion Did you link to the correct diff? I'm not sure how all of those changes would affect triggering the recovery console. A lot of them seem related to other work (some reference generating 4096-bit keys) and are patches against a number of other files.

I personally feel this code is ready to merge, but I'd prefer @tlaurion that you merge it since I submitted it. I'd like to add additional functionality to it but want to wait until this core feature set is merged.

tlaurion:
> @tlaurion Did you link to the correct diff? I'm not sure how all of those changes would affect triggering the recovery console. A lot of them seem related to other work (some reference generating 4096-bit keys) and are patches against a number of other files.
> 
> I personally feel this code is ready to merge, but I'd prefer @tlaurion that you merge it since I submitted it. I'd like to add additional functionality to it but want to wait until this core feature set is merged.
@kyle: 
yep, previous comment was linking specifically to init file, line 22 of another banch i'm depending on.

The init addition i'm referring to is to simply touch /tmp/config, prior to launching recovery shell. Else all modified scripts by this PR tries to incorporate /tmp/config which doesn't exist, since not created yet. As a result, all modified scripts fail ungracefully.

kylerankin:
@tlaurion OK I've added that change.

PR #496, opened 2018-12-13
Improve mirror URLs

strugee:
The new URL automatically redirects to a nearby, current GNU mirror.

Also, the fact that it's HTTPS helps with restrictive outbound firewall policies that disallow plaintext traffic (for example, using Qubes' firewall functionality).


strugee:
I just pushed a couple commits converting most (all?) of the rest of the mirror URLs to HTTPS since it seemed easier to just expand this PR rather than submitting a brand-new one. See also GregorR/musl-cross#72.

strugee:
I just ran a successful clean build with these patches and the patches from GregorR/musl-cross#72 applied, confirming that nothing gets downloaded over HTTPS. (My development environment's firewall is configured to block anything but HTTPS and SSH.)

PR #510, opened 2019-01-26
Gpg2 with all fixes included

tlaurion:
Supersedes [gpg2 branch that is not based on master](https://github.com/osresearch/heads/tree/gpg2) 

@kylerankin @flammit : please review!

@kylerankin please rebase #490 on this branch?
I'll then propose factory resetting options based on #490.


tlaurion:
fixes #435

tlaurion:
So.

This [section is useless in GPG2](https://github.com/osresearch/heads/pull/510/files#diff-49199d392e898d67aa3a302abee66c19R147)

@flammit : 
When using Heads with Whiptail, the user can upgrade the rom with two options: 
1. With a "clean rom": a clean rom won't contain any user public key. When rebooting, @kylerankin's #490 will trigger and[ ask the user if he wants to import a new key](https://github.com/osresearch/heads/pull/490/files#diff-9a66e05586b37d4779ee1eb88802e759R117). On top of this, [with this PR](https://github.com/kylerankin/heads/pull/2), the user is provided with the option[ to Factory reset a Nitrokey/LibremKey](https://github.com/kylerankin/heads/pull/2/files#diff-9a66e05586b37d4779ee1eb88802e759R258), which will[ generate keys, export public key and trustdb, and insert those into a taken running rom export prior to reflashing it back internally into SPI flash](https://github.com/kylerankin/heads/pull/2/files#diff-9a66e05586b37d4779ee1eb88802e759R378).
2. With a new rom: injects current user public key and otrust.txt into running rom prior to flashing it into SPI flash.

On each reboot, [otrust.txt is imported](https://github.com/osresearch/heads/pull/510/files#diff-fca00be063d58bcbb73662ca90f46bd1R10) so that the user's public key is ultimately trusted (to get rid of warnings that confuses users otherwise)

jandryuk:
Should this PR be cleaned up? Off the top of my head, these two commits cancel reach other out:
https://github.com/osresearch/heads/commit/02728ffeabfefe4c2b7e3e86484332f8102dfe5e
https://github.com/osresearch/heads/commit/ee12c3ab37a061d048d77670cdf17e54334668c5

Those two could just be dropped.

Other commits are fix ups that could just be squashed.

tlaurion:
@jandryuk 
> Should this PR be cleaned up? Off the top of my head, these two commits cancel reach other out:
> [02728ff](https://github.com/osresearch/heads/commit/02728ffeabfefe4c2b7e3e86484332f8102dfe5e)
> [ee12c3a](https://github.com/osresearch/heads/commit/ee12c3ab37a061d048d77670cdf17e54334668c5)
> 
> Those two could just be dropped.
Done.

@flammit @kylerankin : something else? Ready to merge?

tlaurion:
@jandryuk @flammit @kylerankin 

I've been playing around the problem of migrating keys that @flammit was questionning over slack.

Let me get the facts straight so we can have a discussion on what we are doing with keys and what we could do:
- We currently [import public keys in text form](https://github.com/osresearch/heads/pull/510/files#diff-fca00be063d58bcbb73662ca90f46bd1R8) (without specifying --homedir. Doesn't cause problem)
- At the same time, we replace trustdb.gpg (which is the same format for gpg1 and gpg2) and [pubring.gpg](https://github.com/osresearch/heads/blob/8dd1082808f8479f5b9eb5c9827deca53e6b4164/initrd/bin/flash-gui.sh#L197)/[pubring.kbx](https://github.com/osresearch/heads/blob/8dd1082808f8479f5b9eb5c9827deca53e6b4164/initrd/bin/flash-gui.sh#L189) in found in running rom when we do a rom upgrade.
- When gpg2 starts, it checks to see if newer format db is present (pubring.kbx) and if not found, creates it. It doesn't look for pubring.gpg at all anymore.

Actually implemented:
1. Device reownership generates a [otrust.txt file](https://github.com/osresearch/heads/pull/511/files#diff-9a66e05586b37d4779ee1eb88802e759R611) and [injects it in rom](https://github.com/osresearch/heads/pull/511/files#diff-9a66e05586b37d4779ee1eb88802e759R625) prior to reflashing it. This works, but adds a new unneeded addtional file, since we currently have trustdb.gpg but aren't using it.

Proposition:
1. We could use trustdb.gpg to ultimately trust generated key in factory reset. At this point, generated trustdb.gpg knows to ultimately trust the key we generated. We could inject the trustdb instead of otrust.txt, and the next instanciation of gpg --homedir /.gnupg will load it. 
 1. That trusting could happen in key-init and be applied to all public keys present in the rom to remove warnings about untrusted keys (iso signing keys included!). That would require to utilately trust each of those public key (`gpg --edit-key keyid`, `trust 5`, `gpg --update-trustdb`)
1. We could then remove otrust export and import code.
1. If pubring.kbx is found, delete pubring.gpg and not reimport it. (migration done.)

What do you think?
We could still merge this PR as is.

kylerankin:
I'd prefer separating any factory-reset dependencies from just having gpg2 be implemented in Heads. Then changes we'd need to make so your factory-reset workflow works would be limited to those PRs.

tlaurion:
@kylerankin : ok. Will seperate those.

tlaurion:
@kylerankin actually, this is already in this state for the current PR.

I would add some changes i've done so that otrust.txt export at key generation is not needed anymore, the trsuting being applied when the armored gpg key is imported.

tlaurion:
@jandryuk @flammit @jandryuk : please comment the code ASAP so we can merge ASAP.
Thanks.

PR #507, opened 2019-01-08
OEM-> User transfer of devices ownership wizard, triggered by the presence of an empty /boot/oem file

tlaurion:
UPDATE: see [QubesOS OEM to user transfer of ownership](https://github.com/osresearch/heads/pull/507#issuecomment-455814228) below for current state.

Advanced menu addition:
GPG card: factory reset, set key-attr to 4096bits, set passwords, generate keys, remove old keys from rom and insert public and trustdb back, and flash it.
Cryptsetup-reencrypt: reencrypt device and change recovery disk passphrase.

Please review. This is first draft. Everything works but it could be more beautiful and menus should probably be seperated.

@kyle: I was thinking of a oem file being present under /boot to engage a wizard or something.

PR #511, opened 2019-01-26
User reownership of OEM shipped hardware wizard, based on /boot/oem file presence

### LINKED ISSUE #475: Create Whiptail menu to guide through OEM device ownership
@kylerankin : I'll put it in my pile by if it is easy for you, don't hesitate!

tlaurion:
EDIT: I will update this PR once the [staging branch](https://github.com/tlaurion/heads/tree/oem_reownership-staging) is considered stable. Staging builds can be downloaded [here](https://gitlab.com/tlaurion/heads-ci/-/jobs).

Straightened #507 replacement. 
Depends on #510. Fixes #475.

Modifies [this PR (Empty Keyring Detection](https://github.com/kylerankin/heads/pull/2)) to interact directly with /boot/oem file, dropped in place by the OEM once TOTP/HOTP firmware is sealed and /boot files are signed with OEM's owned LibremKey/NitroKey Pro v2. The OEM then ships the devices separately. 

As a result, QubesOS or any other Linux distribution can be preinstalled and firmware and boot integrity is attested at first boot by the user received OEM hardware. He then reowns the hardware himself by reencrypting encrypted drive (LUKS container), reowning the TPM, reowning LibremKey/Nitrokey Pro v2, and recreates /boot checksums and sign them with his newly reowned GPG card prior to first booting preinstalled OS.


tlaurion:
@flammit @kylerankin : Here is how the actual ownership looks like from a UX perspective.

The OEM:
- Flashes a clean rom.
- Factory resets LibremKey/Nitrokey Pro v2
- Seals TOTP/HOTP firmware
- Installs OS, makes sure NitroKey/LibremKey is not connected in second stage of QubesOS installation (else QubesOS detects the Nitrokey as a USB keyboard and refuses to offer sys-usb VM creation).
- Selects a default boot option, create checksum file of /boot files, signs them with GPG card
- Finally, open recovery shell and `mount -o remount,rw /boot && touch /boot/oem && mount -o remount,ro /boot && reboot`
- Ships Hardware and GPG card seperatly to the user.

Once the user receives his hardware, he plugs the GPG card in the hardware and boots. He gets the following:
- LibremKey/Nitrokey LED should flash green:


- And then, the hardware reownership begins:
























- The hardware reownership finishes here. His LibremKey/Nitrokey flashes green:

marmarek:
Is it correct that attacker that intercept both laptop and nitrokey could forge integrity protection? If so, it may be worth adding (optional) extra measurement - for example let the user send their public key and provide back encrypted HOTP/TOTP secret, so it would be possible to verify platform integrity with yet another tool. Since the secret is reset during re-ownership process, it shouldn't be such a big problem that it ever leaved measured environment (of course as long as attacker wouldn't be able to extract it too).

Another thing to verify here - OS (especially unverified one), even when booted with unmodified firmware, should not be able to extract TOTP/HTOP secret. This apply to both RAM (all copies of the secret should be wiped from RAM before booting OS) and also unsealing it again from TPM. One way to achieve the latter would be extending relevant PCR with xen+kernel(+initrd?). Maybe it's already there?

andrewdavidwong:
> Is it correct that attacker that intercept both laptop and nitrokey could forge integrity protection? If so, it may be worth adding (optional) extra measurement - for example let the user send their public key and provide back encrypted HOTP/TOTP secret, so it would be possible to verify platform integrity with yet another tool. Since the secret is reset during re-ownership process, it shouldn't be such a big problem that it ever leaved measured environment (of course as long as attacker wouldn't be able to extract it too).
> 
> Another thing to verify here - OS (especially unverified one), even when booted with unmodified firmware, should not be able to extract TOTP/HTOP secret. This apply to both RAM (all copies of the secret should be wiped from RAM before booting OS) and also unsealing it again from TPM. One way to achieve the latter would be extending relevant PCR with xen+kernel(+initrd?). Maybe it's already there?

I really like this suggestion.

Are there step-by-step end user instructions that customers would follow? I'd like to put myself in the shoes of a security-minded potential customer and go through each step of the process to see whether there are any red flags from that point of view. (E.g., "This step leaves me vulnerable to interdiction," or "This step requires unnecessary trust in a third-party." It could be that there are already security measures in place to guard against these things, but potential customers may not be aware of them or understand them.)

tlaurion:
@marmarek First of all, sorry for the delay answering.

So we talk about a couple of things here that should be addressed separately as necessary.
[General information](https://github.com/osresearch/heads-wiki/blob/master/Keys.md)
1. Integrity protection/measure:
   1. TPM firmware integrity measurements and sealing through TPMTOTP
   1. Nitrokey Pro V2/LibremKey firmware integrity measurements and sealing with HOTP on GPG Card
   1. TPM firmware integrity measurements, TPM sealing of LUKS header dump + user defined disk unlock password, stored in LUKS slot 1
1. Secret creation and storage
   1. TPMTOTP secret generation and storage
   1. HOTP secret generation and storage


> Is it correct that attacker that intercept both laptop and nitrokey could forge integrity protection? If so, it may be worth adding (optional) extra measurement - for example let the user send their public key and provide back encrypted HOTP/TOTP secret, so it would be possible to verify platform integrity with yet another tool. Since the secret is reset during re-ownership process, it shouldn't be such a big problem that it ever leaved measured environment (of course as long as attacker wouldn't be able to extract it too).

The Firmware integrity/boot configuration cannot be compromised without the user knowing.
The user is provided with TPMTOTP QR Code through user chosen secure communication channel by the OEM, as are GPG card user's and admin passwords, TPM owner's password and disk recovery password associated with cryptsetup slot0 password change after cryptsetup-reencrypt has reencrypted the drive. This TOTP is generated based on [these measurements](https://github.com/osresearch/heads-wiki/blob/master/Keys.md#tpm-pcrs). As a result, the user can validate upon reception that the 6 digits code showed on screen is the on his chosen smartphone OTP application.

So even if both the LibremKey and firmware were compromised (we will get into that later), the user can still attest integrity by the laptop alone and it's TPM.

The HOTP seal is implemented in [TPMTOTP](https://github.com/osresearch/tpmtotp) and depends on the TPM measurements. [On hardware that doesn't provide a TPM](https://github.com/osresearch/heads/pull/493) (which is not the case here by board configuration) the HOTP measurement is _currently_ based on checksumming an internal SPI dump. To make it short, HOTP is just an easier way of confirming integrity by inserting user's GPG Card and validating green LED flashing instead of manually validating the TOTP code on his phone. Additionally, when the user sets a disk unlock passphrase, TPMTOTP attestation is done without him knowing, since the integrity measurements need to be valid to release the [key to be used as an additional initrd file](https://github.com/osresearch/heads/blob/584c07042ef4898de52ba4981997211c646b75fd/initrd/bin/kexec-insert-key#L43).

As for the possibility of compromising both the firmware and the GPG card, it is near impossible and don't see how it could be possible without user knowing.

The OEM's public GPG key inserted into the ROM is generated on the GPG card and is bound to it for signing operations. That public key fingerprint is shown to the user at each boot. Factory resetting the key would wipe the keys present in the card and reinject a public key different then the one provided by OEM. Any modification to the firmware would change the PCRs measurements and invalidate HOTP/TOTP integrity measurements, and the [TOTP](https://github.com/osresearch/heads/blob/master/initrd/bin/unseal-totp) and [HOTP](https://github.com/osresearch/heads/blob/master/initrd/bin/unseal-hotp) unsealing would fail. The user has only 3 bad attempts before the card locks user's/admin's individual PIN operations from the card, after which he would have to factory reset the card, which would change inserted user's public key in the rom, invalidate[ files signature validation](https://github.com/osresearch/heads/blob/aef5753373bf38854f24f0c88c6e42c627987ab5/initrd/bin/kexec-select-boot#L49) and inform the user of compromise.
> 
> Another thing to verify here - OS (especially unverified one), even when booted with unmodified firmware, should not be able to extract TOTP/HTOP secret. This apply to both RAM (all copies of the secret should be wiped from RAM before booting OS) and also unsealing it again from TPM. One way to achieve the latter would be extending relevant PCR with xen+kernel(+initrd?). Maybe it's already there?

The sealing of HOTP/TOTPM is an operation that requires to reboot. Every time the user attempts to enter recovery, [files are deleted](https://github.com/osresearch/heads/blob/master/initrd/etc/functions#L18). They could be shredded I guess, since the ramfs is not journalized. You think it would be required?

tlaurion:
I wll update[ this branch](https://github.com/tlaurion/heads/tree/oem_reownership-staging) once the staging is considered stable.

tlaurion:
> > Is it correct that attacker that intercept both laptop and nitrokey could forge integrity protection? If so, it may be worth adding (optional) extra measurement - for example let the user send their public key and provide back encrypted HOTP/TOTP secret, so it would be possible to verify platform integrity with yet another tool. Since the secret is reset during re-ownership process, it shouldn't be such a big problem that it ever leaved measured environment (of course as long as attacker wouldn't be able to extract it too).
> > Another thing to verify here - OS (especially unverified one), even when booted with unmodified firmware, should not be able to extract TOTP/HTOP secret. This apply to both RAM (all copies of the secret should be wiped from RAM before booting OS) and also unsealing it again from TPM. One way to achieve the latter would be extending relevant PCR with xen+kernel(+initrd?). Maybe it's already there?
> 
> I really like this suggestion.
> 
> Are there step-by-step end user instructions that customers would follow? I'd like to put myself in the shoes of a security-minded potential customer and go through each step of the process to see whether there are any red flags from that point of view. (E.g., "This step leaves me vulnerable to interdiction," or "This step requires unnecessary trust in a third-party." It could be that there are already security measures in place to guard against these things, but potential customers may not be aware of them or understand them.)

@andrewdavidwong : The user is guided through a wizard for reownership. I could definitely do screenshots in a guide if that is considered required.

marmarek:
> The user is provided with TPMTOTP QR Code through user chosen secure communication channel by the OEM, as are GPG card user's and admin passwords

Ok, this was the part I missed. Looks good.

> The sealing of HOTP/TOTPM is an operation that requires to reboot. Every time the user attempts to enter recovery, [files are deleted](https://github.com/osresearch/heads/blob/master/initrd/etc/functions#L18). They could be shredded I guess, since the ramfs is not journalized. You think it would be required?

Maybe, depending how ramfs (tmpfs?) implements file removal, but I guess the actual content stays in RAM, until used for something else.

And the same before booting the OS (especially the one from USB).

andrewdavidwong:
> > > Is it correct that attacker that intercept both laptop and nitrokey could forge integrity protection? If so, it may be worth adding (optional) extra measurement - for example let the user send their public key and provide back encrypted HOTP/TOTP secret, so it would be possible to verify platform integrity with yet another tool. Since the secret is reset during re-ownership process, it shouldn't be such a big problem that it ever leaved measured environment (of course as long as attacker wouldn't be able to extract it too).
> > > Another thing to verify here - OS (especially unverified one), even when booted with unmodified firmware, should not be able to extract TOTP/HTOP secret. This apply to both RAM (all copies of the secret should be wiped from RAM before booting OS) and also unsealing it again from TPM. One way to achieve the latter would be extending relevant PCR with xen+kernel(+initrd?). Maybe it's already there?
> > 
> > 
> > I really like this suggestion.
> > Are there step-by-step end user instructions that customers would follow? I'd like to put myself in the shoes of a security-minded potential customer and go through each step of the process to see whether there are any red flags from that point of view. (E.g., "This step leaves me vulnerable to interdiction," or "This step requires unnecessary trust in a third-party." It could be that there are already security measures in place to guard against these things, but potential customers may not be aware of them or understand them.)
> 
> @andrewdavidwong : The user is guided through a wizard for reownership. I could definitely do screenshots in a guide if that is considered required.

As a potential customer, I would consider it required. I think that most customers will want to know what to expect and to understand what they're actually buying.

I think that the guide should begin when the user actually places the order or, at least, before the package is received. It would be nice to have some assurance that the entire package wasn't replaced with a malicious one, for example.

tlaurion:
> > The user is provided with TPMTOTP QR Code through user chosen secure communication channel by the OEM, as are GPG card user's and admin passwords
> 
> Ok, this was the part I missed. Looks good.
> 
> > The sealing of HOTP/TOTPM is an operation that requires to reboot. Every time the user attempts to enter recovery, [files are deleted](https://github.com/osresearch/heads/blob/master/initrd/etc/functions#L18). They could be shredded I guess, since the ramfs is not journalized. You think it would be required?
> 
> Maybe, depending how ramfs (tmpfs?) implements file removal, but I guess the actual content stays in RAM, until used for something else.

From what I get, [rootfs](https://wiki.debian.org/rootfs), is an instance of [tmpfs](https://www.kernel.org/doc/Documentation/filesystems/ramfs-rootfs-initramfs.txt) by default:

> What is rootfs?
> ---------------
> 
> Rootfs is a special instance of ramfs (or tmpfs, if that's enabled), which is
> always present in 2.6 systems.  You can't unmount rootfs for approximately the
> same reason you can't kill the init process; rather than having special code
> to check for and handle an empty list, it's smaller and simpler for the kernel
> to just make sure certain lists can't become empty.
> 
> Most systems just mount another filesystem over rootfs and ignore it.  The
> amount of space an empty instance of ramfs takes up is tiny.
> 
> If CONFIG_TMPFS is enabled, rootfs will use tmpfs instead of ramfs by
> default.  To force ramfs, add "rootfstype=ramfs" to the kernel command
> line.

So in our case, `CONFIG_TMPFS=y` is the default, while not visible in savedefconfig, since implicit.

```
mount
rootfs on / type roots (rw,size=4057276k,nr_inodes=1014319)
```

So the secrets of concern are the one happening when sealing/unsealing:
```
user@build-x230-coreboot-48:~/heads$ grep -Rn "/tmp/secret" initrd/
initrd/etc/functions:18:	rm -rf /tmp/secret
initrd/etc/functions:19:	mkdir -p /tmp/secret
grep: initrd/etc/mtab: No such file or directory
Binary file initrd/etc/.functions.swp matches
initrd/bin/kexec-seal-key:9:KEY_FILE="/tmp/secret/secret.key"
initrd/bin/kexec-seal-key:10:TPM_SEALED="/tmp/secret/secret.sealed"
initrd/bin/kexec-seal-key:11:RECOVERY_KEY="/tmp/secret/recovery.key"
initrd/bin/kexec-insert-key:36:INITRD_DIR=/tmp/secret/initrd
initrd/bin/kexec-insert-key:37:SECRET_CPIO=/tmp/secret/initrd.cpio
initrd/bin/unseal-hotp:6:HOTP_SEALED="/tmp/secret/hotp.sealed"
initrd/bin/unseal-hotp:7:HOTP_SECRET="/tmp/secret/hotp.key"
initrd/bin/unseal-totp:6:TOTP_SEALED="/tmp/secret/totp.sealed"
initrd/bin/unseal-totp:7:TOTP_SECRET="/tmp/secret/totp.key"
initrd/bin/kexec-select-boot:305:			-a "$add" -r "$remove" -o "/tmp/secret/initrd.cpio" \
initrd/bin/kexec-unseal-key:11:mkdir -p /tmp/secret
initrd/bin/kexec-unseal-key:13:sealed_file="/tmp/secret/sealed.key"
initrd/bin/kexec-unseal-key:17:	key_file="/tmp/secret/secret.key"
initrd/bin/kexec-unseal-key:41:		rm -f /tmp/secret/sealed || true
initrd/bin/seal-totp:17:TOTP_SECRET="/tmp/secret/totp.key"
initrd/bin/seal-totp:18:TOTP_SEALED="/tmp/secret/totp.sealed"
initrd/bin/seal-libremkey:6:HOTP_SEALED="/tmp/secret/hotp.sealed"
initrd/bin/seal-libremkey:7:HOTP_SECRET="/tmp/secret/hotp.key"
initrd/init:30:mkdir -p /tmp/secret
```

Since we use tmpfs, it seems that using shred would [do the trick](https://security.stackexchange.com/questions/135822/live-os-secure-file-deletion), and is compiled in [busybox](https://github.com/osresearch/heads/blob/master/config/busybox.config#L282).

So basically, it would be to replace `rm` by `shred` here:
```
user@build-x230-coreboot-48:~/heads$ grep -R "/tmp/secret" initrd/|awk -F ":" {'print $1'}|sort|uniq|while read file; do grep -Hn -e "rm " $file 2>&1|grep -i secret;done 
grep: initrd/etc/mtab: No such file or directory
initrd/bin/kexec-unseal-key:41:		rm -f /tmp/secret/sealed || true
initrd/bin/seal-libremkey:33:rm -f "$HOTP_SECRET"
initrd/bin/seal-totp:46:	rm -f "$TOTP_SECRET"
initrd/bin/unseal-hotp:54:	rm -f "$HOTP_SECRET"
initrd/bin/unseal-hotp:58:rm -f "$HOTP_SECRET"
initrd/bin/unseal-totp:24:	rm -f "$TOTP_SECRET"
initrd/bin/unseal-totp:28:rm -f "$TOTP_SECRET"
initrd/etc/functions:18:	rm -rf /tmp/secret
```

> 
> And the same before booting the OS (especially the one from USB).

richpersaud:
(design considerations for future work on platform security lifecycle management, please let me know if there's a better place to post this)

There are supply chain management and support scenarios where an enterprise customer with geographically distributed offices would like new devices to ship from the OEM to a local support consultant that is trusted to install or replace a device at their local office. This reduces the need for IT staff (who may not exist at a small office) to provision devices.  It also provides a business model for local consultants to develop expertise in platform security for customized enterprise workflows.  Topics to consider:

1. OEMs can ship multiple images (e.g. by geo) and versions.  Could there be a public data source which matches image identity with Bill of Materials, including software components that may have been reproducibly built?  This would allow customers to privately or publicly share risk and threat data associated with each image.

2. An enterprise could require platform image customizations, e.g. configs or mgmt VM changes to be made by their trusted consultant.  When a new device is connected to the local office of an enterprise network for provisioning, how can the enterprise prove that the base image originated from the OEM, that the management VM was blessed by the enterprise and the device was customized by their trusted consultant?

3.  During on-site device provisioning, how are device identity keys generated and installed into the device's management VM, to migrate admin authority from the OEM or support consultant to the enterprise's management server?

4. When a device is in service and the customer requires support, can they place an L1 support call to their trusted local consultant, who is granted temporary device access (via enterprise management server) for troubleshooting?  When purchasing their L1 support contract from the local consultant, can they pay an additional fee for L2 support from one or more software vendors which produced software components (e.g. VMs) in the device image?  This enables their trusted local consultant to work directly with software vendors to address local device issues, improving future software versions to address operational support requirements.

5.  When a device is taken out of service, how does the enterprise management server, local IT admin, or trusted support consultant ensure that device keys, config and data are safely removed before the device is recycled or resold?  Key ownership lifecycle has been a topic of discussion at OCP (Open Compute Platform) in the context of hardware and firmware root of trust, to enable OCP servers from hyperscale cloud providers to be resold on the secondary market, rather than incurring the environmental cost of device destruction.

tlaurion:
> (design considerations for future work on platform security lifecycle management, please let me know if there's a better place to post this)
> 
> There are supply chain management and support scenarios where an enterprise customer with geographically distributed offices would like new devices to ship from the OEM to a local support consultant that is trusted to install or replace a device at their local office. This reduces the need for IT staff (who may not exist at a small office) to provision devices. It also provides a business model for local consultants to develop expertise in platform security for customized enterprise workflows. Topics to consider:
> 
>     1. OEMs can ship multiple images (e.g. by geo) and versions.  Could there be a public data source which matches image identity with Bill of Materials, including software components that may have been reproducibly built?  This would allow customers to privately or publicly share risk and threat data associated with each image.
> 
>     2. An enterprise could require platform image customizations, e.g. configs or mgmt VM changes to be made by their trusted consultant.  When a new device is connected to the local office of an enterprise network for provisioning, how can the enterprise prove that the base image originated from the OEM, that the management VM was blessed by the enterprise and the device was customized by their trusted consultant?
> 
>     3. During on-site device provisioning, how are device identity keys generated and installed into the device's management VM, to migrate admin authority from the OEM or support consultant to the enterprise's management server?
> 
>     4. When a device is in service and the customer requires support, can they place an L1 support call to their trusted local consultant, who is granted temporary device access (via enterprise management server) for troubleshooting?  When purchasing their L1 support contract from the local consultant, can they pay an additional fee for L2 support from one or more software vendors which produced software components (e.g. VMs) in the device image?  This enables their trusted local consultant to work directly with software vendors to address local device issues, improving future software versions to address operational support requirements.
> 
>     5. When a device is taken out of service, how does the enterprise management server, local IT admin, or trusted support consultant ensure that device keys, config and data are safely removed before the device is recycled or resold?  Key ownership lifecycle has been a topic of discussion at OCP (Open Compute Platform) in the context of hardware and firmware root of trust, to enable OCP servers from hyperscale cloud providers to be resold on the secondary market, rather than incurring the environmental cost of device destruction.

@richpersaud Those are all really interesting questions that all seem more linked to QubesOS then Heads and linked to this particular pull request.

One thing I tried to push to address most of your points is to [implement an additional remotely manageable AdminVM.](https://github.com/SkypLabs/my-qubes-os-formula/issues/6). The idea behind this is that an AdminVM would have permissions over a really limited and precise scope (deploy additional templates and create and manage additonal qubes, not the one deployed and managed by dom0).


It's not possible [right now](https://github.com/SkypLabs/my-qubes-os-formula/issues/6), so the remote manageability would give dom0 access and total trust from the user to a 3rd party.

You are more then welcome to open that ticket on [QubesOS github issues tracker](https://github.com/QubesOS/qubes-issues/)

richpersaud:
> One thing I tried to push to address most of your points is to [implement an additional remotely manageable AdminVM.](https://github.com/SkypLabs/my-qubes-os-formula/issues/6). The idea behind this is that an AdminVM would have permissions over a really limited and precise scope (deploy additional templates and create and manage additional qubes that solely depend on those templates, not the one deployed and managed by dom0).
> 
> It's not possible [right now](https://github.com/SkypLabs/my-qubes-os-formula/issues/6), so the remote manageability would give dom0 access and total trust from the user to a 3rd party.

@tlaurion Thanks for the helpful pointers.  We did some preliminary work in http://OpenXT.org to support multiple admin VMs: one per "virtual platform" = group of VMs and related policy, with a dedicated VPN or network path to the remote mgmt server.  E.g. OpenXT has a file transfer mechanism (ICBINN) to allow an admin VM to deploy multi-gigabyte virtual disks to dom0.  The access control pieces are not yet implemented to isolate admin VMs.  I'll look at SaltStack and QubesOS admin VMs to see if there is overlap in goals/approach.

richpersaud:
> with all dom0 access that comes with it; requiring a total trust to a 3rd party from the user. 

QubesOS, XenClient, Moka5, Virtual Computer, OpenXT, Bromium, Hysolate and Windows Sandbox enabled multi-VM endpoints.  The next step is multi-MDM (mobile device management).

Microsoft 365 is moving towards a single subscription price that covers the device hardware lease, Windows license, Office license and security updates.  Microsoft Azure Sphere IoT device purchase price includes a license to 10 years of *Linux* security updates, i.e. remote OS management even if the device/app vendor goes out of business.

The boundaries between OEM [firmware](https://www.reddit.com/r/linuxhardware/comments/aohc8x/some_system76_hardware_beginning_to_see_coreboot/), OEM factory image, OS vendor updates, OS-mandated app store and owner-managed enterprise apps are evolving.  Users are one remote update away from a workflow-impacting (feature, security) change.  As supply chain complexity increases, these remote parties will need to be isolated, granted minimal permission, and monitored by the owners of endpoint devices.

tlaurion:
Will make a PR including just the changes related to the reownership process and force push.

tlaurion:
replaced by https://github.com/osresearch/heads/pull/551

PR #518, opened 2019-02-09
GPG2 branch required sed

tlaurion:
sed is required to import owner trust information in trustdb when keys are imported into pubkey.kbx/pubkey.gpg


tlaurion:
That is required [here](https://github.com/osresearch/heads/blob/master/initrd/bin/key-init#L13)

tlaurion:
Merging. It's required for the gpg2 branch to be fully functional.

PR #521, opened 2019-02-11
module/u-root: added build parameter for uroot template

flammit:
Instead of using UROOT_CMDS too specify the packages to include
use UROOT_TEMPLATE=minimal or all, etc.

### DIFF
--- a/modules/u-root
+++ b/modules/u-root
@@ -7,6 +7,7 @@
 u-root_url := github.com/u-root/u-root
 u-root_output := $(build)/$(BOARD)/u-root.cpio
 UROOT_CMDS ?=
+UROOT_TEMPLATE ?=
 
 export GOPATH=$(build)/go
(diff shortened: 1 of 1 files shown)


hugelgupf:
Templates = cmds in u-root. Templates are just textual expansion, so `UROOT_CMDS=minimal` should work without this patch

flammit:
As per slack, right now the Heads u-root module makefile appends the full package path to each command in UROOT_CMDS:

```u-root_src_cmds := $(foreach cmd,$(UROOT_CMDS),github.com/u-root/u-root/cmds/$(cmd))```

so we need another option to deliver unmodified arguments to the u-root build.

tlaurion:
@hugelgupf : mergeable?

tlaurion:
@flammit ?

tlaurion:
@hugelgupf really unaware at this point if linuxboot can be even built on top of coreboot inside of Heads. Is it supposed to?

PR #522, opened 2019-02-11
Add GPG GUI menu

kylerankin:
It makes more logical sense for GPG functions to be split out into their
    own menu instead of being part of the "Flash" menu. This creates a
    gpg-gui.sh script and moves GPG options there while adding a few
    additional features (like listing keys and initial smartcard key
    generation support).


initrd/bin/gpg-gui.sh
```diff
@@ -0,0 +1,244 @@
+#!/bin/sh
+#
+set -e -o pipefail
```
tlaurion: The effect of this is that once one of the command in a pipe fails, the whole line fails.

kylerankin:
@tlaurion I've removed the pipefail statement from gpg-gui.sh.

tlaurion:
@kylerankin : It was my error. You did checks on file presence, which doesn't cause any problems. My attempts to do a while loop on the non presence of a match string was causing the issue. 

Setting pipefail is debattable, but was not causing any issue in your PR.
If you want to put it back in, there is no problem.

Please rebase on master so I can merge. :)

initrd/bin/gpg-gui.sh
```diff
@@ -1,6 +1,5 @@
 #!/bin/sh
 #
-set -e -o pipefail
```
tlaurion: Please add it back in.

kylerankin:
Changed back.

PR #526, opened 2019-02-16
Use shred instead of rm on secret related files.

tlaurion:
Replacing rm with shred for secrets that should not be present in ram when booting OS.
Address the following [concerns](https://github.com/osresearch/heads/pull/511#issuecomment-464363619)

### DIFF
--- a/initrd/bin/kexec-seal-key
+++ b/initrd/bin/kexec-seal-key
@@ -117,7 +117,7 @@ tpm sealfile2 \
 	-ix 7 X \
 || die "Unable to seal secret"
 
-rm -f "$KEY_FILE" \
+shred -n 10 -z -u "$KEY_FILE" 2> /dev/null \
 || die "Failed to delete key file"
 
 # try it without the owner password first
--- a/initrd/bin/kexec-seal-key
+++ b/initrd/bin/kexec-seal-key
@@ -150,5 +150,5 @@ if ! tpm nv_writevalue \
 	|| die "Unable to write sealed secret to NVRAM"
 fi
 
-rm "$TPM_SEALED" \
+shred -n 10 -z -u "$TPM_SEALED" 2> /dev/null \
 || warn "Failed to delete the sealed secret - continuing"
--- a/initrd/bin/kexec-unseal-key
+++ b/initrd/bin/kexec-unseal-key
@@ -38,7 +38,7 @@ for tries in 1 2 3; do
 		-hk 40000000 \
 	; then
 		# should be okay if this fails
-		rm -f /tmp/secret/sealed || true
+		shred -n 10 -z -u /tmp/secret/sealed || true
 		exit 0
 	fi
 
--- a/initrd/bin/kexec-unseal-key
+++ b/initrd/bin/kexec-unseal-key
@@ -38,7 +38,7 @@ for tries in 1 2 3; do
 		-hk 40000000 \
 	; then
 		# should be okay if this fails
-		shred -n 10 -z -u /tmp/secret/sealed || true
+		shred -n 10 -z -u /tmp/secret/sealed 2> /dev/null || true
 		exit 0
 	fi
 
--- a/initrd/bin/seal-libremkey
+++ b/initrd/bin/seal-libremkey
@@ -28,9 +28,9 @@ tpm unsealfile  \
 	-of "$HOTP_SECRET" \
 || die "Unable to unseal HOTP secret"
 
-rm -f "$HOTP_SEALED"
+shred -n 10 -z -u "$HOTP_SEALED"
 secret="`cat $HOTP_SECRET`"
-rm -f "$HOTP_SECRET"
+shred -n 10 -z -u "$HOTP_SECRET"
 
 # Store counter in file instead of TPM for now, as it conflicts with Heads
 # config TPM counter as TPM 1.2 can only increment one counter between reboots
--- a/initrd/bin/seal-libremkey
+++ b/initrd/bin/seal-libremkey
@@ -28,9 +28,9 @@ tpm unsealfile  \
 	-of "$HOTP_SECRET" \
 || die "Unable to unseal HOTP secret"
 
-shred -n 10 -z -u "$HOTP_SEALED"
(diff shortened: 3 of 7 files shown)


tlaurion:
@mamarek: would that adress your concerns?

A critic that was done over [slack](https://u-root.slack.com/archives/C92MNSRC1/p1550684593002800) was that  the key is injected into initrd that is passed to kexec, so an attack [like this](https://0x00sec.org/t/breaking-encryption-hashed-passwords-luks-devices/811) could still be done from dom0, which means game over anyway.

marmarek:
I don't have access to that slack channel, but (verified!) dom0 will have access to LUKS key anyway. But for that to happen, one need to run 1) unmodified firmware 2) provide luks passphrase (actually passphrase to unlock luks key). The key is not loaded into RAM unless both happen, which (hopefully) mean it's legitimate owner accessing the machine.

My concern is HOTP/TOTP key, which is loaded into RAM _before_ the user provide any secret. And it is also the thing authenticating machine to the user, so if it got stolen, attacker could convince the user to provide the password (into now backdoored firmware). So we should ensure that this secret doesn't leak anywhere. Yes, coldboot style attacks are probably still possible, but are much harder to perform (especially if you control the firmware), than just booting random Linux using built-in menu and dumping RAM from there.

As for the actual PR, yes it should be good. I'm not sure if `-n 10` makes sense on RAM (this makes sense mostly on magnetic discs), but also shouldn't hurt.

tlaurion:
> My concern is HOTP/TOTP key, which is loaded into RAM _before_ the user provide any secret. And it is also the thing authenticating machine to the user, so if it got stolen, attacker could convince the user to provide the password (into now backdoored firmware). So we should ensure that this secret doesn't leak anywhere. Yes, coldboot style attacks are probably still possible, but are much harder to perform (especially if you control the firmware), than just booting random Linux using built-in menu and dumping RAM from there.

Let's go high level first on TOTP/HTOP interaction. Exerpt from [here](https://puri.sm/posts/the-librem-key-makes-tamper-detection-easy/) with some comments inline: 

>  Instead of doing away with the TOTP code we are actually using the same exact TPM secret, just in addition to converting it into a TOTP code, we are also combining it with the incrementing counter to generate an HOTP code too. So when you tell Heads that you want to seal a new TOTP secret, if your version of Heads has Librem Key support enabled, in addition to showing you a QR code, it will also prompt you to insert your Librem Key and set up the secret there as well using the libremkey_hotp_initialize program.

The vulnerable part would be when the [QrCode is generated](https://github.com/osresearch/heads/blob/d54727434e7538cf340f6e844d5bc476496d0298/initrd/bin/seal-totp#L87) and when the [HOTP secret is sealed into the Nitrokey Pro v2/LibremKey](https://github.com/osresearch/heads/blob/d54727434e7538cf340f6e844d5bc476496d0298/initrd/bin/seal-libremkey#L64). Those functions are called only when PCRs mismatch and require the user to type TPM/GPG card Admin PIN, a reboot is required after this step.
> 
> The next time you boot, in addition to displaying the TOTP code in the menu as always, it also attempts to communicate with your Librem Key using the libremkey_hotp_verification tool. If the Librem Key isn’t inserted, it gets a specific error and warns the user that the key isn’t plugged in, but it doesn’t stop the user from booting. If the user then plugs in the key and tells Heads to regenerate the code, it can do a second test from the GUI. If you lose your Librem Key or leave it behind somewhere, you can always just fall back to the standard TOTP code + phone approach until you have a new Librem Key to enroll.

The TOTP code generation happens [here](https://github.com/osresearch/heads/blob/d54727434e7538cf340f6e844d5bc476496d0298/initrd/bin/unseal-totp) while the HOTP happens [here](https://github.com/osresearch/heads/blob/d54727434e7538cf340f6e844d5bc476496d0298/initrd/bin/unseal-hotp). They both depends on the TPM, and the secret could be grabbed only if the firmware was modified, which the user would be aware of and asked to[ regenerate TOTP/HOTP secret](https://github.com/tlaurion/heads/blob/oem_reownership-staging/initrd/bin/gui-init#L294).
> 
> After Heads sends the Librem Key an HOTP code, if the code matches what the Librem Key itself generates, it will flash a green LED and return a success code to Heads to display on the laptop screen. If the HOTP codes don’t match, the Librem Key will flash a red LED indefinitely and also send a specific error back to Heads which will cause it to show a red background and error dialog. Note that you shouldn’t trust the Heads UI to display this error, it’s only for convenience. A modified UI could lie to you so you should only trust the Librem Key LED at this phase of the boot process.

Heads sends HOTP code [here](https://github.com/tlaurion/heads/blob/oem_reownership-staging/initrd/bin/gui-init#L238). The TPMTOTP is used to unseal disk unlock key [here](https://github.com/tlaurion/heads/blob/oem_reownership-staging/initrd/bin/kexec-seal-key) only if PCRs matches, the passphrase typed would fail if the firmware was modified. So even if the user looses it's GPG card, doesn't check the TOTP numeric code on his phone, the system won't permit to unlock the disk with disk unlock key and will propose the user to unlock the disk with disk recovery key passphrase.

That is where we are now.
Any concerns, @marmarek ?

marmarek:
Sounds fine!

A little comment: [this shred](https://github.com/osresearch/heads/blob/d54727434e7538cf340f6e844d5bc476496d0298/initrd/bin/seal-totp#L46) should be done also in case of success. In fact, especially in case of success.

tlaurion:
> Sounds fine!
> 
> A little comment: [this shred](https://github.com/osresearch/heads/blob/d54727434e7538cf340f6e844d5bc476496d0298/initrd/bin/seal-totp#L46) should be done also in case of success. In fact, especially in case of success.

@marmarek: [applied](https://github.com/osresearch/heads/pull/526/files#diff-e081635c0c27156840dbde724b1fbf15R50). But again, HOTP/TOTP generation happens when firmware tampering is [detected](https://github.com/tlaurion/heads/blob/oem_reownership-staging/initrd/bin/gui-init#L294) (measurements don't match) and is followed by a forced reboot before being able to use the system.

PR #529, opened 2019-02-19
Added support for EXT2 (via EXT4 driver) and XFS as boot filesystems

cawilliamson:
This is clearly an opinionated pull request and I wont be too upset if it doesn't get merged but I thought since I was making the changes for my own uses - I'd throw a PR in anyway to at least have the discussion.

- There is very little (almost zero) overhead to including the EXT2 config option which adds support for EXT4 to mount and use EXT2 volumes perfectly which some folks prefer to use as their boot filesystem.

- I have also added XFS filesystem support to all configs which appears to already be there for the qemu images but no others. I have a strong preference for running it on my machines as much as possible and since it is now getting the support of RedHat (via Stratis - replacing BtrFS) I think it's a good time to add support for it.

In short:
- Adds support for EXT2 (via EXT4 driver for minimal ovehead)
- Adds support for XFS

I think the EXT* and XFS are the main preferred filesystems for enterprise these days and including BtrFS support would add a more substantial amount of overhead - especially just for boot partition support.


cawilliamson:
I now know why this wasn't included in the first place - XFS is too large to be included in the kernel. Makes sense.

I'll resubmit this for EXT2 support only.

PR #530, opened 2019-02-19
Add support for EXT2 (via the EXT4 driver)

cawilliamson:
This is a simple change to add support for EXT2 as a boot filesystem (provided by the EXT4 driver.)

This option adds no additional overhead since the EXT4 driver is already being built.


paulmenzel:
I’d prefer the commit message summary to use the verb in imperative mood:

> Add support for EXT2 (via the EXT4 driver)

The change does indeed only build the module `ext2`, which seems to be the same as ext4.

cawilliamson:
> I’d prefer the commit message summary to use the verb in imperative mood:
> 
> > Add support for EXT2 (via the EXT4 driver)
> 
> The change does indeed only build the module `ext2`, which seems to be the same as ext4.

I've now fixed both the commit message and PR title in-line with your request. :+1:

paulmenzel:
Perfect. Thank you very much. The people with commit right will chime in soon.

PR #546, opened 2019-04-10
Add GPG smartcard keygen feature

kylerankin:
This change updates the very basic GPG smartcard feature in the GPG GUI
so that it can properly support generating a key from within Heads. It
offers the user the option to copy the generated GPG public key to a USB
thumb drive so it's not lost as well as the option to reflash the
current Heads BIOS with this new public key added to the keyring.

I've moved the common functions required to flash a new ROM with GPG
changes into a shared function at the top of the script.

### DIFF
--- a/initrd/bin/gpg-gui.sh
+++ b/initrd/bin/gpg-gui.sh
@@ -67,6 +67,48 @@ file_selector() {
     exit 1
   fi
 }
+gpg_flash_rom() {
+  cat "$PUBKEY" | gpg --import
+  #update /.gnupg/trustdb.gpg to ultimately trust all user provided public keys
+  gpg --list-keys --fingerprint --with-colons |sed -E -n -e 's/^fpr:::::::::([0-9A-F]+):$/\1:6:/p' |gpg --import-ownertrust
+  gpg --update-trust
+
+  if (cbfs -o /tmp/gpg-gui.rom -l | grep -q "heads/initrd/.gnupg/pubring.kbx"); then
+    cbfs -o /tmp/gpg-gui.rom -d "heads/initrd/.gnupg/pubring.kbx"
+    if (cbfs -o /tmp/gpg-gui.rom -l | grep -q "heads/initrd/.gnupg/pubring.gpg"); then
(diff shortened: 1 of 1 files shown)


tlaurion:
@kylerankin 

Did:
admin
key-attr (asks for password 3x)
generate

Then followed steps. Generation was successful.
Then exited. The public key was not prompted to be inserted in rom.

[Why not go with a factory reset at this point, which results in the same outcome with a lot less complexity for the user](https://gitlab.com/tlaurion/heads/blob/oem_reownership-staging/initrd/bin/factory-reset-nitrokey-libremkey.sh)?

kylerankin:
@tlaurion I suspect you didn't get the prompt because your GPG session didn't exit with an error code of 0, maybe because you typed 'exit' instead of "quit"? I'm not sure.

I opted to go with key generation instead of a full factory reset, so as not to wipe out any HOTP secrets or other information that might be on the key, but I think adding a "factory reset" option to the main menu in gpg-gui.sh that calls your script would be useful for people who want that full reset.

Also, I hadn't reviewed your script yet so I wasn't aware I was reinventing the wheel!

tlaurion:
@kylerankin, that OEM file hijacks the rest of the process continue across multiple boots (gpg factory reset injects key, requires HOTP/TOTP reseal, cascades into Disk Unlock Key to be redone... reboot, TPM sealing, reboot. Then the reencrypt of disk happens, then the Disk Recovery Password and finally TPM reownership.)

It can't be done in a wizard without external state, and that file is unseen by signature validation until the file is deleted. Any concerns?

kylerankin:
@tlaurion fair enough, that's fine with me. I don't have any concerns (as long as it can be bypassed in case the deletion didn't work or the file shows up again accidentally).

tlaurion:
@kylerankin : ping me here when you merged your latest changes

kylerankin:
@tlaurion The "Add a "Factory reset" GPG option" change is pushed.

tlaurion:
@kylerankin: are we ok with both being present as mine in #511? Would be ready to merge.

kylerankin:
@tlaurion Yes I'm OK with both of my options *and* your option being present in the menu.

PR #548, opened 2019-04-15
Detect USB disk dynamically

kylerankin:
Currently Heads relies on a hard-coded config value to determine which
USB disk to mount. This can be problematic when trying to distribute a
pre-built version of Heads that can work on multiple disk
configurations. I've modified the USB mounting script so that it
attempts to detect all USB boot disks present on the system, pick sane
defaults, and prompt the user when there are multiple choices.

I've also removed the USB configuration option from config-gui.sh as
this config option is no longer used.


tlaurion:
@kylerankin : awesome idea. Will merge in my branch when i'm done with OEM. Will get rid of my code that was based on your config_updater that I used to go around that problem, and give you feedback.

PR #551, opened 2019-04-23
OEM reownership required changes for QubesOS certification of the X230

tlaurion:
The main goal behind this OEM Reownership Wizard is to be able to ship QubesOS or any other OS preinstalled into LUKS encrypted containers, without compromising user's confidentiality, privacy nor security, while enforcing tamper-evidence (provable security) of the firmware and /boot components in transit and at rest, without giving the OEM a possibility to hand those out to third parties. 

With Heads measured boot through TOTP/HOTP integrity seals, preinstalled QubesOS trusted hardware shipped with QubesOS can now become a reality. This PR is the last step before QubesOS certifies the PrivacyBeast x230 laptop, by Insurgo Technologies Libres / Open technologies (This is me! Hello! :) ) 

The OEM flashes Heads, [neutralizes ME](https://github.com/osresearch/heads-wiki/blob/master/Clean-the-ME-firmware.md) and reboots. He is asked to insert his public GPG key, which he does with his own dedicated deployment public key associated with his deployment Nitrokey/Librem Key. On reboot, he is asked to reown the TPM, which will trigger HOTP/TOTP resealing from a prompted menu. The OEM has to make sure that the system clock is good and correct GMT time accordingly before going further. (Make sure RTC clock battery is good).

He then boots pre-signed clonezilla or other ISO image with his deployment Nitrokey/Librem Key and reimages the HD with an up-to-date QubesOS disk image, signs boot config files and then fires the reownership wizard, wth a brand new Librem Key/Nitrokey conected, for that hardware secrets to be unique. To do so [he creates a /boot/oem file](https://github.com/osresearch/heads/pull/551/files#diff-d1ac0113778bfe516c8b663a9ebb4cf8R3), which when dropped into place invokes the OEM reownership wizard. 

That wizards looks for an existing [/media/oem-provisining](https://github.com/osresearch/heads/pull/551/files#diff-101665990fd0dc23df213e98668cebeb) file, that if present,  automagically provision Heads prompts to reencrypt LUKS container, change LUKS associated passphrase, factory resets connected  Librem Key/Nitrokey and define user and admin PINS injects associated public key into running rom, and ends by reowning the TPM. EDIT: if file is present but doesnt contain required secrets, the user is prompted to fill them with autogenerated passphrases, or he can manually select one, while the screen is filled with diceware words to choose from.

Then the OEM boots QubesOS, syncs system time. Then go into recovery, and finally [recreates the /boot/oem](https://github.com/osresearch/heads/pull/551/files#diff-d1ac0113778bfe516c8b663a9ebb4cf8R3) for the user to use the same process. That is the last step done by the OEM before shipping the devices separately/altogether. The next reboot invokes the reownership for the user to do the exact same steps described above when the hardware is received, with a a user provided /media/oem-provisioning file or manually entering secrets into prompts. In either way, the OEM needs to provide the Actual Disk Receovery Key by secured communication channels so the user can reown his computer upon reception.

[See video here](https://archive.org/details/oemuserreownership).

[The /media/oem-provisioning](https://github.com/osresearch/heads/pull/551/files#diff-101665990fd0dc23df213e98668cebeb) file can be [automatically generated by a script](https://github.com/osresearch/heads/pull/551/files#diff-a8f27ecaaec2f742c74c6f20e96fb0bd) to provision Heads required secrets from the OEM wizard. 

The OEM shares used oem-provision file to the user by secured communication channel, accompanied with a screen capture of the TOTP QR  code printed on the screen after TPM reset has been accomplished. The user can already scan that TPMTOTP code in his smartphone's OTP application to attest that the laptop has not been tampered with in transit, while thinking about the secrets he will feed to the new hardware when received, let it be manually or through /media/oem-provisioning.

The OEM reownership wizard explains the steps to the user through whiptail messages, the onboarding one being:
```
The received hardware is currently owned by $oem_name.
We will change that.

This Wizard will otherwise make you own your own hardware by:
1-Reencrypting encrypted drive content (LUKS container)
2-Changing LUKS Disk Recovery Key and it's passphrase
3-Resetting your Librem Key/Nitrokey Pro v2 to factory defaults
4-Setting User and Admin PINs into your Librem Key/NitroKey
5-Generating keys in it, exporting public key and
 importing it in the ROM prior to reprogramming the SPI flash with it
6-Taking integrity measures of all /boot files and sign the resulting
 checksum file with your Librem Key/Nitrokey Pro
7-Reowning your Trusted Platform Module (TPM)

Scan the QR Code with an OTP app only after Reset the TPM last phase.

```
Details of this PR. Evolution of #511 
- OEM mode is enforced on HOTP/TOTP sealed clean hardware.
- OEM is enforced by touching /boot/oem
- OEM can be provisioned automatically if /media/oem-provisioning exists
	which feeds all the secrets automatically.
- OEM oem-reownership/generate_diceware-eom-provisioning.sh script to generate oem-provisioning.generated secrets for both OEM and users. (EDIT: Still there but not usefull. Diceware generated secrets is now part of the standard workflow, while user can manually enter secrets that are validated for minimal required length, different for each of them with safe defaults)

OEM features:
- LUKS rencryption script added, which reencrypts and change Disk Recevory Key passphrase
- LibremKey/Nitrokey factory reset script added, which takes user input first and then change key size, generate keys, exports them and flash rom.
	- Todo: Reset code should be added later on. (To unlock Admin. Else, a DoS is possible on a USB GPG key)

Additions:
- Gitlab CI configuration to replicate builds based on fedora-29 container, making sure that each branch keeps his own seperate cache. This is the simpler way I've found so OEM can maintain their own board branches based on Head's git repo.
- Contains uniformaization of langage.
- Some shellcheck improved code
- LUKS Disk Unlock Key works properly
	- LVM volumes and disk devices are automatically detected
- GPG modules were retouched to remove invalid configure options which generated warnings
- LibremKey/Nitrokey is named everywhere for proper credit of partnership for HOTP
... And a lot of other smaller fixes... 

_Sent from my Galaxy S3 using [FastHub-Libre](https://github.com/thermatk/FastHub-Libre/)_


initrd/bin/config-gui.sh
```diff
@@ -55,6 +55,7 @@ while true; do
   whiptail --clear --title "Config Management Menu" \
     --menu "This menu lets you change settings for the current BIOS session.\n\nAll changes will revert after a reboot,\n\nunless you also save them to the running BIOS." 20 90 10 \
     'b' ' Change the /boot device' \
+    'u' ' Change the USB boot device' \
```
kylerankin: This was removed in the USB autodetection PR #548 you just merged.

tlaurion: just saw it at the same time. forced pushed

initrd/bin/kexec-save-default
```diff
@@ -33,106 +33,213 @@ TMP_MENU_FILE="/tmp/kexec/kexec_menu.txt"
...
-		if [ -n "$lvm_volume_group" ]; then
-			save_key_params="$save_key_params -l $lvm_volume_group $key_devices"
-		else
-			save_key_params="$save_key_params $key_devices"
-		fi
-		echo "Running kexec-save-key with params: $save_key_params"
-		kexec-save-key $save_key_params \
-			|| die "Failed to save the disk key"
-	fi
+  if [ ! -s "$KEY_DEVICES" ]; then
```
kylerankin: Does this section *have* to be run when someone is saving a default boot option? I'd much prefer we move TPM LUKS keys elsewhere so this doesn't confuse users who mistakenly select "yes" here to move along with the default boot process, but end up having problems down the road when the TPM gets reset.

Can this move to a separate function dedicated to this purpose or is this something that has to be set when we save the default boot options?

initrd/bin/kexec-select-boot
```diff
@@ -135,7 +135,7 @@ confirm_menu_option() {
 	if [ "$gui_menu" = "y" ]; then
 		whiptail --clear --title "Confirm boot details" \
 			--menu "Confirm the boot details for $name:\n\n$option\n\n" 20 120 8 \
-			-- 'y' "Boot $name" 'd' "Make $name the default" \
+			-- 'd' "Make $name the default" 'y' "Boot $name" \
```
kylerankin: I know why you made this change, but I'm thinking of the common case where someone is booting an alternate option in their boot menu (say a previous kernel, a recovery boot), there's a good chance they *don't* want it to be the default.

tlaurion: @kylerankin If selection a specific option, then they should be aware of the specific use case they are taking and choose to not be a default in the future. If by mistake they choose to hit the default, thy have plenty of opportunities down the path to cancel their choice (Modify disk? Seal with TPM, confirm proposed LUKS/LVM Volume). 

In the case they already chose a default before, and the file /boot/kexec_key_devices.txt is or it's LVM counterpart is found, and select to boot that specific option. 

If you really want the default to be to boot the default only, the only alternative I see is to check if /boot/kexec_key_devices.txt of its LVM countrpart exist and propose to set a new default first, else present the choice to boot first. What you think?

initrd/bin/kexec-save-default
tlaurion: @kylerankin The code change picks up those integrity changes resulting from a rom flash and is dealt with automatically now. As said below, the user saying Y here would have to say yes to change the disk, yes to seal with TPM, and confirm that the disk partition /LVM volume is the desired one, at each step having the chance to cancel and get to the previous emnu permitting them to boot the selection only.

When does a TPM gets reset?

initrd/bin/kexec-select-boot
kylerankin: I keep going back and forth on this one because I see reasons for both being the default. Picking sane default options is hard. I'm fine with the change.

initrd/bin/kexec-save-default
kylerankin: The problem is that a user who *doesn't* already use the TPM for LUKS disk encryption is being prompted to set it up out of the blue when they are saving their default boot option.

*Every* user will set a default boot option, but only a few will use the TPM for LUKS encryption. If we have to prompt people in this script for this option, I'd prefer we only do it if we detect they are already using this feature. Otherwise the user should store their LUKS key in the TPM via an option in the Advanced -> TPM/TOTP/HOTP Options menu.

merge:
could you also add documentation on why we have OEM mode and what it does exactly? thanks!

tlaurion:
> could you also add documentation on why we have OEM mode and what it does exactly? thanks!

@merge you're right. Editing OP.

initrd/bin/gpg-gui.sh
```diff
@@ -76,6 +76,7 @@ while true; do
     'a' ' Add GPG key to standalone BIOS image + flash' \
     'l' ' List GPG keys in your keyring' \
     'g' ' Generate GPG keys on a USB security token' \
+    'F' ' Factory Reset Nitrokey Pro v2/LibremKey GPG Card' \
```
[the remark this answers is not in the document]
tlaurion: @kylerankin : Yes.
This takes input and validates that the same validation made by GPG is made, while pushing it further by permitting automations and augment key-attr to RSA 4096, generating keys.
- GPG Admin PIN is at least 8 but not more then 20 characters, while not containing space, which is a limitation of seal-libremkey.

The OEM reownership wizard implies that the OEM owns the Librem Key/NitroKey first, which is used to sign /boot configuration checksums while a working Admin PIN is required to seal HTOP. This script permits to factory reset and push public key into the ROM and is automatically filled by /media/oem-provisioning file if filled by OEM/User prior putting the /boot/oem file in place.

The two are complimentory IMHO. But this one is required in a Reownership context, where called unattended.

initrd/bin/gui-init
```diff
@@ -92,17 +129,243 @@ update_checksums()
...
+      echo -e "HOTP status: $HOTP\n\nHit Enter to regenerate new TOTP/HOTP codes."
+      read
+      update_totp
+    fi
+    continue
+  fi
+}
+
+mount_boot
+if [ -e /boot/oem ];then
```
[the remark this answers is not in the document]
tlaurion: @kylerankin : good enough if it's in /etc/functions? I still think all functions should be centralized.

initrd/bin/gpg-gui.sh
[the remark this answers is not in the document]
kylerankin: Yes, my main comment is just asking your opinion on how you would label your option here and based on that how you would suggest labeling the two options (or at least the automated one) in PR #546 just so we could make changes either here or in my PR in the naming, so there isn't confusion.

initrd/bin/gui-init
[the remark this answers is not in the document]
kylerankin: It *probably* is good enough if it's there as long as it's well-tested. That's one reason why I initially held back from putting the GUI file selection function there (but I think it's ready to go there now that we call it from a number of scripts). 

We just don't want a syntax error to mean that every script that sources that file breaks.

initrd/bin/kexec-save-default
```diff
@@ -33,106 +33,213 @@ TMP_MENU_FILE="/tmp/kexec/kexec_menu.txt"
...
+      elif [ "$num_devices" -eq 1 ]; then
+        echo "Single Encrypted Disk found at $devices_suggest. Using it."
+        key_devices=$devices_suggest
+      else
+        echo "No encrypted devices found."
+      fi
+
+    save_key="y"
+    else
+      die "User chose to not add Disk Unlock Key to the TPM"
```
tlaurion: @kylerankin : so the culprit is here. Changing it to warning, while moving [the whole LVM and encrypted disk logic on top](https://github.com/osresearch/heads/pull/551/files#diff-08c6ecea92d88d78364ed40007bddc2aR165) reduce duplication of code and permits to select a default without forcing it to be saved into TPM. Testing

tlaurion: @kylerankin : [while moving this part down](https://github.com/osresearch/heads/pull/551/files#diff-08c6ecea92d88d78364ed40007bddc2aR209), outside of the if then else part. Testing

initrd/bin/gpg-gui.sh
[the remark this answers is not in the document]
tlaurion: @kyle: personally I would remove your option `'o' ' OEM Factory reset + auto keygen USB security token'` in #546 unless you find it any particular advantage over this one, and would call mine something like "OEM Factory reset Librem Key/Nitrokey + keygen + flash public key"? 

Let me know what you think.

initrd/bin/gui-init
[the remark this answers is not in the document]
tlaurion: @kylerankin The file selection is called from gui-init only anyway?

initrd/bin/kexec-save-default
tlaurion: @kylerankin : so the code now prompts and defaults now to setting a new default, asks the user if he wants to mofigy the disk, in which case he is asked if he wants to seal it with the TPM. If he refuses that step, I fixed it so that it just saves a default, like one would normally expect.

The point is that a Heads user SHOULD normally want to use the TPM to set a Disk Unlock Key passphrase, different then the Disk Recovery Key passphrase, which if captured, would give that person the possibility to clone the disk and read content at will, which takes only 25 minutes.

One keypoint, probably misunderstood of Heads, is to set such Disk Unlock Key, which depends on the TPM measurements and passphrase to unlock disk content, which if evesdropped, only permit an attacked to access content on the local computer.

Is that ok with you like that?

tlaurion: @kylerankin : this is wy the OEM Wizard insists so much on setting the Disk Unlock Key for it's added confidentiality, privacy and security.

initrd/bin/kexec-save-default
tlaurion: @kylerankin : changed to logic to address and try to conceal to your [concerns here](https://github.com/osresearch/heads/pull/551#discussion_r277890574)

tlaurion:
@merge : if you have any comment, you are more then welcome!

initrd/bin/gpg-gui.sh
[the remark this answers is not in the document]
kylerankin: The advantage to having both, I think, is that my "OEM Factory reset" option is completely unattended. It factory resets the smart card and generates a random key with default settings and then adds the public key to a USB thumb drive (optionally) and to the firmware keyring.

If I understand your function correctly, it's interactive and will require the OEM to enter in information for GPG to use. If that is true, I think there's a reason to offer both.

initrd/bin/gui-init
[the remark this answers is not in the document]
kylerankin: It's now called from gui-init but it is also used both in gpg-gui and flash-gui.

initrd/bin/kexec-save-default
kylerankin: If the OEM wizard does this, it's OK, I'm mostly objecting to prompting the user with this option when they are outside of the OEM wizard and are just trying to save a default boot option in kexec-save-default.

Personally my preference is to shift to allowing users to unlock their disk with their smart card instead as it's both more convenient and more secure and that's where I've been focusing my own efforts recently (even though Qubes's LUKS implementation doesn't have native support for smart card unlock, Debian does now). For users who go that route, being prompted to use the TPM for LUKS when they are just saving a default boot option could confuse them and they could say 'yes' without understanding all of the implications.

kylerankin: @tlaurion If I'm reading your change correctly, it will only prompt the user if they already have kexec_key_devices.txt present. If that's the case please confirm and I'm totally fine with this change.

oem-reownership/oem.example
```diff
@@ -0,0 +1,21 @@
...
+luks_passwd_changed
+#When not present in this file, factory resets Librem Key/NitroKey with: oem_gpg_Admin_PIN, oem_gpg_User_PIN, oem_gpg_real_name, oem_gpg_email, oem_gpg_comment, defined under /media/oem-provisioning.
+#The wizard generates new keypairs, injects the public key and trust database into the ROM and reboots. 
+#Upon reboot, the oem_gpg_Admin_PIN is used to automatically seal HOTP secret into the Librem Key/NitroKey
+gpg_factory_resetted
+#When not present in this file, reown the TPM with oem_TPM_Owner_Password, which is recommended to be the same as oem_gpg_Admin_PIN
+tpm_reowned
+#Once all stages are done, the Wizard removes /boot/oem, and ensures that /boot file checksum file is generated and signed with oem_gpg_User_PIN.
+#
+#Then the wizard finally launches the selection of a new boot default, and asks the user if he wants to modifie the disk and if he wants to define a Disk Unlock Key. In such case, oem_TPM_Owner_Password is used to seal a new TOTP secret, oem_gpg_Admin_PIN is used to seal a new HOTP secret, and oem_gpg_User_PIN is used to sign configuration files checksums that changed under /boot.
```
merge: s/modifie/modify (doesn't look bad to me as long as it doesn't break anything (having unencrypted rootfs or password encrypted lvm)) but unfortunately very little time to review right now -.-)

initrd/bin/gpg-gui.sh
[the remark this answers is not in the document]
tlaurion: @kylerankin 
> The advantage to having both, I think, is that my "OEM Factory reset" option is completely unattended. 
Got it. I don't know if the OEM would just want to generate a key without wiping the key completely and changing key-attr though. A complete reset seems more adapted to me in that case, but my code doesn't check for /media/oem-provisioning but when /boot/oem is also present.

I think both options can cohabit, let me know if you think that my code could complete replace your function by reading existing oem_gpg* variables under /media/oem-provisioning or if you want both to be offered to the user.

Let me know what you think on that.

initrd/bin/gui-init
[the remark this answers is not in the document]
tlaurion: @kylerankin : I have let the file_selector code to be locally defined under initrd/bin/config-gui.sh and initrd/bin/gpg-gui.sh but you're right, those could be moved to /etc/functions since they are exact duplicate of each other. Will do.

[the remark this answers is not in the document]
tlaurion: @kylerankin : removed in both scripts

initrd/bin/kexec-save-default
tlaurion: @kylerankin : unfortunately no, it actually asks the user every time the default changes. It just reuses the LUKS device defined under kexec_key_devices.txt to prompt the user to reuse it and defaults to use it if the user hits enter.

@merge @ kylerankin: Is there any outside way of knowing that the GPG card is used to unlokc the LUKS container that we could check in? The actual choice of not proposing the highest security option offered by Heads because debian-unstable offers LUKS unlocking from a GPG card is not empowering the user to be able to secure its QubesOS installation the best way it is actually possible to enforce for most other Operating  Systems. 

QubesOS will support it when dom0 moves to fedora-29 if I got all the intrications right, but I would need advice onto how to properly deal with this without dropping the most secure option currently offered by Heads for something not currently supported by most Operating Systems, while QubesOS users would be directly benefiting of this enforcement.

oem-reownership/oem.example
tlaurion: @merge : didn't tought of unencrypted rootfs at all and it's not currently supported with actual modifications. Will have to review code for that, while..... Does it make sense to use Heads to protect hardware for which data is not protected at all? There is a use case for that? Heads offers integrity protection for /boot considering rootfs should be encrypted IMHO.

Let me know if you see a requirement for this.

initrd/bin/kexec-save-default
tlaurion: @kylerankin Mmmmmh... Doesn't seem to be really externally dealable with when looking at https://source.puri.sm/pureos/packages/smartcard-key-luks/blob/master/smartcard-key-luks

kylerankin: @tlaurion How about you just wrap this controversial section inside of an if that gets triggered based on an environment variable/config option? You can even default to it being set to on, and I could ship firmware with it off if I wanted to and allow the user to turn it on. That way everyone can control what the default is.

kylerankin: To answer your other question, there's no way from looking at the LUKS key slots to know whether someone is using a smart card to unlock it.

tlaurion: @kylerankin : anything that could be looked at in the current default GRUB used option to nullify the proposition of sealing with TPM?

oem-reownership/oem.example
kylerankin: Yes it makes sense for Heads to protect hardware that doesn't have an encrypted /. Heads doesn't claim to protect / (by default) anyway and more importantly, Heads *can't* protect /. It claims to protect the boot process up to and including loading and executing the kernel. The underlying OS needs to take over at that point with the firm foundation that everything that executed before it is now trusted and not tampered with.

We need to be careful to not have Heads dictate security policy outside of Heads to users.

initrd/bin/kexec-save-default
kylerankin: Unfortunately not, it's an option that's set in cryptsetup config files in /etc/ and loaded into the initrd. You'd have to extract and inspect the initrd for Debian-specific hooks to figure out it's happening.

kylerankin: In summary, LUKS has a standard secret in the key slot, cryptsetup just executes a script inside initrd that uses the smart card to decrypt a file inside the initrd that contains that secret.

initrd/bin/gpg-gui.sh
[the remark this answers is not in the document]
kylerankin: I think I'd prefer at least for now if we offer both options and maybe make it clear in my approach that it's fully-automated while in your case it's interactive. There are use cases for both. I could see a customer wanting the OEM to customize the key with their info, but set it up on their behalf, and I could see other customers that just want something to protect the machine during shipment, but that they would completely replace once they get it.

initrd/bin/kexec-save-default
tlaurion: @kylerankin : modifications in https://github.com/osresearch/heads/pull/551/commits/f5271913c76bc7886fc6d7b722115c1fa6587317

kylerankin: works for me.

oem-reownership/oem.example
tlaurion: @kylerankin @merge   Please review [this](https://github.com/osresearch/heads/pull/551/commits/677f81047551e6ad72333d436dca560b5550b9f2). 
Basically, that skips kexec-save-key code part to go directly [here](https://github.com/osresearch/heads/blob/677f81047551e6ad72333d436dca560b5550b9f2/initrd/bin/kexec-save-default#L187) and continue execution.

initrd/bin/gpg-gui.sh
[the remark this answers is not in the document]
tlaurion: @kylerankin 
Is it ok if I name it "F OEM Factory reset Librem Key/Nitrokey + keygen + flash public key" or you would prefer another name to distinguish it better with yours?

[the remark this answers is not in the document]
tlaurion: Called it "''F' ' Factory Reset Librem Key/Nitrokey Pro v2 GPG Card + keygen + flash'"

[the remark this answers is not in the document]
tlaurion: @kylerankin : I think all points were addressed.

oem-reownership/oem.example
merge: thanks! that's very imortant imo. 1. like kyle said, it must totatlly be up to the runtime kernel and user, what rootfs to mount, and 2. encryption and tamper-protection are fundamentally different things. it totally makes sense to have an unencrypted rootfs, with (maybe) verification/tamper-protection in place (way outside of Heads' scope).

initrd/bin/kexec-save-default
merge: I'm confident you figure out something sane :) but really I think that Heads shouldn't care what the user's runtime kernel does. That's outside of the project's scope.

tlaurion:
Last commit took care of OEM process going forward even if no encrypted partition was found in the reownership process. Was tested with #553 in on Fedora-30 default installation without encryption.

tlaurion:
had to revert bls support since [it was causing regression](https://github.com/osresearch/heads/issues/553#issuecomment-488389778).

tlaurion:
@merge @kylerankin : functional with unencrypted dusk support. Good for review!

tlaurion:
- Made sure checkums are outputed in gitlab CI artifact.zip while being outputed in build console.
- The CI keeps a cache for each branch, which will permit one branch per board.
- Added @heads105 bls (Fedora 30 and others getting away of grub.cfg) support to future proof OEM stable branch for the next year.

tlaurion:
Freeze.

On current state of things, the /boot/oem file being present searches USB device for /media/oem-reownership that currently needs to be fed externally. LUKS encrypted container is currently supported through mount-usb.

If a LUKS encrypted container is present, mount-usb will ask you for it's passphrase, open it then mount it.

In any case, the provisioned secrets will be used in all of the following steps, if present, without prompting the user for them. 
Actual LUKS Disk Recovery passphrase, New LUKS Disk Recovery Passphrase, GPG Admin PIN (Single word validated. Two words break the HOTP sealing), GPG User PIN, TPM passphrase. 

At each step requiring GPG Admin password (after flashing ROM backup including new public key which requires TOTP/HOTP sealing, after reowning TPM which also required HOTP sealing), or the GPG User passphrase (When signing /boot configuration with new reowned key, when signing LUKS header bounded to Disk Unlock Key) those prompts are autofilled, which eases the reowning UX which is near unattended. The user is given explanatory information prior to a step so that he knows what s going on.

I was asked to stop adding changes to this PR so that it gets reviewed for merge. The next changes will be part of another PR once this one is merged.

Next goal: Ask the user for unprovisioned information from encrypted LUKS container or unencrypted /media/oem-provisioning and save them there after validating them. (GPG real name needs >= 5 characters < 30, GPG Admin PIN >8 chars and in single word, GPG User >6 chars, etc) 

So the OEM reownership process can be completely unattended once those information are given only once, without asking the user to provide those secrets from another computer, Heads environment being the safest available anyway. The resulting values will be stored under /media/oem-provisioning once validated, used USB device partition will be kept under /boot/oem so it's not asked at each reboot of the reownership steps. 

At the end of the process, the user will have a backup of his provisioned secrets on the provided USB disk, where the LUKS container will be reencrypted and its passphrase being the same that he provisioned with New Disk Recovery Key passphrase, and done at the same time his installation gets re-encrypted, making sure the OEM cannot hand those secrets to third parties.  

The OEM only has to provide to the user Actual Disk Recovery Passphrase through secured communication channel, with the TOTP QR Code, so that the user can validate integrity of the firmware and boot configuration automatically upon reception, prior to following the EOM Reownership wizard, making sure that the laptop root of trust has not tampered with in transit after having left the OEM.

initrd/bin/mount-usb
```diff
@@ -126,7 +126,7 @@ fi
 #Check if $USB_MOUNT_DEVICE is a LUKS encrypted device.Ff so, luksOpen it and replace $USB_MOUNT_DEVICE with /dev/mapper/usbdev
 is_Luks_luksOpen
 if [ "$1" = "rw" ]; then
-  mount -o rw $USB_MOUNT_DEVICE /media
+  mount -o rw $USB_MOUNT_DEVICE /media || die "Unable to remount /media into Read Write mode. Is the device write protected?" 
 else
-  mount -o ro $USB_MOUNT_DEVICE /media
+  mount -o ro $USB_MOUNT_DEVICE /media || || die "Unable to mount $USB_MOUNT_DEVICE."
```
kylerankin: Typo with this double ||.

Also, do we really need to die here if one can't mount the USB? I'd rather show errors to the user and move on than drop to recovery shells.

tlaurion: I'll do a warn then.

_Sent from my Galaxy S3 using [FastHub-Libre](https://github.com/thermatk/FastHub-Libre/)_

tlaurion: Removed.

tlaurion: > Typo with this double ||.
> 
> Also, do we really need to die here if one can't mount the USB? I'd rather show errors to the user and move on than drop to recovery shells.

@kylerankin You're right. Removed what was not usb related. OEM can launch docking station IDE device by calling "mount /media" anyway. Let's not mix things.

tlaurion:
@merge if you want to test [this rom](https://gitlab.com/tlaurion/heads/-/jobs/211144770), touch /boot/oem following first command in bash history.

Feed /media/oem-provisioning on a trusted usb device with required secrets as per[ oem-provisioning.example](https://github.com/osresearch/heads/pull/551/files#diff-101665990fd0dc23df213e98668cebeb) and enjoy. It will reencrypt your installation and change its encryption key, change its encryption associated main passphrase, reset Nitrokey/Librem Key dongle with maximum key attributes, set name, e-mail and selected PINs, generate public key, insert it in rom, reflash, seal HOTP, sign /boot config and ask set Disk Unlock Key Passphrase for selected new default boot. [You can skip individual step by bypassing them in the /boot/oem file](https://github.com/osresearch/heads/pull/551/files#diff-52355fd5951e7e15d57c7d0e8d78bd28)

I'll do another PR to be able to provide LUKS encrypted partition on shipped USB device, so that the wizard asks the user for Actual LUKS Recovery passphrase to unlock USB drive (first provided by OEM over secured communication channel). 

If /media/oem-provisioning file is found but secrets are missing, the wizard will propose the user to generate diceware passphrases for each of them until he agrees on proposed ones, or enter them manually at each step., reminding him of it's use in Heads and every day operations. The wizard will insist on never fogetting the LUKS Disk Recovery Key passphrase, which will be used to access both USB device backup and his system. 

After which /media/oem-provisioning file will be saved on encrypted partition and the wizard will go on, saving actually used LUKS partition to /boot/oem for next reboots to not ask the user to choose the device at each reboot until the wizard finishes. 

After the LUKS passphrase is changed on both OS and USB drive, the user will be warned that the pasphrase was changed for him to type the new passsphrase on next steps. 

The user will then have a backup of chosen secrets (TPM, Nitrokey Admin and User PINs, TPM), and won't have to type them multiple times during reownership.

merge:
haven't yet have time to flash and test a bit (still planned, but if at least you both test this, please don't wait), but:
* do you have to enable "keysever-helpers", "hkp" and the others for the gpg2 build?
* do you have to enable nls and asm for the libassuan, libgcrypt and npth build?
* do you have to enable "tests" for the libusb build?

a comparison of payload size wouldn't hurt for this change. only increase if absolutely necessary.

typos and whitespaces are in there. we can fix them later.

tlaurion:
Those were unsupported options generating warnings in the build logs. I removed them for clarity only, in the goal of starting fresh to reduce size of binaries for others to play with.

On May 14, 2019 12:47:38 PM UTC, Martin Kepplinger <notifications@github.com> wrote:
>haven't yet have time to flash and test a bit (still planned, but if at
>least you both test this, please don't wait), but:
>* do you have to enable "keysever-helpers", "hkp" and the others for
>the gpg2 build?
>* do you have to enable nls and asm for the libassuan, libgcrypt and
>npth build?
>* do you have to enable "tests" for the libusb build?
>
>a comparison of payload size wouldn't hurt for this change. only
>increase if absolutely necessary.
>
>typos and whitespaces are in there. we can fix them later.
>
>-- 
>You are receiving this because you were mentioned.
>https://github.com/osresearch/heads/pull/551#issuecomment-492222613

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

merge:
quick bold test run :) After flashing I do
* only nitrokey inserted
* "Add a GPG key to the running BIOS" (no usb stick inserted)
* "Factory reset Librem KEy/Nitro Key Pro v2 GPG Card + keygen + flash
  * PCR values shown... then "no usb key found" and back to the menu (ok)
* now I insert a FAT USB stick too, again "Add a GPG key to the running BIOS
  * Again "Factory Reset Librem Key/Nitro Key..."
  * passwords are asked, name and comment -> English Keyboard! Can we let the user set a keyboard layout?
  * But then: Just back to the top-level menu without any confirmation or error -.- No GPG key added...
* I can retry "Add a GPG key to running BIOS" -> "Factory Reset..." -> the same loop that does nothing

* So I try "OEM Factory reset + auto keygen USB security token"
  * "confirm that your GPG card is inserted" (NitroKey Start is inserted)
  * seems to work. (USER-ID "OEM Key (...)")
  * Takes a long time. We should print a notice about waiting if possible. (nitrokey LED is on)
  * finally "key created...". "Would you like to copy..." -> OK
    * Error: "cp: can't create '/media/<hash>.asc': I/O error
      Unable to unseal TOTP secret. Hit enter to continue. GUI-menu: "ERROR: TOTP Generation Failed!"
* There I try "Reset TOTP" and scan the QR code. Then says "Insert Librem Key/NitroKey"
  * "Trying to connect to Librem Key: .." 2 times. Then "unable to find Librem Key/Nitro Key" (Nitro Key Start is inserted though).
  * ENDLESS Loop to generating QR-Code and trying to "connect" to Librem Key/NitroKey
  * I can't even quit or drop to a shell!!!

* Once again I try to "Factory reset Librem KEy/Nitro Key Pro v2 GPG Card + keygen + flash" with ONLY a USB FAT formatted flash stick inserted. Now it seems to "work" and continues with "Please confirm that your GPG card is inserted". So I plug that in too now. It fails because my passphrase is not supported... (only numbers possible?). I try again (unplug NitroKey), use only number for PIN/admin pw. Then "gpg: key generation failed: Card error": "Setting real name, e-mail and comment in GPG failed". (we shouldn't ask for that maybe).

Also, when I hit "Ignore Error and Continue to default boot menu", I get: `cat: can't open '/boot/kexec_hotp_counter': no such file or directory. Unable to read HOTP counter". This shout not at all be required at that point. And then again "Librem Key/Nitrokey not detected" while a NitroKey Start is inserted. And only *after* this, I get to the Heads boot menu with yellow background. Please! This is quite bad.

At this point I give up for now. I've had a quite bad experience. Please feel free to let me reproduce something specific or tell me what I've done wrong. Also, I haven't debugged any of that :)

Either we don't support a NitroKey Start or there's quite something wrong here still :) thanks

merge:
oh, when I "ignore tampering and boot anyways" with red console background, I don't have Linux boot messages and am not asked the disk's LVM encryption passwork and can only GUESS that Linux is probably waiting for me to type. This might not be related to your changes here though...

tlaurion:
Houla.

Nitrokey pro or Librem Key supported only for 4096 keys.

Will address bugd reported that should have been handled though.

Thanks for testing!!!!



On May 16, 2019 6:18:43 AM UTC, Martin Kepplinger <notifications@github.com> wrote:
>quick bold test run :) After flashing I do
>* only nitrokey inserted
>* "Add a GPG key to the running BIOS" (no usb stick inserted)
>* "Factory reset Librem KEy/Nitro Key Pro v2 GPG Card + keygen + flash
>* PCR values shown... then "no usb key found" and back to the menu (ok)
>* now I insert a FAT USB stick too, again "Add a GPG key to the running
>BIOS
>  * Again "Factory Reset Librem Key/Nitro Key..."
>* passwords are asked, name and comment -> English Keyboard! Can we let
>the user set a keyboard layout?
>* But then: Just back to the top-level menu without any confirmation or
>error -.- No GPG key added...
>* I can retry "Add a GPG key to running BIOS" -> "Factory Reset..." ->
>the same loop that does nothing
>
>* So I try "OEM Factory reset + auto keygen USB security token"
>* "confirm that your GPG card is inserted" (NitroKey Start is inserted)
>  * seems to work. (USER-ID "OEM Key (...)")
>* Takes a long time. We should print a notice about waiting if
>possible. (nitrokey LED is on)
>  * finally "key created...". "Would you like to copy..." -> OK
>    * Error: "cp: can't create '/media/<hash>.asc': I/O error
>Unable to unseal TOTP secret. Hit enter to continue. GUI-menu: "ERROR:
>TOTP Generation Failed!"
>* There I try "Reset TOTP" and scan the QR code. Then says "Insert
>Librem Key/NitroKey"
>* "Trying to connect to Librem Key: .." 2 times. Then "unable to find
>Librem Key/Nitro Key" (Nitro Key Start is inserted though).
>* ENDLESS Loop to generating QR-Code and trying to "connect" to Librem
>Key/NitroKey
>  * I can't even quit or drop to a shell!!!
>
>* Once again I try to "Factory reset Librem KEy/Nitro Key Pro v2 GPG
>Card + keygen + flash" with ONLY a USB FAT formatted flash stick
>inserted. Now it seems to "work" and continues with "Please confirm
>that your GPG card is inserted". So I plug that in too now. It fails
>because my passphrase is not supported... (only numbers possible?). I
>try again (unplug NitroKey), use only number for PIN/admin pw. Then
>"gpg: key generation failed: Card error": "Setting real name, e-mail
>and comment in GPG failed". (we shouldn't ask for that maybe).
>
>Also, when I hit "Ignore Error and Continue to default boot menu", I
>get: `cat: can't open '/boot/kexec_hotp_counter': no such file or
>directory. Unable to read HOTP counter". This shout not at all be
>required at that point. And then again "Librem Key/Nitrokey not
>detected" while a NitroKey Start is inserted. And only *after* this, I
>get to the Heads boot menu with yellow background. Please! This is
>quite bad.
>
>At this point I give up for now. I've had a quite bad experience.
>Please feel free to let me reproduce something specific or tell me what
>I've done wrong. Also, I haven't debugged any of that :)
>
>Either we don't support a NitroKey Start or there's quite something
>wrong here still :) thanks
>
>-- 
>You are receiving this because you were mentioned.
>https://github.com/osresearch/heads/pull/551#issuecomment-492933037

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

merge:
> Houla. Nitrokey pro or Librem Key supported only for 4096 keys. Will address bugd reported that should have been handled though. Thanks for testing!!!!

my nitrokey Start supports 4096bit keys, see https://shop.nitrokey.com/shop/product/nitrokey-start-6

tlaurion:
> quick bold test run :) After flashing I do
> 
>     * only nitrokey inserted
> 
>     * "Add a GPG key to the running BIOS" (no usb stick inserted)
> 
>     * "Factory reset Librem KEy/Nitro Key Pro v2 GPG Card + keygen + flash
>       
>       * PCR values shown... then "no usb key found" and back to the menu (ok)
> 
>     * now I insert a FAT USB stick too, again "Add a GPG key to the running BIOS
That was not tested, sorry. On that USB disk, QubesOS ISO is povided, which is too big for FAT. Tested with ext4, ext3, ext2 formated on /dev/sdXY primary partitions on a USB Disk provided to customer.
>       
>       * Again "Factory Reset Librem Key/Nitro Key..."
>       * passwords are asked, name and comment -> English Keyboard! Can we let the user set a keyboard layout?
Would require another PR with keyboards in ROM and a simple file_select based on #555 
>       * But then: Just back to the top-level menu without any confirmation or error -.- No GPG key added...
> 
>     * I can retry "Add a GPG key to running BIOS" -> "Factory Reset..." -> the same loop that does nothing
> 
>     * So I try "OEM Factory reset + auto keygen USB security token"
>       
>       * "confirm that your GPG card is inserted" (NitroKey Start is inserted)
>       * seems to work. (USER-ID "OEM Key (...)")
@kylerankin #546 
>       * Takes a long time. We should print a notice about waiting if possible. (nitrokey LED is on)
>       * finally "key created...". "Would you like to copy..." -> OK
>         
>         * Error: "cp: can't create '/media/.asc': I/O error
>           Unable to unseal TOTP secret. Hit enter to continue. GUI-menu: "ERROR: TOTP Generation Failed!"

Seems like fat is not supported in Linux payload! Will look. My use case is a provided ext4 filesystem, inside or not of a LUKS container(Flash memory cannot really be deleted), on which the user is prompted to mount through mount-usb (#548 with luks support added) when /boot/oem is found and prompts the user to select partition if many are found . USB disk should be inserted there for all OEM reownership. Will look at it and maybe force a loop or go into recovery to prepare such disk? Didn't thought about it.

You laptop needs to be in a sealed state PRIOR to launching OEM reownership.
> 
>     * There I try "Reset TOTP" and scan the QR code. Then says "Insert Librem Key/NitroKey"
Yeah at this step you are out of expected scenario. The OEM is suppoed to already have inserted his public key and sealed the TOTP and HOTP prior to have touched /boot/oem. The goal of the reownership is to pass ownership from OEM -> user. So it's supposed to be in a certified state prior to change hands, as stated in this [comment](https://github.com/osresearch/heads/pull/551#issuecomment-491640979) 

`Your laptop being in a trusted state simulates the known state in which the laptop would be received by the user, with TOTP QR Code and the Actual Disk Recovery passphrase shared by secured communication channel, prior to booting the laptop the first time, state being validated through NitroKey/Librem Key LED flashing green, while its smartphone OTP application will permit him to double check if desired. If in an untrusted state, then it was tampered with in transit.`
>       
>       * "Trying to connect to Librem Key: .." 2 times. Then "unable to find Librem Key/Nitro Key" (Nitro Key Start is inserted though).
>       * ENDLESS Loop to generating QR-Code and trying to "connect" to Librem Key/NitroKey
>       * I can't even quit or drop to a shell!!!
Normal. No gpg key inside ROM if I get it right. So the HOTP passphrase is the one chosen. Note that no space are tolerated here for GPG Admin PIN, which is taken cared of in the Factory Reset option in the OEM reownership, but from what I understand, you never reached it since no public key was found in ROM. from recovery:
mount /boot
mount -o remount,rw /boot
rm /boot/oem
mount -o remount,ro /boot

Then factory Reset your NitroKey start, which will output public key and artifacts on /media/gpg_keys and move directory to timestamped backup directory if found.
> 
>     * Once again I try to "Factory reset Librem KEy/Nitro Key Pro v2 GPG Card + keygen + flash" with ONLY a USB FAT formatted flash stick inserted. Now it seems to "work" and continues with "Please confirm that your GPG card is inserted". So I plug that in too now. It fails because my passphrase is not supported... (only numbers possible?). I try again (unplug NitroKey), use only number for PIN/admin pw. Then "gpg: key generation failed: Card error": "Setting real name, e-mail and comment in GPG failed". (we shouldn't ask for that maybe).
> 
> 
> Also, when I hit "Ignore Error and Continue to default boot menu", I get: `cat: can't open '/boot/kexec_hotp_counter': no such file or directory. Unable to read HOTP counter". 
You went sideways without having a laptop in a known state from a OEM perspective prior to reowne the laptop.

>This shout not at all be required at that point. And then again "Librem Key/Nitrokey not detected" while a NitroKey Start is inserted. And only _after_ this, I get to the Heads boot menu with yellow background. 

> Please! This is quite bad.
Such a bad experience, sorry. 
Prerequisites: 
1- Clean, updated install of QubesOS
> 
> At this point I give up for now. I've had a quite bad experience. Please feel free to let me reproduce something specific or tell me what I've done wrong. Also, I haven't debugged any of that :)
> 
> Either we don't support a NitroKey Start or there's quite something wrong here still :) thanks
Nitrokey Start should then be supported. It all seems to be from FAT partition on USB disk, the lack of public key inside the rom and the lack of HOTP being sealed prior launching the OEM reownership.

BlackMaria:
My $0.25 wrt 
> Takes a long time. We should print a notice about waiting if possible. (nitrokey LED is on)

There are some places where it might be nice to add a wait cursor, to indicate the system is working in the background.  I suggest we add a routine to /etc/functions something like this

``` bash
function wait_cursor(){
  action=$1
  shift
  process=$@
  $process  &    
  pid=$!  

  spin[0]="-"
  spin[1]="\\"
  spin[2]="|"
  spin[3]="/"
  
  echo -n "${action} ${spin[0]}"
  while $( kill -0 $pid )
  do
    for i in "${spin[@]}"
    do
          echo -ne "\b$i"
          sleep 0.1
    done
  done
}

wait_cursor "Factory reset Nitro Key:  "  ${some_gpg_command} 
wait_cursor "Creating GPG Keys ( this should take x minutes ): "  ${some_gpg_command} 

```
Example code hijacked from stackoverflow : [ using-bash-to-display-a-progress-working-indicator](https://stackoverflow.com/questions/12498304/using-bash-to-display-a-progress-working-indicator)

tlaurion:
> My $0.25 wrt
> 
> > Takes a long time. We should print a notice about waiting if possible. (nitrokey LED is on)
> 
> There are some places where it might be nice to add a wait cursor, to indicate the system is working in the background. I suggest we add a routine to /etc/functions something like this
> 
> ```shell
> function wait_cursor(){
>   action=$1
>   shift
>   process=$@
>   $process  &    
>   pid=$!  
> 
>   spin[0]="-"
>   spin[1]="\\"
>   spin[2]="|"
>   spin[3]="/"
>   
>   echo -n "${action} ${spin[0]}"
>   while $( kill -0 $pid )
>   do
>     for i in "${spin[@]}"
>     do
>           echo -ne "\b$i"
>           sleep 0.1
>     done
>   done
> }
> 
> wait_cursor "Factory reset Nitro Key:  "  ${some_gpg_command} 
> wait_cursor "Creating GPG Keys ( this should take x minutes ): "  ${some_gpg_command} 
> ```
> 
> Example code hijacked from stackoverflow : [ using-bash-to-display-a-progress-working-indicator](https://stackoverflow.com/questions/12498304/using-bash-to-display-a-progress-working-indicator)

Could do. While the whiptail menu explicitly says to the user that the process will take 15 minutes and explains why, while the NitroKey / Librem Key led is a steady RED, and gpg outputs "BUSY" on the screen until it finishes its business.

The actual, expected, OEM reownership process needs to be tested by third parties and usability improved from expected scenario. The rest will be addressed as improvements, not bug reports.

I take the mount-usb critic that tries 2 times and falls into a weird state while FAT not being correctly supported. Will look into it, but my insight would be to support exfat instead (to be able to accept 4Gb+ ISO images), which leaves the user to prepare a USB disk properly OR the OEM to provide such disk, which is already covered by the OEM, not the user.

merge:
yes. please put other enhancements in seperate PRs. Please have a look, but I'm sure we currently support mounting FAT filesystems. I never used spaces in passwords.

GPG pub key being included in flash should never be required for having the TOTP sealed in TPM.  It's a different task than verifying /boot. (sorry if that's clear). Therefore an OEM walkthrough procedure could cleanly "fail" in case something with USB or GPG is wrong.

if /boot/kexec_hotp_counter is not there, why not just don't show the HOTP? It's not really an *error* when I, say, unplug my hard disk.

merge:
I've lost track a bit of what changes you have in here, but why not add "gpg key auto-generation + nitrokey/libremkey setup" __only__ first? (Interactively reading the key-passphrases, but only interactive when really necessary).

Because that's what is really needed for users _right now_, in order to trivially "fix" the "no GPG" error. Security-wise this isn't a problem either, after a know-good flash.

Doing more than this (LUKS stuff, automating using an "oem" file, anything else), would be a local change _after_ having this, wouldn't it?

tlaurion:
Status Update (decided to update here and [push there](https://gitlab.com/tlaurion/heads/-/jobs/227728884) to facilitate tracking of changes and peer review, otherwise compromising past reviews efforts.)

- OEM Reownership Wizard is currently bound to the x230 (because of the card reader drivers [mmc/sdhci modules added into linux-x230](https://github.com/osresearch/heads/pull/551/files#diff-2a92039cb02457fb3a0d2172863f19a7R195) and[ x230 board config files](https://github.com/osresearch/heads/pull/551/files#diff-b44c9ef755b9108f779283ae4709949eR28) and more generally, [exported in the linux config](https://github.com/osresearch/heads/pull/551/files#diff-c44c192f63aab4e550c64696ea703787R98) if `CONFIG_LINUX_SDHCI` is defined under board config. That makes it easier for the OEM since the sdcard is already inserted in the laptop and prepared accordingly to the OEM Reownership requirements (luks encrypted partition already setuped, see below.)

- OEM Reownership Wizard is triggered by the presence of /boot/oem file which can [easily be setuped from the recovery](https://github.com/osresearch/heads/pull/551/files#diff-d1ac0113778bfe516c8b663a9ebb4cf8R4). See requirements below.

- OEM Reownership Wizard feeds/pulls its information from a required to be present `/media/oem-provisioning` file when `/boot/oem` is detected. That `oem-provisioning` file needs to be found under a LUKS encrypted container [or else will refuse to store secrets and fallback into manual mode](https://github.com/osresearch/heads/pull/551/files#diff-893be823b299bd80d45c36112bbbf1e3R94). The reasoning behind that implementation choice is that [sdcard, like SSD devices, do not really delete files, thanks to their wear protection mechanism](https://ssd.eff.org/en/module/how-delete-your-data-securely-linux##SSDs). Encryption is the only way to store secrets.

- OEM Reownership then parse /media/oem-provisioning and is [triggered by any missing provisioning information](https://github.com/osresearch/heads/pull/551/files#diff-10d5be21d425b1f3dd8bb331ef7d9ad6R330). For [all missing provisioning information](https://github.com/osresearch/heads/pull/551/files#diff-10d5be21d425b1f3dd8bb331ef7d9ad6R482), the user is [prompted to generate EFF Diceware passphrases until one fits his memory scheme](https://github.com/osresearch/heads/pull/551/files#diff-10d5be21d425b1f3dd8bb331ef7d9ad6R507) (n) or is offered a choice to define one manually, while being suggested to [choose words from 800 shuffled diceware words outputed on the screen](https://github.com/osresearch/heads/pull/551/files#diff-10d5be21d425b1f3dd8bb331ef7d9ad6R816).

Requirements:
- Heads doesn't actually support diskless setups ( [a /boot partition needs to be found per CONFIG_BOOT_DEV](https://github.com/osresearch/heads/pull/551/files#diff-b44c9ef755b9108f779283ae4709949eR38)). Consequently:
- QubesOS needs to be already installed with standard partition scheme or internal disk prepared from Heads `fdisk`. ([LVM scheme not tested but should work too](https://github.com/osresearch/heads/pull/551/files#diff-08c6ecea92d88d78364ed40007bddc2aR138) for setting up Disk Unlock Key) 
- OEM previously flashed [OEM Reownership firmware supporting firmware](https://gitlab.com/tlaurion/heads/-/jobs/227728884).
- Sealing/unsealing needs to be non-faulty (TOTP/HOTP being sealed post to flashing. Can be done with factory default Librem Key/NitroKey Admin PIN if desired: `12345678` )
- Disk Unlock key is suggested to be set (so that `/boot/kexec_key_devices.txt` explicitly states internal encrypted disk partition). Note that Disk Unlock Key support now requires board config [CONFIG_OFFER_TPM_LUKS_DISK_UNLOCK_KEY](https://github.com/osresearch/heads/pull/551/files#diff-b44c9ef755b9108f779283ae4709949eR31), defaulting to `y` for x230 and x230, and `n` for [Librems](https://github.com/osresearch/heads/pull/551/files#diff-551bc5dfe88f76443b8f9d6dac50093bR30) following @kylerankin request above. Else, [if supported only](https://github.com/osresearch/heads/pull/551/files#diff-08c6ecea92d88d78364ed40007bddc2aR78), a [choice will be given to the user ](https://github.com/osresearch/heads/pull/551/files#diff-08c6ecea92d88d78364ed40007bddc2aR163)(`/dev/mmcp0p1 /dev/sda2`)
- Boot integrity (kexec.sig) needs to be sealed by OEM to be shipped Librem Key/Nitrokey (OEM Reownership is suggested to do so)

My suggested/tested OEM ReOwnership process:
- OEM has a dedicated Deployment Nitrokey/Librem Key used to sign clonezilla ISO, from which his public key is inserted into ROM for disk reimaging purposes. 
- The OEM flashes [reproducible OEM firmware image](https://gitlab.com/tlaurion/heads/-/jobs/227728884), [including his public key inside of the rom](https://github.com/osresearch/heads/pull/551/files#diff-d1ac0113778bfe516c8b663a9ebb4cf8R10), or inserts his public key at first Head's boot prompt, since not found.
- The OEM boots clonezilla from a [signed ISO of his own](https://github.com/osresearch/heads/pull/551/files#diff-d1ac0113778bfe516c8b663a9ebb4cf8R18)
- The OEM clones dockstation disk image backup (`mount /dev/sdb1 /media` from recovery prior to relaunch `gui-init` and `boot from USB` from there ([which uses /media mounted device if already mounted](https://github.com/osresearch/heads/pull/551/files#diff-fdfe87f61e3355a5042988dfbe10e9c2R38)), containing a generic LUKS Recovery Key and associated Recovery Key passphrase (herein named actual Disk Recovery Passphrase). QubesOS user is "user" and login passphrase should be OEM's name to easy reownership.
- The OEM prepares a sdcard containing two partitions:
  - The first one being a LUKS encrypted ext4 partition of 32M with actual Disk Recovery passphrase. (Needs to be the same passphrase then QubesOS chosen encryption passprase at installation time). That partition will contain original SPI flash content (original 8mb Intel ME SPI content, original Lenovo 4mb boot room), a prepared /oem-provisioning file (containing his own GPG real name, e-mail, and GPG comment, and cloned actual Disk Recovery Key), reproducible x230-flash.rom and coreboot.rom.
  - A unencrypted ext4 partition, containing QubesOS latest ISO image, latest Tails image and their corresponding signature files, validated with public signatures present inside of Heads for USB boot integrity validation. TODO: generalize external media integration, so that /dev/sdb from dockstation, sdcard or USB devices are supported seemlessly. Meanwhile, a user wanting to boot from sdcard fro memory expansion card slot would have to call `mount-sdcard` then call-back gui-init, or boot from that sdcard from a USB to sdcard converter (which safety is questionnable, at least).
- The OEM [sets the /boot/oem file](https://github.com/osresearch/heads/pull/551/files#diff-d1ac0113778bfe516c8b663a9ebb4cf8R2) himself and launches the OEM reownership wizard with a blank Nitrokey/Librem Key inserted and the prepared sdcard.
- On reboot, The OEM is asked to type the actual Disk Recovery Key (which should be the same for OEM's internal disk and sdcard partitions).
- In the OEM Reownership Wizard, the OEM will be prompted to automatically generate GPG Admin and User Pins, LUKS new Disk Unlock Passphrase (he should take note) and TPM ownership passphrase (which copies GPG Admin PIN). The OEM only needs to take note of the Disk Recovery Key passphrase, which will be the same for internal disk and sdcard encrypted partitions. That information will be shared with the customer and will be asked of the OEM at each reboot of the machine until "reownership" (randomization of deployment) is done.
- The OEM Reownership wizard will now [save those information](https://github.com/osresearch/heads/pull/551/files#diff-10d5be21d425b1f3dd8bb331ef7d9ad6R641) in the /media/oem-provisioning file.
- The OEM Reownership wizard then reencrypts both sdcard and internal disk with actual Disk Recovery Key, then changes LUKS Disk recovery passphrase to chosen Disk Recovery Key passphrase and reboots.
- From now on until OEM Reownership ending, the Wizard will ask upon reboot explicitely to enter the new Disk Recovery Key passphrase. It then factory resets the NitroKey/Librem Key, exports public key on sdcard encrypted partition, flashes the ROM with that public key, reseal TOTP/HOTP at each steps automatically, sets the TPM ownership passphrase to GPG Admin passphrase and at that final prompts, requires the OEM to take a picture of the QRCode to be sent to customer with new Disk Recovery passphrase and QubesOS login passphrase to be changed through secured communication channel.
- The OEM Reownership Wizard is done. 
- The OEM tests the actual boot of QubesOS by typing Disk Unlock Passphrase.
- The OEM goes into Heads recovery, `mount-sdcard`, `mount -o remount,rw /media`, `cd /media` and `cp oem-provisioning oem-provisioning.oem`. He then edits oem-provisioning to delete everything but "oem_luks_new_Disk_Recovery_Key=XXXXX" that he renames to oem_luks_actual_Disk_Recovery_Key=XXXXX" to reflect the actual situation. he commits the changes by `mount -o remount,ro /media`.
- The OEM finally [sets the /boot/oem flag with his OEM name](https://github.com/osresearch/heads/pull/551/files#diff-d1ac0113778bfe516c8b663a9ebb4cf8R2) for the user to be ready to launch the OEM reownership himself upon device selection.

QubesOS customizations:
- sys-usb devices includes "SD Host Controller: Ricoh Co Ltd PCIe SDXC/MMC Controller"
- QubesOS updates through LAN connection.

User reception:
- User receives TPMTOTP QRCode, actual Disk Recovery Key passphrase and QubesOS user passphrase by secured communication channel prior to device reception.
- Upon reception, user boots device, is denied OEM Reownership if not [outlet powered](https://github.com/osresearch/heads/pull/551/files#diff-9a66e05586b37d4779ee1eb88802e759R110). Otherwise, he is invited to feed the Reownership Wizard once with all required information or [can choose to enter required information manually at each step](https://github.com/osresearch/heads/pull/551/files#diff-10d5be21d425b1f3dd8bb331ef7d9ad6R644) if paranoid.
- The user patients 35 minutes to reenrypt internal 250GB OS partition and 30 seconds for sdcard 32Mb partition.
- The user patients 15 minutes for Nitrokey/Librem Key keys generation.
- The user is ready to boot preinstalled QubesOS in less then an hour after device reception and reownership.
- The user boots QubesOS with Disk Unlock Passphrase and changes his QubesOS user password in dom terminal by typing the `passwd` command. 
- The user enjoys presinstalled QubesOS with firmware and /boot integrity attestation.

tlaurion:
> I've lost track a bit of what changes you have in here, but why not add "gpg key auto-generation + nitrokey/libremkey setup" **only** first? (Interactively reading the key-passphrases, but only interactive when really necessary).

@merge : [One sets Librem Key/Nitrokey RSA keysizes to 4096 bits while validating input of PINs](https://github.com/osresearch/heads/blob/2cd5c48d19031520e5e8143b0ec02aef09938bf3/initrd/bin/factory-reset-nitrokey-libremkey.sh)(Setting up an Admin PIN with space would work for GPG but not for HOTP sealing) [while the other is interactive through GPG console and takes the ouput to export public key](https://github.com/osresearch/heads/blob/2cd5c48d19031520e5e8143b0ec02aef09938bf3/initrd/bin/gpg-gui.sh#L184). 

@merge : if you talk about the [automatic generation of oem keys ](https://github.com/osresearch/heads/blob/2cd5c48d19031520e5e8143b0ec02aef09938bf3/initrd/bin/gpg-gui.sh#L78) it was not filling the need of setting keysize and was not changing passwords either, not filling present need.

I preferred to integrate it so that the user experience was controlled and the same for OEM and user factory reset needs, not having to deal with GPG console directly at all. 

Your comments are welcomed here, if you have a better name to [differentiate current approaches](https://github.com/osresearch/heads/pull/551/files#diff-abc72e882ba88e6dbe7eb70ce2dd944bR109)?

> 
> Because that's what is really needed for users _right now_, in order to trivially "fix" the "no GPG" error. Security-wise this isn't a problem either, after a know-good flash.

@merge: I think i've came across a comment of yours in another issue? 
> 
> Doing more than this (LUKS stuff, automating using an "oem" file, anything else), would be a logical change _after_ having this, wouldn't it?
> 
> edit: I overlooked that that's already there, using 123456 and 12345678 for the key, but why not. anyways, that's good IMO. sorry for the noise.

tlaurion:
> quick bold test run :) After flashing I do
> 
>     * only nitrokey inserted
> 
>     * "Add a GPG key to the running BIOS" (no usb stick inserted)
> 
>     * "Factory reset Librem KEy/Nitro Key Pro v2 GPG Card + keygen + flash
>       
>       * PCR values shown... then "no usb key found" and back to the menu (ok)

@merge: Since your last test, [mount_usb logic](https://github.com/osresearch/heads/pull/551/files#diff-abc72e882ba88e6dbe7eb70ce2dd944bR124) has been replaced by [the generic mount-usb, which attempts to mount usb devices unless /media is already mounted](https://github.com/osresearch/heads/pull/551/files#diff-fdfe87f61e3355a5042988dfbe10e9c2R39). This inconsistence is one of the reason why I decided to implement mount-sdcard and put the prepared LUKS encrypted sdcard partition in memory expension slot in the laptop prior to shipping, and to put `oem_provisioning_partition=/dev/mmcblk0p1` in `/boot/oem ` so the user [doesn't have to deal with those manipulation errors](https://github.com/osresearch/heads/pull/551/files#diff-893be823b299bd80d45c36112bbbf1e3R104) either.

> 
>     * now I insert a FAT USB stick too, again "Add a GPG key to the running BIOS
>       
>       * Again "Factory Reset Librem Key/Nitro Key..."

@merge FAT is supported. I checked.

>       * passwords are asked, name and comment -> English Keyboard! Can we let the user set a keyboard layout?

That would be a nice addition to #555 indeed. PR?

>       * But then: Just back to the top-level menu without any confirmation or error -.- No GPG key added...
> 
>     * I can retry "Add a GPG key to running BIOS" -> "Factory Reset..." -> the same loop that does nothing

That was linked to mount_usb calls. That is resolved with present code.
> 
>     * So I try "OEM Factory reset + auto keygen USB security token"
>       
>       * "confirm that your GPG card is inserted" (NitroKey Start is inserted)
>       * seems to work. (USER-ID "OEM Key (...)")
>       * Takes a long time. We should print a notice about waiting if possible. (nitrokey LED is on)
>       * finally "key created...". "Would you like to copy..." -> OK
>         
>         * Error: "cp: can't create '/media/.asc': I/O error
Same here. Device was mounted, but maybe touched and in unreliable kernel state?

>           Unable to unseal TOTP secret. Hit enter to continue. GUI-menu: "ERROR: TOTP Generation Failed!"
> 
>     * There I try "Reset TOTP" and scan the QR code. Then says "Insert Librem Key/NitroKey"
>       
>       * "Trying to connect to Librem Key: .." 2 times. Then "unable to find Librem Key/Nitro Key" (Nitro Key Start is inserted though).
>       * ENDLESS Loop to generating QR-Code and trying to "connect" to Librem Key/NitroKey
>       * I can't even quit or drop to a shell!!!

Really weird state :|
Seems like connected usb devices were flushed? Cannot replicate but never had that in all my tests.
[The HOTP sealing loop](https://github.com/osresearch/heads/blob/2cd5c48d19031520e5e8143b0ec02aef09938bf3/initrd/bin/gui-init#L94) was implemented to circumvent a HOTP sealing that sometimes fail but works on the second blow. The loop you have seems to be from an inconsistent bug in USB, that could have been resolved by removing and reinserting the NitroKey back in. I dunno :|

> 
>     * Once again I try to "Factory reset Librem KEy/Nitro Key Pro v2 GPG Card + keygen + flash" with ONLY a USB FAT formatted flash stick inserted. Now it seems to "work" and continues with "Please confirm that your GPG card is inserted". So I plug that in too now. It fails because my passphrase is not supported... (only numbers possible?). I try again (unplug NitroKey), use only number for PIN/admin pw. Then "gpg: key generation failed: Card error": "Setting real name, e-mail and comment in GPG failed". (we shouldn't ask for that maybe).
> 

As you pointed out, it was the factory default 12345678 from  @kylerankin eom keygen.
> 
> Also, when I hit "Ignore Error and Continue to default boot menu", I get: `cat: can't open '/boot/kexec_hotp_counter': no such file or directory. Unable to read HOTP counter". This shout not at all be required at that point. And then again "Librem Key/Nitrokey not detected" while a NitroKey Start is inserted. And only _after_ this, I get to the Heads boot menu with yellow background. Please! This is quite bad.

Actually, used x230 config here specifies both TPM and LIBREMKEY, so both are required.
(will check for NitroKey Start HOTP support)
> 
> At this point I give up for now. I've had a quite bad experience. Please feel free to let me reproduce something specific or tell me what I've done wrong. Also, I haven't debugged any of that :)
> 
> Either we don't support a NitroKey Start or there's quite something wrong here still :) thanks
@merge I'm really not sure NitroKey start support HOTP actually.
From other tickets, the Pro and the Storage are HOTP compliant. Will look into the Start, but I think this is the issue.

tlaurion:
@merge : The NitroKey Start [doesn't support HOTP](https://github.com/Nitrokey/nitrokey-start-firmware/search?q=hotp&unscoped_q=hotp). 

So in your case, LIBREMKEY should be deactivated in [x230 config](https://github.com/osresearch/heads/pull/551/files#diff-b44c9ef755b9108f779283ae4709949eR24). 

Should I remove the requirement for LIBREMKEY from the default x230 board config?

tlaurion:
WIP. Issues with reencryption with crypsetup 1.7.5 "Cannot open reencryption log file". Will delete and update when done.

tlaurion:
The latest build was corrupted by QEMU coreboot traces in the CI build. Wait for this message to be deleted prior to test CI builds.

merge:
To me it feels like Libremkey shouldnt be a config option at all. Just support it but not require it. Isn't that quite easily doable too?



Am 8. Juni 2019 23:54:38 MESZ schrieb tlaurion <notifications@github.com>:
>@merge : The NitroKey Start [doesn't support
>HOTP](https://github.com/Nitrokey/nitrokey-start-firmware/search?q=hotp&unscoped_q=hotp).
>
>
>So in your case, LIBREMKEY should be deactivated in [x230
>config](https://github.com/osresearch/heads/pull/551/files#diff-b44c9ef755b9108f779283ae4709949eR24).
>
>
>Should I remove the requirement for LIBREMKEY from the default x230
>board config?

-- 
Martin Kepplinger
http://martinkepplinger.com
sent from mobile

tlaurion:
> To me it feels like Libremkey shouldnt be a config option at all. Just support it but not require it. Isn't that quite easily doable too? 
@merge  absolutely. Removing [CONFIG_LIBREMKEY](https://github.com/osresearch/heads/pull/551/files#diff-b44c9ef755b9108f779283ae4709949eR24) is all that is needed.

In the same blow, It might be a good moment to split board into mutliple ones?
My addition here adds kernel modules in the linux-x230 for mmc and sdhci. while adding the [CONFIG_OFFER_TPM_LUKS_DISK_UNLOCK_KEY](https://github.com/osresearch/heads/pull/551/files#diff-b44c9ef755b9108f779283ae4709949eR31) option for all supporting TPM boards.

Not sure how to make image reproducible for OEMs if not by having a board that works out of the box, including Librem Key support, from my CI. I would go for a new board being x230-oem or oem-libremkey, or leave the user to deactivate such option prior to build with proper documentation. 

Let me know what you think.

Or document board options and advise in the doc to review those options, detailing all of them?

kylerankin:
We need CONFIG_LIBREMKEY as an option because when it's enabled, we warn the user at boot to insert their Librem Key if it's not inserted, so they can get boot verification using it. This is an important ease-of-use piece. It also adds additional binaries into the image that some people might want the option to remove it for those older machines that have smaller flash space.

kylerankin:
> The NitroKey Start [doesn't support HOTP](https://github.com/Nitrokey/nitrokey-start-firmware/search?q=hotp&unscoped_q=hotp).

This is one reason why tweaking things in your OEM script to say "Nitrokey/Librem Key" everywhere could make things confusing, because only two of the existing Nitrokey products (Nitrokey Pro v2, and I believe brand new Nitrokey Storage models) support this HOTP function. Every Librem Key supports it, but not every Nitrokey does.

tlaurion:
> > The NitroKey Start [doesn't support HOTP](https://github.com/Nitrokey/nitrokey-start-firmware/search?q=hotp&unscoped_q=hotp).
> 
> This is one reason why tweaking things in your OEM script to say "Nitrokey/Librem Key" everywhere could make things confusing, because only two of the existing Nitrokey products (Nitrokey Pro v2, and I believe brand new Nitrokey Storage models) support this HOTP function. Every Librem Key supports it, but not every Nitrokey does.

Will revert those changes.

tlaurion:
Unfortunately, it's currently [impossible to boot Tails from the sdcard bay on the x230](https://github.com/osresearch/heads/issues/581#issuecomment-501414411). 

Will contact Tails to have to fixed on the next version.

modules/gpg2
```diff
@@ -30,19 +30,12 @@ gpg2_configure := ./configure \
         --disable-regex \
         --disable-doc \
 	--disable-bzip2 \
-	--disable-asm \
```
tlaurion: All gpg and gpg requirements undefined options have been removed.

tlaurion:
Certified.

tlaurion:
@merge @flammit @kylerankin: will you approve this PR or shall I maintain it seperatly in my gitlab account?

tlaurion:
@merge @flammit @kylerankin @osresearch 
The CI reproducible build of x230-libremkey firmware (downloadable artifacts -> coreboot.rom) is [here](https://gitlab.com/tlaurion/heads/-/jobs/246485354).

The [PrivacyBeast_X230_QubesOS_Certified protected branch](https://gitlab.com/tlaurion/heads/commits/PrivacyBeast_X230_QubesOS_Certified) won't be updated for the next year, following QubesOS certification's public announcement. 

Walktrough videos:
1. [How to activate OEM ReOwnership Wizard](https://archive.org/details/activateoemreownership) (OEM ready-to-ship perspective)
1. [The complete OEM -> User ReOwnership walk-through upon hardware reception](https://archive.org/details/oemuserreownership) (User perspective / OEM perspective after cloning his disk image to make deployment unique)

If you see anything ugly, now would be a good time to speak!

tlaurion:
Fixes needed to be done on the stable branch to fix sdcard issues only showing when attempting to reencrypt sdcard content (resulting in IO errors on some cards of the same brand but not others... sdhci-pci and mmc being involved in the problem)

This implied upgrading the kernel to [latest LTS](#599).
Which tackles the [quest for smaller ROM](#590)

Commits onward are [here](https://github.com/tlaurion/heads/commit/05e8152e9e2bb8154c01660e45b31458f6ea8d2f)

CI builds attempts are happening [here](https://gitlab.com/tlaurion/heads/-/jobs/269631843)

tlaurion:
Most of the work here has been upstreamed.

Missing if anyone wants to upstream:
- diceware passphrase generator
- automatic reusal of past used Luis containers

Will be done soon:
- support of encrypted partitions under luks

PR #557, opened 2019-04-30
kexec-select-boot not compatible with Fedora 30

heads105:
Fedora 30 (and optionally 29) supports a new grub config format where each grub entry is in its own config file. The change is documented here: https://fedoraproject.org/wiki/Changes/BootLoaderSpecByDefault

This patch supports parsing the files in /boot/grub2/loader/entries (blsdir) if it exists. To do so, it relies on a default "append" value called kernelopts either from the grub.cfg file or from grubenv. I think I've captured the possible configuration/override options for this.

Tested on Fedora 30 beta and a fresh install of Fedora 30 released today as well as after all updates installed today. If the loader/entries directory is empty or does not exist, this patch should have no effect.

Note that it does reuse functionality from kexec-parse-boot but I chose to separate it into its own file to reduce impact of change. Perhaps it should be refactored at some point.


tlaurion:
includes #556 .

PR #559, opened 2019-05-02
Parse grub config files for Fedora 29/30

heads105:
Have made this fix more conservative. If kexec-parse-boot provides any output at all, then the BLS parsing will not be performed. The /boot/loader/entries directory will only be parsed if there are no grub .cfg files with direct kernel parameters within them.

To proceed to execute kexec-parse-bls, there are two tests: 
```
[ ! -s $option_file ] - $option_file is zero size
[ -d "$bootdir/loader/entries" ] - /boot/loader/entries directory exists
```

I've tested this with Fedora 30, Qubes 4.0.1 (which seems to have a bizarre /boot/loader/entries directory...) and for good measure, Ubuntu 19.04

### DIFF
--- a/initrd/bin/kexec-parse-bls
+++ b/initrd/bin/kexec-parse-bls
@@ -1,0 +1,99 @@
+#!/bin/sh
+set -e -o pipefail
+bootdir="$1"
+file="$2"
+blsdir="$3"
+kernelopts=""
+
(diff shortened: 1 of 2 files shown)


tlaurion:
Tested on QubesOS!

PR #561, opened 2019-05-11
Fix qemu-coreboot board to boot by default

tlaurion:
add "-nic none" to fix https://github.com/osresearch/heads/issues/516#issuecomment-464000222 and boot heads in qemu


tlaurion:
Actually, I didn't encounter the bug in my tests. Feel free to reopen.

Issue #562, opened 2019-05-13
use coreboot vboot measured boot


merge:
the goal here is basically to close #41 . Currently we have https://github.com/osresearch/heads/blob/master/patches/coreboot-4.8.1/0000-measuredboot.patch that allows us to do measurements in initrd. We can still continue to use this on top of a recent release of coreboot and should do so, and update coreboot independently of this task!

Now since https://review.coreboot.org/c/coreboot/+/29547 and https://review.coreboot.org/q/topic:%22LENOVO_VBOOT%22+(status:open%20OR%20status:merged) coreboot allows to measure boot as part of the [vboot](https://www.coreboot.org/git-docs/Intel/vboot.html) scheme. Even if we won't use parts of vboot (keys in GBB), but only the measured-boot part, it would make sense to use a well-define upstream maintained solution.

A different FMAP (fmd file) is being used in that case:
vboot basically supports 3 different parition schemes (FMAP) to choose from, that defines what funcitonality will be available. "RW_A and RW_B" (which would enable updates) is out of scope for us due to size limitation. "RW_A only" would theoretically be possible, but would also impose at least some unnecessary size limitations and also, we would not need to have a pre-defined static "RO" seperate of "RW_A". Our root of trust will be just Heads itself with it's code up until the point where we will enable write-protection for the SPI flash. Basically, our flashrom update-scripts need to run before that. The paritioning scheme to use for use will most likely be "RO only" (even if it's no static read-only, but locks the flash at some point).

I only quickly tried to write an appropriate "vboot_ro.fmd" file (not yet merged upstream). "RO only" still reduces the available CBFS size for coreboot+payload a little, in order to fit into the vboot mechanism and I ran into problems with building because our kernel+initrd are just a _little_ to big. (I think we'll have 6942720 bytes cbfs and we're already filling the currently available 7M):

```
FLASH@0xff400000 0xc00000 {
       SI_ALL@0x0 0x500000 {
               SI_DESC@0x0 0x1000
               SI_GBE@0x1000 0x2000
               SI_ME@0x3000 0x4ed000
       }
       SI_BIOS@0x500000 0x700000 {
               RW_VPD(PRESERVE)@0x0 0x1000
               SMMSTORE(PRESERVE)@0x1000 0x40000

               WP_RO@0x41000 0x6bf000 {
                       FMAP@0x0 0x800
                       RO_FRID@0x800 0x40
                       RO_PADDING@0x840 0x7c0
[...]

Issue #517, opened 2019-02-08
Reduce  libslang.so.2 library footprint


tlaurion:
@kylerankin : doing some tests to reduce libraries footprints, since x200 and other hardware that don't have a TPM also tend to have smaller SPI flash chips (4mb on some, 8mb on others).

Whiptail depends on a 1.1M libslang.so.2 library, which is the biggest of all, with currently deactivated features:
```
--with-z=no \
--with-png=no \
--with-pcre=no \
--with-onig=no \
```

Can we deactivate some other unneeded features? Here is the list:
```
--with-terminfo=no \
--with-readline=no \
--with-slsyswrap=no \
--with-x=no \
--with-pcre=no \
--with-pcrelib=no \
--with-pcreinc=no \
--with-onig=no \
--with-oniglib=no \
--with-oniginc=no \
--with-png=no \
--with-pnglib=no \
--with-pnginc=no \
--with-z=no \
--with-zlib=no \
--with-zinc=no \
--with-iconv=no \
--with-iconvlib=no \
--with-iconvinc=no \
```

Edit: even if all of those are turned off, the library still pounds 1.1M. Any advice?

tlaurion:
@kylerankin : any idea?

tlaurion:
@MrChromebox : any idea here in link with #590 ?

MrChromebox:
@tlaurion unless I'm missing something, the only in-tree board I see including the slang module is qemu-coreboot (both explicitly and via CONFIG_NEWT).

tlaurion:
@MrChromebox and when whiptail is requested without full graphical interface (#453 ) to be able to use gui-init from command line without fbwhiptail for lesser space environements. (Not really a prority as of now).


those are combined space used per gui-init with fbwhiptail, current ticket being without fbwhiptail.
```
790103 832 -rwx------ 1 user user 851968 Jun 28 13:01 ./lib/libcairo.so.2
790113 652 -rwx------ 1 user user 667048 Jun 28 13:01 ./lib/libpixman-1.so.0
790114 200 -rwx------ 1 user user 201096 Jun 28 13:01 ./lib/libpng16.so.16
```

PR #565, opened 2019-05-20
config/linux-x230: remove debug symbols

merge:
Remove debug symbols and dynamic debug support from Linux. This reduces
our resulting bzImage by 100K.

This should help when size becomes even more limited, see #562


merge:
for the record, I flashed that and did a test run.

paulmenzel:
On 20.05.19 13:10, Martin Kepplinger wrote:
> Remove debug symbols and dynamic debug support from Linux. This reduces
> our resulting bzImage by 200K.

How big was the image before?

merge:
thanks for asking, I was wrong and updated the commit message. It's 100K: 3031K -> 2936K. circa.

merge:
With this reduction I was able to do a test-build using coreboot's vboot measured boot. (not without this reduction - we really are at our size limit currently).

The usable cbfs size (fmap see #562) for our vboot-case is not yet defined. This should just be a beginning and we should become even smaller after this!

tlaurion:
#517 is the biggest culprit

merge:
> #517 is the biggest culprit

I see. it is good to have another promising candidate to work on. Unless we need to have debug symbold by default, it doesn't make this change less useful though, right? I'd say let's reduce our size where possible without losing any functionality.

Issue #536, opened 2019-03-17
QubesOS Xen CMDLINE currently not respected


tlaurion:
As of right now, grub.cfg Xen command line looks like this (dom0/Heads: `cat /boot/kexec_default.1.txt`):
`Qubes, with Xen 4.8.5 and Linux 4.14.74-1.pvops.qubes.x86_64|xen|kernel /xen-4.8.5.gz placeholder console=none dom0_mem=min:1024M dom0_mem=max:4096M iommu=no-igfx ucode=scan smt=off |module /vmlinuz-4.14.74-1.pvops.qubes.x86_64 placeholder root=/dev/mapper/qubes_dom0-root ro rd.luks.uuid=luks-caec4f1d-3370-4690-9958-4fa8247d2557 rd.luks.options=discard rd.lvm.lv=qubes_dom0/root rd.lvm.lv=qubes_dom0/swap i915.alpha_support=1 intel_iommu=on rhgb quiet rd.qubes.hide_all_usb|module /initramfs-4.14.74-1.pvops.qubes.x86_64.img`

Currently passed Xen command line (dom0: `xl info | grep commandline`):
`xen_commandline        : no-real-mode reboot=no vga=current`

Expected:
`placeholder console=none dom0_mem=min:1024M dom0_mem=max:4096M iommu=no-igfx ucode=scan smt=off no-real-mode reboot=no vga=current`

tlaurion:
As a result, QubesOS recommended Xen configurations are flushed. This has the following preset consequences:
-  lack of dom0_mem=max:4096M may in many cases prevent starting sys-net and sys-usb (not enough continuous memory)
-  lack of smt=off leave hyper threading enabled, which unfortunately nowadays is a security issue (namely L1TF)

tlaurion:
@kylerankin : please test on Librems: QubesOS provided grub.cfg Xen options include `iommu=no-igfx` and `smt=off`. We may need to duplicate board config `CONFIG_BOOT_KERNEL_REMOVE` and `CONFIG_BOOT_KERNEL_ADD` logic for XEN if those standard options have impact for Purism.

@flammit "Approved" this PR from his comment on slack:
`if it works on your x230 w/ Qubes, i guess it LGTM.  i only ignored the args because I didn’t know what the interaction was between these options and the ones in the Qubes grub`

PR #568, opened 2019-05-21
(WIP) update coreboot

merge:
(WIP) update coreboot to 4.11
    
    DO NOT MERGE! This is for testing only!
    
    Update coreboot and configs from 4.8.1 to 4.11 and port the needed patches.
    
    For an overview of the changes in coreboot, see
    https://doc.coreboot.org/releases/coreboot-4.9-relnotes.html
    https://doc.coreboot.org/releases/coreboot-4.10-relnotes.html
    https://doc.coreboot.org/releases/coreboot-4.11-relnotes.html
    
    TODO:
    * coreboot defconfig update for all boards!
      (configure FSP_USE_REPO?) Needed for skylake/kabylake
    * test


MrChromebox:
coreboot patches 0001/0002/0003 need all instances of 'IS_ENABLED(CONFIG_XXX)' replaced with 'CONFIG(XXX)'.  I count 6 lines / 8 instances that need to be changed across the 3 patches

merge:
> coreboot patches 0001/0002/0003 need all instances of 'IS_ENABLED(CONFIG_XXX)' replaced with 'CONFIG(XXX)'. I count 6 lines / 8 instances that need to be changed across the 3 patches

done that and updated. thanks

tlaurion:
https://github.com/osresearch/heads/issues/605

MrChromebox:
I've been working on this from my end as well, and still seeing issues with coreboot 4.10 + Xen/Qubes. On an x230, Qubes will fail to start sys-net and sys-firmware at boot, same symptoms as issue #536. On the Librems, Xen freezes on the kexec. (there's also a patch needed for all SKL/KBL devices or the system is completely broken)

I'm wondering if we should just jump ahead to 4.11 at this point given it was just tagged

merge:
> 
> I'm wondering if we should just jump ahead to 4.11 at this point given it was just tagged

sure we should; and stick with that as long as it's convenient here.

tlaurion:
@MrChromebox any info on what happens at kexec?

MrChromebox:
> @MrChromebox any info on what happens at kexec?

negative, the screen just locks after printing the kernel command line info and kexec'ing

sylv-io:
hey folks, regarding #605 #616  I'm working on Measured Boot support for mors Sandy/Ivy bridge Thinkpads, where coreboot version 4.11 is necessary. Any progress for the update ?

MrChromebox:
@sylv-io what approach are you taking for measured boot support? Using vboot w/4.11? backporting the patch that separates out measured boot from vboot? adapting the existing Heads measured boot patch?

sylv-io:
> @sylv-io what approach are you taking for measured boot support? Using vboot w/4.11? backporting the patch that separates out measured boot from vboot? adapting the existing Heads measured boot patch?

Using measured boot implementation from  coreboot 4.11. So we can create the missing target boards. Are there any missing patch in 4.11 for heads?

MrChromebox:
@sylv-io so using vboot with A/B booting. We can't use that across the board as some (many) devices lack the necessary room in flash to store multiple copies of the kernel. Also, Qubes doesn't boot with coreboot 4.11 (4.10+ actually) + Heads currently, at least on the devices I've tested.

tlaurion:
@MrChromebox : have you played with https://review.coreboot.org/c/coreboot/+/35077 ?

sylv-io:
> @sylv-io so using vboot with A/B booting. We can't use that across the board as some (many) devices lack the necessary room in flash to store multiple copies of the kernel.

Yes, this is a big problem. Therefore I'm using vboot only with a ro partition and reduce the ME region to the minimal size. In this way it does fit (tested with a x220 with 8MiB).

> Also, Qubes doesn't boot with coreboot 4.11 (4.10+ actually) + Heads currently, at least on the devices I've tested.

too bad.. do you know the reason why ?

MrChromebox:
@tlaurion I have, that patch has gone thru a lot of changes, and can't be applied to coreboot 4.11 as it depends on a lot of commits since the tagged release. I tried an older patchset at the suggestion of the author but it too would require a lot of massaging to apply against 4.11

@sylv-io I'm working on the Qubes issue

tlaurion:
@PatrickRudolph https://github.com/osresearch/heads/pull/568#issuecomment-596651142 ?
Questioning the interest of VBOOT+measuredboot in link with #690 and needed space for 8mb SPI flash.

PatrickRudolph:
> @PatrickRudolph [#568 (comment)](https://github.com/osresearch/heads/pull/568#issuecomment-596651142) ?
> Questioning the interest of VBOOT+measuredboot in link with #690 and needed space for 8mb SPI flash. The more I think about it and Heads use case, the less I see interest in going the VBOOT way at all.

The maximum CBFS size with VBOOT+measured boot is 0x7a0000

zaolin:
At the moment vboot is the only useful solution since 35077 never gets merged and the previous patch of trammel is incomplete. The only solution IMHO is to strip down HEADS as I said before. Wasting more and more space in the SPI flash makes it unusable anyway, even without vboot. I would recommend starting to identify the space consumption of all HEADS components

tlaurion:
> I would recommend starting to identify the space consumption of all HEADS components
> 
#590 
> Update: Getting rid of libgcrypt and replacement for gpg would be a good way to safe 2MB in total.
> See, https://sequoia-pgp.org/ as alternative
Will look, thanks.

tlaurion:
Closing per [agreement](https://github.com/osresearch/heads/pull/721#issuecomment-656783863)

PR #569, opened 2019-05-22
Librem 4.10 update

MrChromebox:
Update coreboot defconfigs, get_blobs scripts, and readme files for Purism Librem devices

should be merged after #568


merge:
These changes are now part of the mentioned #568 PR. In case they change here I can just update them over there.

MrChromebox:
these have been pulled into #568, so no longer needed as a standalone PR

PR #576, opened 2019-05-24
gui-init: reorder, simplify main menu entries and add Power Off entry

merge:
Clean up the main menu to only include menu entries that are expected
to be part of users' everyday routines. On the main menu itself, each option
should be almost trivially easy to understand.

* move "Exit to recovery shell" under Advanced Settings (quite obvious change IMO)
* add "Refresh TOTP/HOTP". Needed by anybody who verifies firmware integrity using TOTP
* keep "Advanced Settings" for the other stuff most people won't care too much about, but rename to "Settings" only. (we don't have "normal" ones, so why have "advanced" ones, right?)
* add "Power Off" at the bottom. IMO this is can be very useful to anybody: This enables users to
only verify the firmware integrity using OTP, and do nothing more. After having left the device out of sight, one might want to do a quick sanity check only.

That's the resulting menu after these changes:
* Default boot
* Refresh TOTP/HOTP
* Settings -->
* Power Off

### DIFF
--- a/initrd/bin/gui-init
+++ b/initrd/bin/gui-init
@@ -176,7 +176,6 @@ while true; do
       'y' ' Default boot' \
       'a' ' Advanced Settings -->' \
       'P' ' Power Off' \
-      'x' ' Exit to recovery shell' \
       2>/tmp/whiptail || recovery "GUI menu failed"
 
     totp_confirm=$(cat /tmp/whiptail)
@@ -191,6 +190,7 @@ while true; do
       'c' ' Change configuration settings -->' \
       'f' ' Flash/Update the BIOS -->' \
       'G' ' GPG Options -->' \
+      'x' ' Exit to recovery shell' \
       'r' ' <-- Return to main menu' \
       2>/tmp/whiptail || recovery "GUI menu failed"
 
--- a/initrd/bin/gui-init
+++ b/initrd/bin/gui-init
@@ -174,6 +174,7 @@ while true; do
     whiptail $MAIN_MENU_BG_COLOR --clear --title "$CONFIG_BOOT_GUI_MENU_NAME" \
       --menu "$date\nTOTP: $TOTP | HOTP: $HOTP" 20 90 10 \
       'y' ' Default boot' \
+      'r' ' Refresh TOTP/HOTP' \
       'a' ' Settings -->' \
       'P' ' Power Off' \
       2>/tmp/whiptail || recovery "GUI menu failed"
@@ -214,7 +215,6 @@ while true; do
       --menu "Select An Option" 20 90 10 \
       'g' ' Generate new TOTP/HOTP secret' \
       'p' ' Reset the TPM' \
-      'r' ' TOTP/HOTP does not match, refresh code' \
       'n' ' TOTP/HOTP does not match after refresh, troubleshoot' \
       'r' ' <-- Return to main menu' \
       2>/tmp/whiptail || recovery "GUI menu failed"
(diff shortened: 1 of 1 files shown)


merge:
Since this is so visible: @MrChromebox @kylerankin @tlaurion any objections or thoughts? thanks.

tlaurion:
Escape refreshes TOTP. 

Another implementation could be to show " poweroff" and "refresh TOTP" only  if LIMBREMKEY isn't exported in board configuration. But I wasn't too successful in creating dynamic whiptail menu in the past, maybe @kylerankin could give insight on this.

merge:
thanks for thst hint. no need to add complexity now IMO

Am 25. Mai 2019 04:43:13 MESZ schrieb tlaurion <notifications@github.com>:
>Escape refreshes TOTP. 
>
>Another implementation could be to show " poweroff" and "refresh TOTP"
>only  if LIMBREMKEY isn't exported in board configuration. But I wasn't
>too successful in creating dynamic whiptail menu in the past, maybe
>@kylerankin could give insight on this.

-- 
Martin Kepplinger
http://martinkepplinger.com
sent from mobile

tlaurion:
Maybe that escape key hint to refresh TOTP, and hitting 'r' before launch of gui-init or generic-init (in main unit) should be documented in a heads keyboard shortcuts?

Keyboard shortcuts
'r': when hit repeatedly prior to GUI, launches Heads recovery shell
'Esc': Refreshes TOTP code in graphical GUI
'Power Button': push and hold for 5-10 seconds on any system to force an immediate power cut-off.

tlaurion:
I don't mind the change personally. @kylerankin? @flammit?

kylerankin:
Does the ESC button actually refresh gui-init? I didn't think fbwhiptail supported hitting keys that get immediately processed. I thought you had to select a GUI option with arrow keys and press Enter.

I guess I don't understand why you'd want to add a "power off" menu option. If you pressed and held a "power off" key, you could just press and hold your actual power off button on your computer. Is there a use case I'm not thinking of where holding your physical power button isn't an option?

MrChromebox:
> Does the ESC button actually refresh gui-init?

annoyingly, yes. IMO, the expected behavior is to return to the previous menu, not return back to the main menu (complete with any pop-up warnings)

> I guess I don't understand why you'd want to add a "power off" menu option.

I think the main reason would be to avoid having to hold the power button for ~10s to power off. Normal pre-OS behavior is to power off immediately (eg, how SeaBIOS behaves) when the power button is pressed, but since Heads is a booted OS, the power button acts accordingly

merge:
> Does the ESC button actually refresh gui-init? I didn't think fbwhiptail supported hitting keys that get immediately processed. I thought you had to select a GUI option with arrow keys and press Enter.
> 
> I guess I don't understand why you'd want to add a "power off" menu option. If you pressed and held a "power off" key, you could just press and hold your actual power off button on your computer. Is there a use case I'm not thinking of where holding your physical power button isn't an option?

It's an option that users generally think of some mean of "last resort" for cutting power, something that might even "damage" the system and so on. I even found myself booting up and powering down from Debian once :)

So this way we support the "Quickly check firmware integritiy only" usecase for everybody.

I'd wouldn't underestimate a "Power Off" Menu Entry, also because of it's simplicity. Heads gets familiar and make an impression of being simple to understand, from the beginnen. "Boot" or "Start" at the top (selected by default), refresh the code and settings below in order to "work with" Heads, and "Power Off" at the bottom. That seems like a way more intuitive layout than we have right now.

merge:
> > Does the ESC button actually refresh gui-init?
> 
> annoyingly, yes. IMO, the expected behavior is to return to the previous menu, not return back to the main menu (complete with any pop-up warnings)

I reverted the change here too. I thought about it, and I think it has it's own value to have useful, but trivial to understand and use, menu entries. It's actually also more consistent UX-wise, not to misuse ESC here. Let's defined that independently, as you said, return back 1, or something (later).

merge:
fyi, i've alway updated this PR's first comment at the top with the resulting main boot menu. I'm running it and IMO it's much more beautiful and _useful_ now. any other concerns?

tlaurion:
@kylerankin no opposition here. You can merge if you have none either.

PR #539, opened 2019-03-20
Add support for T420 and T430, enable FBWhiptail for X220.

SebastianMcMillan:
The T420 support is currently untested, but currently builds and _should_ work.

Blob extraction scripts now autodetect coreboot's ifdtool and me_cleaner.

PR #578, opened 2019-06-07
T420 initial support + X220 FBWhiptail Support

SebastianMcMillan:
Separating my original PR #539. 
This branch introduces T420 support (complete with FBWhiptail support) and modifications to the X220 to support FBWhiptail.

No known new regressions are present, and this does not fix the 50 second delay that is currently present.


boards/t420/t420.config
```diff
@@ -0,0 +1,30 @@
+# Configuration for a x220 running Qubes and other OS
+# The Linux configuration is close enough to the x230
+export CONFIG_COREBOOT=y
+CONFIG_COREBOOT_CONFIG=config/coreboot-x220.config
```
merge: is that correct?

boards/t420/t420.config
```diff
@@ -0,0 +1,30 @@
...
+CONFIG_LINUX_CONFIG=config/linux-x230.config
```
merge: if correct, we should add a comment, why, I think

boards/t420/t420.config
```diff
@@ -1,12 +1,12 @@
-# Configuration for a T420 running Qubes and other OS
-# The Linux configuration is close enough to the x230
+# Configuration for a T420 running Qubes and other OS, X220 is identical to X230 on the Linux Side of things.
 export CONFIG_COREBOOT=y
-CONFIG_COREBOOT_CONFIG=config/coreboot-t420.config
+CONFIG_COREBOOT_CONFIG=config/coreboot-T420.config
```
merge: really uppercase T?

boards/t420/t420.config
```diff
@@ -27,7 +27,7 @@ export CONFIG_TPM=y
 export CONFIG_BOOTSCRIPT=/bin/gui-init
 export CONFIG_BOOT_REQ_HASH=n
 export CONFIG_BOOT_REQ_ROLLBACK=n
-export CONFIG_BOOT_KERNEL_ADD=""
+export CONFIG_BOOT_KERNEL_ADD="intel_iommu=on"
```
merge: why not in the first place, when you add the file?

boards/t420/t420.config
SebastianMcMillan: This is correct. If the already existing X220 is close enough to X230, and T420 is close enough to X220, T420 is close enough to X230. 

Tested on my T420 as working.

boards/t420/t420.config
SebastianMcMillan: I don't quite understand what you mean?

SebastianMcMillan:
> why would you add and revert a commit in 1 PR?
either temporary changes that I didn't mean to pull in. I'm still learning how to use git in larger projects.

> we need to review this seperately and this has to be a different PR. interesting...
This was suggested by you, after all.

> why not do 2 commit when changing 2 board configs?
I was working on troubleshooting two boards at that time.

> This could be squashed into the previous change that adds t420.config
That would require me to figure out squashing, and I do not have access to command line git at this time.

> the commit message should better read "blobs/t420: remove gitignore"
I would change that if I could.

SebastianMcMillan:
This branch's flash.sh changes are reverted to conform with #592's way of supplying flash definitions.

tlaurion:
@SebastianMcMillan Have you had a chance to [test](https://github.com/corna/me_cleaner/wiki/Internal-flashing-with-coreboot#neutralize-and-shrink-intel-me) including ifd and reduce me and changing CBFS size in coreboot config and see if the 60 seconds delay linked to ME disappeared?

SebastianMcMillan:
I've not yet had the chance, but I will get around to it by the end of the
month.

techge:
> @SebastianMcMillan Have you had a chance to test including ifd and reduce me and changing CBFS size in coreboot config and see if the 60 seconds delay linked to ME disappeared?

As mentioned in #539 I followed the guide and truncated the ME, still the boot delay is valid.

techge:
Commenting `CONFIG_MEASURED_BOOT` [here](https://github.com/osresearch/heads/blob/0e1f36be37819ac6b211d8b7834e0edb1bfe6aad/config/coreboot-x220.config#L5) stops the delay. So that means it is a TPM issue, isn't it?

SebastianMcMillan:
I don't believe this is an ME issue. I can use the exact same IFD and ME with stock coreboot and we don't see this issue. Furthermore, using stock ME and IFD still reproduces this issue.

This issue is certainly a 4.8.1 issue. I've tested pre 4.8.1 heads on my X220 and can confirm the issue is not present. At this point, I will wait for heads to be based on 4.11+ and see if it's still an issue.

techge:
I would like to already play around a bit. So may I ask how I could switch to old coreboot in the meantime?

techge:
Okay, I used this PR as a basis for my [own branch](https://github.com/osresearch/heads/pull/578) and combined with the [hint](https://github.com/osresearch/heads/pull/539#issuecomment-576083869) from @tlaurion it actually works fine now for me!

Would @SebastianMcMillan like to separate the x220 stuff as I did and create a new PR as this is tested now, shall we wait or shall I create a PR based on the branch I mentioned above?

SebastianMcMillan:
I'll get on that tonight.

tlaurion:
@SebastianMcMillan ?

SebastianMcMillan:
I apologize on the delay, my T420 is running into some difficulties, and I've been a little caught up with coursework. I am working on testing @tlaurion's suggestion on the X220 as well.

SebastianMcMillan:
T420 and X220 boot delay fixes tested and confirmed working on my end.

PR #579, opened 2019-06-07
X230 flash universal init

SebastianMcMillan:
This change makes the x230-flash.init file into a universal install.init, letting one reuse the install.init for other split SPI flash boards, such as the T430.


boards/x230-flash/x230-flash.config
```diff
@@ -16,7 +16,7 @@ CONFIG_LINUX_USB=y
 CONFIG_LINUX_E1000E=y
 
 export CONFIG_USB_BOOT_DEV="/dev/sdb1"
-export CONFIG_BOOTSCRIPT=/bin/x230-flash.init
+export CONFIG_BOOTSCRIPT=/bin/install.init
```
merge: "updating" and "x230-flash.config" seems to be wrong in the commit message.

initrd/bin/install.init
```diff
@@ -20,7 +20,7 @@ echo ''
 echo 'To install from flash drive:'
 echo ''
 echo '  mount -o ro /dev/sdb1 /media'
-echo '  flashrom-x230.sh /media/x230.rom'
+echo '  flash.sh /media/<rom-file-name>'
 echo ''
```
merge: has already been fixed...

boards/x230-flash/x230-flash.config
merge: oh sorry, I was reading it wrong. the commit message is fine :)

merge:
I would prefer a new <laptop-model>-flash.init script. And if we then see that it really can be exactly the same for different models, we can rename.

Adding support for a new board itself would be nice not to touch others. Is it just me?

initrd/bin/install.init
SebastianMcMillan: it wasn't back when those changes were actually made.

SebastianMcMillan:
Seeing how the x230-flash.init doesn't do anything specific to the x230, I figure it would be easier to make a universal install.init, not only reducing the amount of clutter we have in the initrd, but also is easier to work with.

tlaurion:
@merge flash specifics are defined under flash.sh anyway. Each board addition should be added to it anyway. I like the idea to generalize!

initrd/bin/install.init
tlaurion: > has already been fixed...

@merge, where?

tlaurion:
> I would prefer a new -flash.init script. And if we then see that it really can be exactly the same for different models, we can rename.
> 
> Adding support for a new board itself would be nice not to touch others. Is it just me?

@merge: I would love to better understand your stance following [past comment](https://github.com/osresearch/heads/pull/579#issuecomment-500083390)

initrd/bin/install.init
merge: in https://github.com/osresearch/heads/commit/00bf4740e3c08feb1870f4c7b4dd7823681bf22c

merge:
> > I would prefer a new -flash.init script. And if we then see that it really can be exactly the same for different models, we can rename.
> > Adding support for a new board itself would be nice not to touch others. Is it just me?
> 
> @merge: I would love to better understand your stance following [past comment](https://github.com/osresearch/heads/pull/579#issuecomment-500083390)

I think x230-flash.init can indeed be renamed to something more generic, if we add another "xxxx-flash" board.

tlaurion:
> > > I would prefer a new -flash.init script. And if we then see that it really can be exactly the same for different models, we can rename.
> > > Adding support for a new board itself would be nice not to touch others. Is it just me?
> > 
> > 
> > @merge: I would love to better understand your stance following [past comment](https://github.com/osresearch/heads/pull/579#issuecomment-500083390)
> 
> I think x230-flash.init can indeed be renamed to something more generic, if we add another "xxxx-flash" board

@merge: and my point is that the boards specific flashroms options currently need to be defined under [flash.sh](https://github.com/osresearch/heads/blob/master/initrd/bin/flash.sh#L14), the proper flashrom call being defined per $CONFIG_BOARD variable, put dynamically under /etc/config, sourced by init and all other scripts under Heads.

The only current exception in boards being if the board in question can flash another SPI (eg: iKVM) from within, like in the case of the [kgpe-d16](https://github.com/osresearch/heads/blob/master/initrd/bin/flashrom-kgpe-d16-openbmc.sh)

tlaurion:
@merge @SebastianMcMillan update?

tlaurion:
@SebastianMcMillan please rebase.
@MrChromebox : I think this should be merged, personally.

MrChromebox:
@tlaurion needs a rebase obv since the one commit isn't needed anymore. I think 'install.init' isn't the best name, but I can live with it

SebastianMcMillan:
If the plan is to drop the 2-part flash in favor of flashing both chips externally, then I don't think this PR should be merged.

tlaurion:
@SebastianMcMillan gbe and me blobs are still nit.resolved problems.

boards/x230-flash/x230-flash.config
tlaurion: @SebastianMcMillan This part would still be nice.

SebastianMcMillan:
Now that I've got an X230 again, I'll have to take a look at it sometime
soon.

PR #595, opened 2019-07-22
Adding flash tools in x220 images

spacekookie:
Without this setting, following the install guide [1] becomes
impossible, as the `cbfs` command is missing. This change
mirrors the state of the x230 config, which is required to
build a working system.

[1]: http://osresearch.net/Installing-Heads.html


SebastianMcMillan:
These changes are already present in #578, which also provide other quality of life changes such as FBWhiptail and HOTP support among others.

I could use more testers to validate said changes, however, maybe to actually get some of these improvements into master.

tlaurion:
X220 with fbwhiptail merged

PR #601, opened 2019-08-11
Supermicro x11ssh coreboot configuration

osresearch:
Supermicro X11SSH-F board

The Supermicro X11SSH is a modern server platform that now supports coreboot thanks porting efforts of [9elements](https://9esec.io/blog/first-modern-coreboot-server-platform/) in partnership with [Mullvad's system transparency](https://mullvad.net/en/blog/2019/6/3/system-transparency-future/) and [open firmware future](https://mullvad.net/en/blog/2019/8/7/open-source-firmware-future/) projects.

This is a work in progress due to the switch to the mainline coreboot head, rather than a released version. Serial console, ethernet (IGB) and SATA work. Haven't tried video (AST2400) or TPM.

The only non-public blob necessary is the ME region. The FSP is built from Intel's publically available binaries for Kabylake.


citypw:
heads stuck at boot time and the only output is "ASpeed VGA text mode initalized". But coreboot with seabios seem working perfect so far. Any idea how to fix it?

osresearch:
I haven't tried booting with a monitor (no VGA displays home), so I'm using the serial console. Even with the AMI UEFI and the AST driver there were weird cutouts in the BMC's KVM during boot, although I didn't try to debug it too much.

citypw:
@MrChromebox Thanks for the tips. 

I also noticed that it will be hung for a few mins in coreboot w/ seabios. Will try heads later.

citypw:
I checked where it stuck via serial console. The log is here:

--------------------
***** Normal boot: /bin/generic-init
y) Default[    3.032050] NOHZ: local_softirq_pending 80
[    3.037584] NOHZ: local_softirq_pending 282
[    3.041775] NOHZ: local_softirq_pending 282
[    3.045962] NOHZ: local_softirq_pending 282
[    3.050147] NOHZ: local_softirq_pending 282
[    3.054334] NOHZ: local_softirq_pending 282
[    3.058521] NOHZ: local_softirq_pending 282
[    3.062708] NOHZ: local_softirq_pending 282
[    3.066901] NOHZ: local_softirq_pending 282
[    3.071085] NOHZ: local_softirq_pending 282
 boot
n) TOTP does not match
r) Recovery boot
u) USB boot
m) Boot menu
Unable to retrieve sealed file from TPM NV
!!!!! TOTP code generation failed
New value of PCR[4]: 8a6a96fde1a8dd96271479dc40742b36aba3c2b3
!!!!! Starting recovery shell
--------------------

It entered into the recovery shell eventually. But my USB keyboard ain't working while the serial console seems doesn't work on keyboard input( from HOST side). Any suggestion about workaround?

MrChromebox:
> It entered into the recovery shell eventually. But my USB keyboard ain't working while the serial console seems doesn't work on keyboard input( from HOST side). Any suggestion about workaround?

I added loading of USB kernel modules to gui-init in 4a85c853, but not to normal init -- you'll need to do something similar for the USB keyboard to function on the host

citypw:
Thanks, the keyboard works. And the input on serial console works with disabled "hardware flow control". But the minicom isn't able to display the correct QR code after seal-totp. Any idea to fix it?

MrChromebox:
I'm not sure I ever tested that far - I only used the serial console to debug the GUI not working, and once I loaded the AST DRM kernel module I switched over to that

osresearch:
The board support is now in the coreboot mainline: https://review.coreboot.org/plugins/gitiles/coreboot/+/08aa502d79d04a13c56293021cd66d9c3c270f97

So we can remove the patch and some of the other hackery.

There is also a WIP to support this board with u-bmc, removing even more vendor firmware. https://github.com/u-root/u-bmc/issues/186

MrChromebox:
true, but I'd expect the next baseline to be coreboot 4.10, so will still need the patch to add the board until we rebase on 4.11

citypw:
I changed native graphic init to VGA BIOS which dumped from the fw. The monitor will show heads logo and stuck there so the keyboard isn't working even the initialization of USB keyboard has finished. I'll have to input via the serial console. Is the possible some kernel options needed to be enable since the default config is based on linuxboot?

Yeah, looking forward to see heads's baseline to be the next release of coreboot.

citypw:
I think I misunderstand the linuxboot config which is enabled earlyprint=serial. Just removed that option and problem sovled! Thank you guys.

tlaurion:
> Thanks, the keyboard works. And the input on serial console works with disabled "hardware flow control". But the minicom isn't able to display the correct QR code after seal-totp. Any idea to fix it?

https://github.com/osresearch/heads/issues/603#issuecomment-570952947

tlaurion:
@osresearch : this needs to be rebased to master (got a notification that something changed here, don't see what)

tlaurion:
Probably linked to testing #845

tlaurion:
Abandoned

PR #606, opened 2019-08-21
OEM factory reset

MrChromebox:
Add an improved OEM Factory reset function, which erases/resets the TPM and GPG security token, generates a new GPG key, signs the files in /boot and sets the default boot device / boot selection.

Add a new item in the main options menu for it, and remove the existing GPG reset function since it is superseded by this.

Add a function to detect a clean boot situation, and prompt the user to run the OEM factory reset.

### DIFF
--- a/initrd/bin/gpg-gui.sh
+++ b/initrd/bin/gpg-gui.sh
@@ -164,29 +164,6 @@ gpg_post_gen_mgmt() {
       gpg_flash_rom
   fi
 }
-gpg_sc_oem_reset() {
-  GPG_KEY_NAME=`date +%Y%m%d%H%M%S`
-  # Factory reset GPG card
-  {
-    echo admin
-    echo factory-reset
-    echo y
-    echo yes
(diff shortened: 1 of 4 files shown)


tlaurion:
Are we redoing piece by piece what was already done in OEM reownership wizard pr?

kylerankin:
In this case the existing automated OEM reset wizard in the GPG menu still required a lot of intervention and prompts that weren't necessary so this PR removed the remaining prompts to make it completely non-interactive. Because it did more than just reset GPG (it also needed to re-sign files in /boot etc) it made more sense to move it up one menu as it impacted more than GPG. Other than that from my reading this is just a simplified and further automated form of what was already merged in a few months ago.

kylerankin:
I've tested this and it's much improved over the previous automated GPG reset function I wrote and handles edge cases better while also not bothering the user w/ unnecessary prompts. Merging.

tlaurion:
I see another borrow from PrivacyBeast without setting proper randomized secrets. 

I have a question to Purism. How many ideas will be taken from my branch without proper credit and code rewrites?

PR #607, opened 2019-08-27
TOTP Tweaks

MrChromebox:
Reduce friction when generating a new TOTP/HOTP secret by eliminating an unnecessary 'press enter to continue' prompt following QR code generation, and by attempting to use the default admin PIN set by the OEM factory reset function. Fall back to prompting the user if the default PIN fails.

Update text on TOTP error prompt to provide better guidance for users following the use of the OEM factory reset function

Also, ensure error messages are visible to users before being returned back to the GUI menu from which they came by wrapping existing calls to die()

### DIFF
--- a/initrd/bin/gui-init
+++ b/initrd/bin/gui-init
@@ -90,11 +90,9 @@ prompt_update_checksums()
 }
 update_totp()
 {
-  echo "Scan the QR code to add the new TOTP secret"
+  echo -e "Scan the QR code to add the new TOTP secret...\n\n"
   /bin/seal-totp
   if [ -x /bin/libremkey_hotp_verification ]; then
-    echo "Once you have scanned the QR code, hit Enter to configure your Librem Key"
-    read
     /bin/seal-libremkey
   else
     echo "Once you have scanned the QR code, hit Enter to continue"
--- a/initrd/bin/gui-init
+++ b/initrd/bin/gui-init
@@ -169,8 +169,14 @@ while true; do
       TOTP=`unseal-totp`
       if [ $? -ne 0 ]; then
         whiptail $CONFIG_ERROR_BG_COLOR --clear --title "ERROR: TOTP Generation Failed!" \
-          --menu "ERROR: Heads couldn't generate the TOTP code.\n\nIf this is the first time the system has booted, you should reset the TPM\nand set your own password\n\nIf you just reflashed your BIOS, you'll need to generate a new TOTP secret.\n\nIf you have not just reflashed your BIOS, THIS COULD INDICATE TAMPERING!\n\nHow would you like to proceed?" 30 90 4 \
-          'g' ' Generate new TOTP/HOTP secret' \
+          --menu "    ERROR: Heads couldn't generate the TOTP code.\n
+    If you have just completed a Factory Reset, or just reflashed
+    your BIOS, you should generate a new HOTP/TOTP secret.\n
+    If this is the first time the system has booted, you should
+    reset the TPM and set your own password.\n
+    If you have not just reflashed your BIOS, THIS COULD INDICATE TAMPERING!\n
+    How would you like to proceed?" 30 90 4 \
+          'g' ' Generate new HOTP/TOTP secret' \
           'i' ' Ignore error and continue to default boot menu' \
           'p' ' Reset the TPM' \
           'x' ' Exit to recovery shell' \
--- a/initrd/bin/seal-libremkey
+++ b/initrd/bin/seal-libremkey
@@ -16,17 +16,24 @@ mount_boot()
   fi
 }
 
+fatal_error()
+{
+  echo -e "\nERROR: ${1}; press Enter to continue."
+  read
+  die "$1"
+}
+
 tpm nv_readvalue \
 	-in 4d47 \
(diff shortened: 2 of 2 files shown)


tlaurion:
Not sure about consuming one attempt out of 3 with default oem password here.

On August 27, 2019 2:24:41 AM UTC, MrChromebox <notifications@github.com> wrote:
>Reduce friction when generating a new TOTP/HOTP secret by eliminating
>an unnecessary &#39;press enter to continue&#39; prompt following QR
>code generation, and by attempting to use the default admin PIN set by
>the OEM factory reset function. Fall back to prompting the user if the
>default PIN fails.
>
>Update text on TOTP error prompt to provide better guidance for users
>following the use of the OEM factory reset function
>
>Also, ensure error messages are visible to users before being returned
>back to the GUI menu from which they came by wrapping existing calls to
>die()
>You can view, comment on, or merge this pull request online at:
>
>  https://github.com/osresearch/heads/pull/607
>
>-- Commit Summary --
>
>  * gui-init: update TOTP error prompt
>  * gui-init/seal-libremkey: reduce friction when generating new secret
>
>-- File Changes --
>
>    M initrd/bin/gui-init (14)
>    M initrd/bin/seal-libremkey (47)
>
>-- Patch Links --
>
>https://github.com/osresearch/heads/pull/607.patch
>https://github.com/osresearch/heads/pull/607.diff
>
>-- 
>You are receiving this because you are subscribed to this thread.
>https://github.com/osresearch/heads/pull/607

-- Sent from /e/ Mail

MrChromebox:
> Not sure about consuming one attempt out of 3 with default oem password here.

I'm not seeing any limit on the number of attempts when calling libremkey_hotp_initialize with an admin (vs user) pin. Am I missing something there?

tlaurion:
From my understanding there is 3 attempts before locking specific functions of the key.

3 bad user PINs locks user functions and requires the admin PIN to unlock user functions back to normal.

3 bad admin PINs locks admin and requires the reset PIN to unlock admin functions back to normal.

After it is locked, the only way to unlock admin fonctions is through a reset PIN, not currently serltuped neither in my gpg factory reset script nor in Purism's factory reset code.

No?
I don't have any spare Librem Key to test right now.

On August 30, 2019 4:38:02 AM UTC, MrChromebox <notifications@github.com> wrote:
>> Not sure about consuming one attempt out of 3 with default oem
>password here.
>
>I'm not seeing any limit on the number of attempts when calling
>libremkey_hotp_initialize with an admin (vs user) pin. Am I missing
>something there?
>
>-- 
>You are receiving this because you commented.
>https://github.com/osresearch/heads/pull/607#issuecomment-526452332

-- Sent from /e/ Mail

MrChromebox:
I do see in gnupg's documentation that 3 successive incorrect attempts to login with an admin PIN destroys the card -- which is really, really, bad given we don't warn the user at all, and most are likely completely unaware.

kylerankin:
There are three separate secrets on a Librem Key or Nitrokey:
1. GPG admin PIN on the OpenPGP smart card
2. GPG user PIN on the OpenPGP smart card
3. HOTP admin PIN on the Librem Key/Nitrokey flash chip

3 failures on #1 apparently destroys the smart card to protect the PIN (but I think this is configurable). In this case the HOTP secret and those functions would still work though--it's independent of the smart card
3 failures on #2 locks the user and you have to use the admin PIN to unlock it again to use the GPG secrets. In this case the HOTP secret and those functions would still work though--it's independent of the smart card
3 failures on #3 I believe does nothing. My understanding is that you have unlimited tries against that PIN

tlaurion:
> There are three separate secrets on a Librem Key or Nitrokey:
> 
>     1. GPG admin PIN on the OpenPGP smart card
> 
>     2. GPG user PIN on the OpenPGP smart card
> 
>     3. HOTP admin PIN on the Librem Key/Nitrokey flash chip
> 
> 
> 3 failures on #1 apparently destroys the smart card to protect the PIN (but I think this is configurable). 

[Unless a reset PIN code is set](https://github.com/osresearch/heads/pull/551#issue-272893818).

>In this case the HOTP secret and those functions would still work though--it's independent of the smart card. It's also important to note that the user rarely if ever actually uses the admin GPG PIN--they almost always use the user GPG PIN for operations.
But when resealing secrets, eg. in the case the LUKS header changed when setting a Disk Unlock Key and associated passphrase to be released by the TPM since triggering the same code path.
> 
> 3 failures on #2 locks the user and you have to use the admin PIN to unlock it again to use the GPG secrets. In this case the HOTP secret and those functions would still work though--it's independent of the smart card

While using the same PIN? 
> 
> 3 failures on #3 I believe does nothing. My understanding is that you have unlimited tries against that PIN

I can't test right now. Thanks for validating this.

MrChromebox:
@tlaurion I was able to confirm that 3 incorrect admin pin entries when calling libremkey_hotp_initialize does indeed lock out the LK and require a reset

MrChromebox:
@tlaurion I'm going to drop this PR and resubmit the first commit (to which I don't believe there was any objection) with some others. I'll keep the 2nd commit in our tree for now until I figure out a better way to handle it

PR #625, opened 2019-11-19
kexec-save-default: guard TPM LUKS usage with config option

MrChromebox:
Move the TPM LUKS key storage to its own function instead of prompting users each time they save the default boot option.


tlaurion:
I thought we agreed to put on option in board config to call it automatically for boards who use it.

MrChromebox:
@tlaurion updated

tlaurion:
@osresearch : which other boards should have this in?

tlaurion:
@merge you ok with that?

merge:
> @merge you ok with that?

since it doesn't change things for the x230 for example, I'm ok with that, yes

tlaurion:
Would now miss other boards

MrChromebox:
just the T420, right?

tlaurion:
I'll check. It's hard to make an assumption. Best guest would be to apply to all boards having TPM support but librems.

On February 20, 2020 4:02:52 AM UTC, MrChromebox <notifications@github.com> wrote:
>just the T420, right?
>
>-- 
>You are receiving this because you were mentioned.
>https://github.com/osresearch/heads/pull/625#issuecomment-588598280

-- Sent from /e/ Mail

MrChromebox:
what do you think I did exactly? ;-)

the only boards with 'export CONFIG_TPM=y' outside of the librems are the x220, x230, and now t420

tlaurion:
Sorry didn't look, thought server board would!

Okidoki! Just t420 :)

On February 20, 2020 4:15:52 AM UTC, MrChromebox <notifications@github.com> wrote:
>what do you think I did exactly? ;-)
>
>the only boards with 'export CONFIG_TPM=y' outside of the librems are
>the x220, x230, and now t420
>
>-- 
>You are receiving this because you were mentioned.
>https://github.com/osresearch/heads/pull/625#issuecomment-588600772

-- Sent from /e/ Mail

MrChromebox:
should be good now :)

MrChromebox:
@tlaurion rebased this on master, using your preferred config string, and no change for non-Librem boards. If preferred, we could reverse the config logic (eg, CONFIG_TPM_NO LUKS_UNLOCK) and have the Librems select it. Either way, a pretty simple change to review/merge

tlaurion:
@MrChromebox : It would be preferable that the logic is inversed, so that Purism is responsible to maintain its boards to not use Heads Disk Unlock Key security feature, while all other community boards use it per Heads design.

MrChromebox:
@tlaurion makes sense, I will reverse the logic and push an update

MrChromebox:
@tlaurion logic reversed and patch rebased on Heads master

PR #626, opened 2019-11-19
Update Librem configs

MrChromebox:
- set default boot device to NVMe, fall back to SATA
- remove unusable secondary payloads
- disable IOMMU on iGPU for Linux payload only


paulmenzel:
corein*f*o in first commit message.

Is there an upstream bug report regarding the iGPU IOMMU issue?

MrChromebox:
> Is there an upstream bug report regarding the iGPU IOMMU issue?

no, effectively I'm just making the Librem command line match the x230.

PR #635, opened 2019-12-02
update musl-cross to 1952975

rofl0r:
this should fix issues with compressed ELF header sections.


rofl0r:
(using this PR to get CI to kick in and test the patch i added.)

PR #646, opened 2020-01-02
Add option to choose GnuPG userinfo during OEM reset

alex-nitrokey:
It is already possible to change the default user and admin/TPM PIN during OEM factory reset. It would be nice to also be able to choose a proper name, mail and comment field for the GnuPG keys.

### DIFF
--- a/initrd/bin/oem-factory-reset
+++ b/initrd/bin/oem-factory-reset
@@ -20,6 +20,10 @@ ADMIN_PIN_DEF=12345678
 TPM_PASS_DEF=12345678
 CUSTOM_PASS=""
 
+GPG_USER_NAME="OEM Key"
+GPG_KEY_NAME=`date +%Y%m%d%H%M%S`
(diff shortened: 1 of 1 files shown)


tlaurion:
Then more input validation is needed. For example, the name has a min and max size, e-mail address format needs to be checked, passphrases sizes too, else gpg will fail applying those values at runtime.

tlaurion:
That work has already been done [here](https://gitlab.com/tlaurion/heads/blob/PrivacyBeast_X230_QubesOS_Certified/initrd/bin/factory-reset-libremkey.sh) a while back.

Please take what you need and give me credit for it in your commit messages.

alex-nitrokey:
Sorry, could not do it earlier. Thanks for your valuable hints.

tlaurion:
@alex-nitrokey passés under my radar. Assigning to myself. Will test but looks good to me.

tlaurion:
LGTM. @MrChromebox ?

PR #649, opened 2020-01-07
libksba: fix qsort handler to reproducible sort the string table

osresearch:
They are invoking undefined behaviour by returning odd values to `qsort()`, so it is no wonder it failed to produce reproducible results.


osresearch:
Also added an issue with libgpg-error, although it hasn't been updated since 2016. https://dev.gnupg.org/T4801

PR #653, opened 2020-01-09
Use musl cross make for Heads, Linux, coreboot and edk2

osresearch:
This replaces the musl-cross checkout with either a local build of musl-cross-make or an external one specified via `$(CROSS)` environment variable.  This is really helpful in reducing the size of Heads trees since they no longer need to contain multiple gcc builds.  It also changes the circleci to use a docker image with musl-cross-make pre-built, allowing a somewhat faster CI cycle.

One remaining issue is that the musl-cross-make does need both x86_64-linux-musl and i386-linux-musl since coreboot is a 32-bit build.  `x86_64-linux-musl-gcc -m32` should work, except for a problem with libgcc.a.


.circleci/config.yml
```diff
@@ -2,88 +2,61 @@ version: 2
 jobs:
   build:
     docker:
-      - image: osresearch/heads-ubuntu:16.04
+      - image: osresearch/musl-cross:38e52db
```
tlaurion: Having the toolchain prebuilt is a nice shortcut for testing, but wouldn't we better have a clean checkout of up to date, mainstream docker images (ubuntu:18.04, ubuntu:latest, fedora, etc) so that we can trace build problems from the host toolchain changes from one build to the other and across different images, the default being ubuntu:18:04?

Consequently, from CircleCI at least, anyone wanting to build Heads themselves would simply have to subscribe to the project.

I'm not a CI expert, but the ideal down this path would be to have the same Heads codebase, and have environment variables specifying the host docker image to be used and the board config to build, so that one CircleCI/GitlabCI config can build for all hosts and boards config, requiring only different CI configurations based on the same code to test different hosts and boards.

PR #654, opened 2020-01-12
Update kexec to 2.0.20

flammit:
Depends on #653

Fixes issue with new `musl-cross-make` module (from #653) which causes `kexec` to fail to load the new kernel with an error: `Unhandled rela relocation: R_X86_64_PLT32`.


osresearch:
Is there a problem with the short git hashes? Since the directory is named based on them, it ends up quite long.

Rather than accumulate patch files for old versions and to keep the history of the patches, I've been doing a `git mv` to move it to the new version number.

What happened with the tails and fedora keys?

flammit:
On some machines (can't figure out what causes it), the tar.gz file generated by github contains the short directory name "tpmtotp-18b860f/*" instead of "tpmtotp-18b860fdcf5a55537c8395b891f2b2a5c24fc00a/" (which is expected given the sha256sum of the tar.gz file).

Sorry I should delete the old kexec 2.0.16 patch file.  The previous patch doesn't apply to the new version, so I had to update it.

The Tails and Fedora keys are just the ones with updated key expiration times.

PR #580, opened 2019-06-07
T430 initial support

SebastianMcMillan:
Requires #579 to be merged first.

Introduces T430 support, coupled with t430-flash.

Directly based on X230 configs, currently untested. Made to help troubleshoot #541.

Issue #506, opened 2019-01-05
Possible T430 Port


JeffreyAbbinante:
I might be buying a T430 soon. So, I will try and help port heads to the T430. Assuming I actually follow through on the purchase. If not, I think this is still a good idea and people should be working on it. It should be similar to the X230.

SebastianMcMillan:
Port has been made, but is untested. Feel free to test T430 on my repo.

tlaurion:
@JeffreyAbbinante @SebastianMcMillan : PR welcome based on https://github.com/osresearch/heads/commit/21faf524b9d791a75f1710ada502f21298b8fa77

tlaurion:
#t430

Issue #274, opened 2017-10-17
coreboot option table is not used.


persmule:
It seems that coreboot option table is not used in current configs. I do not know whether it is a consideration about security, but it removes the capability about tuning those options. (e.g. to assign more UMA for gpu than default 32MiB)
Enabling both USE_OPTION_TABLE and STATIC_OPTION_TABLE (Load default configuration values into CMOS on each boot) may be a better choice, as it allows these default values stored in CBFS being tuned via nvramtool(1) after building and before flashing.

osresearch:
Is this still relevant? Please re-open if we should enable the option table.

PR #656, opened 2020-01-18
T430 Support

flawedworld:
Building upon what @SebastianMcMillan created in #580 I have created a patch for coreboot to backport TPM functionality to the T430, allowing for the usage of coreboot 4.8.1I have also updated his work to allow the new musl cross compiler.

TOTP and TPM functionality appear to be wholly functional. GPG based actions however results in all actions failing with an 'End of file' error. I have tried resetting the NitroKey and attempted RSA 4096 and RSA 2048 and do not appear to be able to perform any GPG related actions. No signing actions are possible.

Thus I am at a loss of what to do. Suggestions appreciated.

UPDATE 31/01/2020: GPG issues all solved, t430 is fully functional but t430-flash has minor script issue


seitokaichou:
I also have a T430, already disassembled, and would like to help with this. 
>Building upon what @SebastianMcMillan created in #580 I have created a patch for coreboot to backport TPM functionality to the T430, allowing for the usage of coreboot 4.8.1

Nice! I hate to ask, but what's blocking https://github.com/osresearch/heads/pull/568? Wouldn't waiting for that be more practical in the long term (If I understand correctly, the coreboot 4.11 native vboot might have different issues on this device than the current in-house measured boot code, rendering this effort moot once that gets merged)? 

Are there any IRC or slack channels you hang out in? Would love to get in touch and solve this together.

tlaurion:
@flawedworld Temporary fix for gpg signing was upstreamed (GPG_TTY was asking tty to return actual termnial, which currently doesn't. #665 still needs fix for other non single terminal machines (servers) )

flawedworld:
@tlaurion I will merge in commits from master into my branch and build tomorrow.

flawedworld:
@seitokaichou great! coreboot 4.11 porting time will take likely a while still, this port is fine until it is done, I am not experienced enough unfortunately to assist with the whole port to 4.11 Given that the T430 much like the T530 and the rest of the **30 series is like the X230, I do not expect to see any major issues arise in future. I am present on the u-root slack if you wish to PM me there.

flawedworld:
@tlaurion One issue with t430-flash is that install.init is not invoked on startup, if i try to execute it once in ash by doing /bin/inistall.init I get permission denied. However if I perform each command manually I can successfully flash the rom made from 't430' and have 0 issues there.

So I can happily say the port for actual core heads is done, its just the flash script that needs a fix, which seems trivial hopefully

flawedworld:
@seitokaichou If you wish to assist, I reccomend following the steps in my previous comment and in theory you should be able to get functional heads on T430

tlaurion:
@sebastianMcMillan, can you give your comments?

SebastianMcMillan:
This is a decent stop-gap solutions until 4.11 is used. Here's my notes:
1. This uses commits of both #580 as well as #579. This means that this PR effectively generalizes the flash phase init file as proposed in #579. Is that intended?
2. install.init doesn't have proper permissions. Maybe you need to give it a chmod +x on your build system to work.
3. You enable the ability to change CMOS values. How do we measure changes to them?

tlaurion:
@flawedworld @SebastianMcMillan 
1- Please keep changes minimal and base your branch on top of other users branches, so that rebasing permits you to incorporate changes without including them personally.
2- @SebastianMcMillan is right.
3- I'm not a fan of that CMOS pull request @flawedworld. Unless proven otherwise, I do not see how this is needed with the added risk of adding another source of persistence and externally modifiable hardware behavior without a direct possibility of measurement. As specified [here](https://github.com/osresearch/heads/issues/274), if someone comes up with a need for that and proposes code to measure modified CBFS regions, that would be mergeable.

flawedworld:
All changes made do not intrude on x230 files, we use the x230-linux config, i see no need to change it for the t430 as all the **30 series are highly similar in configuration.

@tlaurion The ability to modify CMOS does not affect the SPI flash chip thus measured boot is unaffected and any values we can alter are limited to that which coreboot limits us to hence because of this I do not see a risk. These values are not executable and just a list of glorified true or false so I believe it is safe.

If despite this it is still seen as a risk, we can do what #274 discusses and use STATIC_OPTION_TABLE meaning we load CMOS values in at boottime and if a user wishes to edit values they must do so at compile time.

As @SebastianMcMillan stated, file permissions have been updated. Flash init is no longer generalized. This can be a separate pull.

If this all looks okay, I will start cleaning up commits.

tlaurion:
Seems good! Try to give @SebastianMcMillan as much credit for his contributions!

flawedworld:
@tlaurion Of course! @SebastianMcMillan You have been an absolute pleasure to work with!

tlaurion:
@flawedworld @SebastianMcMillan : resolve conflict and good to merge too if confirmed working!

tlaurion:
#506

SebastianMcMillan:
@tlaurion conflicts have been resolved, awaiting @flawedworld to test and confirm working.

MrChromebox:
looks like this could use a rebase on master

tlaurion:
@flawedworld @petabyteboy : can you rebuild and confirm that that board compiles and works as intended?

flawedworld:
Main 't430' build confirmed working. Going to test t430-flash soon.

The following was tested:
Adding GPG Key
TPM Reset and TOTP generation
GPG Signing of /boot
Setting /boot
Booting
Changing of CMOS values in OS, then seeing them reset to compile time defined settings after shutdown

flawedworld:
@tlaurion t430-flash fully working, the following has been tested:
Mounting USB
Execution of flash.sh to flash 't430' sucessfully

HOWEVER, a minor visual issue was noticed by me when using a modded 1080p IPS panel. This works by using a eDP to LVDS adapter between the panel and the motherboard for those who aren't aware of its workings. Everything still works however part of the screen is cut off at the bottom and there is general glitchiness. As evidenced by commits I attemped to fix this but 7c3d8a2 would not compile. Help needed, but I do not deem this to be important as the main 't430' build works perfectly with this modded panel and the user is only going to be in 't430-flash' for mere seconds anyway.

flawedworld:
@tlaurion comments please?

tlaurion:
@flawedworld  seems to work!
You have made a clean build?

`make BOARD=t430 real.clean && rm -rf crossgcc build/*  && make BOARD=t430` to arrive to your working results? To me it looks all good, still not convinced about [this](https://github.com/osresearch/heads/pull/656/files#diff-e396d974843635153bef2ef1a98a08f2R26-R27) but users seems to want it, so I won't go against it.

As a sidenote, if boards customizations are required, let it be to have LIBREM_KEY, or specify added linux kernel modules (sdcard required modules etc), then boards customizations alone would be required (different board config matching different linux config, specific for that specialized board). No problem here, but I see specialized boards as being required in the future.

Good for me! @SebastianMcMillan ? Will merge if no one has opposition tomorrow.

tlaurion:
@flawedworld 
> HOWEVER, a minor visual issue was noticed by me when using a modded 1080p IPS panel. This works by using a eDP to LVDS adapter between the panel and the motherboard for those who aren't aware of its workings. Everything still works however part of the screen is cut off at the bottom and there is general glitchiness. As evidenced by commits I attemped to fix this but [7c3d8a2](https://github.com/osresearch/heads/commit/7c3d8a29b1b8196cefd936684311f2ae9166df3d) would not compile. Help needed, but I do not deem this to be important as the main 't430' build works perfectly with this modded panel and the user is only going to be in 't430-flash' for mere seconds anyway.

Please open another issue with compilation error trace you encounter from a clean build.

SebastianMcMillan:
I've gotten reports that even in the main t430 build there's screen garble, and said T430 doesn't have the FHD mod either.

SebastianMcMillan:
@tlaurion alright, the graphics issue is resolved. Ready for a final review.

flawedworld:
@tlaurion Checks failed due to TLS issue on Circle CI, likely ca-certificates will need update or something like that.

tlaurion:
@flawedworld : did a lot of 

> @tlaurion Checks failed due to TLS issue on Circle CI, likely ca-certificates will need update or something like that.

@flawedworld: rerun that pipeline from failed point and it should work. Played in the past days with both circleCI and GitlabCI and didn't encounter the issue your CircleCI environement unfortunately encountered when it tried to download a package.

OkashiOdayakana:
Any progress on this?

tlaurion:
@flawedworld ?

MrChromebox:
configs need to be adjusted for flashrom 1.2, and IMO would be nice to see a lot of these commits squashed and refactored now that things are working properly.

flawedworld:
@tlaurion Yeah been getting super sidetracked recently, I am going to neaten things up hopefully and will update.

flawedworld:
@tlaurion commits neatened, tested build, successfully able to flash.

Only changes are the increase in CBFS size and flashrom changes. The original commits for this pull may be viewed at: https://github.com/fibreblazer/heads/commits/master_old

flawedworld:
@MrChromebox I will amend the commit to add in newlines and I will remove the commented lines in flash configs, I kept them for easy reference between x230 and t430 but will nuke them.

As for point 3, I feel that this is better placed as part of 4.12. If 4.12 gets merged in before this I will update this pull.

MrChromebox:
@flawedworld sounds good to me, will approve once updated

flawedworld:
@MrChromebox Right I believe this updated commit should do the trick.

MrChromebox:
@flawedworld 0060-enable-tpm-on-t430.patch is still missing the trailing newline, otherwise looks good

flawedworld:
oh? I better check that, my apologies!

flawedworld:
@MrChromebox I believe this should do it.

PR #658, opened 2020-01-22
flash.sh: Add progress bar

MrChromebox:
Show progress of flashrom reads/writes when flashing by means
of a progress bar, adapted from the Librem coreboot flashing script.

Also provides a time-stamped flashrom log in /tmp to aid diagnosis in the
event that a firmware update fails.

### DIFF
--- a/initrd/bin/flash.sh
+++ b/initrd/bin/flash.sh
@@ -15,6 +15,89 @@ case "$CONFIG_FLASHROM_OPTIONS" in
   ;;
 esac
 
+flashrom_progress() {
+    local current=0
+    local total_bytes=0
+    local percent=0
+    local IN=''
+    local spin='-\|/'
+    local spin_idx=0
+    local progressbar=''
+    local progressbar2=''
+    local status='init'
+    local prev_word=''
+    local prev_prev_word=''
+
+    progressbar2=$(for i in `seq 48` ; do echo -ne ' ' ; done)
+    echo -e "\nInitializing internal Flash Programmer"
+    while true ; do
+        prev_prev_word=$prev_word
+        prev_word=$IN
+        read -r -d' ' IN || break
+        if [ "$total_bytes" != "0" ]; then
+            current=$(echo "$IN" | grep -E -o '0x[0-9a-f]+-0x[0-9a-f]+:.*' | grep -E -o "0x[0-9a-f]+" | tail -n 1)
+            if [ "${current}" != "" ]; then
+                percent=$((100 * (current + 1) / total_bytes))
+                pct1=$((percent / 2))
+                pct2=$((49 - percent / 2))
+                progressbar=$(for i in `seq $pct1 2>/dev/null` ; do echo -ne '#' ; done)
+                progressbar2=$(for i in `seq $pct2 2>/dev/null` ; do echo -ne ' ' ; done)
+            fi
+        else
+            if [ "$prev_prev_word"  == "Reading" ] && [ "$IN" == "bytes" ]; then
+                # flashrom may read the descriptor first, so ensure total_bytes is at least 4MB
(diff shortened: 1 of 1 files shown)


MrChromebox:
updated with some error suppression for `seq` since occasionally it would puke around 99%/1%

MrChromebox:
resulting output (hopefully this works):

flashrom progress bar

MrChromebox:
@flammit can you review / test on top of flashrom 1.2 please?

tlaurion:
Adding gpg key and flash resulted in a 171% flashing progress at 100%.
Same when reapplying ROM with settings on a ROM where BIOS region of size of 7100000 is applied.

@MrChromebox

MrChromebox:
I'll take a look at this tomorrow or day after, didn't test w/ '--ifd -i BIOS' but shouldn't make a difference...hmm

tlaurion:
@MrChromebox sorry for bad picture, was a race condition.
signal-2020-06-28-170310

MrChromebox:
@tlaurion think I fixed it, but can't reproduce so will need you to re-test

MrChromebox:
> Result is the same as before and seems to be caused per BIOS region only being flashed, not whole flash.

ok, I'll modify one of my devices here to only flash BIOS region and see if I can reproduce/resolve

tlaurion:
@MrChromebox

MrChromebox:
@tlaurion will look into this coming week

MrChromebox:
@tlaurion I'm not able to reproduce this anomaly you're seeing, even on my x230. The Total flash size reflects the size of the BIOS region (~12MB on x230-hotp-maximized) and the counter never runs past 100%. Can you retry with a fresh build?

tlaurion:
It works with maximized builds because opaque 12mb is taken into calculation. 

Problem arises for boards not flashing the whole combined space. That is, if --ifd --image bios is specified

tlaurion:
@mrchromebox: One thing I miss is the ability that we had from the console to see that the image that is attempted to be flashed equals what is already on the SPI flash. Good news is, reflashing the same rom that is on SPI should keep the HOTP and TOTP values consistent if keeping settings. So should not be a problem.

Tested working as of now:
- x230-hotp-verification. Report is good and doesn't go over 100% even if for legacy boards, only the BIOS region is asked to be flashed. Initial size shown is of 12mb though, but then only the BIOS region is considered for internal, non-displayed calculations, but it works. Reflashing same firmware over also keeps TOTP/HOTP valid.
- x230-hotp-maximized: same! Reflashing same firmware over also keeps TOTP/HOTP valid.

@MrChromebox : should we still say, explicitely, to the user if the rom flashed is the same that is on SPI?
Otherwise, ready to merge.

tlaurion:
@MrChromebox actually, looking closer at the code, there is supposed to be a confirmation saying that the [rom is identical](https://github.com/osresearch/heads/pull/658/files#diff-e33f7a1603ea94b6a0ff385404e9860ff48e1c5afe2ed5c8e68207d3b39186ccR81), which I never got. Might want to look into that

MrChromebox:
@tlaurion that part of the code is indeed working, which can be verified by flashing the same ROM from the recovery shell twice in a row: 
PXL_20211217_211732016

Flashing the same ROM as is running on the system currently will not generate the 'identical image' message because the RW_MRC_CACHE coreboot FMAP region is populated with the RAM training data on the first boot after flashing (or if the RAM modules change), and when flashing a new ROM, that data is cleared.

tlaurion:
@MrChromebox Perfect, thanks for the reminder. 
RW_MRC_CACHE will always be different from initial flash from trained memory updated data.

So that goes to relaying into HOTP/OTP to confirm that internally/externally flashed rom is the same as it was before.

Merging!

PR #659, opened 2020-01-22
mount-usb: provide useful output when loading kernel modules

MrChromebox:
Provide user some info as to what is heppening, otherwise the
system appears to be hung for 5s with no indication why.

Side note: do we really need a 5s delay after this?

### DIFF
--- a/initrd/bin/mount-usb
+++ b/initrd/bin/mount-usb
@@ -5,9 +5,16 @@
 enable_usb
 
 if ! lsmod | grep -q usb_storage; then
-  insmod /lib/modules/usb-storage.ko \
+  count=$(ls /dev/sd* 2>/dev/null | wc -l)
+  timeout=0
+  echo "Scanning for USB storage devices..."
+  insmod /lib/modules/usb-storage.ko >/dev/null 2>&1 \
   || die "usb_storage: module load failed"
-  sleep 5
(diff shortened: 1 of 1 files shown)


tlaurion:
@MrChromebox : good catch for the sleep of 5 seconds, I was asking myself where that delay came from.

Well, couldn't we watch for /dev expected devices to appear instead of waiting 5 seconds?
Those are /dev/sdX that are expected to exist after usb-storage loads?

Something that would sleep 5 seconds at worst, checking at each second if /dev/* changed and exits if it did?

@osresearch : Not sure why we wait 5 seconds here but for scsi or some slower devices who are really long to init from USB. Use case?

MrChromebox:
> Well, couldn't we watch for /dev expected devices to appear instead of waiting 5 seconds?
Those are /dev/sdX that are expected to exist after usb-storage loads?

perhaps something like:
```
count=$(ls /dev/sd* 2>dev/null | wc -l)
timeout=0
while ($count == $(ls /dev/sd* 2>dev/null | wc -l)) {
    [[ timeout > 4 ]] && break
    read -t 1
    timeout=$timeout+1
}
```

MrChromebox:
updated based on earlier comments, and threw in some spurious error suppression to boot

PR #663, opened 2020-01-26
init: fix invalid GPG_TTY variable

flammit:
busyboy tty isn't working after the musl-cross-make change so
revert to known good value.


tlaurion:
@flammit : /dev/console is local tty per gathered documentation. Would that break kgpe-d16 and other server based boards?

MrChromebox:
fix confirmed on Librem laptops, need to test on server still

PR #666, opened 2020-01-29
flash-gui: set unset variable USB_FAILED

MrChromebox:
Not setting USB_FAILED when call to mount-usb succeeds results
in a spurious 'sh: 0 unknown operand' error printed to console.


kylerankin:
Minimal fix, looks good. Merging.

PR #676, opened 2020-02-18
WiP thin-provisioning-tools inclusion

tlaurion:
thin-provisioning-tools inclusion inside of Heads through additional CONFIG_THIN-ROVISIONING-TOOLS board option

... Not functional at the moment since i'm unable to make thin-provisioning-tools to build with musl-cross (it sticks with g++ whatever I do)

New generic board addition: x230-privacybeast
- thin-provisioning-tools and libraries are too big to be added under x230 without neutering ME and expending ifd consequently (which is why those minimal binaries are provided)
- THIN-PROVISIONING-TOOLS board option present
- linux config adding thin provisioning support and dm snapshots
- coreboot config pointing to expended ifd descriptor and neutered ME (x230 Lenovo 2.76 extraction)

Changes related to thin-provisioning-tools
- lvm2 module now supports Thin provisioning
- libaio and libexpat dependencies seem to build correctly (untested because...)
- thin-provisioning-tools module currently cannot build against musl-cross, depends on g++ and don't look into /install for its dependencies
- Makefile modified to build if configured in board

Bonus: Board config generates valid top and bottom ROM images to be flashed once from external flasher, while coreboot.rom can be flashed internally. At term, the [flash scripts depending on board specified flashrom options](https://github.com/osresearch/heads/pull/592) PR merge will permit this board to R/W whole ROM, not only the BIOS region.

If someone has input... issue here: https://github.com/jthornber/thin-provisioning-tools/issues/131


tlaurion:
To replicate problem, follow heads wiki building guide to install dependencies, add libtool, expat and libaio (autoconf?) and do:
make BOARD=x230-privacybeast

tlaurion:
@flammit @osresearch @kylerankin : your opinion would be needed [here](https://github.com/osresearch/heads/issues/307#issuecomment-588365157). 

This PR includes blobs to free 5M of space in the rom and produces bottom and top valid images, which could permit CI to prebuild those images without nasty scripts under blobs directory.

tlaurion:
Apparently not needed per [wyng](https://github.com/tasket/wyng-backup/issues/54)

Issue #673, opened 2020-02-16
ISO boot from master hardcoded to $CONFIG_USB_BOOT_DEV?


tlaurion:
Bug reported by user not on github:

> 
> ISO BOOT ISSUES TROUBLESHOOTING LOG & FIX (may be worth posting the fix
> somewhere, at least?)
> 
> Tried the steps you suggested (mount-usb && usb-init).
> 
> `mount-usb` fails, "resource busy," because it's already mounted
> 
> `usb-init` brings up the USB GUI
> 
> Qubes verifies fine
> 
> Then I get the usual errors, exact sequence:
> ```
> +++ Mounting ISO and booting
> !!!!! Something failed during USB boot
> ```
> 
> Then I drop out to the recovery shell with an error and the shell
> doesn't work:
> ```
> /bin/ash: can't set tty process group: Not a tty
> [1]+ Done   usb-init
> # ls
> 	can't set tty process group: not a tty
> # ls
> 	ash: l: not found
> 
> ```
> 
> Reboot and drop to shell directly.
> 
> Try and see if my kexec_iso/Qubes...iso/ settings are messed up. Rename
> kexec_iso/Qubes..iso/ to kexec_iso/Qubes..iso.bak/
> 
> Mount works, but I still get the usual error with usb-init.
> 
> Reboot to fix the shell.
> 
> Add a *lot* of echo statements to kexec-init-iso to find the error.
> 
> Turns out it fails getting DEV_UUID for /dev/sdb1
> 
> Reboot, try booting via ISO via GUI, get the usual error, drop to the
> recovery shell. `mount` reveals the media with the ISO is mounted as
> /dev/sdc1
> 
> Output of blkid command (as used in kexec-init-iso) is empty
> 
> $CONFIG_USB_BOOT_DEV appears to be hard-set to /dev/sdb1 in /tmp/config
> when it needs to be /dev/sdc1
> 
> So go back and change the CONFIG_USB_BOOT_DEV variable in
> boards/x230/x230.config
> 
> Make BOARD=x230 and reflash
> 
> ISO verified booting now works!
> 
> POTENTIAL LONG-TERM SOLUTION:
> eliminate the CONFIG_USB_BOOT_DEV variable from the board config
> entirely. The USB boot device should be discovered dynamically.
> 
> OPEN QUESTION: x230 hardware is well-understood. Why is it different on
> my board and only my board?
> 
> Maybe it's because I have a YubiKey plugged in at boot?
> 
[...]

PR #685, opened 2020-02-20
Eliminate use of CONFIG_USB_BOOT_DEV

MrChromebox:
mount-usb switched to dynamic USB device detection a while back,
so eliminate instances of CONFIG_BOOT_USB_DEV, and derive the
mounted USB device from /etc/mtab in the one place where it's
actually needed (usb-scan). Clean up areas around calls to
mount-usb for clarity/readability.

Addresses issue #673

Test: Build Librem 13v4, boot ISO file on USB


MrChromebox:
needs good/thorough testing, pushed just to get feedback on approach

tlaurion:
Had this debate prior of change of behavior with Kyle a while back but the goal of having a fallback to board configured device was to be able to boot from dock station hard drive or second hard drive to facilitate OEM tasks or recovery from fast drive having lots of ISOs. I like the approach and will continue to boot in recovery, mount device in /media and run gui-init again, but now it won't be mount /media. It will be mount /dev/sdb1 /media. Will probably put it in history file to facilitate backstage tasks.

I would personally love a generic "boot from other media", where usb drives would be considered as any other block devices after loading usb modules through lsblk or similar. Will check to see if I can propose something in that direction. 

On February 20, 2020 4:31:26 AM UTC, MrChromebox <notifications@github.com> wrote:
>needs good/thorough testing, pushed just to get feedback on approach
>
>-- 
>You are receiving this because your review was requested.
>https://github.com/osresearch/heads/pull/685#issuecomment-588603974

-- Sent from /e/ Mail

tlaurion:
@MrChromebox : 
what about doing a simple `stat -c % /sys/block/* | grep pci`
And form there, remove CONFIG_BOOT_DEV from the output to create a "Boot from other devices" instead of Boot from USB devices? 

The removing hardcoded CONFIG_BOOT_USB_DEV would become totally irrelevant in all use cases.

MrChromebox:
> And form there, remove CONFIG_BOOT_DEV from the output to create a "Boot from other devices" instead of Boot from USB devices?

CONFIG_BOOT_DEV isn't necessarily what's mounted on /boot, esp on Librems where there can be multiple SATA drives or SATA + NVMe. would need to extract from /etc/mtab like I'm doing

right now, USB booting has two very different paths: signed ISO file, and unsigned kernel. what we'd need to do is separate out the former, and then group the latter as part of the new unsigned 'Boot from other devices' option

Issue #690, opened 2020-03-09
LibremKey/Nitrokey with x220


eganonoa:
With the current master branch, which includes the changes to the x220 configs, amongst others, a new CBFS size ("to fix 50 seconds boot delay problems"), I found that you cannot build Heads with the additional libremkey/nitrokey module because there isn't sufficient space. 

Reverting to the prior CBFS size (0x7e8000) fixes that problem. With that change made to the coreboot-x220.config file and CONFIG_LIBREMKEY=y added to x220.config, my x220 boots incredibly quickly and everything works. To be honest, it appears to boot faster than the x230s I have used with either Heads or Pureboot. So it doesn't seem to bring back any 50 second boot delay problem that there was before (this is my first time with an x220 in heads so I don't know about the boot delay).

In addition, even before using the libremkey/nitrokey module, I found that current master cannot build Heads on an x220 that properly stores the GPG key. Each time I would reboot it would ask me to flash it afresh. Given the above, I suspect that this is also related to the CBFS size.

All of which leads me to believe that the CBFS size from "T420 initial support + X220 FBWhiptail Support #578" merged into the master branch on February 19 isn't right for the x220 and that it should revert to a CBFS size of 0x7e8000.

tlaurion:
@SebastianMcMillan : can you do a clean build with prior proposed settings (add CONFIG_LIBREMKEY in board config and CBFS correction in coreboot config) and confirm 50 seconds delay is gone doing so?
```
tar cvzf archives.tar.gz ./build/coreboot-4.8.1/util/crossgcc/tarballs/ ./packages ./build/musl-cross-38e52db8358c043ae82b346a2e6e66bc86a53bc1/sources/ && 
make BOARD=x220 real.clean && rm -rf crossgcc ./build/* && tar zxvf archives.tar.gz && make BOARD=x220
```

tlaurion:
Or, it would be time to create x220-librekey board with different coreboot CBFS  size?

eganonoa:
> Or, it would be time to create x220-libremkey board with different coreboot CBFS size?

[...]

PR #693, opened 2020-03-09
Fix X220 and T420 CBFS sizes

SebastianMcMillan:
Add some room in the CBFS to actually save GPG keys, as well as have room to add libremkey support on libremkey-enabled builds. As of right now, they cannot save GPG keys due to the lack of space. Addresses #690.

We cannot simply revert the sizes due to the 50 second delay being seemingly related to the CBFS size being too large.

The values provided here do work on my T420 and X220, do not cause said delay, and will enable GPG keys being saved as well as libremkey support presumably working on libremkey-enabled builds (someone with one needs to test to confirm it, I get all the prompts and such, I just don't have a libremkey to test if behavior is as expected).


tlaurion:
@SebastianMcMillan https://github.com/osresearch/heads/pull/568#issuecomment-596995130

eganonoa:
#response_container_BBPPID{font-family: initial; font-size:initial; color: initial;}  Well re the bios, I was using a modified version of UEFI 1.46 and ECP 1.24. This is the latest Bios for the x220 (27 Jun 2018) and will have no further updates. So I think you can comfortably move forward with the understanding that the latest Bios is the one to use.                                                                                                                                                                           From: notifications@github.comSent: 10 March 2020 19:20To: heads@noreply.github.comReply to: reply@reply.github.comCc: ralphbunche@gmail.com; mention@noreply.github.comSubject: Re: [osresearch/heads] Fix X220 and T420 CBFS sizes (#693)  @SebastianMcMillan  @eganonoa  #568 (comment)
Once again. Most updated Lenovo BIOS update needs to be applied, providing consistent original Intel ME, needs to present to extract consistent blobs and offer the same IFD/CBFS regions. The guideline referred here is in goal of having consistent upgradeable CBFS regions (required to flash internally, which means only reflashing the BIOS section determined by coreboot in CBFS configuration section of coreboot config which needs to match IFD descriptor in blob).
If we go with a first, required, externally flashed coreboot containing well carved CBFS region, containing modified IFD blob (expended) and neutered+deactivated ME(reduced maximally), then Heads provided flash scripts could be modified in next ROM upgrade to flash whole ROM images, not only --ifd --image bios region. That would permit maximal available and usable space for all Heads needs for a board, in a reproducible manner (thorugh #307 acceptance of plan)
Else, we are stuck at using consistent CBFS sizes that would match ifd original descriptor size and CBFS regions matching non-neutered ME. Hope I made this point clear this time.
This same situation will be valid for all Ivy/Sandy bridges without mrc, fsp and fully neutered Intel ME (no syslib, no kernel in ME), while to a wider extent, to all existing and future coreboot Intel board inclusion (and probably AMD too, we'll see) in Heads with less freeable space in ME and less available space in Coreboot since relying on more required blobs to be present for initialization of hardware.
All those concerns should probably be in another ticket. I'll cherrypick those later on.

tlaurion:
@SebastianMcMillan @eganonoa : See https://github.com/osresearch/heads/issues/690#issuecomment-600006494

SebastianMcMillan:
@tlaurion you able to review this and merge the changes? My X220 and T420 were both on latest stock firmwares and this the size in this PR are as high as they can go for now. Further work will need to be done to maximize the space of these systems, but that'll be a ways away, and as of right now they don't even function in master, so this is better than nothing.

PR #715, opened 2020-05-04
CIs: pass CircleCI to debian:bullseye docker image

tlaurion:
- Provide logs.tar.gz and cpios to facilitate debugging of reproducibility issues
- Provide x230-flash.rom to be able to externally flash heads for newcomers to heads along with coreboot.rom which is to be internally flashable, untouching ME Gbe and other things (unmodified CBFS in coreboot)
- Delete build/make directory in CI, else there is issues when host's make binary versions mismatches, which is a fix for older host build systems (ubuntu 16.04 original CI of @osresearch which hides cross-make-gcc onto docker dependency for which we don't know what exactly was put in the host build dystem docker image to produce actual results.)

This is a first step toward https://github.com/osresearch/heads/issues/571

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,14 +2,22 @@ version: 2
 jobs:
   build:
     docker:
-      - image: osresearch/musl-cross:38e52db
+      - image: debian:bullseye
     steps:
+      - run:
+          name: Install dependencies
+          command: |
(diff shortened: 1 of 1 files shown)


tlaurion:
@osresearch @flammit @MrChromebox any opposition?

PR #709, opened 2020-04-24
Enable vboot on Lenovo devices for measured boot

PatrickRudolph:

PR #703, opened 2020-04-12
Add xx30-maximized and xx30-hotp-maximized boards (11.5mb flashable BIOS regions, reproducible me.bin and generated gbe.bin and totally externally and internally flashable roms)

tlaurion:
**_REQUIRES_**:
- External flashing with individual `maximized-top.rom` on top SPI chip and `maximized-bottom.rom` counterpart on bottom SPI flash chip.
**_or_**
- Internal flashing over initially unlocked ifd (did you run `ifdtool -u` on your bottom.rom stock backup prior of reflashing it?).
If so, internal flashing is possible through a manual `flashrom --force --noverify-all -p internal -w maximized.rom` (12mb rom) from command line to switch from xx30 to xx30-maximized.

Adds:
- blobs/xx30/me.bin (98k, ROMP and BUP only modules, cannot be removed), shared among all xx30 boards (Helix (Type 3xxx), T430, T430i, T430s, T430si, T431s, T530, T530i, W530, X1 Carbon (Type 34xx), X1 Helix (Type 3xxx), X1 Helix (Type 3xxx) 3G, X230, X230i, X230 Tablet, X230i Tablet, X230s
- blobs/xx30/ifd.bin (~~Might be generalizeable to ifd.bin: haven't tested on above boards~~ _EDIT: Reported working_)
- adds coreboot config which [specifies maximal CBFS space usable inside of Heads (11.5Mb instead of 7Mb)](https://github.com/osresearch/heads/pull/703/files#diff-935e48e65275b2fcaaeedb57a41098b9R8)
- ~~GitlabCI~~ (**EDIT: changed their caching system. Unusable for our use case now**) and CirecleCI modifications, ~~changing the default board compiled to be x230-external-flash (instead of x230)~~ EDIT: no replace. 
[...]

Issue #571, opened 2019-05-22
Maintain actual CI to confirm reproducibility of roms across distributions


tlaurion:
Reproducible builds CI based on Ubuntu docker are not working for a long while now:

@osresearch
https://circleci.com/gh/osresearch/heads

@nemanjan00
https://gitlab.com/nemanjan00/heads-ci/

Does anyone have a working one?

Fedora-29 based one does work, but lacks hooks to build from master commits:
https://gitlab.com/tlaurion/heads

Any help appreciated in fixing any of those problems!

 

_Sent from my Galaxy S3 using [FastHub-Libre](https://github.com/thermatk/FastHub-Libre/)_

BlackMaria:
my $0.25. circleci probably has some restrictions or limits that we are hitting. I cant verify since I don't have access to the account.

1. Space:  it may be related to the amount of disk being used.
1. Threads: tons of threads are spawned when using `make -j 4`. I would suggest removing the -j 4 and see what happens.
   * IMHO we don't really need it to build multi-thread anyway.

NB I have built master in a loop for a weekend and never hit any deadlocks etc.

tlaurion:
@osresearch : that would be lovely if fixed. 
Meanwhile, my gitlabci ubuntu builder [works](https://gitlab.com/tlaurion/heads/tree/Ubuntu-PrivacyBeast_X230_QubesOS_Certified)

tlaurion:
An temporary fix for this was introduced by @osresearch [here](https://github.com/osresearch/heads/pull/653#pullrequestreview-341233982), using a specific docker image to have bootsrapping tools encompassed in a docker image.

While non-ideal, this works.

tlaurion:
Good news, builds are now reproducible again since https://github.com/osresearch/heads/commit/6b485ed89f533f7c264543ac333c36ed0d19d715

Any people experienced enough to propose CircleCI and GitlabCI configurations to be in tree so that CircleCI and GitlabCI could automatically build for differnt hosts and boards upon passed environment variables?  ([CircleCI](https://circleci.com/docs/2.0/env-vars/#using-bash_env-to-set-environment-variables))

- docker image (ex: ubuntu:18.04, defaulting to it)
  - depending on host image type (ex: ubuntu) install proper dependencies (apt, dnf, others...)
- board configuration (ex: x230, defaulting to it)

[...]

PR #721, opened 2020-05-14
Add option to build against coreboot 4.12

MrChromebox:
Update Heads to allow building with coreboot 4.12:

* update version and hashes for coreboot and coreboot-blobs
* use coreboot's own toolchain (adapted from #709)

Build/boot tested on Librem 13v2/13v4/15v3/15v4, x230, qemu-coreboot-fbwhiptail 

Outstanding issues:
 * Qubes/Xen fails to boot on SKL/KBL Librems (fine on x230 and out of tree WHL-based Librem Mini)
 * Individual boards can be migrated to coreboot 4.12 in subsequent PRs

### DIFF
--- a/modules/coreboot
+++ b/modules/coreboot
@@ -3,6 +3,9 @@ modules-$(CONFIG_COREBOOT) += coreboot
 ifeq "$(CONFIG_COREBOOT_VERSION)" "4.8.1"
 coreboot_version := 4.8.1
 coreboot_hash := f0ddf4db0628c1fe1e8348c40084d9cbeb5771400c963fd419cda3995b69ad23
+else ifeq "$(CONFIG_COREBOOT_VERSION)" "4.12"
+coreboot_version := 4.12
+coreboot_hash := edcad000ee9b73183c396ea76155629b3d27c693e0f1ae83e3424c4d936e2be2
 else
 $(error "$(BOARD): does not specify coreboot version under CONFIG_COREBOOT_VERSION")
 endif 
@@ -55,7 +58,16 @@ coreboot_target := \
 coreboot_output := coreboot.rom
 coreboot_depend += linux initrd $(musl_dep)
 
-$(build)/$(coreboot_dir)/.configured: $(COREBOOT_IASL)
+ifeq "$(CONFIG_COREBOOT_VERSION)" "4.8.1"
+COREBOOT_TOOLCHAIN=""
+$(COREBOOT_TOOLCHAIN):
+else
+COREBOOT_TOOLCHAIN="$(build)/$(coreboot_base_dir)/.xcompile"
+$(COREBOOT_TOOLCHAIN): $(build)/$(coreboot_base_dir)/.canary
+	$(MAKE) -C "$(build)/$(coreboot_base_dir)" CPUS=`nproc` crossgcc-i386
+endif
+
+$(build)/$(coreboot_dir)/.configured: $(COREBOOT_IASL) $(COREBOOT_TOOLCHAIN)
 $(COREBOOT_IASL): $(build)/$(coreboot_base_dir)/.canary
 	$(MAKE) -C "$(build)/$(coreboot_base_dir)" CPUS=`nproc` iasl
 
@@ -98,10 +110,17 @@ modules-y += coreboot-blobs
 
 coreboot-blobs_version := $(coreboot_version)
 coreboot-blobs_tar := coreboot-blobs-$(coreboot-blobs_version).tar.xz
+coreboot-blobs_url := https://www.coreboot.org/releases/$(coreboot-blobs_tar)
+
+ifeq "$(CONFIG_COREBOOT_VERSION)" "4.8.1"
 coreboot-blobs_tar_opt := --strip 3
 coreboot-blobs_dir := coreboot-$(coreboot-blobs_version)/3rdparty/blobs
-coreboot-blobs_url := https://www.coreboot.org/releases/$(coreboot-blobs_tar)
 coreboot-blobs_hash := 18aa509ae3af005a05d7b1e0b0246dc640249c14fc828f5144b6fd20bb10e295
+else
+coreboot-blobs_tar_opt := --strip 2
+coreboot-blobs_dir := coreboot-$(coreboot-blobs_version)/3rdparty
+coreboot-blobs_hash := 4735ee6850d55d1e65dee8b08cc9b28b8af00b42acf31365f5d9545406579104
+endif
 
 ## there is nothing to build for the blobs, this should be
 ## made easier to make happen


SebastianMcMillan:
CBFS size on T420 and X220 is already at absolute maximum (pending PR).

MrChromebox:
@SebastianMcMillan #693 you mean? I'll rebase on top of that once merged

MrChromebox:
@SebastianMcMillan all good now I think

tlaurion:
This PR is the followup of #721

tlaurion:
@MrChromebox @SebastianMcMillan  : 4.12 is populating exactly the same as [here](https://github.com/osresearch/heads-wiki/blob/master/Keys.md#tpm-pcrs)? 

Will try to test in the next days on x230.

MrChromebox:
> This PR is the followup of #721

bit of a circular reference :) Assume you meant a different PR?

tlaurion:
@MrChromebox correcte past reference.

Well I do not know if it was expected to work but from my results, neither of the following are populated PCRs:
```
0: Boot block
1: ROM stage
3: Heads Linux kernel and initrd
4: Boot mode (0 during /init, then recovery or normal-boot)
5: Heads Linux kernel modules
```
[Ref](https://github.com/osresearch/heads-wiki/blob/master/Keys.md#tpm-pcrs) 

@MrChromebox only done functional testing here, setting a default boot option and mystyping disk unlock key passphrase to have `pcrs` called from `/etc/functions`. I do not know a functional working testing path outside of this since fgoing to the recovery invalidates the measurements by extending them.

Was it thought to be functional and a simple repackaging of #709?
@patrickRudolph?

Screenshot:
signal-attachment-2020-05-18-223758

tlaurion:
@MrChromebox :
I see patches removed to coreboot but none added from #709 ?

MrChromebox:
@tlaurion I don't see any coreboot-related patches in #709, with the exception of the SMM patch, which is already merged in coreboot 4.12. 

And with 4.12, measured boot is decoupled from vboot, so vboot isn't necessary (though it would certainly make sense to rebase #709 to add vboot in addition to measured boot).

PatrickRudolph:
> @MrChromebox corrected past reference. Questining it now.
> 
> Well I do not know if it was expected to work but from my results, neither of the following are populated PCRs:
> 
> ```
> 0: Boot block
> 1: ROM stage
> 3: Heads Linux kernel and initrd
> 4: Boot mode (0 during /init, then recovery or normal-boot)
> 5: Heads Linux kernel modules
> ```
> 
> [Ref](https://github.com/osresearch/heads-wiki/blob/master/Keys.md#tpm-pcrs)
> 
> @MrChromebox only done functional testing here, setting a default boot option and mystyping disk unlock key passphrase to have `pcrs` called from `/etc/functions`. I do not know a functional working testing path outside of this since going to the recovery invalidates the measurements by extending them, while going forward just trust measurements to be valid. This is not the case here.
> 
> Was it thought to be functional and a simple repackaging of #709?
> @PatrickRudolph?
> 
> Screenshot:
> signal-attachment-2020-05-18-223758

The PCRs are extended like it's documented here: https://doc.coreboot.org/security/vboot/measured_boot.html
That's not how heads extens the PCRs, so you need another patch on top of 4.12 to keep the current behaviour.

tlaurion:
@MrChromebox : i'm not sure everybody will agree here, but since patches are applied to versions and commit ids now, I do not think removing patches is a good idea.

coreboot 4.8.1 patches could stay in repo forever.
coreboot 4.11 patches, 4.12 patches...

We could force boards to use specific versions of coreboot (kgpe-d16) instead of deleting boards.

@PatrickRudolph @MrChromebox : I'm not sure what are the next steps here.
@MrChromebox : 4.11 patches for vboot+measured boot are [here](https://github.com/osresearch/heads/blob/8824eba378d5a1c1ed95b1aad527c537d557e4fc/patches/coreboot-4.11/0100-mb-lenovo-Add-support-for-VBOOT-on-12MiB-devices.patch)

MrChromebox:
@tlaurion are you suggesting we keep older versions of modules and patches in tree and buildable? that's a nightmare to maintain w/r/t build tools etc. If anything, branch off for older devices and they can be maintained by interested parties, but let's not carry everything forward forever

The coreboot 4.11+vboot PR has the same issue with PCRs being used differently by coreboot.

I'll take a look at coreboot 4.11+'s measured boot and see if I can extend the PCRs to match the previous implementation, but won't have time to do that until next week

tlaurion:
@MrChromebox I was talking about a way to say:
- [ ] Coreboot version in board config
- [ ] keeping multiple coreboot modules with version appended
- [ ] keeping patches for coreboot specific versions

This way, someone wanting to build kgpe-d16 could still make BOARD=kgpe-d16 in CI when #571 goes forward without having to branch. Heads would continue to support boards, while if support is brought back by patches or newer coreboot version, boards could simply specify newer coreboot version.

Note: 
- Patches are applied at module package decompression. So adding a new patch requires to delete `build/module_ver` else patch is not applied.

The outcome of this would be to be able to produce reproducible board roms as artifacts. As of today, the x230-hotp-board is build from CircleCI (debian) and GitlabCI (Fedora).

MrChromebox:
@tlaurion but it's not just keeping coreboot 4.8.1 and patches, it's maintaining the toolchain, and likely a fixed version of flashrom and the kernel as well. If we decide we want to fix those for all boards, then there's an exponential increase in work to maintain, and an inertia to keep boards on older versions if they work there. 

I feel like the way coreboot does things with periodic versioned releases, and dropping non-compliant boards after a new release is the only sane way to do things. Tag/branch a release before we move to coreboot 4.12, and let legacy boards live there.

tlaurion:
What Is propose is to keep:

- [ ] Board specifying coreboot + version
- [ ] module/coreboot_version
- [ ] keep patches just like they are right now, being applied for commit id/version

Limitations:
- Patches are applied at module's package decompression only. So adding new patches will require CI's cache clearance and/or manual build/coreboot-* removal if the patches list is different.

Effect:
- kgpe-d16 board and other boards being dropped by coreboot could still be built by CI, while other boards still maintained would require board version switch upon being tested functional to change versioning to newer version. I do not think other modules require the same principle, while the implementated logic for coreboot could probably be applied to linux module, so the danger of breaking a board would be limited, while augmenting a tiny bit the maintainership needed per contributors to test module upgrades.

@MrChromebox : thoughts?

MrChromebox:
my main concern is with the toolchain, and changes to it now affecting 2 (or more) versions of coreboot

AdmerStroh:
Hi. Compiling of this completes succesfull, but flashing the rom gives me no video output on my x230 with modded fhd display even if I swap the data.vbt and gma-mainboard.ads in the coreboot files with their patched equivalents. It used to work with the merge:coreboot_next patch. Any ideas what could cause this? How can I get this to work on the x230 with fhd mod?

MrChromebox:
@AdmerStroh does coreboot 4.12 + SeaBIOS/Tianocore work properly with the FHD modifications? I don't know what changes might have affected the FHD mod between 4.11 and 4.12, but I'm guessing something on the libgfxinit side. What's the modded gma-mainboard file look like?

AdmerStroh:
@MrChromebox It does (the latest version of Skulls works just fine without any additional modifications). The modded gma-mainboard file has just LVDS output removed (see https://review.coreboot.org/c/coreboot/+/28950/10/src/mainboard/lenovo/x230/variants/x230_fhd/gma-mainboard.ads). It is strange regarding that stock coreboot 4.12 works just fine.. Maybe someone with a fhd-modded x230 could test if it works for them?..

MrChromebox:
@AdmerStroh is the screen backlight coming on? the default config for the x230 w/coreboot 4.12 under Heads should leave libgfxinit as the default display init. Unfortunately I only have a standard x230 here to test, not a FHD one

AdmerStroh:
@MrChromebox The backlight is coming on and can be dimmed via press on the power button. I will try to force output over the mini-dp port and get something on an external monitor. I will report the results.

AdmerStroh:
@MrChromebox unfortunately, even with only HDMI1 in gma-mainboard file enabled, no video output over external monitor (mini-dp). Maybe I made a fault merging your patch? For now, I will just use skulls or revert back to 4.11. Maybe someone more experienced than me with a x230 fhd mod could try and share their experience. Nonetheless thank you!

tlaurion:
> @MrChromebox unfortunately, even with only HDMI1 in gma-mainboard file enabled, no video output over external monitor (mini-dp). Maybe I made a fault merging your patch? For now, I will just use skulls or revert back to 4.11. Maybe someone more experienced than me with a x230 fhd mod could try and share their experience. Nonetheless thank you!

@AdmerStroh : my intuition here may be how you cleaned me.
Have you followed https://github.com/osresearch/heads-wiki/blob/master/Clean-the-ME-firmware.md ?

AdmerStroh:
@tlaurion I used the skulls script with the me-cleaning option on the bottom spi chip. Will it be reinstalled when I internally flash the full 12mb rom image? By the way, compiling and externally flashing the 4MB x230-flash rom with this patch works and I get a console output

tlaurion:
@AdmerStroh @MrChromebox  thus is why I'm wondering where your absence of console comes from full rom image if flash works.

AdmerStroh:
@tlaurion @MrChromebox Ok. I could succesfully resolve my issue by applying this patch https://github.com/osresearch/heads/pull/703 and flashing both the top and the bottom image externally. So probably tlaurion was right with the me not properly cleaned in the bottom chip.. Thank you very much for the help :)

tlaurion:
> @tlaurion @MrChromebox Ok. I could succesfully resolve my issue by applying this patch #703 and flashing both the top and the bottom image externally. So probably tlaurion was right with the me not properly cleaned in the bottom chip.. Thank you very much for the help :)
@AdmerStroh : please make sure you read the big fat warning I just added [here](https://github.com/osresearch/heads/pull/703#issuecomment-637719023)

If you intend to flash through x230-flash image, you are not supposed to play with #703 at all. #703 is WiP and is not ready, hence why it is not merged in master.

tlaurion:
@MrChromebox @PatrickRudolph : I attacked https://github.com/osresearch/heads/pull/721#issuecomment-633277184

Now the question is what MeasuredBoot approach we take.
- [ ] #562 
- [ ] #721 
- [ ] [Adapt Heads PCR to coreboot/ patch coreboot to use Heads PCRS](https://github.com/osresearch/heads/pull/709#issuecomment-646995150)

tlaurion:
@PatrickRudolph @MrChromebox : What are the next steps? We also have #568 that should be closed, I think.

MrChromebox:
@tlaurion agreed on closing #568.

I think we get #776 in, then we can add support for newer coreboot versions using coreboot measured boot for now (with or without vboot). and then also work on adapting Heads PCRs if we want to. But supporting multiple coreboot versions first seems like a necessary first step

siro20:
I guess adapting heads PCRs is a hard requirement for 4.12 to keep all boards in sync, measuring the same part of the firmware into specific PCRs.
Please note there's a reason why (old) boards have been dropped from the coreboot codebase. The primary reason was that it's not maintained.

tlaurion:
@MrChromebox @PatrickRudolph : updated https://github.com/osresearch/heads/pull/721#issuecomment-630546662 for clarity and asked for feedback https://github.com/osresearch/heads/pull/709#issuecomment-657131720

osresearch:
Adapting the official vboot/coreboot PCR usage makes some sense to me, since maintaining coreboot patches is quite a bit of work and adding the TPM drivers to the various stages is a somewhat invasive change.

Using one PCR for all of the stages also seems fine, especially if there is an event log that validates what went into the measurements.  Having separate ones in the Heads design was mostly out of laziness on my part for ease of precomputation.

If Bootguard is enabled (in modes 2, 3, or 5), my understanding that PCR0 will be extended with the CRTM (the ACM and the IBB). Does this conflict with vboot's use of PCR0?

tlaurion:
> If Bootguard is enabled (in modes 2, 3, or 5), my understanding that PCR0 will be extended with the CRTM (the ACM and the IBB). Does this conflict with vboot's use of PCR0?
@PatrickRudolph @pietrushnic : Insights?

PatrickRudolph:
> > If Bootguard is enabled (in modes 2, 3, or 5), my understanding that PCR0 will be extended with the CRTM (the ACM and the IBB). Does this conflict with vboot's use of PCR0?
> 
> @PatrickRudolph @pietrushnic : Insights?

coreboot's measured boot doesn't use PCR0, so it doesn't conflict as
* coreboot measured all stages into PCR2 
* all runtime data that might change (due to user interaction or hardware configuration) into PCR3
* PCR0 and PCR1 are reserved for Google's chromeos vboot and are usually 0 on non chromeos enabled devices.

We could change Heads PCRs to be more similar to coreboot's measurements:
* Remove Bootblock (PCR0) and ROM stage (PCR1) and rename to firmware and payload (PCR2)
* Remove Heads Linux kernel and initrd (PCR3) as now part of PCR2
* Add "Runtime data" as PCR3
* Keep PCR 4 and 5

tlaurion:
@PatrickRudolph excellent!

MrChromebox:
so, assume we would want to adjust Heads' PCRs prior to this change going in?

PatrickRudolph:
We need to either rework the existing coreboot patches introducing measured boot on top of 4.12 to keep current PCR registers or change heads, as otherwise some of the PCRs are zero and others filled with the wrong measurements as seen here [https://github.com/osresearch/heads/pull/721#issuecomment-630546662](https://github.com/osresearch/heads/pull/721#issuecomment-630546662)

tlaurion:
@MrChromebox @PatrickRudolph 

Here are the calls to pcr binary from heads:
```
user@x230-master:~/heads$ grep -Rn pcr initrd/
initrd/bin/kexec-seal-key:85:luks_pcr=`tpm calcfuturepcr -ix 16 -if /tmp/luksDump.txt`
initrd/bin/kexec-seal-key:90:  pcr_5="X"
initrd/bin/kexec-seal-key:92:  pcr_5="0000000000000000000000000000000000000000"
initrd/bin/kexec-seal-key:115:	-ix 5 $pcr_5 \
initrd/bin/kexec-seal-key:116:	-ix 6 $luks_pcr \
initrd/bin/wget-measure.sh:13:	die "Usage: $0 pcr-index url"
initrd/bin/kexec-unseal-key:45:	pcrs
grep: initrd/etc/mtab: No such file or directory
initrd/etc/functions:43:pcrs() {
initrd/etc/functions:44:	head -8 /sys/class/tpm/tpm0/pcrs
initrd/sbin/insmod:27:if [ ! -r /sys/class/tpm/tpm0/pcrs -o ! -x /bin/tpm ]; then
```

While the coreboot patch is the one responsible to measure specifics:
```
user@x230-master:~/heads$ grep -Rn tlcl_measure patches/coreboot-4.8.1/
patches/coreboot-4.8.1/0009-Add-heads-TPM-measurements-to-Skylake-Kabylake.patch:21:+		tlcl_measure(2, (const void*) start, size);
patches/coreboot-4.8.1/0009-Add-heads-TPM-measurements-to-Skylake-Kabylake.patch:54:+		tlcl_measure(0, bootblock, bootblock_size);
patches/coreboot-4.8.1/0009-Add-heads-TPM-measurements-to-Skylake-Kabylake.patch:56:+		tlcl_measure(1, _romstage, _eromstage - _romstage);
patches/coreboot-4.8.1/0009-Add-heads-TPM-measurements-to-Skylake-Kabylake.patch:69:+		tlcl_measure(1, (const void*) start, size);
patches/coreboot-4.8.1/0009-Add-heads-TPM-measurements-to-Skylake-Kabylake.patch:90:+		tlcl_measure(1, (const void*) dest, size);
patches/coreboot-4.8.1/0030-sandybridge.patch:25:+		tlcl_measure(0, bootblock, bootblock_size);
patches/coreboot-4.8.1/0030-sandybridge.patch:28:+		tlcl_measure(1, &_romstage, &_eromstage - &_romstage);
patches/coreboot-4.8.1/0030-sandybridge.patch:55:+		tlcl_measure(2, (const void*) start, size);
patches/coreboot-4.8.1/0000-measuredboot.patch:170:+		tlcl_measure(3, (const void*) start, size);
patches/coreboot-4.8.1/0000-measuredboot.patch:435:+uint32_t tlcl_measure(int pcr_num, const void * start, size_t len);
patches/coreboot-4.8.1/0000-measuredboot.patch:459:+uint32_t tlcl_measure(int pcr_num, const void * start, size_t len)
```

tlaurion:
> We need to either rework the existing coreboot patches introducing measured boot on top of 4.12 to keep current PCR registers or change heads, as otherwise some of the PCRs are zero and others filled with the wrong measurements as seen here [#721 (comment)](https://github.com/osresearch/heads/pull/721#issuecomment-630546662)

@PatrickRudolph we go in rounds here. @osresearch agrees on changing heads to match coreboot. Details are https://github.com/osresearch/heads/pull/721#issuecomment-663992910 to modify this PR or #709

MrChromebox:
@tlaurion by 'changing Heads to match coreboot' do we mean adjusting the Heads patches for coreboot 4.8.1 as @PatrickRudolph described above?

tlaurion:
@MrChromebox from et understanding there soule be patches needed for 4.8.1 ti match coreboot measurements basically all going into PCR 2, modify heads TPM extend scripts to measure changed PCR and then change boards one by one from coreboot 4.8.1 to newer coreboot versions after testing

tlaurion:
@MrChromebox : this shows the way https://github.com/osresearch/heads/pull/795/files#diff-8d39c7575f026b40d86f213a4fcea2d6R6-R10

tlaurion:
linked to #721 and #709. Please confirm irrelevance for closure @MrChromebox

tlaurion:
https://github.com/osresearch/heads/commit/28d3b7c89c114fccdf937f55ccce4397fa9338d1 being merged, #721 #709 are next.
@MrChromebox @PatrickRudolph

tlaurion:
@MrChromebox ok. So logic is inverted and fallsback to 4.12 default for boards not specifying other version.
Testing

tlaurion:
@MrChromebox : please rebase on master and do not delete coreboot 4.8.1 patches!

tlaurion:
@MrChromebox : not specifying the coreboot version under test board x230-hotp-verification resulted in this error:
```
make[1]: Entering directory '/root/project'
modules/coreboot:10: *** "x230-hotp-verification: does not specify coreboot version under CONFIG_COREBOOT_VERSION".  Stop.
make[1]: Leaving directory '/root/project'
make: *** [Makefile:587: all] Error 2

```https://app.circleci.com/pipelines/github/tlaurion/heads/292/workflows/8cb5121b-dc5d-4099-8aed-ad80cd1cdea0/jobs/317/parallel-runs/0/steps/0-123

Doing manual tests locally.

MrChromebox:
@tlaurion yeah was about to say, the behavior w/r/t specifying the coreboot version is unchanged. No logic inverted, no fallback

tlaurion:
Ok, CI system up and running for x230 x230-hotp-verification, t430, t430-hotp verification, qemu-coreboot and qemu-coreboot-fbwhiptail here with artifacts:

https://app.circleci.com/pipelines/github/tlaurion/heads/298/workflows/795e66c6-4ad4-4c1d-bf08-c439330124dd/jobs/324

I had to patch this PR. @MrChromebox Please review
[patch.txt](https://github.com/osresearch/heads/files/5065457/patch.txt)

MrChromebox:
@tlaurion I don't see why you need to patch anything outside of the board/coreboot configs, unless your fork isn't up to date. The 'libremkey-hotp-verification' module doesn't exist anymore so that tells me it isn't.

Any reason the Librem boards (which need to stay on 4.8.1 for now) aren't being built (without blobs) as part of CI testing?

tlaurion:
@PatrickRudolph : unfortunately, the x230-hotp-verification build here (and I suspect all other xx30 boards) still suffer from #770 .
Interestingly enough, #765 is gone. Any idea? Rest is feature pair in terms of measurements and basic testing works as a direct replacement.

Personally, as for @jan23 and the general community, #770 is a show stopper.

MrChromebox:
also, even if issues migrating boards to 4.12, we should still go ahead and merge this, since no impact to existing boards and it will allow adding of other new boards (like the Librem Mini) which require 4.12

PatrickRudolph:
I don't see why #770 would block this PR.

tlaurion:
> @tlaurion I don't see why you need to patch anything outside of the board/coreboot configs, unless your fork isn't up to date. The 'libremkey-hotp-verification' module doesn't exist anymore so that tells me it isn't.

@MrChromebox rebased my branch over osresearch/master just to make sure. Here is comparison over this branch 
https://github.com/MrChromebox/heads/compare/coreboot-4.12...tlaurion:x230-hotp-verification_coreboot412?expand=1

> Any reason the Librem boards (which need to stay on 4.8.1 for now) aren't being built (without blobs) as part of CI testing?

Just because I didn't added them yet and thought that blobs needed to be present as specified into blobs dir and extracted from full roms, just like the xx20 boards. With #798, adding those boards into CIs configs will not add much build time. Want to do provide a PR? Please put boards in blocks with similarities of coreboot+linux near of each other in CircleCI and GitlabCI configs.

tlaurion:
@MrChromebox added missing boards in latest commit https://github.com/MrChromebox/heads/commit/d6438384b3c5b841906825296760d75f0d7b8935

Building here: https://app.circleci.com/pipelines/github/tlaurion/heads/300/workflows/ef62c2d4-12ff-41ef-9cac-de6b867381d9/jobs/326

tlaurion:
@MrChromebox 
> Any reason the Librem boards (which need to stay on 4.8.1 for now) aren't being built (without blobs) as part of CI testing?

But the build will obviously fail because blobs are not provided in tree (https://github.com/osresearch/heads/issues/307) 
https://github.com/osresearch/heads/blob/master/config/coreboot-librem13v2.config#L13-L23
https://github.com/osresearch/heads/blob/master/config/coreboot-librem13v4.config#L13-L23
https://github.com/osresearch/heads/blob/master/config/coreboot-librem15v3.config#L13-L23
https://github.com/osresearch/heads/blob/master/config/coreboot-librem15v4.config#L13-L23

So I will revert that change once that point is proven.

MrChromebox:
@tlaurion comparing branches isn't helpful when commits get merged out of order (changing the hashes) and there is no unique ID associated with a given commit. We should think about adding one as part of a commit hook like coreboot does.

I still maintain that this PR is completely sufficient as-is. It touches a single file and only adds the option to build against coreboot 4.12. Building boards using it and the changes needed there should be handed in separate PRs and not affect this one

tlaurion:
@MrChromebox agreed. [This](https://github.com/MrChromebox/heads/compare/coreboot-4.12...tlaurion:x230-hotp-verification_coreboot412?expand=1#diff-2d4ed2860b55bb308eacc61e6c18a86fR8) needs to be changed and it builds for all current boards.

MrChromebox:
> @MrChromebox agreed. [This](https://github.com/MrChromebox/heads/compare/coreboot-4.12...tlaurion:x230-hotp-verification_coreboot412?expand=1#diff-2d4ed2860b55bb308eacc61e6c18a86fR8) needs to be changed and it builds for all current boards.

well, as part of migrating boards to 4.12, we should update the defconfigs as well -- increasing the CBFS size as needed, dropping config items that make no sense, etc

PR #722, opened 2020-05-18
board: x230-hotp-verification (includes libremkey-hotp-verification)

tlaurion:
Add x230-hotp-verification board with libremkey-hotp-verification in (supports remote attestation for Nitrokey Pro 2, Nitrokey Storage 2 and Librem Key)

- Includes CircleCI config (that will need to be reverted prior of merge) that permitted to create the artifacts that can be tested [here](https://app.circleci.com/pipelines/github/tlaurion/heads/158/workflows/b6e5fd19-72d4-485c-850c-f3a4a305c44e/jobs/169) so that board can be tested.

Maybe the way to go is to add all boards that can be built without external dependencies in https://github.com/osresearch/heads/issues/571 ? Each board takes 15 minutes when CircleCI cache was built from initial build, which can be reused to build other boards configs, as tested previously.


tlaurion:
Forced pushed so that buildsystems now builds for x230-flash, x230, and x230-hotp-verification

MrChromebox:
why did this need a separate coreboot config? it's the same as the regular x230 config, just cleaned up and with a larger CBFS.

tlaurion:
>why did this need a separate coreboot config? it's the same as the regular x230 config, just cleaned up and with a larger CBFS.

@MrChromebox : each board needs to be coupled with a coreboot specific configuration, since modules are built in the board configuration directory, and coreboot config needs to point in the directory for initrd and other files it needs to point to.

linux configurations can be shared across a board configurations, while coreboot configurations, right now, needs to be duplicated.

This board configuration takes into account your comments that were raised into #703. 
The x230 mother board configuration is still as it was originally.

MrChromebox:
ah ok, thanks for clarifying -- didn't realize coreboot configs couldn't be shared currently

PR #729, opened 2020-05-24
CIs: restrict cache, clean past build artifacts.

tlaurion:
Testing PR autobuilding at the same time (Want to see if PR builds automatically and produces equivalent artifacts

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -48,7 +48,7 @@ jobs:
       - run:
           name: qemu-coreboot
           command: |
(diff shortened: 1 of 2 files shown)


tlaurion:
The conclusion of this PoC is that a PR is checked against CI only, while the commit is tested against both CIs.

CircleCI is[ referred with green check](https://circleci.com/gh/tlaurion/heads/194?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)
GitlabCI will be tested upon commit, but passed [here](https://gitlab.com/tlaurion/heads/-/jobs/565613615/)

PR #747, opened 2020-06-09
oem-factory-reset: Handle non-TPM case

MrChromebox:
Guard TPM-related bits by checking CONFIG_TPM = y


tlaurion:
tested on kgpe-d16. Works

PR #759, opened 2020-06-24
Add partition Label to list of mount-usb

alex-nitrokey:
This helps people to differentiate their USB drives, if they have multiple devices inserted (e.g. Nitrokey Storage + USB drive).

### DIFF
--- a/initrd/bin/mount-usb
+++ b/initrd/bin/mount-usb
@@ -55,10 +55,10 @@ if [ -z ${USB_MOUNT_DEVICE} ]; then
     # remove block device from list if numeric partitions exist, since not bootable
     let USB_NUM_PARTITIONS=`ls -1 $i* | wc -l`-1
     if [ ${USB_NUM_PARTITIONS} -eq 0 ]; then
-      echo $i $(blkid | grep $i | grep -o 'LABEL=".*"' | sed 's/\ UUID.*$//') >> /tmp/usb_disk_list
+      echo $(blkid | grep $i | grep -o 'LABEL=".*"' | cut -f2 -d '"') $i >> /tmp/usb_disk_list
     else
       for j in $(ls $i* | tail -${USB_NUM_PARTITIONS}); do
-        echo $j $(blkid | grep $j | grep -o 'LABEL=".*"' | sed 's/\ UUID.*$//') >> /tmp/usb_disk_list
+        echo $(blkid | grep $j | grep -o 'LABEL=".*"' | cut -f2 -d '"') $j >> /tmp/usb_disk_list
       done
     fi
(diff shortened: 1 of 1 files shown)


initrd/bin/mount-usb
```diff
@@ -55,9 +55,11 @@ if [ -z ${USB_MOUNT_DEVICE} ]; then
     # remove block device from list if numeric partitions exist, since not bootable
     let USB_NUM_PARTITIONS=`ls -1 $i* | wc -l`-1
     if [ ${USB_NUM_PARTITIONS} -eq 0 ]; then
-      echo $i >> /tmp/usb_disk_list
+      echo $i $(blkid | grep $i | grep -o 'LABEL=".*"' | sed 's/\ UUID.*$//') >> /tmp/usb_disk_list
```
MrChromebox, line 58: alternate formatting:  
echo "$i ($(blkid | grep $i | grep -o 'LABEL=".*"' | cut -f2 -d '"'))" >> /tmp/usb_disk_list

could even swap the order/positioning to put the label first

tlaurion: @MrChromebox : yes!!!! Lable first would be awesome

tlaurion:
@MrChromebox this goes straight back to my mindgame of having external boot devices managed altogether instead of usb and other no main boot devices. Another issue once this one is merged?

MrChromebox:
@tlaurion adding labels seems like a good thing regardless, so I'm good with doing that as a follow on.

For this, just need to decide on preferred formatting:
orig -- /dev/<device>_LABEL="<label>"
mine -- /dev/<device>_(<label>)

alex-nitrokey:
I just realized that the last commit need a addition here https://github.com/osresearch/heads/pull/759/commits/20b07dd1b380151270431665f6088926e075d3d7#diff-fdfe87f61e3355a5042988dfbe10e9c2R100 because otherwise we try to mount the label. Will adapt the PR accordingly...

tlaurion:
Fix still needed, mounts label as reported in last comment

MrChromebox:
thinking about this some more, might be better to keep the device first, simply so they are all aligned. Esp for the case of a device without a label, it would put things really out of whack.

MrChromebox:
after even more pondering, might not be a bad idea to do this for /boot device selection as well. I'll put together a PR for that

alex-nitrokey:
> after even more pondering, might not be a bad idea to do this for /boot device selection as well. I'll put together a PR for that

Meanwhile I at least reversed the order here so that the PR should work now again.

MrChromebox:
@tlaurion this is good to merge

PR #760, opened 2020-06-24
Add systeminfo menu option

alex-nitrokey:
This PR enables users to check the build-in hardware (CPU, RAM size, size of disk).

### DIFF
--- a/initrd/bin/gui-init
+++ b/initrd/bin/gui-init
@@ -233,6 +233,7 @@ while true; do
       'c' ' Change configuration settings -->' \
       'f' ' Flash/Update the BIOS -->' \
       'G' ' GPG Options -->' \
+      'S' ' Systeminfo' \
       'F' ' OEM Factory Reset -->' \
       'x' ' Exit to recovery shell' \
       'r' ' <-- Return to main menu' \
@@ -371,6 +372,12 @@ while true; do
 
   if [ "$totp_confirm" = "G" ]; then
     gpg-gui.sh
+    continue
+  fi
+
+  if [ "$totp_confirm" = "S" ]; then
+    whiptail --title 'Systeminfo' \
+      --msgbox "$(cat /proc/meminfo | grep 'MemTotal' | sed 's/MemTotal:/Mem:/')\n\n$(cat /proc/cpuinfo | grep 'model name' | uniq | sed 's/model name.*:/CPU:/')\n\n$(fdisk -l | grep '/dev/sd.:' | sed 's/B,.*/B/')" 16 60
     continue
   fi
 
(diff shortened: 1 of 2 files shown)


initrd/bin/gui-init
```diff
@@ -374,6 +375,12 @@ while true; do
     continue
   fi
 
+  if [ "$totp_confirm" = "S" ]; then
+    whiptail --title 'Systeminfo' \
+      --msgbox "$(cat /proc/meminfo | grep 'MemTotal' | sed 's/MemTotal:/Mem:/')\n\n$(cat /proc/cpuinfo | grep 'model name' | uniq | sed 's/model name.*:/CPU:/')\n\n$(fdisk -l | grep '/dev/sd.:' | sed 's/B,.*/B/')" 16 60
```
MrChromebox, line 380: would be useful to display the Heads/firmware version as well.
also, should include NVMe disks in the fdisk call

MrChromebox:
I adapted this a bit, moving to the main menu, adding the firmware version, and reformatting for a slightly cleaner appearance:

https://source.puri.sm/coreboot/purism-heads/-/commit/88708862be71e3b96d542c1ca044d6ff3788d1b9

alex-nitrokey:
Thanks a lot for you improvement!

The MEMTOTAL showed 4MB for me. I found that unsuitable ;)

The FW version is "CBET4000" for me. Can you explain how this is supposed to work/what it should show and if it is suitable for heads anyway?

MrChromebox:
> The MEMTOTAL showed 4MB for me. I found that unsuitable ;)

it should have been 'GB' not 'MB' ;-)

> The FW version is "CBET4000" for me. Can you explain how this is supposed to work/what it should show and if it is suitable for heads anyway?

on Thinkpads, coreboot uses the thinkpad-i2c driver to populate the SMBIOS tables with the serials and other info provided by the EC, as necessary to ensure the thinkpad-acpi driver works under linux (see src/drivers/i2c/at24rf08c). the SMBIOS 'bios-version' field is one of those that gets overwritten/prepended, and possible that using a space delimiter is the issue here -- I was just wanting to drop the date since it's always going to be the epoch date when built with 'TIMELESS'.

On devices not using this coreboot driver, the result will be whatever is defined in the boards' coreboot config under CONFIG_LOCALVERSION.

MrChromebox:
@tlaurion this will likely need a follow-up patch to handle firmware version strings which contain spaces, since right now the first space is used as the delimiter to chop off the date at the end (since it's meaningless for timeless builds)

MrChromebox:
also, since it was switched from GB to MB, the value isn't correct anymore. 32070 MB != 32GB. I'll push a fix

MrChromebox:
pushed #767

PR #756, opened 2020-06-20
x230-htop-verification: coreboot CBFS_SIZE changed to 0x710000

tlaurion:
as per https://github.com/osresearch/heads/pull/748/files#diff-540d604c350e8fd52c25559b642795fbR5

Adresses: https://github.com/osresearch/heads/issues/737#issuecomment-646991153

PR #748, opened 2020-06-10
Update hotp-verification (fixes #746)

alex-nitrokey:
- [x] use newer version of hotp-verification module to avoid patch
- [x] testing new version on board

What do you think, shall I make two separate PRs instead?

(will fix #746)

Issue #746, opened 2020-06-09
Upgrade nitrokey-hotp-verification module


tlaurion:
@szszszsz upgrade module/nitrokey-hotp-verification with https://github.com/Nitrokey/nitrokey-hotp-verification/commit/1cf1e1f3601b79692b81d409597136b8ff7c723a

tlaurion:
@szszszsz : I would recommend renaming the module to nitrokey-hotp-verification and Makefile accordingly so that users know who is the clear maintainer of that module.

tlaurion:
I will do the housekeeping for other boards.

szszszsz:
Indeed. Maybe the more generic name could be used, like `security-key-hotp-verification`, but that's a detail.

cc: @kylerankin @jans23 @Nitrokey

kylerankin:
I like the idea of renaming it to be a generic name like
'security-key-hotp-verification' since now both Librem Keys and Nitrokeys
work, and hopefully one day even more keys could work!

tlaurion:
The token words that were accepted were: 
- "USB security dongle" to talk about GPG supporting cards, as opposed to keys which are known to be the numeric result of them per unmerged PR of Heads documentation.
- "HOTP USB security dongle" to talk about supported HOTP devices under heads.

alex-nitrokey:
I would like to try to add a PR these days that makes sure that 

* HOTP USB Security dongle is default (with a `CONFIG_HOTPKEY` or alike)
* there is a possibility to use a branding (either Nitrokey or Libremkey) via `CONFIG_HOTPKEY_BRANDING` or alike

while it is totally annoying to handle multiple branding in code (it really is!) I think it has a major advantage if users know which device to put in (for both, users of Librem Notebook and Libremkey and users that use another combination).

What do you think about this approach?

tlaurion:
> I would like to try to add a PR these days that makes sure that
> 
>     * HOTP USB Security dongle is default (with a `CONFIG_HOTPKEY` or alike)

CONFIG_HOTP_USB_SECURITY_DONGLE seems fit.
> 
>     * there is a possibility to use a branding (either Nitrokey or Libremkey) via `CONFIG_HOTPKEY_BRANDING` or alike

As previously discussed in Nitrokey attempt to detect the keys per hardware ID, there was this idea of having the branding stored inside of /boot prior of signing with GPG.

[...]

PR #761, opened 2020-06-24
Fix branding issue with HOTP USB Security Dongles

alex-nitrokey:
This PR fixes  the issue of having the terminology `Librem Key` everywhere in the code while there are more possible keys. On the same time we want to make it easy for users to know which key to insert. Thus, it is ensured that heads asks for the correct branding after initialization by storing the info in `/boot/kexec_hotp_key`.

This PR already includes #748 and thus #756 because otherwise it would not function correctly.

Things still needed:

- [x]  Include upstream changes ([Nitrokey/nitrokey-hotp-verification](https://github.com/Nitrokey/nitrokey-hotp-verification)) once they are ready
- [x] Fix all usage of current `librem_hotp_verification` PATH accordingly

In best case I can add these tomorrow already.

Please note that I intentionally used other terminology for the variables than discussed in #746 as they felt more suitable for me. Please consider my suggestion and let me know if shall change `CONFIG_HOTPKEY` to `CONFIG_HOTP_USB_SECURITY_DONGLE` instead.


alex-nitrokey:
Ready and tested. Reviews are welcome :)

I am sorry that it took so long and I hope you like this solution.

tlaurion:
Testing reproducibility from local build and https://app.circleci.com/pipelines/github/tlaurion/heads/255/workflows/8d927eec-1090-4e19-98c2-d2d20682fa4a/jobs/277

tlaurion:
`NitroKey CircleCI`:
`31e0a25e4dcf2efb2e4f541050db894709c376750bc224b8bc884df74a7e5fcd  ./bin/hotp_verification`
Local build on debian-10:
`ec8c73bbfd7380b7262164d50d8bf25e430b37f71b61543974a7d576d95c1127  ./bin/hotp_verification`

`

tlaurion:
@szszszsz we still have reproducibility issues?

tlaurion:
Randomly chosen binary:

[Nitrokey CI](https://app.circleci.com/jobs/github/Nitrokey/heads/81):
`9ffac1032dc6805d0a45200d2d9e9543928ecc6555a6c47c7f6a174cb24bfe77  ./lib/libqrencode.so.3`

Local:
`9ffac1032dc6805d0a45200d2d9e9543928ecc6555a6c47c7f6a174cb24bfe77  ./lib/libqrencode.so.3`

tlaurion:
Building and functionally working, though. :)

tlaurion:
Note that testing with a Librem Key still asks for user to add Nitrokey.
How is that going be be dealt with finally? Are we expected to duplicate all boards config to have their _Nitrokey _LibremKey derivatives so that it is specified specifically in board configs?

@kylerankin @alex-nitrokey

tlaurion:
@alex-nitrokey : Im confortable with `CONFIG_HOTPKEY`

alex-nitrokey:
> Note that testing with a Librem Key still asks for user to insert his Nitrokey. Not sure Purism will be happy with that merge.

Erm... yes, I guess it is the version with the older firmware were there was a vid:pid error, right? As this PR is checking the vid to decide which device is used, it is difficult to do anything about this issue with the older firmware without breaking the idea of an automatic discovery :(

> we still have reproducibility issues?

Will have a look at it with @szszszsz

alex-nitrokey:
besides `hotp_verfication` there is a reproducibility problem with

* `./lib/libgcrypt.so.20` and
* `./lib/libassuan.so.0`

`hotp_initialize` buils correctly though.

szszszsz:
@tlaurion @alex-nitrokey 
We should have no repro issues anymore - investigating. Queued job on [Gitlab CI](https://gitlab.com/szszszsz/heads-nitrokey/-/jobs/617403404).

alex-nitrokey:
I tested the new circleci hashes with a local build (debian) and they matches for all hotp stuff (though not for the above mentioned).

szszszsz:
Removed `pkg-config` and Git version use. Rebuilds queued.

szszszsz:
@tlaurion @alex-nitrokey 
Looks like this change helped. Both builds have `deb80848041f503` hashes now for `./bin/hotp_verification`. Confirmed as well with local build following Gitlab CI config file.

Regarding https://github.com/osresearch/heads/pull/761#issuecomment-651617176 nothing comes to my mind.

szszszsz:
Hi @alex-nitrokey @tlaurion !
- I believe this PR is ready for merge. If not, what's left here to finish?
- I think the repro build issue mentioned in https://github.com/osresearch/heads/pull/761#issuecomment-651617176 is not introduced by this PR. Could you confirm or deny @alex-nitrokey ?

alex-nitrokey:
> I believe this PR is ready for merge. If not, what's left here to finish?

I considered it done, yes.

> I think the repro build issue mentioned in #761 (comment) is not introduced by this PR. Could you confirm or deny @alex-nitrokey ?

Exactly, it consists independently. Probably after updating gpg2, but I am just guessing here.

initrd/bin/seal-hotpkey
```diff
@@ -49,26 +57,35 @@ mount_boot
...
 
-libremkey_hotp_initialize "$admin_pin" $HOTP_SECRET $counter_value 
+hotp_initialize "$admin_pin" $HOTP_SECRET $counter_value $HOTPKEY_BRANDING
 if [ $? -ne 0 ]; then
   echo -e "\n"
   read -s -p "Error setting HOTP secret, re-enter Admin PIN and try again: " admin_pin
   echo -e "\n"
-  if ! libremkey_hotp_initialize "$admin_pin" $HOTP_SECRET $counter_value ; then
+  if ! hotp_initialize "$admin_pin" $HOTP_SECRET $counter_value $HOTPKEY_BRANDING ; then
     # don't leak key on failure
```
szszszsz, line 89: Just noticed: here `$HOTPKEY_BRANDING` should be quoted I believe. Otherwise it looks good.
@alex-nitrokey What do you think?

alex-nitrokey: thanks, fixed.

initrd/bin/seal-hotpkey
```diff
@@ -49,26 +57,35 @@ mount_boot
...
     shred -n 10 -z -u "$HOTP_SECRET" 2> /dev/null
-    die "Unable to find Librem Key"
+    die "Unable to find $HOTPKEY_BRANDING"
   fi
 fi
 
+# Set HOTP USB Security Dongle branding based on VID
+if [ ! $(lsusb | grep -q "20a0:") ]; then
+	HOTPKEY_BRANDING="Nitrokey"
+elif [ ! $(lsusb | grep -q "316d:") ]; then
```
MrChromebox, line 73: why are these '!' ? seems to make the logic incorrect

MrChromebox, line 73: not sure why, but this ends up showing my Librem Key as a Nitrokey.
using:
`elif $(lsusb | grep -q "316d:" ; then`
works correctly however

alex-nitrokey: The negation is used because grep returns 0 if it found something.

Older Librem Keys had the same VID as the Nitrokeys because of a mistake done when building the firmware for it, see https://github.com/osresearch/heads/pull/761#issuecomment-651598181

I could not think of a way to prevent that :(

So I am wondering, what is the return value of `lsusb | grep -q "316d:"` for you? If it is `0` the patch should work as intended. Does it?

szszszsz: In other words first batches of the Librem Key had incorrectly set VID:PID same as Nitrokey Pro has, hence the confusion.
Perhaps they have changed the USB names only back then (might not show up in the `lsusb`, but maybe `dmesg`) - that would already help to differentiate between models, otherwise I do not see how to make it possible.

MrChromebox: > So I am wondering, what is the return value of lsusb | grep -q "316d:" for you? If it is 0 the patch should work as intended. Does it?

0, and no the patch does not work correctly with my LK. It identifies as a NK without the change I mentioned above

MrChromebox: pushed #781 to fix

Issue #584, opened 2019-06-16
Kexec-parse-boot can't handle arch install usb


stayinside:
The file located in `/arch/boot/syslinux` has the entry `LINUX boot/x86_64/vmlinuz` and is parsed by kexec-parse-boot as `Boot Arch Linux (x86_64) |elf|kernel /arch/boot/syslinux/boot/x86_64/vmlinuz`. The actual path is `/arch/boot/x86_64/vmlinuz`. This throws and error when going through the menu and is unable to boot.

PR #762, opened 2020-06-26
Tweak syslinux parsing code to be compatible with new Arch isos

tslilc:
Hi there,

Thanks for all your work on this great project! Much like #584, i was frustrated by the inability to boot my recovery iso of choice. The below patch aims to address two things:

- Arch (202005 anyway) provides multiple parameters to initrd via commas: `INITRD boot/intel_ucode.img,boot/amd_ucode.img,boot/x86_64/archiso.img`. These were not being separately parsed.

- As mentioned in #584, the root directory in `appenddir` was being incorrectly given. I'm less certain about this, but it would appear that syslinux expects a layout as appears on the Arch iso, so that regex does the (correct AFAICT) thing. This works for Arch, but i'm not actually sure what the requirement is. I'm happy to work more on this if someone can point me to more syslinux bootloaders or some docs.


tlaurion:
@tslilc : any reason why your fixes are not integrated into `fix_path()` ?

Tested your change against :

- QubesOS 4.0 latest release: no regression.
- Fedora-Workstation-Live 30_Beta: no regression
- Clonezilla: no regression.

tlaurion:
@tslilc Also, please create a branch in your repo so that its easy to differentiate (you master will move, a branch should not)

Thanks! :)

tslilc:
@tlaurion Thanks for looking this over and testing! 

- `fix_path()` is called multiple times in multiple places. While it would be possible to add the hack `echo $appenddir | cut -d\/ -f -2` there conditional upon `"$state" = "syslinux_entry"`, i thought it was conceptually clearer to put this in `syslinux_end()`

- right, sorry about the branch thing, this is my first time contributing here. That makes sense.

Finally, i read (part of) the [syslinux docs](https://wiki.syslinux.org/wiki/index.php?title=Config) to try to understand the correct way to parse these files. Unfortunately the current implementation is a hack, which makes my `$appenddir` fix a hack on top of a hack. 

At present, `scan_options()` in `kexec-select-boot` does essentially

    for i in `find $bootdir -name "$config"`; do (... parse stuff...)

but this will not in general work for {SYS,ISO}LINUX boots. In particular, the [docs about configuration loading](https://wiki.syslinux.org/wiki/index.php?title=Configuration_location_and_name) say that SYSLINUX tries, in order, 

1. $root/isolinux.cfg 
2. $root/boot/isolinux/syslinux.cfg
3. searching for the same files in "$root/isolinux/", "$root/boot/syslinux/", "$root/syslinux/", and "/", in this order. 

This is only true of version 4.03+. This is not compatible with the (presumably alphabetic, possibly undefined) ordering produced by `find`. The reason we should care about this is because we cannot statelessly and correctly parse "*.cfg" files because:

there are "global" directives which must be remembered through parsing. One such directive, visible in the Arch iso in "isolinux/isolinux.cfg" is `APPEND /arch/`. The [docs](https://wiki.syslinux.org/wiki/index.php?title=Config#APPEND) are bizarrely vague and confusing about what `APPEND` is supposed to do saying things like "The options are added at the very beginning of the kernel command line" (append? beginning?) and "If you enter multiple APPEND statements in a single LABEL entry, only the last one will be used." (so not append at all). Also no comment as to how this interacts with INITRD lines. 

Anyway, if i had to guess, i'd say this is the magic entry for the Arch iso which sets the directories correctly, and does properly what my `cut` hack above is trying for.

Frankly this whole SYSLINUX thing is a disaster. 

If our goal is to make Arch (2020.05.01) iso work, this will do it. It might even fix others, and shouldn't break non-pathological {ISO,SYS}LINUX setups. But, Arch anyway actually includes a (BLS)[https://systemd.io/BOOT_LOADER_SPECIFICATION/] file, which has complete and correct information and we *could* parse that without messing with `$appenddir` above -- unfortunately, the current BLS implementation is tuned for grub and would need some reworking for {ISO,SYS}LINUX.

In summary:
a. the paring of commas is unrelated and should be merged anyway
b. we could implement this `cut` and `$appenddir` hack, but it would be just that
c. there's no way, presently, for `kexec-parse-boot` to actually work correctly for {SYS,ISO}LINUX by the nature of its implementation and call site
d. BLS is a viable solution for this use case, but the implementation needs to be revised


So ... what would you like to see?

tlaurion:
Hmmm looking.

If I understand it [well](https://github.com/archlinux/archiso/blob/master/configs/baseline/syslinux/syslinux.cfg#L8) the APPEND statement is a simple equivalent to the `CONFIG_BOOT_KERNEL_ADD` statement that can be added inside of [board config](https://github.com/osresearch/heads/blob/master/boards/x230-hotp-verification/x230-hotp-verification.config#L34) or added into [default boot options](https://github.com/osresearch/heads-wiki/blob/master/Boot.md) which is already [prepended prior of kernel arguments](https://github.com/osresearch/heads/blob/7a755a879ec241f59f30eabe88f6e99a2fdc80f0/initrd/bin/kexec-boot#L32).

[BLS fixes were merged in here](https://github.com/osresearch/heads/pull/559)

Are those information changing your train of thoughts on possible implementations?

tslilc:
Thanks for looking over that wall of text!

When i was talking about the APPEND statement, i meant [this one](https://github.com/archlinux/archiso/blob/master/configs/baseline/isolinux/isolinux.cfg#L6), not the one you linked to. According to the docs this is supposed to be parsed first. I think this is unrelated to the board config or default boot options, and i expect that this is about fixing the relative paths in the Arch iso which appear, for example, [here](https://github.com/archlinux/archiso/blob/master/configs/baseline/syslinux/syslinux.cfg#L6).

The point about global state and configuration file order was that we should do *something* with `APPEND /arch/` first, before we look at the other files.

As for BLS, as i might have mentioned above, `kexec-parse-bls` was written with *only* Grub in mind. See for instance the stuff [around these lines](https://github.com/osresearch/heads/blob/master/initrd/bin/kexec-parse-bls#L87) which fusses about `grubenv` -- something not present on a {SYS,ISO}LINUX boot medium.

Now that i look at it, `kexec-parse-bls` has the same "multiple initrd stipulations" problem that `kexec-parse-boot` does (fixed in the case of {ISO,SYS}LINUX with the commas thing). This would not be a difficult fix.

So yeah, i'm actually looking for direction from this project. We could try to implement any of
1. a conformant {SYS,ISO}LINUX parser (yuck), which would involve redoing `kexec-select-boot` to hand off the logic to the parser in the {SYS,ISO}LINUX case so that processing order is correct
2. a conformant BLS parser and use BLS before {SYS,ISO}LINUX
3. keep the hacks above and hope they're not too fragile

Evidently 3. is easiest, 2. is probably the best compromise on effort and expected payoff, and 1. is the only actual correct solution.

tlaurion:
I think 2 is best, while per precedent test 3 works

tlaurion:
@mrchromebox? Thoughts?

MrChromebox:
@tlaurion get 3) in now, so that things work for the users that need them, and work on improving 2) to support grub and sys/isolinux. Unless someone really wants to tackle 1), but doesn't seem that way

tlaurion:
@tslilc you have your plan 😃

tlaurion:
@tslilc pinging you to see if you have update or blockers?

tslilc:
@tlaurion thanks for the ping. I had completely forgotten about this. There are no blockers, and i think where we left this i had to give the PR against a non-master branch. I'm not sure how to do that just yet, so i'll look into that and hopefully have the patch ready soon.

tslilc:
Wow, ok, apparently this PR now contains every commit since i last forked. Ridiculous. I'll close this and open another.

tlaurion:
Revive of this PR under #1120

PR #763, opened 2020-06-30
Delete AES keys of Nitrokey Storage after reset

alex-nitrokey:
If smartcard Nitrokey Storage was factory-reset, we delete AES keys on
it as well.

Explaination: After oem-factory-reset was started the AES on the Nitrokey Storage that is used for the encrypted volume and the password safe is is not usable anymore because the smart card was factory-reset. To make it usable, a user needs to delete it via Nitrokey App. By doing so, the HOTP secret is deleted as well, resulting in a bad warning in Heads. Therefore, we are resetting AES key right after factory-reset with hotp_verification

This PR depends on #761 and thus can only be merged if this is merged (this is why this looks huge, but isn't). I opened this PR as a reminder to include [these small lines](https://github.com/osresearch/heads/commit/30236ffbaaea83acadaf10f3337376a0affa22fc) in future.


alex-nitrokey:
This commit is now ready to merge. Tested it today with current master merged into the PR.

To sum up the above: Users of NK Storage would need to create new AES keys themselves and thus would need to regenerate HOTP key again afterwards. With this PR the AES keys on the Storage are regenerated during oem-factory-reset.

Issue #764, opened 2020-06-30
oem-factory-reset failing during gpg factory-reset


alex-nitrokey:
The oem-factory-reset is failing for me if started directly after starting the device (e.g. after flashing), but is working fine if you call `oem-factory-reset` out of the recovery shell.

Did anybody experience this as well? I couldn't see the reason yet. It seems to fail at [this position](https://github.com/osresearch/heads/blob/8dc5b7616aec2586c4a3bd59ac8cfca1cfa01e91/initrd/bin/oem-factory-reset#L71).

MrChromebox:
it's failing for me regardless if started from the boot prompt or manually from the menu; I believe this started when #646 was merged

alex-nitrokey:
> I believe this started when #646 was merged

I was thinking the same - yet, I know it did work :dagger: 

> it's failing for me regardless if started from the boot prompt or manually from the menu;

please try going into recovery shell and type `oem-factory-reset` from there. This should work. Though the GnuPG commands are output which was not the case in the past. This is why I believe that updating gpg2 could be the cause here.

MrChromebox:
> please try going into recovery shell and type oem-factory-reset from there. This should work.

yes, see my edit :)

> This is why I believe that updating gpg2 could be the cause here.

will have to try reverting that to isolate, then can diagnose

tlaurion:
@alex-nitrokey : I would prefer having mount-usb from preceded with a prompt prior of exporting the keys and keep factory-reset outputing normally to /media if not already mounted.

I will merge my changes to that script in subsequent PR for the reownership wizard to work with upstreamed code, and to me it doesn't really make sense to have a factory reset without exporting the public key.

techge:
> @alex-nitrokey : I would prefer having mount-usb from preceded with a prompt prior of exporting the keys and keep factory-reset outputing normally to /media if not already mounted.
I will merge my changes to that script in subsequent PR for the reownership wizard to work with upstreamed code, and to me it doesn't really make sense to have a factory reset without exporting the public key.

This was intended for #766 , wasn't it?

It is just a suggestion. In our experience many people do not care about the GnuPG keys and just want the boot measurement, thus having a pubkey is not always needed or even confusing and the factory-reset aborts if you do not put in an usb drive... But I totally see that this might be controversial. I just wanted to suggest the change at least as we have implemented for NitroPad.

PS: do not be confused, I am logged in with another account (thus, not alex-nitrokey).

[...]

PR #766, opened 2020-06-30
Make gnupg pubkey export an option only in oem-factory-reset

alex-nitrokey:
Since #758 is merged, users have a option to export GnuPG pubkey if
necessary. Thus, we they do not need to insert a USB drive during
factory reset. Until now the whole process failed just because a user
did not provide a USB drive instead.

This shall be fixed by this commit


tlaurion:
As wrongly wrote (sorry) in #764 :
>@alex-nitrokey : I would prefer having mount-usb from preceded with a prompt prior of exporting the keys and keep factory-reset outputing normally to /media if not already mounted.

>I will merge my changes to that script in subsequent PR for the reownership wizard to work with upstreamed code, and to me it doesn't really make sense to have a factory reset without exporting the public key.

So in this case, I would set the default to N in the factory-reset script prompt, detect if there is USB drive and not fail but continue?

alex-nitrokey:
> So in this case, I would set the default to N in the factory-reset script prompt, detect if there is USB drive and not fail but continue?

Sounds good to me!

alex-nitrokey:
To add: I am preparing a commit accordingly.

alex-nitrokey:
This should be fine now :)

tlaurion:
@alex-nitrokey : conflict needs to be resolved prior of merging and testing

alex-nitrokey:
Should be good to go again now :)

tlaurion:
Works as intended.

PR #776, opened 2020-07-10
coreboot: 4.8.1 fixed in Makefile, coreboot module and board configs

tlaurion:
This is to facilitate newer coreboot version integration and testing (#721 #709) without breaking working and untested boards to newer coreboot release, until each board config is confirmed to work with newer version. 
The same approach is to be taken to migrate kernel version.

[We have to choose for #721 and #709 if Heads PCRs have to change for coreboot (this is shared code) or if coreboot source code should be patched to conserve Heads measuring scheme](https://github.com/osresearch/heads/pull/709#issuecomment-646995150).

Let me know what you think:
@patrickrudolph @MrChromebox


MrChromebox:
this makes sense to me, and seems like a reasonable way to keep support for boards that coreboot no longer supports, and that we need to bump the coreboot version to support newer boards (eg, librem mini). Probably should do this for the kernel version too.

PR #778, opened 2020-07-10
USB / GUI Tweaks

MrChromebox:
A handful of small UI improvements:

* Move duplicated functions usb_mount() and file_selector() into new gui_functions script.

* Modify usb_mount to always unmount /media before user selection of USB device for updating firmware, saving keys, etc. and to handle a user cancellation differently from a mount failure.

* Improve readability of prompts in flash-gui.sh.

* Add 'fold' utility in busybox config

* Use fold to wrap error messages in oem-factory-reset

Tested on a Librem 13v4


tlaurion:
fold is a really interesting approach. Would it be an easy replacement to put inside of all whiptail msg, without dealing with newlines inserted manually?

Will test this with other pull requests in next test batch.

PR #784, opened 2020-07-29
Fix OEM factory reset and setting of default boot with F32

MrChromebox:
Handful of small fixes to address bugs in oem-factory-reset and kexec-select-boot which manifested with an install of Fedora 32:

* oem-factory-reset needs to parse BLS spec grub files as well as standard ones
* oem-factory-reset needs to set the index of kexec_default.{i}.txt rather than always using 1
* oem-factory-reset and kexec-select-boot should use -s vs -r to test if any valid boot entries produced by kexec-parse-boot


tlaurion:
@MrChromebox : Iso detached signed manually (to mitigate #785) and it works. Merging

PR #749, opened 2020-06-11
WIP: Upgrade Cryptsetup 2.3 (fixes #669)

alex-nitrokey:
I tried around a bit to get cryptsetup 2.3 working. Unfortunately, I just can't find a solution to properly link libuuid with the library. I hope somebody here knows better than me...

(see [CI build](https://app.circleci.com/pipelines/github/Nitrokey/heads/40/workflows/7c87a1ec-e186-45b3-9d8a-a106988e5014/jobs/43) for output)
```
/bin/bash ./libtool  --tag=CC   --mode=link /root/project/crossgcc/bin/x86_64-linux-musl-gcc -fdebug-prefix-map=/root/project=heads -gno-record-gcc-switches -D__MUSL__ -I/root/project/install/include -L/root/project/install/lib  -Wall  -g -O2   -o libcrypto_backend.la  lib/crypto_backend/libcrypto_backend_la-crypto_cipher_kernel.lo lib/crypto_backend/libcrypto_backend_la-crypto_storage.lo lib/crypto_backend/libcrypto_backend_la-pbkdf_check.lo lib/crypto_backend/libcrypto_backend_la-crc32.lo lib/crypto_backend/libcrypto_backend_la-argon2_generic.lo lib/crypto_backend/libcrypto_backend_la-cipher_generic.lo lib/crypto_backend/libcrypto_backend_la-cipher_check.lo    lib/crypto_backend/libcrypto_backend_la-crypto_kernel.lo  lib/crypto_backend/libcrypto_backend_la-pbkdf2_generic.lo libargon2.la 
libtool: link: (cd .libs/libcrypto_backend.lax/libargon2.a && /root/project/crossgcc/bin/x86_64-linux-musl-ar x "/root/project/build/cryptsetup-2.3.3/./.libs/libargon2.a")
[...]

PR #788, opened 2020-07-30
Fix kexec to work with Debian Installer (fixes #699)

### LINKED ISSUE #699: USB-Boot Debian installer
Did anybody successfully boot a Debian installation transferred via dd on a USB drive? For example [this image](https://caesar.ftp.acc.umu.se/debian-cd/current/amd64/iso-cd/debian-10.3.0-amd64-netinst.iso).

It appears to me that one can not kexec into the provided kernel/initram of the iso, neither on heads nor an another system. Is this a known bummer or is there anything one can do about it? Or maybe the system message are just not transferred?

### what I does
trying to usb-boot to a Debian installation USB drive, prepared via dd and the install iso

### what does happen
You see the kexec message that it is booting the kernel, nothing happens afterwards (the screen freezes).

### what I expect
booting into the Debian installer

alex-nitrokey:
This patch was reported by @bemoody in issue #699

Tested via `BOARD=x230-hotp-verification` on a Thinkpad x230

### DIFF
--- a/patches/kexec-2.0.20.patch
+++ b/patches/kexec-2.0.20.patch
@@ -1,7 +1,7 @@
-diff --git ./Makefile.in ./Makefile.in
+diff --git a/Makefile.in b/Makefile.in
 index fb01134..bf1973e 100644
---- ./Makefile.in
-+++ ./Makefile.in
+--- a/Makefile.in
++++ b/Makefile.in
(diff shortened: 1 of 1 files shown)


tlaurion:
@alex-nitrokey @MrChromebox Awesome news!!!! 
Does this resolve QubesOS 4.1 ISO boot issue too? (not talking about #749 here which still needs fixing).

PR #792, opened 2020-08-06
blobs/x220|t420: Fix extract script

PatrickRudolph:
Command returns a list of utilities found. This can happen if multiple
coreboot folders are present.

Use only one to fix a crash in the following lines.

Test: Being able to extract blobs when two coreboot folders are present,
      both containing an IFDTOOL.

Signed-off-by: Patrick Rudolph <patrick.rudolph@9elements.com>

### DIFF
--- a/blobs/t420/extract.sh
+++ b/blobs/t420/extract.sh
@@ -28,7 +28,7 @@ while getopts ":f:m:i:" opt; do
 done
 
 if [ -z "$MECLEAN" ]; then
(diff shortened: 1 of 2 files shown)


SebastianMcMillan:
Just a note: the T420 script is identical since all I did was copy the X220 script into the T420 folder. Maybe that would be good to fix in this PR as well?

PatrickRudolph:
done.

Issue #769, opened 2020-07-02
CircleCI: seperate caches based on core modules measurements


tlaurion:
The idea here (to replicate on #768 GitlabCI):
- Have ./crossgcc unique compiler cache based on modules/compiler file being measured
- Have ./build/* unique build cache based on modules/* files being measured

That way:
- bash scripts modification won't require to rebuild any module
- When modules/* change, only the build cache will be remade and made unique
- When ./crossgcc requires to be remade, crossgcc unique cache will be remade and unique

As of right now, the cache is made a single time and never remade, based on:
```
- save_cache:
          key: heads-{{ .Environment.CIRCLE_USERNAME }}-{{ .Environment.CACHE_VERSION }}
```

and 

```
 - restore_cache:
          key: heads-{{ .Environment.CIRCLE_USERNAME }}-{{ .Environment.CACHE_VERSION }}

```

Failed attempts to specialize CircleCI caches happened [here](https://app.circleci.com/pipelines/github/tlaurion/heads?branch=CircleCI_specialized_caches)

tlaurion:
Note: not only modules should be measured to create a distinct cache key (generation/extraction in CIs), but also the patches linked to modules, else they will never be applied.

So the best key would probably be:

- [ ] ./crossgcc being cached if ./modules/musl-cross and ./patches/musl-cross* are still the same (longest to build, mostly never changed. First level of cache restored if second level fails to still speedup builds by1+hour)
- [ ] ./modules/* ./patches/* directories content digest (because patches are applied only at module decompression and never revalidated) resulting in /build being cached and restored if measurements are still valid

This simple CI hack would speed up rebuilding most Heads changes, which are mostly scripts.

tlaurion:
This ticket priority just got promoted since #795 requires a cache of coreboot 4.11 while other platforms requires a cache of coreboot 4.8.1 with different patches being applied.... Will try to wrap my head around https://app.circleci.com/pipelines/github/tlaurion/heads?branch=CircleCI_specialized_caches and add https://github.com/osresearch/heads/issues/769#issuecomment-671476773 requirements

tlaurion:
Second attempt here: 
https://github.com/tlaurion/heads/commit/9ed4690af02e9bf62c6cbe52c39e6afc7e5e332f

Failing result here: https://app.circleci.com/pipelines/github/tlaurion/heads/283/workflows/b3f64a9b-25f9-4194-9884-849a1fd9f897/jobs/308

```
[...]

PR #798, opened 2020-08-11
CircleCI : specialized caches

tlaurion:
This changes the way CircleCI handles cache and limit the possibility of cache corruptions and removes the CircleCI owner bounding that was there previously.

Now:
- If modules/* and patches/* have the same hashes, that cache is restored. It saves the new cache if the hashes are different. This cache is usefull for most changes happening in Heads, which are security policy related and linked to scripts changes. If modules or hashes changes and no cache exists, if will restore the next cache.
- If musl-cross module and applied patches have the same hashes of an existing cache, that cache is restored. It is saved once for a single hash which means that the module version (commit) and applied patches are the same. This permits to economize one hour of build time for each build and rarely ever change.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,9 +24,9 @@ jobs:
       - restore_cache:
           keys:
             #Restore existing cache for modules checksums validated to be exactly the same as in github current commit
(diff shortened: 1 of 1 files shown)


tlaurion:
Fixes #769

PR #805, opened 2020-08-18
Support Multiple Kernel Options

MrChromebox:
Add support for multiple kernel versions, following the same pattern as used for coreboot. 
Set existing kernel version as the default for all existing boards.

Add option to use kernel 4.19.139 (current LTS version).
Duplicate existing patches from 4.14.62 as they all apply cleanly.

Subsequent PRs can move boards to newer kernel as needed/desired.


tlaurion:
@MrChromebox 
We will go a bit more conservative (learning from merging #721 which broke 4.8.1 builds now under #807) and switch boards one by one this time, making sure that we won't break anything. 

#807 CircleCI needs to be fixed first with current kernel prior of being able to CircleCI with this PR.

MrChromebox:
@tlaurion intention isn't to switch boards, as likely no benefit for most existing ones -- though we may eventually want to move to a 5.x kernel universally. This is needed to support some other Librem boards I want to upstream (Librem Mini, Librem server L1UM) which require a newer kernel, either for platform support or performance reasons.

tlaurion:
@MrChromebox 
> @tlaurion intention isn't to switch boards, as likely no benefit for most existing ones -- though we may eventually want to move to a 5.x kernel universally. This is needed to support some other Librem boards I want to upstream (Librem Mini, Librem server L1UM) which require a newer kernel, either for platform support or performance reasons.

Got it. So this PR is to fixate linux kernel version to actual version so that we are able to switch board configs one by one. Read to fast. Will stil ltest when having a functional CI again....

tlaurion:
Building here https://app.circleci.com/pipelines/github/tlaurion/heads/334/workflows/5eb74b45-46db-4f13-b789-803237929377/jobs/362

MrChromebox:
rebased

tlaurion:
Passed build https://app.circleci.com/pipelines/github/tlaurion/heads/335/workflows/f70d9506-3c41-42c6-8161-41e874bca62a/jobs/363

Merging

Issue #807, opened 2020-08-19
CircleCI not able to build Coreboot 4.8.1 after coreboot module change


tlaurion:
@MrChromebox 
Building [without cache](https://app.circleci.com/pipelines/github/tlaurion/heads/315/workflows/3a003bc2-676d-489f-8e0d-d3189e9223e0/jobs/343) since commit  c6a82a246411f091ccebdc683da83fafd989b08a results in:
```
gcc -I/root/project/build/coreboot-4.8.1/util/sconfig -Ix230-flash/util/sconfig -o x230-flash/util/sconfig/sconfig x230-flash/util/sconfig/lex.yy.o x230-flash/util/sconfig/sconfig.tab.o x230-flash/util/sconfig/main.o
/usr/bin/ld: x230-flash/util/sconfig/main.o:/root/project/build/coreboot-4.8.1/util/sconfig/sconfig.h:80: multiple definition of `head'; x230-flash/util/sconfig/sconfig.tab.o:/root/project/build/coreboot-4.8.1/util/sconfig/sconfig.h:80: first defined here
collect2: error: ld returned 1 exit status
make[2]: *** [util/sconfig/Makefile.inc:43: x230-flash/util/sconfig/sconfig] Error 1
rm x230-flash/util/cbfstool/fmd_scanner.c x230-flash/util/cbfstool/fmd_parser.c
make[2]: Leaving directory '/root/project/build/coreboot-4.8.1'
make[1]: *** [Makefile:382: /root/project/build/coreboot-4.8.1/x230-flash/.build] Error 1
make[1]: Leaving directory '/root/project'
make: *** [Makefile:587: all] Error 2
```

Any idea?

tlaurion:
Same happened back there at that commit [from CircleCI](https://app.circleci.com/pipelines/github/osresearch/heads/196/workflows/168e94c2-be36-47bd-8847-80a0be4e4b63/jobs/209)

MrChromebox:
what was the resolution when the issue first surfaced back then?

tlaurion:
The following successful tests were to test 4.12.

I'm trying this https://app.circleci.com/pipelines/github/tlaurion/heads/316/workflows/d636d5c4-d490-40c3-bfab-ac5e4aee95b5/jobs/344

tlaurion:
[...]

PR #808, opened 2020-08-20
Fix CircleCI build problems

### LINKED ISSUE #799: Makefile: make 4.2.1 is not respecting Heads logic verification for already extracted packages
We can see the result here when a cache is restored without hacks:
https://app.circleci.com/pipelines/github/tlaurion/heads/293/workflows/e50cfc8b-c61f-474b-b2de-476e3cee845d/jobs/319/parallel-runs/0/steps/0-107

We see here that the build fails since make was already extracted there and the patch was already applied.
@osresearch ?

My hack was to delete that directory prior of building when qemu-coreboot was first built [here](url), but it would be nice if make was also respecting Heads build system logic.

tlaurion:
- Fix issue #807
- Esthetical modules/coreboot changes to facilitate future coreboot maintainership putting hashes of coreboot and coreboot-blobs together
- Fix #799 with implementation of @osresearch's recommendation.

@MrChromebox : please review

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 jobs:
   build:
     docker:
-      - image: debian:bullseye
+      - image: debian:10
     steps:
       - run:
           name: Install dependencies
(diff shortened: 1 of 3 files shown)


tlaurion:
Build success. merging

PR #810, opened 2020-08-21
flash.sh: Fix serial number persistence

MrChromebox:
If the serial_number file exists in cbfs but has zero size,
adding it to the new ROM to be flashed will fail badly.
Check for non-zero size before adding to new ROM.


tlaurion:
@MrChromebox linked to flashtools having been upgraded.  That PR should also upgrade flashtool to latest commit

MrChromebox:
@tlaurion yes. but points to flashtools not being fixed to a specific version/commit. That's a separate PR I suppose ;-)

PR #832, opened 2020-09-14
CircleCI: readd librem_mini

tlaurion:
while making sure that if a board build f…ails, all logfiles modified in the last minute are outputted on the screen prior to really failing and exiting.


tlaurion:
@MrChromebox build was successful here: https://app.circleci.com/pipelines/github/tlaurion/heads/383/workflows/ad2a4ae1-803f-4f7e-848e-e4f76204ee5a/jobs/412

Please test artifact:
https://412-103208611-gh.circle-artifacts.com/0/build/librem_mini/coreboot.rom

MrChromebox:
@tlaurion it boots but isn't functional, seems to be missing a lot of handling for the non-TPM case

tlaurion:
@MrChromebox 
`export CONFIG_TPM=n` is not present in board config?

tlaurion:
@MrChromebox : you need two board configs, one with TPM and one without, just like in https://github.com/osresearch/heads/pull/832/commits/004e36f8604423050f8e71d60281375b4db4684b ?

tlaurion:
So following functional https://app.circleci.com/pipelines/github/tlaurion/heads/387/workflows/498762bd-a640-4bb0-9cf0-ba7d66d71abb/jobs/416

MrChromebox:
@tlaurion the Librem Mini doesn't have a TPM, so the w/TPM build isn't useful. The no-TPM build works better than the initial one (I was able to perform a factory reset), but still fails due to some functions expecting a TPM and not handling the lack of one properly (like generating a new secret)

tlaurion:
@MrChromebox : You want me to delete the standard board and replace with the NoTPM board?

This PR shows that building coreboot 4.12 and coreboot 4.8.1 from CI works. 
I would want to have this board onboard (build), with improvements coming with future PR with fixes related with TPM-less (non-current to Heads current use cases) merged.

AFAIK Librem Key/Nitrokey Pro/ Nitrokey Storage HOTP support for TPM less board is stalled with proposed improvements here:
https://github.com/osresearch/heads/pull/493

MrChromebox:
@tlaurion please do. I must have missed the explicit export when upstreaming since #493 (or some version thereof) is in Purism's tree and so explicitly setting TPM=n isn't necessary.

w/r/t #493, what needs to happen there to move it forward? Or, are there smaller chunks within it that could be easily accepted now to improve the non-TPM case?

tlaurion:
> @tlaurion please do. I must have missed the explicit export when upstreaming since #493 (or some version thereof) is in Purism's tree and so explicitly setting TPM=n isn't necessary.
> 
> w/r/t #493, what needs to happen there to move it forward? Or, are there smaller chunks within it that could be easily accepted now to improve the non-TPM case?

@MrChromebox: I am not sure about this implementation, TBH. We are talking about TPM being auto-detected and used if found, and codepath changing if not, right?

I think those discussions should happen in seperate PRs.

So for the current one. You want me to delete librem_mini board dir and associated coreboot config and keep the librem_mini-NoTPM board and coreboot config?

Please resume discussion inside of #493 and do PR to propose dynamic TPM detection upstream. We want the less changes between upstream Heads and sidetracked implementations, else we do not know what we are talking about.

MrChromebox:
> So for the current one. You want me to delete librem_mini board dir and associated coreboot config and keep the librem_mini-NoTPM board and coreboot config?

correct

> Please resume discussion inside of #493 and do PR to propose dynamic TPM detection upstream. We want the less changes between upstream Heads and sidetracked implementations, else we do not know what we are talking about

copy that

Issue #771, opened 2020-07-05
GPG keypair generation from Heads, then copied to USB Security dongle


tlaurion:
Current customer concerns:
- What do I do if I loose the USB Security dongle. Worst : drown USB security dongle in pool, toilet.
- What do I do if I lock myself out of the USB Security dongle Admin PIN?

Solutions:
- [ ] Have the keypair generated inside of Heads (/tmp) then offered to be [copied to USB Security dongle](https://docs.puri.sm/Librem_Key/Getting_Started/User_Manual.html#generate-gpg-keys-on-your-computer) with a big fat warning if destination backuppartition mountpoint is not encrypted? (I would deny such export)
- [ ] Have the factory reset code also generate a Reset Code (that permits to unlock Admin PIN)

EDIT: implementation plan here: https://github.com/osresearch/heads/issues/771#issuecomment-740850208

tlaurion:
@alex-nitrokey @kylerankin @MrChromebox : Thoughts?

alex-nitrokey:
As far as I can see, the function of the dongle is to detect unauthorized alternation of the firmware/everything in /boot. So, losing/breaking the dongle does solely mean that the user needs a new device to be able to detect such change on the system again. The problem arises if the dongle gets lost during a situation in which the pc was unattended and exposed as well.

One could implement a feature to prepare two dongles at once during oem factory reset to have a backup device.

Imho it is not useful to let the key be exported if you like to make Heads secure and easy to use for the innocent. If you really like a full backup of your GnuPG keys you already can prepare your dongle first and then provide the public key to Heads - though this is for the more experienced users.

I do not see how rising the PIN size should help here?

MrChromebox:
a second dongle isn't going to be a drop-in replacement since the counters will be out of sync, and you'd still have to generate a new secret with the backup dongle.

techge:
> a second dongle isn't going to be a drop-in replacement since the counters will be out of sync

A right, forgot that.

> and you'd still have to generate a new secret with the backup dongle.

This would have been solvable by copying the secret to both dongles, I guess.

tlaurion:
The goal here is really to be loss-proof from encrypted mails perspective. I think it is ok to accept the user to have to validate with Disk Unlock Key code and or TOTP and not update dom0 before having secrets sealed through HOTP on new device with options provided below.

Menu options:
- [ ] Heads keypair generation + key copied to smartcard + PINs generation + Reseal HOTP secret
[...]

Issue #453, opened 2018-09-17
Heads port to Lenovo G505s


tlaurion:
Someone has this laptop?

awokd:
Affirmative, and contact info for a few other owners. What's up?

tlaurion:
@awokd : some of the owners are developers and would be willing to port Heads to them? Would you tag those owners or invite them here? I've ported ot he KGPE-D16, would gladly port Heads to hte G505s if I had access to one.

Have nice and cheap providers to buy from to suggest? Some of you could lend me the hardware for the time being of the porting?

awokd:
- No TPM, is that OK for Heads?
- A10-5750M 4 core CPU with built in GPU, some models have an additional external GPU
- Not aware of any formal providers, can only find in used market

If TPM isn't required and there's no serious C level coding involved, I can give it a try. Might need some hand-holding, though. Do you have a step-by-step of how you ported Heads to that KPGE-D16? G505s ought to be pretty similar. If no TPM doesn't quash the idea, I'll drop a line to others and point them here to see if there's any developer availability.

tlaurion:
TPM is not a show stopper, while better.
Librem/Nitrokey Pro v2 keys can validate the integrity of the firmware in a
similar way externally.

Check the porting guide on heads-wiki project from osresearch's github.

Can give you a hand but won't be able to debug. Working Linux and coreboot
configurations are required. You can copy the kgpe-d16 board config as a
start.

tlaurion:
@awokd : Coreboot 4.8.1 hasn't been merged into Heads yet...
```
git remote add kylerankin https://github.com/kylerankin/heads
git remote add flammit https://github.com/flammit/heads
git merge flammit/coreboot-4.8 
make BOARD=x230 modules.clean
make BOARD=x230 -j4
```

tlaurion:
@awokd: coreboot 4.8.1 is merged.
Is there any advancements?

awokd:
Thank you, I will try it again when I get a few!

zamaudio:
I used to have this laptop and co-authored the coreboot port for it.  Unfortunately I no longer have the hardware and it's quite rare around here even on the second hand market :(

awokd:
Damien Zammit wrote on 12/27/18 4:02 AM:
> I used to have this laptop and co-authored the coreboot port for it.  Unfortunately I no longer have the hardware and it's quite rare around here even on the second hand market :(
> 

[...]

PR #326, opened 2018-02-24
Introduce io386 to heads and use it to finalize chipset at runtime (PR0 initial work)

persmule:
On some newer platforms of intel (confirmed on nehalem, sandy/ivy
bridge), coreboot after commit [2ac149d294af795710eb4bb20f093e9920604abd](https://github.com/coreboot/coreboot/commit/2ac149d294af795710eb4bb20f093e9920604abd)
registers an SMI to lockdown some registers on the chipset, as well
as access to the SPI flash, optionally. The SMI will always be triggered
by coreboot during S3 resume, but can be triggered by either coreboot
or the payload during normal boot path.

Enabling lockdown access to SPI flash will effectly write-protect it,
but there is no runtime option for coreboot to control it, so letting
coreboot to trigger such SMI will leave the owner of the machine lost
any possibility to program the SPI flash with its own OS, and becomes
a nightmare if the machine is uneasy to disassemble, so a scheme could
be implement, in which the SMI to lockdown chipset and SPI flash is left
for a payload to trigger, and temporarily disabling such triggering in
order to program the SPI flash needs authentication.

I have implemented a passcode-protected runtime-disableable lockdown
with grub, described [here](https://github.com/hardenedlinux/Debian-GNU-Linux-Profiles/blob/master/docs/hardened_boot/grub-for-coreboot.md#update-for-coreboot-after-commit-2ac149d294af795710eb4bb20f093e9920604abd). In order to implement a similar scheme for
Heads, I wrote [io386](https://github.com/hardenedlinux/io386).

With this commit, io386 will be called before entering boot routine
to trigger the SMI to finalize the chipset and write protect the SPI
flash at the same time. Entering recovery shell will leave the flash
writable.

(The authentication routine implemented in previous revisions has been
split as an independent commit.)

PR #836, opened 2020-09-16
Use a USB security key as a TPM work-alike in the absence of a physical TPM

MrChromebox:
On machines without a TPM, we'd still like some way for the BIOS to
attest that it has not been modified. With a USB security key, we can have
the BIOS use its own ROM measurement converted to a SHA256sum and truncated
so it fits within an HOTP secret. Like with a TPM, a malicious BIOS with
access to the correct measurements can send pre-known good measurements
to the USB security key.

This approach provides one big drawback in that we have to truncate the
SHA256sum to 20 characters so that it fits within the limitations of
HOTP secrets. This means the possibility of collisions is much higher
but again, an attacker could also capture and spoof an existing ROM's
measurements if they have prior access to it, either with this approach
or with a TPM.

This supersedes #493

### DIFF
--- a/initrd/bin/flash.sh
+++ b/initrd/bin/flash.sh
@@ -8,7 +8,7 @@ set -e -o pipefail
 
 case "$CONFIG_FLASHROM_OPTIONS" in
   -* )
-    echo "Board $CONFIG_BOARD detected, continuing..."
+    [ "$1" != "-s" ] && echo "Board $CONFIG_BOARD detected, continuing..."
   ;;
   * )
     die "ERROR: No board has been configured!\n\nEach board requires specific flashrom options and it's unsafe to flash without them.\n\nAborting."
@@ -30,6 +30,10 @@ flash_rom() {
     else
       die "$ROM: Read inconsistent"
     fi
+  elif [ "$SHA" -eq 1 ]; then
+    flashrom $CONFIG_FLASHROM_OPTIONS -r "${ROM}" 1&>2 >/dev/null \
+    || die "$ROM: Read failed"
+    sha256sum ${ROM} | cut -f1 -d ' '
   else
     cp "$ROM" /tmp/${CONFIG_BOARD}.rom
     sha256sum /tmp/${CONFIG_BOARD}.rom
@@ -52,20 +56,29 @@ flash_rom() {
 if [ "$1" == "-c" ]; then
   CLEAN=1
   READ=0
+  SHA=0
   ROM="$2"
 elif [ "$1" == "-r" ]; then
   CLEAN=0
   READ=1
+  SHA=0
+  ROM="$2"
+  touch $ROM
+elif [ "$1" == "-s" ]; then
+  CLEAN=0
+  READ=0
+  SHA=1
   ROM="$2"
   touch $ROM
 else
   CLEAN=0
   READ=0
+  SHA=0
   ROM="$1"
 fi
 
 if [ ! -e "$ROM" ]; then
-	die "Usage: $0 [-c|-r] <path_to_image.rom>"
+	die "Usage: $0 [-c|-r|-s] <path_to_image.rom>"
 fi
 
 flash_rom $ROM
--- a/initrd/etc/functions
+++ b/initrd/etc/functions
@@ -267,6 +267,19 @@ combine_configs() {
 	cat /etc/config* > /tmp/config
 }
 
+# Generate secret value using first 40 chars of ROM SHA256 hash
+secret_from_rom_hash() {
+	local ROM_IMAGE="/tmp/coreboot-notpm.rom"
+
+	echo -e "\nTPM not detected; measuring ROM directly\n" 1>&2
+	# use a previously-copied image if it exists
+	if [ -f ${ROM_IMAGE} ]; then
+		sha256sum ${ROM_IMAGE} | cut -f1 -d ' ' | cut -c 1-40 | tr -d '\n'
+	else
+		flash.sh -s ${ROM_IMAGE} | cut -c 1-40 | tr -d '\n'
+	fi
+}
+
 update_checksums()
 {
 	# clear screen
--- a/initrd/init
+++ b/initrd/init
@@ -44,6 +44,13 @@ hwclock -l -s
 . /etc/functions
 . /etc/config
 
+# set CONFIG_TPM dynamically before init
+if [ -e /dev/tpm0 ]; then
+  export CONFIG_TPM='y'
+else
+  export CONFIG_TPM='n'
+fi
+
 if [ "$CONFIG_COREBOOT" = "y" ]; then
 	/bin/cbfs-init
 fi
@@ -87,6 +94,13 @@ if [ "$boot_option" = "r" ]; then
 	fi
 	exec /bin/ash
 	exit
+fi
+
+# Override CONFIG_TPM and persist via user config
+if [ -e /dev/tpm0 ]; then
+  echo "CONFIG_TPM=y" >> /etc/config.user
+else
+  echo "CONFIG_TPM=n" >> /etc/config.user
 fi
 
 combine_configs
--- a/initrd/bin/seal-hotpkey
+++ b/initrd/bin/seal-hotpkey
@@ -24,19 +24,24 @@ else
 	HOTPKEY_BRANDING="HOTP USB Security Dongle"
 fi
 
-tpm nv_readvalue \
-	-in 4d47 \
-	-sz 312 \
-	-of "$HOTP_SEALED" \
-|| die "Unable to retrieve sealed file from TPM NV"
+if [ "$CONFIG_TPM" = "y" ]; then
+	tpm nv_readvalue \
+		-in 4d47 \
+		-sz 312 \
+		-of "$HOTP_SEALED" \
+	|| die "Unable to retrieve sealed file from TPM NV"
 
-tpm unsealfile  \
-	-hk 40000000 \
-	-if "$HOTP_SEALED" \
-	-of "$HOTP_SECRET" \
-|| die "Unable to unseal HOTP secret"
+	tpm unsealfile  \
+		-hk 40000000 \
+		-if "$HOTP_SEALED" \
+		-of "$HOTP_SECRET" \
+	|| die "Unable to unseal HOTP secret"
 
-shred -n 10 -z -u "$HOTP_SEALED" 2> /dev/null
+	shred -n 10 -z -u "$HOTP_SEALED" 2> /dev/null
+else
+	# without a TPM, use the first 20 characters of the ROM SHA256sum 
+	secret_from_rom_hash > $HOTP_SECRET
+fi
 
 # Store counter in file instead of TPM for now, as it conflicts with Heads
 # config TPM counter as TPM 1.2 can only increment one counter between reboots
--- a/initrd/bin/seal-totp
+++ b/initrd/bin/seal-totp
@@ -17,71 +17,78 @@ fi
 TOTP_SECRET="/tmp/secret/totp.key"
 TOTP_SEALED="/tmp/secret/totp.sealed"
 
-dd \
-	if=/dev/urandom \
-	of="$TOTP_SECRET" \
-	count=1 \
-	bs=20 \
-	2>/dev/null \
-|| die "Unable to generate 20 random bytes"
+if [ "$CONFIG_TPM" = "y" ]; then
+	dd \
+		if=/dev/urandom \
+		of="$TOTP_SECRET" \
+		count=1 \
+		bs=20 \
+		2>/dev/null \
+	|| die "Unable to generate 20 random bytes"
 
-secret="`base32 < $TOTP_SECRET`"
+	secret="`base32 < $TOTP_SECRET`"
 
-# Use the current values of the PCRs, which will be read
-# from the TPM as part of the sealing ("X").
-# PCR4 == 0 means that we are still in the boot process and
-# not a recovery shell.
-# should this read the storage root key?
-if ! tpm sealfile2 \
-	-if "$TOTP_SECRET" \
-	-of "$TOTP_SEALED" \
-	-hk 40000000 \
-	-ix 0 X \
-	-ix 1 X \
-	-ix 2 X \
-	-ix 3 X \
-	-ix 4 0000000000000000000000000000000000000000 \
-	-ix 7 X \
-; then
+	# Use the current values of the PCRs, which will be read
+	# from the TPM as part of the sealing ("X").
+	# PCR4 == 0 means that we are still in the boot process and
+	# not a recovery shell.
+	# should this read the storage root key?
+	if ! tpm sealfile2 \
+		-if "$TOTP_SECRET" \
+		-of "$TOTP_SEALED" \
+		-hk 40000000 \
+		-ix 0 X \
+		-ix 1 X \
+		-ix 2 X \
+		-ix 3 X \
+		-ix 4 0000000000000000000000000000000000000000 \
+		-ix 7 X \
+	; then
+		shred -n 10 -z -u "$TOTP_SECRET" 2> /dev/null
+		die "Unable to seal secret"
+	fi
+
 	shred -n 10 -z -u "$TOTP_SECRET" 2> /dev/null
-	die "Unable to seal secret"
-fi
-
-shred -n 10 -z -u "$TOTP_SECRET" 2> /dev/null
 
 
-# to create an nvram space we need the TPM owner password
-# and the TPM physical presence must be asserted.
-#
-# The permissions are 0 since there is nothing special
-# about the sealed file
-tpm physicalpresence -s \
-|| warn "Warning: Unable to assert physical presence"
+	# to create an nvram space we need the TPM owner password
+	# and the TPM physical presence must be asserted.
+	#
+	# The permissions are 0 since there is nothing special
+	# about the sealed file
+	tpm physicalpresence -s \
+	|| warn "Warning: Unable to assert physical presence"
 
-# Try to write it without the password first, and then create
-# the NVRAM space using the owner password if it fails for some reason.
-if ! tpm nv_writevalue \
-	-in $TPM_NVRAM_SPACE \
-	-if "$TOTP_SEALED" \
-; then
-	warn 'NVRAM space does not exist? Owner password is required'
-	read -s -p "TPM Owner password: " tpm_password
-	echo
-
-	tpm nv_definespace \
-		-in $TPM_NVRAM_SPACE \
-		-sz 312 \
-		-pwdo "$tpm_password" \
-		-per 0 \
-	|| die "Unable to define NVRAM space"
-
-	tpm nv_writevalue \
+	# Try to write it without the password first, and then create
+	# the NVRAM space using the owner password if it fails for some reason.
+	if ! tpm nv_writevalue \
 		-in $TPM_NVRAM_SPACE \
 		-if "$TOTP_SEALED" \
-	|| die "Unable to write sealed secret to NVRAM"
-fi
+	; then
+		warn 'NVRAM space does not exist? Owner password is required'
+		read -s -p "TPM Owner password: " tpm_password
+		echo
 
-shred -n 10 -z -u "$TOTP_SEALED" 2> /dev/null
+		tpm nv_definespace \
+			-in $TPM_NVRAM_SPACE \
+			-sz 312 \
+			-pwdo "$tpm_password" \
+			-per 0 \
+		|| die "Unable to define NVRAM space"
+
+		tpm nv_writevalue \
+			-in $TPM_NVRAM_SPACE \
+			-if "$TOTP_SEALED" \
+		|| die "Unable to write sealed secret to NVRAM"
+	fi
+
+	shred -n 10 -z -u "$TOTP_SEALED" 2> /dev/null
+else
+	# without a TPM, use the first 20 characters of the ROM SHA256sum 
+	secret_from_rom_hash > $TOTP_SECRET
+	secret="`base32 < $TOTP_SECRET`"
+	shred -n 10 -z -u "$TOTP_SECRET" 2> /dev/null
+fi
 
 url="otpauth://totp/$HOST?secret=$secret"
 secret=""
--- a/initrd/bin/unseal-hotp
+++ b/initrd/bin/unseal-hotp
@@ -36,19 +36,24 @@ fi
 
 #counter_value=$(printf "%d" 0x${counter_value})
 
-tpm nv_readvalue \
-	-in 4d47 \
-	-sz 312 \
-	-of "$HOTP_SEALED" \
-|| die "Unable to retrieve sealed file from TPM NV"
+if [ "$CONFIG_TPM" = "y" ]; then
+	tpm nv_readvalue \
+		-in 4d47 \
+		-sz 312 \
+		-of "$HOTP_SEALED" \
+	|| die "Unable to retrieve sealed file from TPM NV"
 
-tpm unsealfile  \
-	-hk 40000000 \
-	-if "$HOTP_SEALED" \
-	-of "$HOTP_SECRET" \
-|| die "Unable to unseal HOTP secret"
+	tpm unsealfile  \
+		-hk 40000000 \
+		-if "$HOTP_SEALED" \
+		-of "$HOTP_SECRET" \
+	|| die "Unable to unseal HOTP secret"
 
-shred -n 10 -z -u "$HOTP_SEALED" 2> /dev/null
+	shred -n 10 -z -u "$HOTP_SEALED" 2> /dev/null
+else
+	# without a TPM, use the first 20 characters of the ROM SHA256sum 
+	secret_from_rom_hash > $HOTP_SECRET
+fi
 
 if ! hotp $counter_value < "$HOTP_SECRET"; then
 	shred -n 10 -z -u "$HOTP_SECRET" 2> /dev/null


tlaurion:
@MrChromebox 
 20 chars limitation was fixed a long time ago in firmware, exposed here: https://github.com/osresearch/heads/pull/493#issuecomment-487379379 so no need to truncate prematurely sha256sum to 20 characters in newer upstreamed and produced Lirem Keys/Nitrokey Pro versions manufactureed since 2019.

Questions of how to deal with old keys is here: https://github.com/osresearch/heads/pull/493#issuecomment-487379379 with returns if users really want this from both Purism (@kylerankin ) /Nitrokey? (@jans23) ? 

Else a big fat warning could be given to users if using a USB Security dongle's firmware (detectable) requiring to truncate secret in insecure ways, giving users a risk choice they should be clear about prior of using such technology for tamper evidence.

tlaurion:
@MrChromebox : i'm so confused by the build failing. Logs (now capture) saying that the compiler was killed. Too memory hungry?!

tlaurion:
Not a cryptographer here, but my intuition is that it might be better to use sha1sum vs sha256sum (tpm 1.1 measurements are sha1 anyway, so would be equivalent) resulting in hash that would fit under those 40 chars buffer that we have available. 

Better to truncate a 64 char into 40 (or worse, twenty...) or use lower checksuming techniques to fill buffer available?

initrd/bin/flash.sh
```diff
@@ -30,6 +30,10 @@ flash_rom() {
     else
       die "$ROM: Read inconsistent"
     fi
+  elif [ "$SHA" -eq 1 ]; then
+    flashrom $CONFIG_FLASHROM_OPTIONS -r "${ROM}" 1&>2 >/dev/null \
+    || die "$ROM: Read failed"
+    sha256sum ${ROM} | cut -f1 -d ' '
```
tlaurion, line 36: sha1sum fitting in available 40 chars?

initrd/etc/functions
```diff
@@ -261,6 +261,19 @@ combine_configs() {
...
 }
 
+# Generate secret value using first 20 chars of ROM SHA256 hash
+secret_from_rom_hash() {
+	local ROM_IMAGE="/tmp/coreboot-notpm.rom"
+
+	echo -e "\nTPM not detected; measuring ROM directly\n" 1>&2
+	# use a previously-copied image if it exists
+	if [ -f ${ROM_IMAGE} ]; then
+		sha256sum ${ROM_IMAGE} | cut -f1 -d ' ' | cut -c 1-20 | tr -d '\n'
```
tlaurion, line 271: sha1sum fitting in 40 char?

initrd/etc/functions
```diff
@@ -261,6 +261,19 @@ combine_configs() {
...
+	else
+		flash.sh -s ${ROM_IMAGE} | cut -c 1-20 | tr -d '\n'
```
tlaurion, line 273: No need to cut if sha1sum instead of sha256sum and firmware 0.8+?

initrd/init
```diff
@@ -44,6 +44,13 @@ hwclock -l -s
 . /etc/functions
 . /etc/config
 
+# set CONFIG_TPM dynamically before init
+if [ -e /dev/tpm0 ]; then
```
tlaurion, line 48: What if TPM lines are cut? What is the behavior and conditional codepath effects here?

initrd/init
```diff
@@ -89,6 +96,13 @@ if [ "$boot_option" = "r" ]; then
 	exit
 fi
 
+# Override CONFIG_TPM and persist via user config
+if [ -e /dev/tpm0 ]; then
+  echo "CONFIG_TPM=y" >> /etc/config.user
+else
+  echo "CONFIG_TPM=n" >> /etc/config.user
```
tlaurion, line 103: Not measured. No change detected.

initrd/init
```diff
@@ -89,6 +96,13 @@ if [ "$boot_option" = "r" ]; then
...
+fi
+
 combine_configs
```
tlaurion, line 106: Combined in ram, no change detected.

jans23:
> Questions of how to deal with old keys is here: #493 (comment) with returns if users really want this from both Purism (@kylerankin ) /Nitrokey? (@jans23) ?

Loosing backward compatibility to old Nitrokeys is ok for me.

kylerankin:
My preference is to detect the firmware version and default to the
strongest method their hardware/firmware happens to support instead of
presenting users with scary prompts or expecting them to understand the
relative strengths of hashing algorithms.

MrChromebox:
@tlaurion @jans23:  Kyle and I tested here, and all our of LKs are using firmware >= 0.8, so they all support 40-byte keys. So we're fine dropping support for/breaking older NK/LK firmwares.

I'll update the PR to use 40-byte truncation; Kyle and I agree that SHA256 truncated is still better than SHA1 not.

initrd/init
MrChromebox: I'd assume that a TPM with cut lines would behave exactly like a device without a TPM, since there would be no communication and /dev/tpm0 would not exist

tlaurion: @MrChromebox my point here is that since the codepath is dynamic and not set inside of ROM inside of /etc/config anymore, cutting the lines of TPM would simply result in a different codepath without the user knowing the his TPM lines were cut, and from the LIBREM KEY/ Nitrokey Pro Nitrokey Storage being used for validation. I'm not sure this is desirable.

MrChromebox: the user would absolutely be notified, since there would be text indicating no TPM on the main menu, and LK/NK verification would fail as the secret would have changed from the TPM-backed one to the hash-based one.

tlaurion:
@MrChromebox could you rebase on master to trigger a build please?

tlaurion:
@MrChromebox Is the librem_mini taking a change in its board config, or should you provide an additional board config added in CircleCI to be able to test this platform and being pushed at the same time to give example for future platform integrations?

tlaurion:
@jans23 @MrChromebox @alex-nitrokey @kylerankin:

Shouldn't this feature be activated per additional boards, by explicitely specifying its configuration flags?
`export CONFIG_TPM=n` && `export CONFIG_HOTPKEY=y` (We have no TPM on platform but have a HOTPKEY compatible USB Security dongle, documented as well, to do TPM similar remote attestation)

MrChromebox:
> @MrChromebox Is the librem_mini taking a change in its board config, or should you provide an additional board config added in CircleCI to be able to test this platform and being pushed at the same time to give example for future platform integrations?

I'm not understanding this question at all. There is a single librem_mini board config. It does not have a TPM. It will always use a HOTP key.

MrChromebox:
part of the issue here is that there is no handling for boards without a TPM (or with a deactivated vTPM) -- they just fail miserably because /dev/tpm0 doesn't exist. I'm happy to split this into two PRs -- handing of boards without a TPM, and handling of boards without a TPM and with a HOTP key, but we need to define what the former looks like first

tlaurion:
> > @MrChromebox Is the librem_mini taking a change in its board config, or should you provide an additional board config added in CircleCI to be able to test this platform and being pushed at the same time to give example for future platform integrations?
> 
> I'm not understanding this question at all. There is a single librem_mini board config. It does not have a TPM. It will always use a HOTP key.

Sorry for the confusion @MrChromebox, I wrote [this](https://github.com/osresearch/heads/pull/836#issuecomment-710593094) prior of [reviewing board](https://github.com/osresearch/heads/pull/836#issuecomment-710607555).
 
After reviewing board config, my point here is to clearly state what makes it different as a board in terms of configuration: without a TPM while requiring HOTPKEY supported USB Security dongle.

tlaurion:
> part of the issue here is that there is no handling for boards without a TPM (or with a deactivated vTPM) -- they just fail miserably because /dev/tpm0 doesn't exist. I'm happy to split this into two PRs -- handing of boards without a TPM, and handling of boards without a TPM and with a HOTP key, but we need to define what the former looks like first

@MrChromebox Agreed that there are some codepaths not dealing properly with boards without a TPM. My tests back in 2018 with the [KGPE-D16](https://github.com/tlaurion/heads/commit/100751d2683342495a9ff269a6952d26f1b0153f) and x200 were with generic-init, so I have not thoroughly tested newer gui-init post https://github.com/osresearch/heads/issues/477

tlaurion:
We will have to reflect on a way to disable hid functions of USB Security dongles inside of heads guys!
@MrChromebox @jans23 @szszszsz @alex-nitrokey @daringer @kylerankin

szszszsz:
Hi @tlaurion ! Could you elaborate on disabling the HID interface mentioned in https://github.com/osresearch/heads/pull/836#issuecomment-711176858? I do not see such need coming from this ticket context, unless I am missing something.

tlaurion:
@szszszsz it's not specific to this PR, but HID without TPM released Disk Unlock Key opens the door to rubber ducky attacks where Heads could be instructed to go into recovery, sed files and reenter GUI, showing measurements as being good and where Heads scripts are not validated for integrity. 

The only protection Heads offers on this is that [entering recovery](https://github.com/osresearch/heads/blob/master/initrd/etc/functions#L13-L42) changes [PCR4](https://github.com/osresearch/heads-wiki/blob/master/Keys.md#tpm-pcrs), invalidating TPM released content. When TPM + TPM released Disk Unlock Key codepath is active, the user, even if receiving `HOTP:Success`, would still have to enter TPM nvram passphrase to be able to boot system default. 

In the current PR, if someone manages to create a rubber ducky with HID programmed to enter keypresses, the screen would flicker, and shell scripts could be modified without the user being aware of it. Actual codepath would validate the firmware parts, not actual ramfs scripts, and boot into default boot option without TPM measurements being invalidated.

My point here is if there is no TPM released Disk Unlock key + Actual PR + USB Security dongle, hid functions could invalidate Heads security without the user knowing, even more for platform activating [USB early in init to support USB keyboards](https://github.com/osresearch/heads/blob/master/initrd/init#L67-L70).

alex-nitrokey:
I do not know enough about it yet, to have a solution at hand, but honestly I am thinking about looking at this topic for some time. I was thinking about looking at how [usbguard](https://usbguard.github.io/) is working and if we can include something like this. On a clean boot there should be no new device plugged into the heads machine other than those present during initialization as we do not want any fiddling by an malicious device anyway. To conclude imho a blocking of unknown devices is a good idea anyway.

szszszsz:
1. To add to https://github.com/osresearch/heads/pull/836#issuecomment-712713397, perhaps we could add device authentication via OpenPGP to avoid vid:pid:sn triple spoofing, then accept its input.
2. Another idea is to tunnel the HOTP requests to the USB Security dongle over CCID instead of HID.

tlaurion:
My point here is that a rubber ducky specially crafted to fit Librem Key/Nitrokey Pro 3D printed shells to have green led flashing to simulate functionning could break the protection offered by this PR if hid is not blocked. Otherwise, vid:pid:sn triple spoofing can make KO this security mechanism.

>     1. To add to [#836 (comment)](https://github.com/osresearch/heads/pull/836#issuecomment-712713397), perhaps we could add device authentication via OpenPGP to avoid vid:pid:sn triple spoofing, then accept its input.

Authentication is a good idea only if #771 is implemented and user has a backup of his private key he can reinject at will in a USB Securty dongle replacement if prior is lost, else he locks himself out of Heads. Not desirable.
> 
>     2. Another idea is to tunnel the HOTP requests to the USB Security dongle over CCID instead of HID.

I think this one is the better approach. While HID still needs to be somehow disabled inside of Heads (no input accepted from USB Security dongle) else sed is going to win. 

3. Another approach here would be to implement not only /boot measurements, but all shell scripts measurements prior of permitting boot. #862  is going that path setting the basis. We could definitely have an extended validation inside of Heads; reusing the concept of hashes, produced by the build system as of now as an example, and be able, at the moment of booting system default, to revalidate that ramfs scripts are still valid.

Thoughts? @MrChromebox @jans23 @szszszsz @alex-nitrokey @daringer @kylerankin:

MrChromebox:
> I think this one is the better approach. While HID still needs to be somehow disabled inside of Heads (no input accepted from USB Security dongle) else sed is going to win.

agreed. Problem is, how do we do that effectively? Using a blocklist for HID would mean newer security dongles would require a Heads update to be fully secure

tlaurion:
> > I think this one is the better approach. While HID still needs to be somehow disabled inside of Heads (no input accepted from USB Security dongle) else sed is going to win.
> 
> agreed. Problem is, how do we do that effectively? Using a blocklist for HID would mean newer security dongles would require a Heads update to be fully secure

Well... FWUPD firmware upgrades are around the corner, in all cases. :) And newer produced USB Security dongles should be provided after newer firmwares being deployed on hardware in all case. HID ID can be faked, btw. I have no solution for that problem, unfortunately, as of now. Disk Unlock Key released by the TPM is the true last line of protection here. I have no idea how to replicate such solution in a TPM-less platform, since USB device cannot be considered truely in the root of trust, being accessible too late to accomplish the same result. :|

tlaurion:
Any update here @szszszsz @jans23 @kylerankin ?

tlaurion:
@MrChromebox @szszszsz I also thought that HOTP secret was increased upstream so that there is no need of truncation under Heads?

szszszsz:
It is indeed @tlaurion: 
- current maximum HOTP secret length implemented in firmware for both Nitrokey Pro / Librem Key and Nitrokey Storage is 40 bytes;
- length in nitrokey-hotp-verification appears to be hardcoded to 20 bytes -  [nitrokey-hotp-verification/blob/master/operations.c#L55], which is currently maximum handled if not patched in Heads; this should be easily changeable.

cc @jans23 

[nitrokey-hotp-verification/blob/master/operations.c#L55]: https://github.com/Nitrokey/nitrokey-hotp-verification/blob/master/operations.c#L55

MrChromebox:
> current maximum HOTP secret length implemented in firmware for both Nitrokey Pro / Librem Key and Nitrokey Storage is 40 bytes

but that's not the case for all shipped Nitro/Librem keys, is it? I'd need to see what the earliest firmware is we shipped on the LK and if it supports 40-byte HOTP secrets

tlaurion:
@MrChromebox https://github.com/osresearch/heads/pull/493#issuecomment-487379379

MrChromebox:
@tlaurion yes so I just need to figure out if we shipped any LKs with firmware 0.7 or earlier

tlaurion:
@MrChromebox : or as stated https://github.com/osresearch/heads/pull/493#issuecomment-495834923
react to firmware version and trim only then.

tlaurion:
@szszszsz and on HID issue?

MrChromebox:
@tlaurion I'm not sure how we got back on the 20 vs 40-byte truncation debate, but looking at the current patch set, it truncates to 40 bytes, which means I've already confirmed that it's not an issue with the firmware on any shipped Librem Keys

tlaurion:
@MrChromebox I see no patch in heads tree against https://github.com/Nitrokey/nitrokey-hotp-verification/blob/master/operations.c#L55 as confirmed from @szszszsz [here](https://github.com/Nitrokey/nitrokey-hotp-verification/blob/master/operations.c#L55) nor PR upstream nor change of heads module to point to a version of hotp-verification that supports 40bytes?

MrChromebox:
@tlaurion you're right, I'm sorry, I thought you had tagged me in the comment above due to this patch truncating to 20 bytes, which it does not. The 40-byte secret generated by secret_from_rom_hash() is still truncated by nitrokey-hotp-verification when using the set command, but that's a separate (if related) issue from this PR

tlaurion:
@MrChromebox one doesn't work as intended without a fix in the other.

szszszsz:
https://github.com/osresearch/heads/pull/836#issuecomment-752805580 @tlaurion 
> @szszszsz and on HID issue?

I understand you mean here tunneling the current HID commands through CCID interface. No update yet. 
cc @jans23

daringer:
ok, so one of our main issues is HID, so I had a naive idea:
* we compile usb-hid as a module
* during heads boot we first use `gpg` to validate the security key using some asymmetric sign operation, thus we know the security key is "the right one"
* once this is done we load `hid` into the kernel
* HID devices will pop-up, if there is only one with the HID-ID we are looking for, things are fine, if some device is spoofing it we would need something to validate that the CCID device we were talking to previously is the same as the one we see popping up via HID (maybe heads could generate a random token, save it to the security key using the "user DOs", and realize something to read the same DO via HID, what would then proof that the HID connection is actually referring to the same device)

what do you guys think about that?

jans23:
Sorry for high jacking this discussion but at the same time all relevant people are subscribed already.

I understand this issue is about a solution to mitigate evil-maid attacks without using a TPM. In this case, which mechanism prevents an evil maid on-site to perform the following attack?

1. Read the flash
2. Compute the firmware's hash (which is used as the OTP secret)
3. Prepare an "evil" firmware and configure it with the legitimate OTP secret
4. Flash her own "evil" firmware

Subsequently the evil firmware should be able to calculate and return legitimate OTP codes and thus appearing not-tampered.

tlaurion:
@jans23 Absolutely.

> I understand this issue is about a solution to mitigate evil-maid attacks without using a TPM. In this case, which mechanism prevents an evil maid on-site to perform the following attack?
> 
>     1. Read the flash
> 
>     2. Compute the firmware's hash (which is used as the OTP secret)
> 
>     3. Prepare an "evil" firmware and configure it with the legitimate OTP secret
> 
>     4. Flash her own "evil" firmware
> 
> 
> Subsequently the evil firmware should be able to calculate and return legitimate OTP codes and thus appearing not-tampered.


The most simple attack, as you theorized here, would practically be:
1- read the rom and measure it with same algorithm (cut, trim)
1- [replace this code block with precedent obtained hash](https://github.com/osresearch/heads/pull/836/files#diff-0fd54c445ad40620392abb0e6defb6974df6798254e326594d6300cc86058e24R277-R280)

Done.

jans23:
Since you confirm this attack would be possible. What are the threats Measured Boot without TPM is protecting against? Why to use it instead of a non-Measured Boot firmware?

tlaurion:
@jans23 I am not in a position of being able to defend the current TPM-less implementation.

tlaurion:
For the current implementation, a silly idea that comes to my mind would be a flash backup inside of /boot, being signed and checked against, where again, more advanced attack could also fake the gpg validation if rom is compromised. By lack of real RoT, I have no idea how to really make it reliable.

tlaurion:
Other approach that was thought of before would be to have the output hashes.txt dumped under /boot, and encompassed by detached signing. It complicates the faking of result, while still possible under TPM-less attestation, as said numerous times previously: the measurements is happening way too late. 

Another silly idea here, where impractical, would be to provide to TPM-less owners a detached signed ISO to boot from to do the same validation in another RO environement (iomem=relaxed) to verify the hashes on a USB stick. But as you can see, we are reducing convenience for security, where this PR focuses on convenience while reducing security.

tlaurion:
@jans23 the problem of being able to internally backup and reflash a tampered ROM is linked to the absence of #361 and the absence of a #771 implementation.

Once those are merged, the actual TPM-less implementation would be less risky to [be so easily hackable](https://github.com/osresearch/heads/pull/836#issuecomment-757956802); or would at least require external reflash to be successful (bypassing authentication to access recovery shell).

MrChromebox:
> @MrChromebox one doesn't work as intended without a fix in the other.

true, but fixing the truncation on the hotp-verification side is out of scope for this PR

kylerankin:
As has already been mentioned, this are certainly attacks on this approach.

The purpose of this PR is not to reduce security in the name of
convenience, but instead to try to extend *some* level of protection that
Heads offers to hardware that physically does not have a TPM. The goal is
not to remove that protection from hardware that has a TPM. At Purism we
initially set up Heads on laptops with a TPM and those and all future
laptops included a TPM, but we created this patch initially so we could
*backport* Heads to older laptops that would otherwise support Heads but
didn't come with a discrete TPM. We also currently use it for our Mini
hardware until such point that it includes a TPM.

Without such a patch, that hardware would not be able to perform a check
for firmware tampering at all. With the patch, there is still a risk from a
targeted Evil Maid attack, but the user gets the other protections Heads
offers for kernel/initrd changes, as well as possibly detecting firmware
changes from a less-sophisticated attacker.

Again, the ideal is to use hardware with a TPM. You can think of this PR
like others that aim to extend Heads support of some kind to more hardware.
There are of course approaches to further harden the non-TPM use case (as
well as the TPM use case), such as locking the recovery console that should
certainly be looked into, but in my opinion those are separate issues from
this patch.

MrChromebox:
@tlaurion I've pushed a PR against nitrokey-hotp-verification here:
https://github.com/Nitrokey/nitrokey-hotp-verification/pull/20

I can push a patch as PR here as well (which I've already tested locally on a Librem 13v4 w/LK), or we can wait for above PR to be merged and then bump the module version.

tlaurion:
@MrChromebox https://github.com/Nitrokey/nitrokey-hotp-verification/commit/b44abf1a6a33185b428322492b0bfa0b83cd7f86 was merged.

MrChromebox:
> @MrChromebox [Nitrokey/nitrokey-hotp-verification@b44abf1](https://github.com/Nitrokey/nitrokey-hotp-verification/commit/b44abf1a6a33185b428322492b0bfa0b83cd7f86) was merged.

PR #962 opened :)

MrChromebox:
@tlaurion so now that the secret length is handled, back to core of this PR: as Kyle pointed out, we're not reducing the security of anything, but adding some level of support/security to boards which don't have a TPM

tlaurion:
I can merge, while being against it.

- [ ] [Simple attack defeating completely the purpose of this PR, promoting false sense of security](https://github.com/osresearch/heads/pull/836#issuecomment-758197835
)
- [ ] [Mitigations to current risks of usage not yet implemented, leaving a false sense of security since tampered rom can easily be flashed from recovery and never upgraded again](https://github.com/osresearch/heads/pull/836#issuecomment-758225148)

I will leave merging this to a vote. 

If majority of the implicated people here thumbs up this comment, I merge. Otherwise it stalls until proper fixes are implemented, I also include people that are asking for that feature for platforms without TPM, but don't seem to understand the current risks linked to current implementation and believe that it would be equivalent to a TPM platform.

@MrChromebox @daringer @jans23 @kylerankin @szszszsz @osresearch @fhvyhjriur @Tonux599 @Thrilleratplay @irelativism @blobless

MrChromebox:
@tlaurion I don't see how access to the recovery shell is relevant. You can flash a tampered ROM from the GUI all the same

tlaurion:
> @tlaurion I don't see how access to the recovery shell is relevant. You can flash a tampered ROM from the GUI all the same

True that. Then that codepath should implement the same authentication mechanism.

daringer:
As also stated by before, I would love to have this *version* of heads on any machine compared to a closed firmware, think here we're all together, right? So, I also would suggest to leave this decision to the user. 

In particular I would vote **yes**, if we would extend this PR with a more prominent way to communicate the expected security to the user. Some simple ideas:
* clearly tell the user during boot that this machine is not runnning a TPM-supported Heads (as part of `CONFIG_BOARD_NAME` ?), something like (auto-)appending "TPM-less". An alternative would be directly including this information inside `initrd/bin/gui-init`  -> `MAIN_MENU_TITLE`
* Document this TPM-less Heads variant (inside the wiki? or somewhere else) and its drawbacks compared to the TPM-supported Heads.
* I am not the most creative guy, but how about a name extension for "Heads" for this specific variant "Heads-lite", "Heads-slim"

I know this may sound trivial, but I totally second the point that a TPM-less Heads installation should not leave the impression of being as secure as a TPM-supported Heads. From my point of view it is very important to be as transparent to the user as possible. So best-case we would enable the user to judge all alone, whether this is enough security for his or her application/use-case.

tlaurion:
> From my point of view it is very important to be as transparent to the user as possible.

@daringer: We then take into consideration that the user is able to understand and make the right choice based on implementation differences. If Purism states somewhere on their website how easy it is currently to bypass the security implementation of this PR, then I also have no problem. I am currently waiting for a decision here to work on t400 x200 and other GM45 based boards which have no TPM. As stated before, reflashing through heads is really easy. Reflashing a tampered ROM, without authentication, would take a prepared Evil maid less then a minute... While the user will still think he's protected because he runs open source Head's/Pureboot TM firmware. This is reputation based. There is not a lot of people that will go down to the PR discussions and even the wikis. This is why I have asked people to vote. I'm against, but this is no monarchy.

daringer:
> The flaws, both theoretical(HID keyboard Rubber Ducky going in recovery, sedding bash scripts in a blink of an eye and going back to gui) and now the practically proven flaws we now all know would be easy to deploy, is all known prior of merging. 

No discussion from my side, strong and true arguments. On the other side, as you also stated this decision is not exclusively a technical decision. So, to not exaggerate this too much, I will vote for yes and hereby briefly outline why:
* **TPM2** (or in general future hardware) will introduce other drawbacks (e.g., ME cannot be disabled, or is even needed...) so Heads will sooner or later have to cope with different levels of security 
* **Added security** is imho better than no-added security (as long as the level of security is transparent)
* **Keeping development resources and knowledge together inside Heads**, i.e., non-tpm improvements will increase security for both: tpm-driven and non-tpm Heads installations (as otherwise likely (worst-case even closed) forks will partly split resources)

Therefore my likely not too heavy vote is :+1:

lrvick:
Having read the entire thread and caught up, my (potentially wrong) TLDR of both sides is:

1. A TPM-Free solution lacks any defense against evil maid attacks all historically supported divices which include a TPM which theatens to make uses assume mistake equivilant UX of different devices for equivilant security, which is bad and must be avoided.
2. If a TPM-free solution can offer damage control to someone rooted by a remote adversary that stock firmware can not offer, then it has merit.

On point 1:

I think boot integrity solutions need to support more than binary pass/fail states to a user if not all potential configurations are equally secure. 

Android for example solves the UX issues of different levels of boot trust with strict color codes:

- YELLOW: Warning screen for LOCKED devices with custom root of trust set
- ORANGE: Warning screen for UNLOCKED devices
- RED (eio): Warning screen for dm-verity corruption

A user used to seeing green that now sees any other color would hopefully pause to read.

I will also note that older Android devices with weaker verified boot inforcements still get GREEN if all checks possible on the given hardware are satisfied. This is reasonable as the boot level responsible is read only and not user/attacker updatable and if -anything- in the boot path is updated, an OS wipe is forced as the TEE will no longer cooperate with the FDE handshake.

If heads could similarly inform a user accurately with clear UX of various security modes, alternate modes are OK IMO.

E.g.
"Your boot stack is reasonably well defended against tampering by remote or local adversaries" in TPM operation.
"Your boot stack is reasonably well defended against tampering by remote adversaries" in TPM-free operation.

Heads in TPM-free mode however can't really promise even the latter statement though.

As I undersand current implementations, the following two attacks are potentially viable when there is no TPM in play:

1. A remote adversary could flash malicious Heads firmware that always shows verified boot was successful. Worse, it could DFU the nitrokey on next enumeration and flash a small/fast firmware that does nothing but blink green.
2. A remote adversary could flash a firmware that blinks green -and- behaves as a usb rubber ducky on next enumeration with no hardware modifications.

Unless there are strong defenses to both of theses attacks in tpm-less operations I feel Heads offers little to no value there.

Honestly these attacks could be bad even when a TPM -is- in play if the firmware does not have a physical write protect and should be addressed either way, perhaps elsewhere. ( I have some ideas).

I am in favor of Heads supporting tpm-less devices only if paired with strong and immutable UX showing what classes of attack they are defended against beyond what stock BIOs would, but It it is not clear to me if this is really the case ATM.

Someone steer me straight if I am way off base here!

jans23:
To me it's important that users are informed well. Such information during configuration of Heads would be good. Informing users who buy a system would be seller's responsibility.

I vote for "yes". 

In order to understand the whole system's security: [This article](https://www.golem.de/news/librem-mini-v2-im-test-der-kleine-graue-mit-dem-freien-bios-2011-152017-3.html) (in German) states that Librem Mini's flash chip could only be written at the very first boot after flashing. Is this true and is it a property of the chip itself or is another aspect of the system guaranteeing it?

MrChromebox:
>  This article (in German) states that Librem Mini's flash chip could only be written at the very first boot after flashing. Is this true and is it a property of the chip itself or is another aspect of the system guaranteeing it?

no idea where they got that idea from, but it would make updates nearly impossible (obviously) and I can't think of how someone would even implement such a restriction

tlaurion:
I was pointed out to Purism's [article](https://puri.sm/posts/coreboot-news-new-script-pre-built-binaries-and-pureboot-on-non-tpm-laptops/) by a technical journalist who was inquiring about the validity of the claim:

`[...]you can now use a Librem Key in place of your TPM chip and get similar protection against tampering!`

Will that blob post have an erratum?

initrd/etc/functions
```diff
@@ -267,6 +267,19 @@ combine_configs() {
 	cat /etc/config* > /tmp/config
 }
 
+# Generate secret value using first 40 chars of ROM SHA256 hash
+secret_from_rom_hash() {
+	local ROM_IMAGE="/tmp/coreboot-notpm.rom"
+
+	echo -e "\nTPM not detected; measuring ROM directly\n" 1>&2
+	# use a previously-copied image if it exists
```
tlaurion, line 275: @MrChromebox @daringer @jans23 @kylerankin @szszszsz @osresearch @fhvyhjriur @Tonux599 @Thrilleratplay @irelativism @blobless @lrvick

1. Replace this by hardcoded value obtained from backuped flash. (apply same recipe to obtain hash)
2. Do whatever you want inside of ROM secured by Nitrokey Pro/Nitrokey Storage/Librem Key for which "[you can now use a Librem Key in place of your TPM chip and get similar protection against tampering!"](https://web.archive.org/web/20201108000216/https://puri.sm/posts/coreboot-news-new-script-pre-built-binaries-and-pureboot-on-non-tpm-laptops/)
3- Succeed HOTP challenge at every boot until next firmware upgrade.

We merge?

initrd/etc/functions
```diff
@@ -267,6 +267,19 @@ combine_configs() {
...
+	if [ -f ${ROM_IMAGE} ]; then
+		sha256sum ${ROM_IMAGE} | cut -f1 -d ' ' | cut -c 1-40 | tr -d '\n'
+	else
+		flash.sh -s ${ROM_IMAGE} | cut -c 1-40 | tr -d '\n'
+	fi
+}
```
tlaurion, line 281: @MrChromebox @daringer @jans23 @kylerankin @szszszsz @osresearch @fhvyhjriur @Tonux599 @Thrilleratplay @irelativism @blobless @lrvick

1. Boot hardware secured by Librem Key/Nitrokey Pro/Nitrokey Storage on TPM-less supported devices. User owning those devices trusted marketing speech and left said hardware unattended because remotely attested with the USB Security dongle they key in their pockets. Press 'r' frenetically until reaching recovery shell. Plug formatted usb thumbdrive. Keep calm.
2. Run the following from shell prompt/a script: `mount-usb rw && flash.sh -r /media/backuped.rom && umount /media && poweroff`. Wonder around, get back your USB dongle when screen goes off. Leave the scene with USB thumbdrive.
3. Craft your own evil maid aimed rom at leasure. Tamper also with /boot content! If there is enough space in SPI flash of the hardware in question, add as much stuff as you need! make scripts to copy them under /boot if not existing, modify Heads scripts to your liking so that signed digest verification lies... If you want, you can craft an evil xen, kernel, initrd and grub config if you'd like! Be creative.
4. Obtain secret from rom at leasure, used as remote attestation secret in the HOTP challenge with user's Nitrokey Pro/Nitrokey Storage/Librem Key, by running the following on obtained backuped.rom: `sha256sum backuped.rom | cut -f1 -d ' ' | cut -c 1-40 | tr -d '\n'`
5. Replace the content of `secret_from_rom_hash` function commented here with the value obtained from precedent step, simply replacing `secret` in between the following quotes: `echo "secret"`
6. Build your crafted rom. Put it back on usb thumbdrive as file `new.rom`
7. Redo step 1 on unattended hardware because trusted by user that the rom is protected with his USB Security dongle kept securely with him.
8. Run the following from shell/script: `mount-usb rw && flash.sh /media/new.rom && umount /media && poweroff`. This will take more time then step 2, since writing takes more time. Keep looking discretely at the screen for it to go dark. Get back your thumbdrive and leave for good.
9- Be confident that on every next boot, HOTP challenge will succeed over Nitrokey Pro/Nitrokey Storage/Librem Key... until the user decides to upgrade firmware on said trusted hardware. Repeat at will on that hardware. Code was merged upstream under Heads. Keep in mind that all TPM-less hardware out there now supported by Heads can be hacked easily the same way by users bragging about their security dongles and the hardware you know doesn't have a TPM. Keep track of supported devices upstream, own them and test locally for guaranteed success!

We merge?

jans23:
Until recently I assumed the [stated flash chip protectio](https://github.com/osresearch/heads/pull/836#issuecomment-770614660)n would be in place. Now I doubt this is true. In the case there is no flash chip protection I vote with "no".

initrd/etc/functions
daringer: hilarious (although realistic) story @tlaurion , my favorite: 
>  Play dumb. Be cute. Be prepared, plan accordingly

but seriously: seeing which misconceptions this (not even upstream) patch has already produced,
I think I also re-cast my vote to "no". I still see a gap between Heads and coreboot+seabios in terms of security (no password, no hotp, not even the chance to check the firmware, no /boot signing), but have to admit that the whole: Heads-implies-another-level-of-security argument, is stronger than I expected.

rugk:
Backlink German article on the security of TPM-less Heads and Purism's Librem Mini including this discussion:
https://www.golem.de/news/purism-sicherheit-zweiter-klasse-im-librem-mini-2103-155007.html

tlaurion:
Any followup on recommendations https://github.com/osresearch/heads/pull/836#issuecomment-761893255 so that changes applied from Purism fork can be merged upstream here? 

I think we all agree  that TPM + HOTP is better then TPM-less (with simple oneliner hack). The question now is how to make the user aware in total transparency of the differences of the approaches and where.

tlaurion:
@kylerankin @MrChromebox update on last comment?

MrChromebox:
@tlaurion see my comments in #326. SMM locking via the payload isn't viable in Skylake and newer platforms

tlaurion:
@kylerankin @MrChromebox coule you write something on the difference between tpm enforced attestation vs USB security dongle without TPM support on heads-wiki so we can finally merge this in?

kylerankin:
@tlaurion Which section of the wiki did you want me to submit a PR? I don't
have a strong opinion myself but also I don't know where this fits best.

tlaurion:
@kylerankin threat modeling section would be great?

tlaurion:
> 1. To add to [Use a USB security key as a TPM work-alike in the absence of a physical TPM #836 (comment)](https://github.com/osresearch/heads/pull/836#issuecomment-712713397), perhaps we could add device authentication via OpenPGP to avoid vid:pid:sn triple spoofing, then accept its input.
> 
>     2. Another idea is to tunnel the HOTP requests to the USB Security dongle over CCID instead of HID.

@szszszsz : https://github.com/osresearch/heads/pull/1145 removes USB_HID from kernel, and it doesn'tt impact HOTP/GPG functions. Am I missing something?

tlaurion:
@MrChromebox (not relevant to this ticket but still important since it was talked here in the past):
> @tlaurion see my comments in #326. SMM locking via the payload isn't viable in Skylake and newer platforms
This platform is actually [enforcing platform locking](https://github.com/hardenedvault/vaultboot/commit/4506fc26d0d9aed1e15030bbf8cd49cede6c4d49#diff-fd87943b0e476011d16937056fdc249138c6a68ab09c081d9171259f6c60ed54) through patches (VaultBoot forks of Heads)

irelativism:
you should look into something called TBM (TBM!=TPM), trusted boot module a project by whithebox systems and also supported by nlnet

irelativism:
https://nlnet.nl/project/trustedbootmodule/
https://src.whiteboxsystems.nl/TBM
https://media.ccc.de/v/SHA2017-280-tbm_trusted_boot_module

tlaurion:
@irelativism Thanks for the pointer. The challenge here, as with any additional external components doing measurements, is to have that device connected at a low level enough so that, in Heads case, coreboot can initialize such device soon enough (boot bock preferred) to be able to have such component measure itself and next stages of the firmware.

On current existing devices, such components require simple communication protocols (LPC/I2C) to be able to talk with components existing on such buses. I checked the links you pointed out. This is linked to another discussion that I reveisited in the past months by [mikebdp2](https://github.com/mikebdp2) under https://github.com/osresearch/heads/issues/717 which back in time, missed his point and edited original reply under https://github.com/osresearch/heads/issues/717#issuecomment-628016131. It is true that a serial connection under early USB init could be accomplished under coreboot, and such USB device could be used to act as a USB alternative to TPM.

I welcome further discussions on the matter to happen under https://github.com/osresearch/heads/issues/717, not here.

tlaurion:
@JonathonHall-Purism : I think this is time to revisit this PR.

A lot of things changed since this PR was first opened under https://github.com/osresearch/heads/pull/493
- USB HID was a blocker. It is not exposed anymore from Nitrokey/Librem Key firmware somewhere after v.0.10?
- Heads now requires the board configurations to specify if a usb keyboard is desired/required, and adds USB HID support only if requested per board config (to mitigate rubber ducky attacks)
- HOTP secret size was increased in both firmware and in software

@kylerankin I requested in past comments a threat modeling note under heads wiki from Purism [here](https://github.com/osresearch/heads/pull/836#issuecomment-991469434)

----
The reason I'm poking again here is that 
- Talos II coreboot port is coming along https://github.com/osresearch/heads/pull/1002
- KGPE-D16 coreboot upstreaming effort is continuing

As opposed to laptops that simply come/doesn't come with a TPM, those motherboards (as for Purism's and Nitrokey's mini (running on UEFI and tianocore as of now @jans23) would still require something along the lines of dynamic codepaths for TPM/no-tpm, as currently implemented in this PR at https://github.com/osresearch/heads/pull/836/files#diff-63d474ce25ab525343c0d061b286b1cd45199cd14fdf9ea535f396fd8be5de45R47. Using Heads (today) without a TPM is not so relevant, while having HOTP is better then nothing, even though nothing can prove thatthe firmwre wasn't tampered with to produce and feed the right measurement to HOTP alone without a TPM. But we would still need to be able to have boards that support TPM to be able to dynamically set `CONFIG_TPM=n` if no TPM is detected, instead of supporting multiple unique board configurations named `-tpm/-no-tpm`.

@jans23 @SergiiDmytruk @MrChromebox @JonathonHall-Purism @szszszsz  : https://github.com/osresearch/heads/pull/836/files#diff-63d474ce25ab525343c0d061b286b1cd45199cd14fdf9ea535f396fd8be5de45R47 could be reused seperately, of course,  but since the current states of Nitrokey Pro/Nitrokey Storage/Librem Key firmware and Heads have evolved to mitigate criticisms of this PR over time (some risks still existing for boards having `CONFIG_HOTPKEY=y`  and `export CONFIG_USB_KEYBOARD` without a TPM detecting recovery shell access) I think it is time to rediscuss this PR in the goal of merging it, with proper thread modeling notes in board configurations themselves and in the threat model section of the wiki.


@SergiiDmytruk : Meanwhile, maybe a seperate PR might be needed to simply have TPM detected and having CONFIG_TPM deactivated at runtime as implemented under https://github.com/osresearch/heads/pull/836/files#diff-63d474ce25ab525343c0d061b286b1cd45199cd14fdf9ea535f396fd8be5de45R47 so that Talos II can have a single workstation board config and server board config, the same applying to kgpe-d16 boards reupstreaming efforts. Note that kgpe-d16 is having ps2 for keyboard/mouse, while Talos II is requiring USB for input in workstation mode, so that risks exposed above would also apply if another workstation board configuration is brought online to support add support to `CONFIG_HOTP`, since export CONFIG_USB_KEYBOARD is required. But again, the Talos II with/without TPM (and without `CONFIG_HOTP`) is a different story here: the point of mitigating rubber ducky attacks while still using USB Security dongle to validate firmware integrity being at stake here. And having a TPM+HOTP not being the same as having HOTP alone without a TPM (which we all agreed in previous discussions here that those were not equivalent outside of marketing speech).

initrd/init
tlaurion: Point here is what is in cbfs is measured, but not the result of combine_configs.

tlaurion:
Merged as part of #1419.

Requires to not have a TPM and HOTP.
Interesting for #453 and #925 #925 #934

harlyboy0178 (approved):

PR #842, opened 2020-09-25
Add 'CPUS' as make BOARD=X argument

tlaurion:
This is a nifty way permitting us to limit ressource allocation if desired (CIs).

From experiementations, there is no way to limit ressource consumption on CircleCI but to pay to have more resources.
In our use case, with cache now being built when boards builds successfully for the same modules/* and patches/* hashes, a cache is built. It fallsback to taking the last available built cache for musl-cross-make having been built for its own specific module signature and existing patches.

The problem we had recently in CI was linked to the addition of coreboot 4.12 board (librem_mini-NoTPM) which, for some obscure reason, failed building with OOM being invoked. GCC expects each thread to have 1GB available, and CircleCI's guaranteed resource allocation is of 4Gb ram, with 32 softcores. So Heads magically taking nproc made explode the coreboot build toolstack for the first time.

This PR introduces the possibility of passing CPUS=X on command line, where that variable is then passed to all submodules, where nproc is taken if not defined. This seems to resolve the current failed PR builds and already passed previous failing point https://app.circleci.com/pipelines/github/tlaurion/heads/447/workflows/d1138321-3a1d-4ffb-9165-e8803d9648ac/jobs/480, where the successfull build will produce a new cache that will be picked up with following builds not touching modules/* nor patches/*

@MrChromebox

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -60,6 +60,24 @@ jobs:
 #          path: build/qemu-linuxboot/hashes.txt
 
       - run:
+          name: librem_mini-NoTPM
+          command: |
+            rm -rf build/librem_mini-NoTPM/* build/log/* && make CPUS=4 \
+                V=1 \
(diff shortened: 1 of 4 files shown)


MrChromebox (approved):
if it fixes CI building under coreboot 4.12 then LGTM

PR #845, opened 2020-09-26
modules/coreboot: Only do coreboot version check if we're building boot (fixes #841)

### LINKED ISSUE #841: linuxboot based boards not building anymore

NHellFire:
Tested builds of qemu-linuxboot and qemu-coreboot, both succeeded.


tlaurion:
@NHellFire Ci fails because #842 not being merged. This will need to be rebased ton top of master once merged in.

MrChromebox (commented):
wait, why is coreboot being included as a module for a Linuxboot build anyway? Seems like that's the issue that needs to be fixed

NHellFire:
> wait, why is coreboot being included as a module for a Linuxboot build anyway? Seems like that's the issue that needs to be fixed

Top Makefile includes all modules (00559de, 1c64e4c). It was fine before the version check was added as it only contained variables.

tlaurion:
@MrChromebox I like the fix. Rebasing on master with linixbootboarf added in CI should show the fix working.

tlaurion:
@NHellFire can you add the board in CircleCI and rebase on master?

NHellFire:
> @NHellFire can you add the board in CircleCI and rebase on master?

I've rebased (and fixed showing the error).
Currently, readding linuxboot back to CI will just fail as Debian 10 uses gcc 8.3 (linuxboot/edk2#1).

tlaurion:
@NHellFire So if I understand well, this will work once linuxboot has your PR merged upstream, while linuxboot will not offer reproducibility since not bound to particular commit ID as other modules. Actually, it seems that [linuxboot is the only module still not being bound to a commit id](https://github.com/osresearch/heads/blob/master/modules/linuxboot#L3-L5)

NHellFire:
CI should pass now. Local build passed with gcc 9.3.0 and binutils 2.34 (Ubuntu 20.04.1).

tlaurion:
@NHellFire not sure about the hack implemented in the linuxboot module to apply patches outside of the root Makefile logic.
Why are those patch not directly applied into linuxboot upstream, or for a specific linuxboot git commit as other modules?

if we look at [musl-cross-make module](https://github.com/osresearch/heads/blob/master/modules/musl-cross#L7-L11), being dependend on a specific commit for reproducible builds in the future on a specific Heads commit being rebuild, the following module specifies what to download as opposed to downloading master which cannot be replicated in the future if master changes.

As a result for that specific musl-cross-make git version, no patch are currently applied (no patch exists for that specific module-hash), since no patches relative to that specific commit id exits (Duh. there is a past artifact for musl-cross lying around, will fix that).

Does that make sense?

In the present case, linuxboot should be fixed to commit id https://github.com/linuxboot/linuxboot/commit/10be0dc2652e1fc8c11d419ca2c4ff93920d4165 into module, and your linuxboot-ed2k patches should be applyable directly inside of the decompressed archive for that specific commit ID, where the patch would be in a file patches/linuxboot-10be0dc2652e1fc8c11d419ca2c4ff93920d4165.patch or in seperate files under patches/linuxboot-https://github.com/linuxboot-10be0dc2652e1fc8c11d419ca2c4ff93920d4165/1.patch etc following root Makefile patch application rule.

tlaurion:
@NHellFire ?

NHellFire:
I'll see if I can get to this in the next few days. I've been away from
home and a bit busy.

tlaurion:
@NHellFire please tag me for followups

NHellFire:
@tlaurion Can you check the latest revision?
Thanks

boards/winterfell/winterfell.config
```diff
@@ -52,7 +52,7 @@ dxe_offset := 860000
 dxe_size := 6a0000
 flash-dxe: $(build)/$(BOARD)/linuxboot.rom
 	( echo u$(dxe_offset) $(dxe_size) ; \
-	pv $(build)/linuxboot-git/build/$(BOARD)/dxe.vol \
+	pv $(build)/linuxboot-b5376a441e8e85cbf722e943bb8294958e87c784/build/$(BOARD)/dxe.vol \
```
tlaurion, line 55: Ok, I get what you are doing here.
You can get the build directory by reusing Makefile included module.

In this case: `$(linuxboot_base_dir)`

.circleci/config.yml
```diff
@@ -34,30 +34,26 @@ jobs:
...
-#      - store-artifacts:
-#          path: build/qemu-linuxboot/linuxboot.rom
-#      - store-artifacts:
-#          path: build/qemu-linuxboot/hashes.txt
+      - run:
+          name: qemu-linuxboot-edk2
+          command: |
+            make \
+                BOARD=qemu-linuxboot \
+                `/bin/pwd`/build/linuxboot-b5376a441e8e85cbf722e943bb8294958e87c784/build/qemu/.configured \
```
tlaurion, line 42: That will break in future linuxboot version bump. Why static?
`$(linuxboot_base_dir)/build/qemu/.configured` ?

modules/coreboot
```diff
@@ -1,5 +1,6 @@
 modules-$(CONFIG_COREBOOT) += coreboot
 
+ifeq "$(CONFIG_COREBOOT)" "y"
 ifeq "$(CONFIG_COREBOOT_VERSION)" "4.8.1"
```
tlaurion, line 4: Indentation should shift

modules/linuxboot
```diff
@@ -77,3 +83,25 @@ linuxboot.run: $(build)/$(BOARD)/linuxboot.rom
...
+
+linuxboot_depends += linuxboot-edk2
+modules-$(CONFIG_LINUXBOOT) += linuxboot-edk2
+
+linuxboot-edk2_version := b7c299e4799012e8a20958a68693ab95b7626aa9
+linuxboot-edk2_hash := 78f48f46414b019b5fcd2d7049fd79e955d25f2965123d0bbdf7226564db33e5
+
+linuxboot-edk2_tar := linuxboot-edk2-$(linuxboot-edk2_version).tar.gz
+linuxboot-edk2_url := https://github.com/linuxboot/edk2/archive/$(linuxboot-edk2_version).tar.gz
+linuxboot-edk2_dir := $(linuxboot_base_dir)/edk2
```
tlaurion, line 99: so that is the env variable to reuse `$linuxboot-edk2_dir`

tlaurion (changes requested):

tlaurion:
@NHellFire I see [that requests for change](https://github.com/osresearch/heads/pull/845#pullrequestreview-513159754) in this PR are still valid?

modules/linuxboot
```diff
@@ -21,7 +27,7 @@ linuxboot_configure := \
 	if [ "$(linuxboot_board)" = "qemu" ]; then \
 		echo >&2 "Pre-building edk2 OVMF" ; \
 		( cd $(build)/$(linuxboot_base_dir)/edk2/OvmfPkg ; \
-		./build.sh -n $$CPUS \
+		./build.sh -n $(CPUS) \
```
synackd, line 30: @tlaurion What do you say to making this change a separate PR and merging it in? I was about to make one fixing this before seeing this change in this PR.

tlaurion: @synackd please do. Same should be made under coreboot module

synackd: In hindsight, I probably could have combined #940 and #941 into one PR, though it does categorically separate the linuxboot and coreboot modules...

tlaurion: Have you tested it? The idea here is to pass CPUS to command line
make BOARD= x230 CPUS=2

There is no such invocation as CPUS as a command. 
@synackd?

tlaurion:
@NHellFire @synackd ping into cleaning this PR?

tlaurion:
> @NHellFire not sure about the hack implemented in the linuxboot module to apply patches outside of the root Makefile logic.
> Why are those patch not directly applied into linuxboot upstream, or for a specific linuxboot git commit as other modules?

@NHellFire patches are under patches/module-version and applied automatically, if modules respect module statements. 
I cannot merge modules pinning to specific version. See requested changes above.

modules/linuxboot
synackd: These changes can be retracted since they are already merged as of #940.

tlaurion:
Linked with https://github.com/osresearch/heads/issues/971#issuecomment-773353107

PR #849, opened 2020-10-02
musl-cross: remove old patch artifact of the musl-cross era

tlaurion:
Remove unused patch.
musl-cross-make runs without heads patches.

@MrChromebox


MrChromebox:
looks like you need to kill the patch application code for the module since it doesn't handle the case of no patches

MrChromebox (approved):

PR #795, opened 2020-08-07
(WIP) KGPE-D16 migration to coreboot 4.11 (community effort)

tlaurion:
This is based on #472 and will be very verbose in the commits.
Similar work will be required for any board leaving coreboot 4.8.1 to go 4.11+
Similar work will be required to upgrade kernel to newer version so that each board can be tested prior of merging changes, permitting to keep coreboot+linux modules while we migrate.

Comments and partitipation welcome in the goal of fixing #740

PR #867, opened 2020-10-23
(WIP) KGPE-D16 Coreboot 4.11 + Measured Boot

Tonux599:
This is based upon #795 (credit to @tlaurion) but focuses on measured boot without vboot.

I would appreciate feedback from the community, notably that `./patches/coreboot-4.11/0020-kgpe-d16_measured-boot-support.patch` is implemented correctly.

Also note that this error occurs when microcode is included in the CBFS:
```
TPM: pcr 2 measure ff7b9e88 @ 12684: 1077e1af95dc35d5661d54b0ba72652fa0d5aeec
src/drivers/pc80/tpm/tis.c:521 unexpected TPM status 0xff
src/drivers/pc80/tpm/tis.c:717 failed sending data to TPM
TPM: command 0x14 send/receive failed: 0x10000001
TPM: pcr 2 measure ff7bd088 @ 7876: 983867f34d8599e779cf9e1454f8f4518b5628ac
src/drivers/pc80/tpm/tis.c:521 unexpected TPM status 0xff
src/drivers/pc80/tpm/tis.c:717 failed sending data to TPM
TPM: command 0x14 send/receive failed: 0x10000001
TPM: pcr 2 measure ff7b9e88 @ 12684: 1077e1af95dc35d5661d54b0ba72652fa0d5aeec
src/drivers/pc80/tpm/tis.c:521 unexpected TPM status 0xff
src/drivers/pc80/tpm/tis.c:717 failed sending data to TPM
TPM: command 0x14 send/receive failed: 0x10000001
TPM: pcr 2 measure ff7bd088 @ 7876: 983867f34d8599e779cf9e1454f8f4518b5628ac
src/drivers/pc80/tpm/tis.c:521 unexpected TPM status 0xff
src/drivers/pc80/tpm/tis.c:717 failed sending data to TPM
TPM: command 0x14 send/receive failed: 0x10000001
TPM: pcr 2 measure ff7b9e88 @ 12684: 1077e1af95dc35d5661d54b0ba72652fa0d5aeec
src/drivers/pc80/tpm/tis.c:521 unexpected TPM status 0xff
src/drivers/pc80/tpm/tis.c:717 failed sending data to TPM
TPM: command 0x14 send/receive failed: 0x10000001
TPM: pcr 2 measure ff7bd088 @ 7876: 983867f34d8599e779cf9e1454f8f4518b5628ac
src/drivers/pc80/tpm/tis.c:521 unexpected TPM status 0xff
src/drivers/pc80/tpm/tis.c:717 failed sending data to TPM
TPM: command 0x14 send/receive failed: 0x10000001
Found TPM SLB9635 TT 1.2 by Infineon
TPM: Startup
TPM: command 0x99 returned 0x0
TPM: Asserting physical presence
TPM: command 0x4000000a returned 0x0
TPM: command 0x65 returned 0x0
TPM: flags disable=0, deactivated=0, nvlocked=1
TPM: setup succeeded
TPM: pcr 2 measure fffff800 @ 2048: 80909df13fd6a372cca9837b538971e375d3f4d4
TPM: command 0x14 returned 0x0
TPM: pcr 2 measure ff000300 @ 1048576: dcfbe783c5d139c5481253b8c8ee8909b49297ca
TPM: command 0x14 returned 0x0
TPM: pcr 2 measure 00e00000 @ 213776: 6f279a6cc1687a5b3170d80cca7a3147e59037ca
TPM: command 0x14 returned 0x0
TPM: pcr 2 measure 00e00000 @ 1162424: 2ea6887b5d072fb004b2115a320621b8f32f2ac4
TPM: command 0x14 returned 0x0
Found TPM SLB9635 TT 1.2 by Infineon
TPM: Startup
TPM: command 0x99 returned 0x26
TPM: Can't run startup command.
TPM: pcr 2 measure ff042988 @ 9895: 75fe16ebfce00b4becdc5b7f6f31feec791b87c6
TPM: command 0x14 returned 0x0
\_SB.PCI0.LPC.TPM: LPC TPM PNP: 004e.0
TPM: pcr 2 measure ff041b78 @ 3524: a1615718de98d6d619812c6b9fd3c3f5177bc137
TPM: command 0x14 returned 0x0
TPM: pcr 2 measure 00090000 @ 4224: d74d67b4e9bc8dc6d3536082e78dcae391fa1503
TPM: command 0x14 returned 0x0
TPM: pcr 2 measure 01000000 @ 3772880: 35e974ae974845687c8128055661fe4005e5b970
TPM: command 0x14 returned 0x0
TPM: pcr 2 measure 00040000 @ 180: 31fae3ad7f3a0f2ee6fa1afeaa370ba152dc2662
TPM: command 0x14 returned 0x0
TPM: pcr 2 measure 00091000 @ 25: 1a0c772ebbab34217f5ee534c8e01c94fd8d3bd6
TPM: command 0x14 returned 0x0
TPM: pcr 2 measure 04000000 @ 4041216: 5cef02f227841f10622c4ca0adbb14843d6c4b97
TPM: command 0x14 returned 0x0
```
However when microcode is removed such as:
```
cbfstool coreboot.rom remove -n microcode_amd.bin
cbfstool coreboot.rom remove -n microcode_amd_fam15h.bin
```
The above error does not occur:
```
Found TPM SLB9635 TT 1.2 by Infineon
TPM: Startup
TPM: command 0x99 returned 0x0
TPM: Asserting physical presence
TPM: command 0x4000000a returned 0x0
TPM: command 0x65 returned 0x0
TPM: flags disable=0, deactivated=0, nvlocked=1
TPM: setup succeeded
TPM: pcr 2 measure fffff800 @ 2048: 80909df13fd6a372cca9837b538971e375d3f4d4
TPM: command 0x14 returned 0x0
TPM: pcr 2 measure ff000300 @ 1048576: dcfbe783c5d139c5481253b8c8ee8909b49297ca
TPM: command 0x14 returned 0x0
TPM: pcr 2 measure 00e00000 @ 213776: 6f279a6cc1687a5b3170d80cca7a3147e59037ca
TPM: command 0x14 returned 0x0
TPM: pcr 2 measure 00e00000 @ 1162424: 2ea6887b5d072fb004b2115a320621b8f32f2ac4
TPM: command 0x14 returned 0x0
Found TPM SLB9635 TT 1.2 by Infineon
TPM: Startup
TPM: command 0x99 returned 0x26
TPM: Can't run startup command.
TPM: pcr 2 measure ff042988 @ 9895: 75fe16ebfce00b4becdc5b7f6f31feec791b87c6
TPM: command 0x14 returned 0x0
\_SB.PCI0.LPC.TPM: LPC TPM PNP: 004e.0
TPM: pcr 2 measure ff041b78 @ 3524: a1615718de98d6d619812c6b9fd3c3f5177bc137
TPM: command 0x14 returned 0x0
TPM: pcr 2 measure 00090000 @ 4224: d74d67b4e9bc8dc6d3536082e78dcae391fa1503
TPM: command 0x14 returned 0x0
TPM: pcr 2 measure 01000000 @ 3772880: 35e974ae974845687c8128055661fe4005e5b970
TPM: command 0x14 returned 0x0
TPM: pcr 2 measure 00040000 @ 180: 31fae3ad7f3a0f2ee6fa1afeaa370ba152dc2662
TPM: command 0x14 returned 0x0
TPM: pcr 2 measure 00091000 @ 25: 1a0c772ebbab34217f5ee534c8e01c94fd8d3bd6
TPM: command 0x14 returned 0x0
TPM: pcr 2 measure 04000000 @ 4041216: 5cef02f227841f10622c4ca0adbb14843d6c4b97
TPM: command 0x14 returned 0x0
```
I think this is because the TPM is not initialised early enough but will test more soon.


tlaurion:
I will use your branch and push it on my own github so that the built is made, since my github heads is followed by CircleCI which trigger builds for defined boards under .circleci/config.yml

I will do edit .circleci/config.yml for it to only build kgpe-d16 boards touched by your branch.
@

tlaurion:
@Tonux599 actually, decided to start by just replicating your branch to trigger a build. (Rebasing resuted iun chaos)
https://app.circleci.com/pipelines/github/tlaurion/heads/534/workflows/67a1e987-9e7d-4ff4-b06b-a471466f701a/jobs/578

tlaurion:
@Tonux599 actually, decided to start by just replicating your branch to trigger a build. (Rebasing resuted in chaos)
https://app.circleci.com/pipelines/github/tlaurion/heads/534/workflows/67a1e987-9e7d-4ff4-b06b-a471466f701a/jobs/578

tlaurion:
Build resulting of merge conflict resolution is happening here: https://app.circleci.com/pipelines/github/tlaurion/heads/536/workflows/823f5132-5586-40d1-be03-3fbd55e01e9e/jobs/580

From branch https://github.com/tlaurion/heads/tree/tonux_kgpe-d16_411_measured-boot_rebased

Tonux599:
@tlaurion so `kgpe-d16_workstation` built so, yay. Looks like `kgpe-d16_workstation-usb_keyboard` failed due to malformed make command.

Tonux599:
Success [here](https://app.circleci.com/pipelines/github/Tonux599/heads/5/workflows/0f965b57-a688-4f0c-9951-1be65d212e43/jobs/9). Will bring this PR on par with that.

tlaurion:
@Tonux599 https://app.circleci.com/pipelines/github/tlaurion/heads/541/workflows/9d19e4e0-a6ab-4d68-bfa1-f47d61b9b760/jobs/588 has passed.

The question here is to know if the microcodes are applied by QubesOS and if the system requires them or not at this point.
You can cherry pick or rebase force push https://github.com/tlaurion/heads/tree/tonux_kgpe-d16_411_measured-boot_rebased on top of here and that would replace #795 :) or I can push my branch over #795 with your commits if build is successful.

I think we should remove microcodes, but I haven't tested the builds myself yet.

Tonux599:
@tlaurion I think we may have made the same changes in each of our branch as this PR now builds as is but I can still force push yours to this if you desire. Let me know :) 

Regarding the microcode, gimme a few days as I'll see if I can fix that error. My concern being that if microcode is not measured **and** not included in the CBFS, an attacker could implement malicious microcode without the end user being aware.

tlaurion:
Todo:
- remove LOCAL_VERSION in coreboot config to be compliant with internally kept version scheme for FWUPD upgrades

tlaurion:
> @tlaurion I think we may have made the same changes in each of our branch as this PR now builds as is but I can still force push yours to this if you desire. Let me know :)
> 
> Regarding the microcode, gimme a few days as I'll see if I can fix that error. My concern being that if microcode is not measured **and** not included in the CBFS, an attacker could implement malicious microcode without the end user being aware.

Well, microcode injection still requires AMD signatures on binaries pushed to CPU so I am not sure that is a real risk. My recommendation was to disable microcode inclusion in ROM. But if you think you can fix this, no prob.

https://github.com/tlaurion/heads/tree/tonux_kgpe-d16_411_measured-boot_rebased is just rebased on master with a more clean CircleCI config, that could be merged as is, since all baords were built successfully, while cache is still being made and uploaded as we speak (with same signature that will be reused for all modules/* and patches/* being the same).

boards/kgpe-d16_server-whiptail/kgpe-d16_server-whiptail.config
```diff
@@ -0,0 +1,75 @@
+# Configuration for a kgpe-d16_server with whiptail, permitting usage of gui-init on console without FB graphic
+# per https://www.raptorengineering.com/coreboot/kgpe-d16-bmc-port-status.php work, with patches merged into Heads
+#
+# Tested: ASMBV4 reprogrammed per https://www.raptorengineering.com/coreboot/kgpe-d16-bmc-port-status.php
+#
+# Status:
+# - ASpeed text output given on VGA
+# - All console output rerirected to sttyS0
+# - No TPM support while 20-1 TPM 1.2 chips can be ordered. Report status in Heads
```
tlaurion, line 9: Maybe give specs here?

boards/kgpe-d16_server/kgpe-d16_server.config
```diff
@@ -0,0 +1,59 @@
+# Configuration for a kgpe-d16_server
+# per https://www.raptorengineering.com/coreboot/kgpe-d16-bmc-port-status.php work, with patches merged into Heads
+#
+# Tested: ASMBV4 reprogrammed per https://www.raptorengineering.com/coreboot/kgpe-d16-bmc-port-status.php
+#
+# Status:
+# - ASpeed text output given on VGA
+# - All console output rerirected to sttyS0
+# - No TPM support while 20-1 TPM 1.2 chips can be ordered. Report status in Heads
```
tlaurion, line 9: same

config/coreboot-kgpe-d16_server-whiptail.config
```diff
@@ -0,0 +1,19 @@
+CONFIG_LOCALVERSION="heads"
```
tlaurion, line 1: remove

config/coreboot-kgpe-d16_server.config
```diff
@@ -1,3 +1,4 @@
+CONFIG_LOCALVERSION="heads"
```
tlaurion, line 1: remove

config/coreboot-kgpe-d16_workstation-usb_keyboard.config
```diff
@@ -0,0 +1,19 @@
+CONFIG_LOCALVERSION="heads"
```
tlaurion, line 1: remove

config/coreboot-kgpe-d16_workstation.config
```diff
@@ -0,0 +1,19 @@
+CONFIG_LOCALVERSION="heads"
```
tlaurion, line 1: remove

config/linux-kgpe-d16_server.config
```diff
@@ -0,0 +1,326 @@
+CONFIG_LOCALVERSION="-heads"
```
tlaurion, line 1: remove

tlaurion, line 1: cp config/linux.. to build/linux/..
cd build/linux..
make savedefconfig
cp defconfig ../../config/linux..

for all linux configs

tlaurion:
@MrChromebox : what is the impact of those coreboot 4.11 patches on the librem_l1um ?

boards/kgpe-d16_server-whiptail/kgpe-d16_server-whiptail.config
tlaurion: TPM SLB9635 TT 1.2 by Infineon

tlaurion:
> Found TPM SLB9635 TT 1.2 by Infineon
> TPM: Startup
> TPM: command 0x99 returned 0x0
> TPM: Asserting physical presence
> TPM: command 0x4000000a returned 0x0
> TPM: command 0x65 returned 0x0
> TPM: flags disable=0, deactivated=0, nvlocked=1
> TPM: setup succeeded
> TPM: pcr 2 measure fffff800 @ 2048: 80909df13fd6a372cca9837b538971e375d3f4d4
> TPM: command 0x14 returned 0x0
> TPM: pcr 2 measure ff000300 @ 1048576: dcfbe783c5d139c5481253b8c8ee8909b49297ca
> TPM: command 0x14 returned 0x0
> TPM: pcr 2 measure 00e00000 @ 213776: 6f279a6cc1687a5b3170d80cca7a3147e59037ca
> TPM: command 0x14 returned 0x0
> TPM: pcr 2 measure 00e00000 @ 1162424: 2ea6887b5d072fb004b2115a320621b8f32f2ac4
> TPM: command 0x14 returned 0x0
> Found TPM SLB9635 TT 1.2 by Infineon
> TPM: Startup
> TPM: command 0x99 returned 0x26
> TPM: Can't run startup command.
> TPM: pcr 2 measure ff042988 @ 9895: 75fe16ebfce00b4becdc5b7f6f31feec791b87c6
> TPM: command 0x14 returned 0x0
> \_SB.PCI0.LPC.TPM: LPC TPM PNP: 004e.0
> TPM: pcr 2 measure ff041b78 @ 3524: a1615718de98d6d619812c6b9fd3c3f5177bc137
> TPM: command 0x14 returned 0x0
> TPM: pcr 2 measure 00090000 @ 4224: d74d67b4e9bc8dc6d3536082e78dcae391fa1503
> TPM: command 0x14 returned 0x0
> TPM: pcr 2 measure 01000000 @ 3772880: 35e974ae974845687c8128055661fe4005e5b970
> TPM: command 0x14 returned 0x0
> TPM: pcr 2 measure 00040000 @ 180: 31fae3ad7f3a0f2ee6fa1afeaa370ba152dc2662
> TPM: command 0x14 returned 0x0
> TPM: pcr 2 measure 00091000 @ 25: 1a0c772ebbab34217f5ee534c8e01c94fd8d3bd6
> TPM: command 0x14 returned 0x0
> TPM: pcr 2 measure 04000000 @ 4041216: 5cef02f227841f10622c4ca0adbb14843d6c4b97
> TPM: command 0x14 returned 0x0

@Tonux599 
I'm confused on why:
> Found TPM SLB9635 TT 1.2 by Infineon
> TPM: Startup

Is shown twice?

tlaurion:
@Tonux599 see https://github.com/tlaurion/heads/commit/4d6cee761ae204095568d8653b04c9eb48057bea

tlaurion:
> This is based upon #795 (credit to @tlaurion) but focuses on measured boot without vboot.
> 
> I would appreciate feedback from the community, notably that `./patches/coreboot-4.11/0020-kgpe-d16_measured-boot-support.patch` is implemented correctly.
> 
> Also note that this error occurs when microcode is included in the CBFS:
> 
> ```
> TPM: pcr 2 measure ff7b9e88 @ 12684: 1077e1af95dc35d5661d54b0ba72652fa0d5aeec
> src/drivers/pc80/tpm/tis.c:521 unexpected TPM status 0xff
> src/drivers/pc80/tpm/tis.c:717 failed sending data to TPM
> TPM: command 0x14 send/receive failed: 0x10000001
> TPM: pcr 2 measure ff7bd088 @ 7876: 983867f34d8599e779cf9e1454f8f4518b5628ac
> src/drivers/pc80/tpm/tis.c:521 unexpected TPM status 0xff
> src/drivers/pc80/tpm/tis.c:717 failed sending data to TPM
> TPM: command 0x14 send/receive failed: 0x10000001
> TPM: pcr 2 measure ff7b9e88 @ 12684: 1077e1af95dc35d5661d54b0ba72652fa0d5aeec
> src/drivers/pc80/tpm/tis.c:521 unexpected TPM status 0xff
> src/drivers/pc80/tpm/tis.c:717 failed sending data to TPM
> TPM: command 0x14 send/receive failed: 0x10000001
> TPM: pcr 2 measure ff7bd088 @ 7876: 983867f34d8599e779cf9e1454f8f4518b5628ac
> src/drivers/pc80/tpm/tis.c:521 unexpected TPM status 0xff
> src/drivers/pc80/tpm/tis.c:717 failed sending data to TPM
> TPM: command 0x14 send/receive failed: 0x10000001
> TPM: pcr 2 measure ff7b9e88 @ 12684: 1077e1af95dc35d5661d54b0ba72652fa0d5aeec
> src/drivers/pc80/tpm/tis.c:521 unexpected TPM status 0xff
> src/drivers/pc80/tpm/tis.c:717 failed sending data to TPM
> TPM: command 0x14 send/receive failed: 0x10000001
> TPM: pcr 2 measure ff7bd088 @ 7876: 983867f34d8599e779cf9e1454f8f4518b5628ac
> src/drivers/pc80/tpm/tis.c:521 unexpected TPM status 0xff
> src/drivers/pc80/tpm/tis.c:717 failed sending data to TPM
> TPM: command 0x14 send/receive failed: 0x10000001
> Found TPM SLB9635 TT 1.2 by Infineon
> TPM: Startup
> TPM: command 0x99 returned 0x0
> TPM: Asserting physical presence
> TPM: command 0x4000000a returned 0x0
> TPM: command 0x65 returned 0x0
> TPM: flags disable=0, deactivated=0, nvlocked=1
> TPM: setup succeeded
> TPM: pcr 2 measure fffff800 @ 2048: 80909df13fd6a372cca9837b538971e375d3f4d4
> TPM: command 0x14 returned 0x0
> TPM: pcr 2 measure ff000300 @ 1048576: dcfbe783c5d139c5481253b8c8ee8909b49297ca
> TPM: command 0x14 returned 0x0
> TPM: pcr 2 measure 00e00000 @ 213776: 6f279a6cc1687a5b3170d80cca7a3147e59037ca
> TPM: command 0x14 returned 0x0
> TPM: pcr 2 measure 00e00000 @ 1162424: 2ea6887b5d072fb004b2115a320621b8f32f2ac4
> TPM: command 0x14 returned 0x0
> Found TPM SLB9635 TT 1.2 by Infineon
> TPM: Startup
> TPM: command 0x99 returned 0x26
> TPM: Can't run startup command.
> TPM: pcr 2 measure ff042988 @ 9895: 75fe16ebfce00b4becdc5b7f6f31feec791b87c6
> TPM: command 0x14 returned 0x0
> \_SB.PCI0.LPC.TPM: LPC TPM PNP: 004e.0
> TPM: pcr 2 measure ff041b78 @ 3524: a1615718de98d6d619812c6b9fd3c3f5177bc137
> TPM: command 0x14 returned 0x0
> TPM: pcr 2 measure 00090000 @ 4224: d74d67b4e9bc8dc6d3536082e78dcae391fa1503
> TPM: command 0x14 returned 0x0
> TPM: pcr 2 measure 01000000 @ 3772880: 35e974ae974845687c8128055661fe4005e5b970
> TPM: command 0x14 returned 0x0
> TPM: pcr 2 measure 00040000 @ 180: 31fae3ad7f3a0f2ee6fa1afeaa370ba152dc2662
> TPM: command 0x14 returned 0x0
> TPM: pcr 2 measure 00091000 @ 25: 1a0c772ebbab34217f5ee534c8e01c94fd8d3bd6
> TPM: command 0x14 returned 0x0
> TPM: pcr 2 measure 04000000 @ 4041216: 5cef02f227841f10622c4ca0adbb14843d6c4b97
> TPM: command 0x14 returned 0x0
> ```
> 
> However when microcode is removed such as:
> 
> ```
> cbfstool coreboot.rom remove -n microcode_amd.bin
> cbfstool coreboot.rom remove -n microcode_amd_fam15h.bin
> ```
> 
> The above error does not occur:
> 
> ```
> Found TPM SLB9635 TT 1.2 by Infineon
> TPM: Startup
> TPM: command 0x99 returned 0x0
> TPM: Asserting physical presence
> TPM: command 0x4000000a returned 0x0
> TPM: command 0x65 returned 0x0
> TPM: flags disable=0, deactivated=0, nvlocked=1
> TPM: setup succeeded
> TPM: pcr 2 measure fffff800 @ 2048: 80909df13fd6a372cca9837b538971e375d3f4d4
> TPM: command 0x14 returned 0x0
> TPM: pcr 2 measure ff000300 @ 1048576: dcfbe783c5d139c5481253b8c8ee8909b49297ca
> TPM: command 0x14 returned 0x0
> TPM: pcr 2 measure 00e00000 @ 213776: 6f279a6cc1687a5b3170d80cca7a3147e59037ca
> TPM: command 0x14 returned 0x0
> TPM: pcr 2 measure 00e00000 @ 1162424: 2ea6887b5d072fb004b2115a320621b8f32f2ac4
> TPM: command 0x14 returned 0x0
> Found TPM SLB9635 TT 1.2 by Infineon
> TPM: Startup
> TPM: command 0x99 returned 0x26
> TPM: Can't run startup command.
> TPM: pcr 2 measure ff042988 @ 9895: 75fe16ebfce00b4becdc5b7f6f31feec791b87c6
> TPM: command 0x14 returned 0x0
> \_SB.PCI0.LPC.TPM: LPC TPM PNP: 004e.0
> TPM: pcr 2 measure ff041b78 @ 3524: a1615718de98d6d619812c6b9fd3c3f5177bc137
> TPM: command 0x14 returned 0x0
> TPM: pcr 2 measure 00090000 @ 4224: d74d67b4e9bc8dc6d3536082e78dcae391fa1503
> TPM: command 0x14 returned 0x0
> TPM: pcr 2 measure 01000000 @ 3772880: 35e974ae974845687c8128055661fe4005e5b970
> TPM: command 0x14 returned 0x0
> TPM: pcr 2 measure 00040000 @ 180: 31fae3ad7f3a0f2ee6fa1afeaa370ba152dc2662
> TPM: command 0x14 returned 0x0
> TPM: pcr 2 measure 00091000 @ 25: 1a0c772ebbab34217f5ee534c8e01c94fd8d3bd6
> TPM: command 0x14 returned 0x0
> TPM: pcr 2 measure 04000000 @ 4041216: 5cef02f227841f10622c4ca0adbb14843d6c4b97
> TPM: command 0x14 returned 0x0
> ```
> 
> I think this is because the TPM is not initialised early enough but will test more soon.

@miczyg1 @pietrushnic any idea on modyfying https://github.com/osresearch/heads/blob/0782f77b9b6f93168decb95f7f4bb09453482272/patches/coreboot-4.11/0020-kgpe-d16_measured-boot-support.patch and https://github.com/osresearch/heads/blob/0782f77b9b6f93168decb95f7f4bb09453482272/patches/coreboot-4.11/0021-kgpe-d16_c-environment_bootblock.patch so that TPM is initialized early enough to measure microcode and to fix TPM being initialized twice?

MrChromebox:
> @MrChromebox : what is the impact of those coreboot 4.11 patches on the librem_l1um ?

the only one which affects anything outside of the KGPE-D16 itself is the C environment patch, and that shouldn't be problematic. I'd want to test and verify to be sure though.

More generally though, feel like this 53-commit patch set could be squashed to ~5 commits and make reviewing a bit easier

Tonux599:
> @Tonux599
> I'm confused on why:
> 
> > Found TPM SLB9635 TT 1.2 by Infineon
> > TPM: Startup
> 
> Is shown twice?

@tlaurion romstage is called three times when the board is booted, with the first two times a soft reset is called. It's likely that the romstage doesn't reach the TPM initialisation on its first iteration and does so on the second and third. This behaviour has been persistent throughout my work on this board.

Tonux599:
> More generally though, feel like this 53-commit patch set could be squashed to ~5 commits and make reviewing a bit easier

If @tlaurion is happy to squash his circleci commits, I'm happy to squash all my commits that are directly to this board.

miczyg1:
@tlaurion what is worth checking is whether the TPM associated memory and IO cycles are decoded by southbridge.  Like this on apu2:
https://github.com/coreboot/coreboot/blob/master/src/southbridge/amd/pi/hudson/bootblock.c#L76 (although the registers may be different on the KGPE-D16). What is the revision of coreboot so that I can check the state of TPM cycles decoding? Is it pure 4.11 tag with your patches on top of it or 4.11_branch?

Tonux599:
> @tlaurion what is worth checking is whether the TPM associated memory and IO cycles are decoded by southbridge. Like this on apu2:
> https://github.com/coreboot/coreboot/blob/master/src/southbridge/amd/pi/hudson/bootblock.c#L76 (although the registers may be different on the KGPE-D16). What is the revision of coreboot so that I can check the state of TPM cycles decoding? Is it pure 4.11 tag with your patches on top of it or 4.11_branch?

Pure 4.11 tag with patches on top as far as I know.
@miczyg1 I have fixed the TPM error in the OP by having the TPM initialise as early as possible, which is after `sb7xx_51xx_early_setup()` and moving `update_microcode(val);` to run after that. So it reads as below. Can you please advise if this is sensible, adequate or stupid. Thanks.
```
	/* run _early_setup before soft-reset. */
	sr5650_early_setup();
	sb7xx_51xx_early_setup();

	if (CONFIG(MEASURED_BOOT) && CONFIG(LPC_TPM)) {
		tpm_setup(0);
		tlcl_lib_init();
	}

	update_microcode(val);

	if (CONFIG(LOGICAL_CPUS)) {
		/* Core0 on each node is configured. Now setup any additional cores. */
		printk(BIOS_DEBUG, "start_other_cores()\n");
		start_other_cores(bsp_apicid);
		post_code(0x37);
		wait_all_other_cores_started(bsp_apicid);
	}
```

miczyg1:
I see that there is the AP startup soon after TPM startup `start_other_cores(bsp_apicid)`.  I just hope the APs do not go the same path and do not perform the tpm_setup over and over again. You will not see the messages about multiple TPM setup routines because the output from APs is squelched. I would add boot_cpu() to the condition:

`if (CONFIG(MEASURED_BOOT) && CONFIG(LPC_TPM) && boot_cpu())` 

This will additionally require an include of smp/node.h (if not yet added)

Tonux599:
> I see that there is the AP startup soon after TPM startup `start_other_cores(bsp_apicid)`. I just hope the APs do not go the same path and do not perform the tpm_setup over and over again. You will not see the messages about multiple TPM setup routines because the output from APs is squelched. I would add boot_cpu() to the condition:
> 
> `if (CONFIG(MEASURED_BOOT) && CONFIG(LPC_TPM) && boot_cpu())`
> 
> This will additionally require an include of smp/node.h (if not yet added)

@miczyg1 And are your suggestions in reference to my comment above or how things are currently implemented at this stage in the branch?

tlaurion:
>>What is the revision of coreboot so that I can check the state of TPM cycles decoding? Is it pure 4.11 tag with your patches on top of it or 4.11_branch?
> 
> Pure 4.11 tag with patches on top as far as I know.

@Tonux599 @miczyg1 : correct. This is pure coreboot 4.11 + https://github.com/Tonux599/heads/tree/kgpe-d16_411_measured-boot/patches/coreboot-4.11 applied on top (where manual path from @Tonux599 is not in branch. You should update your PR to ease commenting this out and ease collaboration!)

Tonux599:
> where manual patch from @Tonux599 is not in branch. You should update your PR to ease commenting this out and ease collaboration!

Not sure what patch your referring to, my local branch is identical this one and would prefer to discuss logic of implementation before pushing changes that are potentially worse than the current implementation.

miczyg1:
> > I see that there is the AP startup soon after TPM startup `start_other_cores(bsp_apicid)`. I just hope the APs do not go the same path and do not perform the tpm_setup over and over again. You will not see the messages about multiple TPM setup routines because the output from APs is squelched. I would add boot_cpu() to the condition:
> > `if (CONFIG(MEASURED_BOOT) && CONFIG(LPC_TPM) && boot_cpu())`
> > This will additionally require an include of smp/node.h (if not yet added)
> 
> @miczyg1 And are your suggestions in reference to my comment above or how things are currently implemented at this stage in the branch?
> 
> Edit: So implementing your suggestions to my comment results in the TPM being initialised 3 times. Implementing it to the current state of the branch results in the TPM only being initialised once, but errors when measuring microcode as its not initialised early enough.
> 
> Edit Edit: @miczyg1 I can always move `update_microcode(val);` to later but would appreciate knowing how sensible that is.

Talking specifically about the fragment pasted here: https://github.com/osresearch/heads/pull/867#issuecomment-716497223
Initializing the TPM 3 times should not cause problems if done sequentially, but if all APs are launched at the same time? Imagine all cores trying to write to TPM at the same time, fighting for the access to the TPM. Not only the TPM will get commands in wrong order but also may end up in wrong state.

The microcode update is ok, the APs will execute the update_microcode function and apply the microcode patch as desired. coreboot uses spin locks, etc. to avoid the race conditions there.

Yet I have to analyze the path the other cores will go, because we have two CPUs.

Tonux599:
> Talking specifically about the fragment pasted here: [#867 (comment)](https://github.com/osresearch/heads/pull/867#issuecomment-716497223)
> Initializing the TPM 3 times should not cause problems if done sequentially, but if all APs are launched at the same time? Imagine all cores trying to write to TPM at the same time, fighting for the access to the TPM. Not only the TPM will get commands in wrong order but also may end up in wrong state.
> 
> The microcode update is ok, the APs will execute the update_microcode function and apply the microcode patch as desired. coreboot uses spin locks, etc. to avoid the race conditions there.
> 
> Yet I have to analyze the path the other cores will go, because we have two CPUs.

Understood, thanks for explaining. I'll implement and update this PR and we can go from there. Any suggestions on checking that all cores are/are not fighting for access to the TPM at the same time? I'll enable `CONFIG_DEBUG_TPM` locally to see if it reports any errors or initialisations happening in quick succession.

tlaurion:
> More generally though, feel like this 53-commit patch set could be squashed to ~5 commits and make reviewing a bit easier
@MrChromebox : this is linked to an old debate for which I kinda disagree, once again, in the approach of community based projects.

Squasing commits of a PR at merge time resolves the issue, while limiting giving proper credits to people having contributed. Also, squashing commits, and removing original branches who lead to understanding of the problem, like in this PR, based on an older one, which is based on master's past 4.8.1 port, makes it extremely difficult to newcomers to understand what it means to adapt a board, linux and coreboot configs to make it work with Heads and all implied problems.

I would personally, once this PR is known good, only squash at commit, where I give permission to @Tonux599 to take my changes into CircleCI and make them his own. We have to have a reflexion here on how to do proper code review in PR, not deleting branches, so that newcomers can understand what is the process of bringing new board configs into Heads project. I agree it adds non-relevant commits, while exposing the path that was done to arrive to the end result. Removing those commits requires other people to do the same trial and errors without having clear source of information to base themselves on. While code review can happen directly on changes files, without caring on the specific commits themselves, but the combined result of those commit.

That is my personal opinion.

MrChromebox:
@tlaurion I'm not a fan of squashing all commits at merge time, since it does lose information, and makes it harder to cleanly revert a problematic section. I'm simply suggesting that this 53-patch train can be made more clear by squashing into say 3 commits (CircleCI changes, config changes, coreboot patches). Having signed-off-by lines solves attribution issues. The PR itself provides more than adequate history to the path taken to arrive at the final solution, while the squashed commits provide a clear indication as to what that final solution is.

boards/kgpe-d16_server-whiptail/kgpe-d16_server-whiptail.config
Tonux599: Done. tlaurion/heads@4d6cee761ae204095568d8653b04c9eb48057bea cherry picked.

boards/kgpe-d16_server/kgpe-d16_server.config
Tonux599: Done. tlaurion/heads@4d6cee761ae204095568d8653b04c9eb48057bea cherry picked.

config/coreboot-kgpe-d16_server-whiptail.config
Tonux599: Done. tlaurion/heads@4d6cee761ae204095568d8653b04c9eb48057bea cherry picked.

config/coreboot-kgpe-d16_server.config
Tonux599: Done. tlaurion/heads@4d6cee761ae204095568d8653b04c9eb48057bea cherry picked.

config/coreboot-kgpe-d16_workstation-usb_keyboard.config
Tonux599: Done. tlaurion/heads@4d6cee761ae204095568d8653b04c9eb48057bea cherry picked.

config/coreboot-kgpe-d16_workstation.config
Tonux599: Done. tlaurion/heads@4d6cee761ae204095568d8653b04c9eb48057bea cherry picked.

tlaurion:
@MrChromebox : so plan here would be to create another PR, pointing to past PR and issues, and do a cleaner PR.
Perfect. Will do!

config/linux-kgpe-d16_server.config
Tonux599: Fixed with 2de4e7a4a5e31b6ea2486564de776fb731e89731

MrChromebox:
@tlaurion personally I would just force-push a cleaned set of commits here, but marking this PR as WIP and opening another with the cleaned commits and pointing to this would be fine too.

Tonux599:
So enabling `CONFIG_DEBUG_TPM` results in a (presumably) never-ending loop of `lpc_tpm: Read reg 0x18 returns 0x40`. Strange this lock-up doesn't happen when `CONFIG_DEBUG_TPM=n` or undefined. Full log below from the serial port from the point of power button. Dunno if this helps you diagnose @miczyg1 

<details><summary>Log</summary>
<p>

```coreboot--TIMELESS--LESSTIME--Heads-v0.2.0-985-g2de4e7a4-dirty Thu Jan  1 00:00:00 UTC 1970 romstage starting (log level: 8)...
Initial stack pointer: 000dffb8
CPU APICID 00 start flag set
BSP Family_Model: 00600f12
*sysinfo range: [000c2f60,000cd4cc]
bsp_apicid = 00
cpu_init_detectedx = 00000000
sb700 reset flags: 0000
cpuSetAMDMSR FMAP: Found "FLASH" version 1.1 at 0x0.
FMAP: base = 0xff000000 size = 0x1000000 #areas = 3
FMAP: area COREBOOT found @ 200 (16776704 bytes)
CBFS: Locating 'cmos_layout.bin'
CBFS: Found @ offset 41d80 size dc4
FMAP: area COREBOOT found @ 200 (16776704 bytes)
CBFS: Locating 'cmos_layout.bin'
CBFS: Found @ offset 41d80 size dc4
 done
Enter amd_ht_init
AMD_CB_EventNotify: INFO: HT_EVENT_COH_NODE_DISCOVERED: node 0 link 2 new node: 1
AMD_CB_EventNotify: INFO: HT_EVENT_COH_NODE_DISCOVERED: node 0 link 0 new node: 2
AMD_CB_EventNotify: INFO: HT_EVENT_COH_NODE_DISCOVERED: node 0 link 3 new node: 3
FMAP: area COREBOOT found @ 200 (16776704 bytes)
CBFS: Locating 'cmos_layout.bin'
CBFS: Found @ offset 41d80 size dc4
Forcing HT links to isochronous mode due to enabled IOMMU
FMAP: area COREBOOT found @ 200 (16776704 bytes)
CBFS: Locating 'cmos_layout.bin'
CBFS: Found @ offset 41d80 size dc4
Exit amd_ht_init
amd_ht_fixup
amd_ht_fixup: node 0 (internal node ID 0): disabling defective HT link (L3 connected: 1)
amd_ht_fixup: node 1 (internal node ID 1): disabling defective HT link (L3 connected: 1)
amd_ht_fixup: node 2 (internal node ID 0): disabling defective HT link (L3 connected: 1)
amd_ht_fixup: node 3 (internal node ID 1): disabling defective HT link (L3 connected: 1)
cpuSetAMDPCI 00 done
cpuSetAMDPCI 01 done
cpuSetAMDPCI 02 done
cpuSetAMDPCI 03 done
Prep FID/VID Node:00
  F3x80: e20be281
  F3x84: 01e200e2
  F3xD4: c3312f1a
  F3xD8: 03000016
  F3xDC: 05475631
Prep FID/VID Node:01
  F3x80: e20be281
  F3x84: 01e200e2
  F3xD4: c3312f1a
  F3xD8: 03000016
  F3xDC: 05475631
Prep FID/VID Node:02
  F3x80: e20be281
  F3x84: 01e200e2
  F3xD4: c3312f1a
  F3xD8: 03000016
  F3xDC: 05475631
Prep FID/VID Node:03
  F3x80: e20be281
  F3x84: 01e200e2
  F3xD4: c3312f1a
  F3xD8: 03000016
  F3xDC: 05475631
setup_remote_node: 01 done
Start node 01 done.
setup_remote_node: 02 done
Start node 02 done.
setup_remote_node: 03 done
Start node 03 done.
core0 started:  01 02 03
sr5650_early_setup()
get_cpu_rev EAX=0x600f12.
CPU Rev is Fam 15.
NB Revision is A12.
fam10_optimization()
sr5650_por_init
FMAP: area COREBOOT found @ 200 (16776704 bytes)
CBFS: Locating 'cmos_layout.bin'
CBFS: Found @ offset 41d80 size dc4
Enabling IOMMU
sb700_early_setup()
FMAP: area COREBOOT found @ 200 (16776704 bytes)
CBFS: Locating 'cmos_layout.bin'
CBFS: Found @ offset 41d80 size dc4
sb700_devices_por_init()
sb700_devices_por_init(): SMBus Device, BDF:0-20-0
SMBus controller enabled, sb revision is A15
sb700_devices_por_init: Disabling ISA DMA support
sb700_devices_por_init(): IDE Device, BDF:0-20-1
sb700_devices_por_init(): LPC Device, BDF:0-20-3
sb700_devices_por_init(): P2P Bridge, BDF:0-20-4
sb700_devices_por_init(): SATA Device, BDF:0-17-0
FMAP: area COREBOOT found @ 200 (16776704 bytes)
CBFS: Locating 'cmos_layout.bin'
CBFS: Found @ offset 41d80 size dc4
sb700_pmio_por_init()
lpc_tpm: Read reg 0xf00 returns 0xb15d1
Found TPM SLB9635 TT 1.2 by Infineon
lpc_tpm: Read reg 0x0 returns 0x81
lpc_tpm: Write reg 0x0 with 0x2
lpc_tpm: Read reg 0x0 returns 0xa1
lpc_tpm: Write reg 0x18 with 0x40
lpc_tpm: Read reg 0x18 returns 0x0
lpc_tpm: Write reg 0x18 with 0x40
lpc_tpm: Read reg 0x18 returns 0x0
lpc_tpm: Read reg 0x18 returns 0x40
TPM: Startup
lpc_tpm: Read reg 0x18 returns 0x40
lpc_tpm: Read reg 0x19 returns 0x8
lpc_tpm: Write reg 0x18 with 0x0
lpc_tpm: Write reg 0x18 with 0xc1
lpc_tpm: Write reg 0x18 with 0x0
lpc_tpm: Write reg 0x18 with 0x0
lpc_tpm: Write reg 0x18 with 0x0
lpc_tpm: Write reg 0x18 with 0xc
lpc_tpm: Write reg 0x18 with 0x0
lpc_tpm: Write reg 0x18 with 0x0
lpc_tpm: Read reg 0x18 returns 0x8
lpc_tpm: Read reg 0x18 returns 0x88
lpc_tpm: Read reg 0x18 returns 0x88
lpc_tpm: Read reg 0x19 returns 0x8
lpc_tpm: Write reg 0x18 with 0x0
lpc_tpm: Write reg 0x18 with 0x99
lpc_tpm: Write reg 0x18 with 0x0
lpc_tpm: Read reg 0x18 returns 0x8
lpc_tpm: Read reg 0x18 returns 0x88
lpc_tpm: Read reg 0x18 returns 0x88
lpc_tpm: Read reg 0x19 returns 0x8
lpc_tpm: Write reg 0x18 with 0x1
lpc_tpm: Read reg 0x18 returns 0x8
lpc_tpm: Read reg 0x18 returns 0x80
lpc_tpm: Read reg 0x18 returns 0x80
lpc_tpm: Write reg 0x18 with 0x20
lpc_tpm: Read reg 0x18 returns 0x80
lpc_tpm: Read reg 0x18 returns 0x0
lpc_tpm: Read reg 0x18 returns 0x0
lpc_tpm: Read reg 0x18 returns 0x0
lpc_tpm: Read reg 0x18 returns 0x0
lpc_tpm: Read reg 0x18 returns 0x0
lpc_tpm: Read reg 0x18 returns 0x0
lpc_tpm: Read reg 0x18 returns 0x0
lpc_tpm: Read reg 0x18 returns 0x0
lpc_tpm: Read reg 0x18 returns 0x0
lpc_tpm: Read reg 0x18 returns 0x0
lpc_tpm: Read reg 0x18 returns 0x0
lpc_tpm: Read reg 0x18 returns 0x0
lpc_tpm: Read reg 0x18 returns 0x0
lpc_tpm: Read reg 0x18 returns 0x40
lpc_tpm: Read reg 0x18 returns 0x40
lpc_tpm: Read reg 0x18 returns 0x0
lpc_tpm: Read reg 0x18 returns 0x0
lpc_tpm: Read reg 0x18 returns 0x0
lpc_tpm: Read reg 0x18 returns 0x0
lpc_tpm: Read reg 0x18 returns 0x0
lpc_tpm: Read reg 0x18 returns 0x0
lpc_tpm: Read reg 0x18 returns 0x40
lpc_tpm: Read reg 0x18 returns 0x40
lpc_tpm: Read reg 0x18 returns 0x40
lpc_tpm: Read reg 0x18 returns 0x40
lpc_tpm: Read reg 0x18 returns 0x40
lpc_tpm: Read reg 0x18 returns 0x40
lpc_tpm: Read reg 0x18 returns 0x40
lpc_tpm: Read reg 0x18 returns 0x40
lpc_tpm: Read reg 0x18 returns 0x40
lpc_tpm: Read reg 0x18 returns 0x40
lpc_tpm: Read reg 0x18 returns 0x40
lpc_tpm: Read reg 0x18 returns 0x40
lpc_tpm: Read reg 0x18 returns 0x40
lpc_tpm: Read reg 0x18 returns 0x40
lpc_tpm: Read reg 0x18 returns 0x40
lpc_tpm: Read reg 0x18 returns 0x40
{and on and on}
```

</p>
</details>

miczyg1:
@Tonux599 you may either disable the debug or try the delay workaround option available in the TPM menu in coreboot. Hope it helps.

Tonux599:
Hi @miczyg1, so `CONFIG_TPM_RDRESP_NEED_DELAY` is enabled anyway as I experienced errors without this in the past. While disabling the debug does work, I'm wondering if these messages are linked to the 'fight for access all at once' concern you raised?

miczyg1:
Hi @Tonux599 debug may introduce some discrepancies in the communication with TPM. If it works without debug there is no point in trying to keep it enabled. This issue shouldn't be related to the "fight for access" any longer, since only the BSP core initializes TPM.

Tonux599:
Understood, thanks for explaining again! I infer from your answer then there is no concerns for that issue any more. Would you say that `0020-kgpe-d16_measured-boot-support.patch` is fit for purpose?

miczyg1:
Briefly looking at the patches, I would (almost) say yes. However, I am not that familiar with heads measured boot model so can't say much about the measuring policy (where something should be measured). The hardcoded bootblock size and base address look a little bit ugly. Wouldn't it be better to locate it with CBFS API?

```
size_t bootblock_size = 0;
const void *bootblock = cbfs_boot_map_with_leak("bootblock", 1, &bootblock_size);
```

`1` means CBFS_TYPE_BOOTBLOCK (although not defined anywhere except cbfstool). It will return the pointer to the bootblock contents as well as size passed via the bootblock_size reference. No more hardcodes needed. This will need cbfs.h include.

Tonux599:
> Briefly looking at the patches, I would (almost) say yes. However, I am not that familiar with heads measured boot model so can't say much about the measuring policy (where something should be measured).

@osresearch ?

>The hardcoded bootblock size and base address look a little bit ugly. Wouldn't it be better to locate it with CBFS API?
> 
> ```
> size_t bootblock_size = 0;
> const void *bootblock = cbfs_boot_map_with_leak("bootblock", 1, &bootblock_size);
> ```
> 
> `1` means CBFS_TYPE_BOOTBLOCK (although not defined anywhere except cbfstool). It will return the pointer to the bootblock contents as well as size passed via the bootblock_size reference. No more hardcodes needed. This will need cbfs.h include.

Implemented and tested, thanks :+1:.

tlaurion:
@Tonux599 : the measuring policy was recently modified into heads to match VBOOT measured scheme as documented under official documentation [website](http://osresearch.net/Keys/#tpm-pcrs) and link to this [commit](https://github.com/osresearch/heads-wiki/commit/8e8a50eb1348a214c3b2db92ceaa96ee0e52a907)

Referring to changes that happened [here under Heads](https://patch-diff.githubusercontent.com/raw/osresearch/heads/pull/793.patch) to match [coreboot 4.12 measuring](https://user-images.githubusercontent.com/827570/82279087-b2da2000-9959-11ea-8020-6ff36e947576.jpeg) as discussed [here](https://github.com/osresearch/heads/pull/721#issuecomment-630546662)

Let me know if this requires more clarification!

osresearch:
>[...]The hardcoded bootblock size and base address look a little bit ugly. Wouldn't it be better to locate it with CBFS API?

One issue with trusting the external structure is if an attacker can modify the CBFS struct to point to a clean copy, which is not used by the reset vector since it will use the `0xFFFFFFF0` legacy address, then the measurement might not reflect the reality of what was booted. 

Although, an attacker in that position could also modify the bootblock unless there is an ACM that does the PCR0 measurement, so it is probably acceptable.

tlaurion:
@miczyg1 @Tonux599 per https://github.com/osresearch/heads/pull/867#issuecomment-721283300 it is then better to use hardcoded address?

Tonux599:
> @miczyg1 @Tonux599 per [#867 (comment)](https://github.com/osresearch/heads/pull/867#issuecomment-721283300) it is then better to use hardcoded address?

I'm not qualified but I'm happy to revert my commit that implemeted @miczyg1's suggestion if advised to by maintainers and collaborators of this project.

tlaurion:
@Tonux599 please have a look https://github.com/osresearch/heads-wiki/pull/52#issuecomment-721915936

Tonux599:
> @Tonux599 : the measuring policy was recently modified into heads to match VBOOT measured scheme as documented under official documentation [website](http://osresearch.net/Keys/#tpm-pcrs) and link to this [commit](https://github.com/osresearch/heads-wiki/commit/8e8a50eb1348a214c3b2db92ceaa96ee0e52a907)
> 
> Referring to changes that happened [here under Heads](https://patch-diff.githubusercontent.com/raw/osresearch/heads/pull/793.patch) to match [coreboot 4.12 measuring](https://user-images.githubusercontent.com/827570/82279087-b2da2000-9959-11ea-8020-6ff36e947576.jpeg) as discussed [here](https://github.com/osresearch/heads/pull/721#issuecomment-630546662)
> 
> Let me know if this requires more clarification!

> @Tonux599 please have a look [osresearch/heads-wiki#52 (comment)](https://github.com/osresearch/heads-wiki/pull/52#issuecomment-721915936)

@tlaurion You've confused me a bit, as this PR stands everything is extended into PCR2 just like other boards.

tlaurion:
> > @Tonux599 : the measuring policy was recently modified into heads to match VBOOT measured scheme as documented under official documentation [website](http://osresearch.net/Keys/#tpm-pcrs) and link to this [commit](https://github.com/osresearch/heads-wiki/commit/8e8a50eb1348a214c3b2db92ceaa96ee0e52a907)
> > Referring to changes that happened [here under Heads](https://patch-diff.githubusercontent.com/raw/osresearch/heads/pull/793.patch) to match [coreboot 4.12 measuring](https://user-images.githubusercontent.com/827570/82279087-b2da2000-9959-11ea-8020-6ff36e947576.jpeg) as discussed [here](https://github.com/osresearch/heads/pull/721#issuecomment-630546662)
> > Let me know if this requires more clarification!
> 
> > @Tonux599 please have a look [osresearch/heads-wiki#52 (comment)](https://github.com/osresearch/heads-wiki/pull/52#issuecomment-721915936)
> 
> @tlaurion You've confused me a bit, as this PR stands everything is extended into PCR2 just like other boards.

@Tonux599 : sorry to have confused you. Just wanted to point to, and have your feedback, on what should be documented to permit new dev onboarding on https://github.com/osresearch/heads-wiki/pull/52#issuecomment-721915936

miczyg1:
> > [...]The hardcoded bootblock size and base address look a little bit ugly. Wouldn't it be better to locate it with CBFS API?
> 
> One issue with trusting the external structure is if an attacker can modify the CBFS struct to point to a clean copy, which is not used by the reset vector since it will use the `0xFFFFFFF0` legacy address, then the measurement might not reflect the reality of what was booted.
> 
> Although, an attacker in that position could also modify the bootblock unless there is an ACM that does the PCR0 measurement, so it is probably acceptable.

If CBFS is modified then it doesn't matter if you use legacy reset vector or CBFS API or hardcoded addresses. The measured code will not be the one which was booted in all cases... Using hardcoded 0xFFFFF800 isn't better either. Somebody can place a modified bootblock there too. The CBFS API will return this address if nothing was modified. How do you trust the CBFS API to locate next stages then? How do you trust the romstage or ramstage that is executed next after bootblock?

If the flash is not protected in any way (either ACM or WP pin) what is the trust anchor here? Or what are the measurements for? If somebody replaces this bootblock, it should be reflected by the measurements. Maybe I am biased with how coreboot is developed: "if some functionality in the coreboot exists, use it instead of hardcoding or reimplementing the wheel".  @Tonux599 I am sorry if my advice was bad.

miczyg1:
For example one may modify just a single letter in CBFS file header, for example `fallback/ramstage` -> `fallback/rumstage` and CBFS API won't find your ramstage. The boot process will simply stop. And it can be done without modifying a single bit in the ramstage code.

Tonux599:
> [...] @Tonux599 I am sorry if my advice was bad.

@miczyg1 There is no need to apologise as I hold no difference in value of your advice or @osresearch's advice. At this stage I'm merely nothing more than someone who has some free time to attempt to bring a board that they own into support with heads.

I will, however, make adjustments to my contributions (unless I have a moral objection) in line with projects maintainers wishes (which is a little unclear at the moment, clarification on @tlaurion's and @osresearch's role in the project could be laid out in a `CONTRIBUTING.md`, which would also aid in the effort of onboard new devs into this project).

But back to the subject, I would encourage @miczyg1, @osresearch and really anyone whom have a good understanding coreboot to continue arguing, at which point I would expect maintainers to weigh each side of the argument and reach a decision. I would then implement that.

tlaurion:
@miczyg1 @osresearch : I'm kinda confused right now.

- Heads (@flammit ) implemented changes into [flashtools](https://github.com/osresearch/heads/blob/master/modules/flashtools) to permit additions of files into CBFS free space at Heads runtime, relying on flashrom to take a backup of the ROM and inject the changes back.
- cbfs is now [used to inject user's public key inside of the rom ](https://github.com/osresearch/heads/blob/6a3bb5897af350de72adf89ff643176656f7f257/initrd/bin/gpg-gui.sh) when reowning/owing a USB Security dongle after key generations
- cbfs is also used to[ inject current user's configuration](https://github.com/osresearch/heads/blob/6a3bb5897af350de72adf89ff643176656f7f257/initrd/bin/config-gui.sh) ([which is injected in build time](https://github.com/osresearch/heads/blob/master/Makefile#L591-L599)) into cbfs if the user wants to override the config (kernel passed options and other configs)
-Heads [measures those cbfs modified files prior of using them and modifies PCR7 accordingly](https://github.com/osresearch/heads/blob/584c07042ef4898de52ba4981997211c646b75fd/initrd/bin/cbfs-init#L5-L29) early in [init](https://github.com/osresearch/heads/blob/master/initrd/init#L47-L49)

[Consequently ](http://osresearch.net/Keys/#tpm-pcrs) PCRs measurements for each PCRs are:
- 2: Boot block, ROM stage, RAM stage, Heads linux kernel and initrd
- 4: Boot mode (0 during /init, then recovery or normal-boot)
- 5: Heads Linux kernel modules
- 6: Drive LUKS headers
- 7: Heads user-specific config files

Any problem/improvements that can happen here if there is no ACM?

miczyg1:
> @miczyg1 @osresearch : I'm kinda confused right now.
> 
> * Heads (@flammit ) implemented changes into [flashtools](https://github.com/osresearch/heads/blob/master/modules/flashtools) to permit additions of files into CBFS free space at Heads runtime, relying on flashrom to take a backup of the ROM and inject the changes back.
> * cbfs is now [used to inject user's public key inside of the rom ](https://github.com/osresearch/heads/blob/6a3bb5897af350de72adf89ff643176656f7f257/initrd/bin/gpg-gui.sh) when reowning/owing a USB Security dongle after key generations
> * cbfs is also used to[ inject current user's configuration](https://github.com/osresearch/heads/blob/6a3bb5897af350de72adf89ff643176656f7f257/initrd/bin/config-gui.sh) ([which is injected in build time](https://github.com/osresearch/heads/blob/master/Makefile#L591-L599)) into cbfs if the user wants to override the config (kernel passed options and other configs)
> * Heads [measures those cbfs modified files prior of using them and modifies PCR7 accordingly](https://github.com/osresearch/heads/blob/584c07042ef4898de52ba4981997211c646b75fd/initrd/bin/cbfs-init#L5-L29) early in [init](https://github.com/osresearch/heads/blob/master/initrd/init#L47-L49)

These references are not really related to the discussion. The mentioned mechanisms are heads specific and it is all about heads how they will be handled. They do not have any particular impact on the coreboot stage localization.

If I am not mistaken the CBFS API responsible for locating coreboot stage during boot are intact. So for me it is pointless to hardcode the bootblock address and size if the CBFS API can locate it and return the size. The security threat does not change by using CBFS API here, if someone modifies the flash on the fly, it doesn't matter what method was used (hardcoding or CBFS API). The same CBFS API is used to locate coreboot stages during boot, someone may put a fake/malicious ramstage or other
stage to execute own code. What is important is that everything is measured. As long as the measurements are ok, one should be fine (at least in this particular case). If the SPI flash content was replaced, the measurements should reflect that.

> [Consequently ](http://osresearch.net/Keys/#tpm-pcrs) PCRs measurements for each PCRs are:
> 
> * 2: Boot block, ROM stage, RAM stage, Heads linux kernel and initrd
> * 4: Boot mode (0 during /init, then recovery or normal-boot)
> * 5: Heads Linux kernel modules
> * 6: Drive LUKS headers
> * 7: Heads user-specific config files
> 
No particular comments with that. It looks good.

pietrushnic:
Bootblock is not protected by any means more to that IIRC it measures itself. Bootblock can easily lie about itself and this is an attack vector.

> One issue with trusting the external structure is if an attacker can modify the CBFS struct to point to a clean copy, which is not used by the reset vector since it will use the 0xFFFFFFF0 legacy address, then the measurement might not reflect the reality of what was booted.

@osresearch if I understand the only way to do that is somehow trick SPI reading (spispy?), but do we really have to use such a sophisticated mechanism when we have no S-CRTM?

Also, I think @miczyg1 got it right and it doesn't matter if we hardcode or use CBFS API if this CBFS API is the problem we should distrust it in all cases not only in bootblock. Of course, hardcoding all addresses will eventually cause maintenance issues that's why if there is no difference in threat model we should use a more convenient mechanism.

Tonux599:
So just to note, I've been using kgpe-d16_workstation for about a week now with an nvidia card and Qubes 4.0 and can confirm that all appears stable and heads specific function (i.e. adding PGP keys to flash) work without error.

@osresearch I'm inclined to keep @miczyg1's implementation of locating bootblock unless you have anything further to argue? You input on @miczyg1 and @pietrushnic have said would really be appreciated.

Tonux599:
Force push due to email change (email updated on both GitHub and PGP key). HEAD is verified but commits before appear unverified. When clean-up + squash take place all commits will be signed and verified correctly.

Tonux599:
@tlaurion please see the latest four commits [here](https://github.com/Tonux599/heads/commits/kgpe-d16_411_measured-boot-testing) which is squashed based more on authors as opposed to features (things ended up getting messy). If you're happy with this I can force push into this PR

config/linux-kgpe-d16_server.config
tlaurion, line 1: Must delete

Tonux599: @tlaurion delete in `linux-kgpe-d16_server-whiptail.config` and `linux-kgpe-d16_workstation.config` also?

MrChromebox: > Must delete

this is in the Linux config, not coreboot, so it's still used if present. I don't think we ever discussed removing it from the kernel build string

tlaurion: Absolutely right. Will review only with coffee from now on.

Tonux599:
Rebased. Ready imo @tlaurion

tlaurion:
Reported functional for workstation. Lets just remember that past working commit was https://github.com/osresearch/heads/commit/014e59210d72fc9577eb31540fabf87d83e7f92a prior of moving the kgpe-d16 boards to 4.11!

Awesome work @Tonux599 !!!!

blobless:
I can confirm it works with CONFIG_TPM=n. Excellent job @Tonux599

Tonux599:
> I can confirm it works with CONFIG_TPM=n. Excellent job @Tonux599

Thanks @blobless! There are TPM's available if you choose to use one. I use 'Asus 90-C1B0AU-00XBN0VZ' with no issues.

PR #868, opened 2020-10-23
Small GUI Tweaks

MrChromebox:
A handful of small GUI improvements:

- gpg_gui: use 'and' vs '+' in menu listings
- gui_init: use consistent notation for main menu
- Specify 'friendly' board name vs overriding CONFIG_BOOT_GUI_MENU_NAME
- gui-init: add board name, kernel version to System Info
- flash-gui: improve readability of ROM filename
- Drop duplicate board-specific background color configs

Signed-off-by: Matt DeVillier <matt.devillier@puri.sm>

PR #869, opened 2020-10-23
Improve handling of missing or empty boot disk

MrChromebox:
In gui-init, check for presence of CONFIG_BOOT_DEV, and if missing or unable to be mounted, present the user with a menu offering the option to select another disk, boot from USB, continue to main menu, or drop to a recovery shell.

In oem-factory-reset, if an installed OS is not detected, then skip setting the default boot device or generating /boot checksums.

These changes facilitate ease of use for users when installing a new/empty boot disk, or for delivering an OEM system with preconfigured keys but no installed OS.

### DIFF
--- a/initrd/bin/gui-init
+++ b/initrd/bin/gui-init
@@ -14,27 +14,41 @@ mount_boot()
   
   # Mount local disk if it is not already mounted
   while ! grep -q /boot /proc/mounts ; do
-    # ensure default boot device is set
-    if [ ! -e "$CONFIG_BOOT_DEV" ]; then
-      if (whiptail $BG_COLOR_ERROR --clear --title "ERROR: $CONFIG_BOOT_DEV missing!" \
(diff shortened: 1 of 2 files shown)


tlaurion:
@MrChromebox : looks good to me! needs rebasing.
I also think that the OEM factory reset would fit better/be duplicated in the GPG menu, so that in case of no GPG key being present, the OEM factory reset option would be presented to the user as a possibility instead of just having the choice to go manual.

MrChromebox:
> @MrChromebox : looks good to me! needs rebasing.

done

> I also think that the OEM factory reset would fit better/be duplicated in the GPG menu, so that in case of no GPG key being present, the OEM factory reset option would be presented to the user as a possibility instead of just having the choice to go manual.

handle in a separate PR?

tlaurion:
@MrChromebox Small conflict after having merged #868

MrChromebox:
> @MrChromebox Small conflict after having merged #868

rebased

PR #889, opened 2020-11-03
add chip model for t420 flashrom options

alexmaloteaux:
on t420 (at least on my board) flash.sh and flashrom fails to read / write the bios chip unless the exact chip model is provided to flashrom.

here is a screenshot taken from heads on a t420

sc


tlaurion:
@alexmaloteaux Is it ok if I squash your pull request (2 commits)? Your commits aren't signed. So the result would be me taking ownership of your change.

We basically just want https://github.com/osresearch/heads/pull/889/commits/100461555ebc7cf7f1002ab6eecceca1a50e714a where https://github.com/osresearch/heads/pull/889/commits/527f5931cfc6fb969bc333ac1811d2b1f1944681 is irrelevant (a lot of different SPI chips are present under the hood of the x220)

alexmaloteaux:
i squashed it , force-push and setup gpg too (then reforce push :) )

tlaurion:
Merging. @alexmaloteaux so I take for granted that you applied https://github.com/osresearch/heads/pull/877/files instructions to t420 and it worked? So I should add t420 modifications there?

PR #903, opened 2020-11-20
Add support to denote boot device via label or uuid

root-hardenedvault:
Now one could say CONFIG_BOOT_DEV="LABEL=boot" in board level config.


tlaurion:
@MrChromebox I love it!

MrChromebox:
well, we need to adjust a lot of other handling to accommodate this in other areas: automatic boot detection, user selection of boot device from GUI, etc. unless we want it to only be a board config level feature

PR #906, opened 2020-11-24
Default to 4096 bit for OEM factory reset (fixes #831)

### LINKED ISSUE #831: Have the GPG factory reset functions propose/default to RSA 4096 bits keys, not 2048
@alex-nitrokey @MrChromebox : what approach you prefer?

techge:
Fixes #831

### DIFF
--- a/initrd/bin/oem-factory-reset
+++ b/initrd/bin/oem-factory-reset
@@ -20,6 +20,8 @@ ADMIN_PIN_DEF=12345678
 TPM_PASS_DEF=12345678
 CUSTOM_PASS=""
 
+RSA_KEY_LENGTH=4096
+
 GPG_USER_NAME="OEM Key"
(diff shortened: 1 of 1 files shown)


tlaurion:
@techge @flammit 
This is excatly what i'm talking about.

@flammit should I merge? this forces gpg2 module usage and implies that all boards will use gpg2 module in the future.
Precedently of our discussion, I would merge. Now I just can't.

your input welcome.

flammit:
I don't understand why there would be an issue merging this.  Heads currently already assumes that the gpg installed is gpg2, so there's no harm in committing this.

In the future if it is possible to have either gpg1 or gpg2 be an included module on a board, then _that_ PR that introduces gpg1 should also update the logic here so that the additions of this PR (or any other feature that requires gpg2) are only run when the gpg2 module is used.

tlaurion:
Allright! Merging! Happy merge day!

PR #907, opened 2020-11-24
safeboot boot loader config

osresearch:
The safeboot board configuration is not for a normal coreboot design; it is delivered either as an EFI executable on the local disk or as a PXE image for booting from the network.  This means that it does not have the normal size restrictions and can include fancier tools like cryptsetup 2, bash, openssl, etc.


tlaurion:
Abandoned

Issue #885, opened 2020-11-02
Improvements to scripts under initrd


Thrilleratplay:
During the course of debugging #872, correcting [shellcheck](https://github.com/koalaman/shellcheck) errors and warnings, a number of issues presented that made verifying the changes difficult.  Examples of this were errors that existed in the master branch (although these may have been unique to using `qemu-coreboot-fwwhiptail`), needing to [unquote variables used as whiptail parameters](https://github.com/osresearch/heads/pull/872/commits/85ce46435991396648a3ba1ca6e7176eeeee6946) then adding shellcheck exceptions, as many scripts do not have `.sh` extensions there stances where it is difficult to determine if a function or script is being invoked, and not owning a Librem/Nitro Key to be able to test and of those scripts.  Given how fragile the existing initrd script is, I believe a different approach is needed.

The original plan was to correct shellcheck errors and warning then progress to cleaning up and tightening the scripts.  At this time, it seems that these should be done together; focusing on improving, cleaning up, documenting, normalizing individual functionality to minimize the amount of testing and reduce the potential for introducing bugs.

A few suggestions:

- [ ] [enable CONFIG_BASH_IS_ASH](https://github.com/osresearch/heads/pull/872/commits/4802f713455aae88497d25d0d80cb59d59605c48)
- [ ]  [Give all script files `.sh` exetension](https://github.com/osresearch/heads/pull/872/commits/6a81b16942dadb4ab239ec6bf1cc2f5634e2d813).  Be sure to add [`.sh` extension to references in the script](https://github.com/osresearch/heads/pull/872/commits/4c3a8d891835c3309c7424fba7e3c16078178ce0).
- [ ]  [Replace sheband with `#!/bin/bash`](https://github.com/osresearch/heads/pull/872/commits/7740f5328632ce966c95cbcbcbdf1de933f496e0)
- [ ]  Use function to generate/normalize y/n menus
- [ ]  use environmental variables via `/etc/profile` instead of `/tmp/config` (is there a reason?)
- [ ]  Pass shellcheck without any no ignore statements
- [ ]  No global variables.  Either are defined or documented as an environmental variable.
- [ ]  Functions much return value or return true.
- [ ]  Remove .ash_history, replace examples in ash_history with alias or bash scripts
- [ ]  Ensure all user facing bash scripts have usage menus.
[...]

PR #909, opened 2020-11-26
feat(busybox): set CONFIG_BASH_IS_ASH

Thrilleratplay:
enable CONFIG_BASH_IS_ASH

Part of #885 and should be included with #900


tlaurion:
@Thrilleratplay : impact?

tlaurion:
https://www.in-ulm.de/~mascheck/various/ash/#busybox :
```
compile time option ("ASH_BASH_COMPAT", active per default) for several bash compatibility features:
option pipefail, substring and replacement parameter expansion ${x:y:z} and ${x/y/z}, [[, source, $'...', &>
```

Thrilleratplay:
@tlaurion From everything I can determine, none.  This seems to only creates a symlink from `/bin/ash` to `/bin/bash`.  `CONFIG_ASH_BASH_COMPAT` is a different config, providing Bash like comparability in Ash, and is already enabled.  It is because `CONFIG_ASH_BASH_COMPAT` is enabled that `CONFIG_BASH_IS_ASH` has no impact.  Allowing scripts with shebangs `#!/bin/bash` to be used in initrd scripts and making shellcheck, which does not have ash compatibility, usable.

tlaurion:
@Thrilleratplay haha!

PR #877, opened 2020-10-29
WiP (NOT FUNCTIONAL): xx20 blobs extraction scripts fixes

tlaurion:
me_cleaner called to neuter, deactivate, trim ME and output reduced ME under blob dir, modify both ME and BIOS regions accordingly to be able to accept CONFIG_CBFS_SIZE=0x750000 defined under coreboot configs (attempt to fix #870)

PR #876, opened 2020-10-28
Upgrade to cryptsetup 2.3 and make cryptsetup1/cryptsetup2 optionals

root-hardenedvault:
continue the work of #749 :
.la files turn out to be unnecessary for linking. They may even hinder the linking sometimes. By removing them from `install`, the cryptsetup finally gets built and working.

PR #912, opened 2020-11-27
xx20 blobs fix

Thrilleratplay:
Replacement for PR #877 

* Added GBE blob
* Added IFD blob already modified for minified ME
* added `me7_update_parser.py` ([more information on what it does](https://github.com/Thrilleratplay/me7_update_parser)) script which creates a minified cleaned version of ME7 from the  xx02 ME update file.
* updates download script to automate download of ME update, extraction and processing.
* replaced consolidated t420 and x220 blob directories `under blobs\xx20`
* Added whiptail gui and hotp from x230 board config.
* Updated circleci build

I have externally flashed the artifact from my [cicleci build 18](https://app.circleci.com/pipelines/github/Thrilleratplay/heads/18/workflows/50db207d-b04d-4a0e-a0cf-681ea8bab0e8/jobs/18/parallel-runs/0) on a x220 and it boots to the Heads gui.


boards/t420/t420.config
```diff
@@ -1,11 +1,23 @@
-# Configuration for a T420 running Qubes and other OS, T420 is identical to X230 on the Linux Side of things.
+# Configuration for a T420 running Qubes and other Linux Based OSes (through kexec)
+#
+# Includes
+# - Deactivated+neutered ME and expended consequent IFD BIOS regions
+# - Forged 00:DE:AD:C0:FF:EE MAC address  (if not extracting gbe.bin from backup with blobs/xx30/extract.sh)
```
tlaurion, line 5: xx20

blobs/xx20/readme.md
```diff
@@ -0,0 +1,59 @@
...
+-rw-r--r-- 1 tom users  8192 Nov 23 18:40 gbe.bin
+-rw-r--r-- 1 tom users  4096 Nov 23 18:58 ifd.bin
+-rw-r--r-- 1 tom users 86016 Nov 26 17:04 me.bin
+
+Manually regenerate gbe.bin:
+blobs/x220/gbe.bin is generated per bincfg from the following coreboot patch: https://review.coreboot.org/c/coreboot/+/44510
+And then by following those instructions:
+# Use this target to generate GbE for X220/x230
+gen-gbe-82579LM:
+	cd build/coreboot-4.8.1/util/bincfg/
```
tlaurion, line 49: To reduce maintainership, don't fixate dir to coreboot 4.8.1

tlaurion (changes requested):
@Thrilleratplay small required changes

boards/t420/t420.config
```diff
@@ -35,3 +61,7 @@ export CONFIG_BOOT_KERNEL_REMOVE="quiet"
 export CONFIG_BOOT_DEV="/dev/sda1"
 export CONFIG_BOARD_NAME="ThinkPad T420"
 export CONFIG_FLASHROM_OPTIONS="--force --noverify-all -p internal:ich_spi_mode=hwseq --ifd --image bios"
```
tlaurion, line 63: We do not just write BIOS region with those boards, and that should come with a big fat warning with osresearch.net changes. We want to write the whole ROM internally as we do externally.

`--ifd --image bios` not desired

boards/x220/x220.config
```diff
@@ -35,3 +61,7 @@ export CONFIG_BOOT_KERNEL_REMOVE="quiet"
 export CONFIG_BOOT_DEV="/dev/sda1"
 export CONFIG_BOARD_NAME="ThinkPad X220"
 export CONFIG_FLASHROM_OPTIONS="--force --noverify-all -p internal:ich_spi_mode=hwseq --ifd --image bios"
```
tlaurion, line 63: Same here

We do not just write BIOS region with those boards, and that should come with a big fat warning with osresearch.net changes. We want to write the whole ROM internally as we do externally.

`--ifd --image bios` not desired

tlaurion (changes requested):
no more --ifd specified regions here. We want whole flash, both for external flash and internal flash. 

This is kind of the goal here :)
We also have to remind new comers that flashing internally this rom IS NOT  compatible with prior versions. (Which is why I named x230-external-flash board, expecting to remove x230 board altogether when the PR is approved. Not sure about this migration path, but I think we have to limit bricks of users who were using xx20 boards before. My thought process here was that if x230 disappears in board dir, that would be a clear, non-documentation needed path for builders to be forced to realize that board changed, where past board phased out. Thoughts welcome.)

boards/t420/t420.config
Thrilleratplay: Corrected.  Also corrected in x220.config

tlaurion:
Not sure about reusing xx20 previous boards definition. They should be phased out and replaced, so past builders will notice directly that they cannot `make BOARD=x220` anymore, and realize that a x220-external-flash exists, requiring him to flash externally (since we do not know their IFD state regions and should not imply that they are good). 

This is why I named x230-external-flash board, expecting to remove x230 board altogether when the PR is approved. Not sure about this migration path, but I think we have to limit bricks of users who were using xx20 boards before. My thought process here was that if x230 disappears in board dir, that would be a clear, non-documentation needed path for builders to be forced to realize that board changed, where past board phased out. Thoughts welcome.

blobs/xx20/readme.md
Thrilleratplay: This was copied from #703 ([here](https://github.com/osresearch/heads/pull/703/files#diff-2713f9bba9c6bd279025474d76958633751480c8655f7c2ac77471498e6f7ea4R51)). It is included as of coreboot-4.13, should it still be added as a general coreboot parch?

Thrilleratplay:
@tlaurion While I agree with most of the internal/external flashing comments, there seems to two different points being argued.  One where the current user's layout is preserved by creating boards `x220-external-flash`/`t420-external-flash` and one where it doesn't matter as the internal flash is updating all regions of the rom.

There does not seem to be a good answer.  Currently, the internal flash scripts update only the BIOS region.  By using the internal flash scripts with this rom as is, likely will result in an error.  Flashing externally, means they will lose their configuration and need to set it up again.

I am thinking a modified version of `flashrom.sh` that can be made that can be included on the same USB as this new version of Heads rom, that copying the PGP key into the new CBFS and flashes the entire rom.

Thrilleratplay:
@tlaurion Generally speaking for this PR and #703, I think it is best to consolidate the board variations into single models (if possible).  Instead of `x230-nkstorecli`, ` x230-hotp-verification `,  `x230-flash`, there is only `x230` which includes the nkstorecli and hotp options with the IFD, ME and GBE in one 12mb file.  An extra task to split it into 8mb and 4mb chunks, complete with sha256, will allow users to either build from scratch or download the artifacts then flash externally on each chip or internally with the full 12mb rom, all with one build.  For the x220/t420, it would be the same but there is no need split the file, there is only one 8mb chip.


Initially, this will be a headache as users will need to take special steps to update the entire rom, as the current `flash.sh` only updates the bios but can be mitigated with a update script (as noted in my previous comment).

boards/t420/t420.config
Thrilleratplay: Updated.

tlaurion:
@Thrilleratplay 
I see that you updated the flashrom commands. That is insufficient.

My comments were targeting the following.
1- User is having x220 board compiled from master.
2- User builds x220 board from the result of this merge.
3- User attempts to flash internally the new ROM. This means he attempts to only flash BIOS region, which is supposed to match CBFS coreboot defined space. Which is supposed to match IFD expended BIOS region, resulting of reduced ME region. This is not the case. Behavior?
4- The user will be able to flash internally, the whole flash, only after having already flashed externally. Brick guaranteed. where if x220 and x220-external-flash were seperated, and where x220 would be deprecated, and deleted, and replaced with x220-external-flash, we would do proper prevention.

I agree all those x230 boards should be merged inside of x230-external-flash. Where the scenario is more incertain for the xx20 boards. This is where specialization of boards started for the different x230 boards.

Once again:
- x230 actual board does not include HOTP support. Adding HOTP support requires, in code paths, to have a USB Security dongle supporting HOTP. This is not common to all use cases. x230-external-flash contains HOTP (USB Security dongle remote administration), NKSTORECLI (For new reownership mode i'm working on but not fast enough in my taste), dropbear, ethernet drivers etc since there is no space constraints for the moment.

I think the paths should be different and should differenciate those use cases too:
- We should not force HOTP based remote attestation based on USB Security dongle upon users. TPM based TPMTOTP is perfectly fine for most tech people not desiring to buy a USB Security dongle.

I just know that replacing the x220 doesn't seem the good path. 
For your script, its a chicken/egg problem. Since we do not know if user really unlocked ifd with ifdtool, nor neutered ME correctly (might have just deactivated it without neutering it) we cannot imply that IFD matches recuced ME region and expended BIOS region, nor that we can overwrite IFD, unless the first flash is external with such board produced ROM.

This is why I would phased out xx20 board for their `xx20-external-flash` counterparts, and would continue to have `-htop` counterparts, since they imply different use cases and code paths, not just size related issues nor prior external flash.

Let's chat live?

Thrilleratplay:
@tlaurion Ok, I think I understand what you are try to accomplish.  `xx20-external-flash` boards are intended to be used as externally flashed roms in addition to x220,  the same way `x230` and `x230-flash` are built together (I just noticed that in #703 that those are `x230-external-flash` and t`430-external-flash`).  It is a new board to distinguish it from the previous layout.  That makes sense if the other variants are being removed.  My only complaint is the names `x230-external-flash` and `x230-flash` are too similar, what about `xx30-external-flash` was something along the lines of `xx30-complete-rom`? The `xx20` boards can follow the same nomenclature.

As for HOTP, if it needs to be a separate board build, that is fine.  I assumed it wasn't included due to space restrictions and was being readded based on `CONFIG_HOTPKEY=y` being added to[to the xx30-external-flashboards](https://github.com/osresearch/heads/pull/703/files#diff-d7e056deb459d8cce5d5988269fea48446083e3b542f0e4e7a25ba254db4fe15R39).

tlaurion:
@Thrilleratplay 

If we move away of `x230-flash` and `x230` bundle, their counterparts could continue to be named `x230-external-flash` since there wouldn't be any nomenclature imprecision, since both boards (`x230-flash` and `x230-external-flash`) would be short lived.

The plan is to phase out two parts boards altogether, while enforcing clear distinction that one are not compatible with the other.

As for `-hotp` boards, they implement and bind the user to have bought another piece of hardware, which we do not want to render a prerequisite for Heads. It actually and simply replace TPMTOTP remote attestation on smarphone dependency into looking that your USB Security dongle is flashing green for the same result.

As for implementing a generic xx30 board, there are differences on the coreboot config level for the t430 and x230 since the t430 has integrated graphics + external GPU, and there was requests to enforce those differences with OPTIONAL_ROM support if I recall well. Once again, since I do not own those boards, its hard for me to tell if they should be joined or not. I just recall that I was against merging those changes inside of the x230 since they were not relevant.

So my recommendation stays

tlaurion:
( When we agree on nomenclature, I will modify xx30 PR accordingly. Same logic applying to all xx20 and xx30, and will even try to coordinate merging of those two pull requests.)

Considering this, I would not mv board files. But would create another PR to deprecate them soon after. Before integrating #876 or attemtping #562 inclusion and debugging, which would better work on top of this anyway.

blobs/xx20/readme.md
tlaurion: Hmmm. I don't think so if next version for xx30 and xx20 are 4.13!

tlaurion:
@merge Lazy here sorry to poke you. But in Skulls instructions, do you FORCE users to unlock IFD? Could a simple instruction be pushed to current users that if they intend to flash x230-external-flash, t430-external-flash, x220-external-flash t420-external-flash and all other future coming sandy / ivy bridge based upraders have to flash with a manual command from command line not only flashing BIOS region but whole flash?

tlaurion:
Note to myself, before merging both xx20 and xx30 branches, ethernet bringup script should randomize MAC address prior of trying to enable link.

tlaurion:
@Thrilleratplay hmmm your PR gots corrupted. bad rebase upon master?

tlaurion:
[Also](https://circleci.com/gh/Thrilleratplay/heads/22?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)
```
#!/bin/bash -eo pipefail
rm -rf build/x220-external-flash/* build/log/* && make CPUS=4 V=1 BOARD=x220-external-flash || touch /tmp/failed_build

2020-11-28 04:12:45+00:00 Wrong gawk detected: 
Makefile:109: *** /root/project/boards/x220-external-flash/x220-external-flash.config: board configuration does not exist.  Stop.

CircleCI received exit code 0
```

Thrilleratplay:
@tlaurion Huh.  Not sure how that happened but rebased against heads/master

tlaurion:
> @merge Lazy here sorry to poke you. But in Skulls instructions, do you FORCE users to unlock IFD? Could a simple instruction be pushed to current users that if they intend to flash x230-external-flash, t430-external-flash, x220-external-flash t420-external-flash and all other future coming sandy / ivy bridge based upraders have to flash with a manual command from command line not only flashing BIOS region but whole flash?

@n4ru Same question would apply to your users coming from 1vyrain. The goal here is to deprecate x230 and x220 (all xx30 and xx20 boards) which relies on unmodified IFD which specifies BIOS region that is too minimalist. As we discussed precedently in other 1vyrain tickets, there is no possibility to trick environment to flash modified IFD and ME regions with neutered+deactivated+reduced ME space, have ifd modified accordinly so that BIOS region is maximized.

What would be your thoughts and migration paths? New boards on Heads will arrive soon, requiring unlocked ifd to be able to wholly internally flash rom upgrades, also from fwupd. 

@merge @n4ru: Will you issue a big fat warning to users that outside reprogramming is necessary for users to flash the xx20-maximized and xx30-maximized board produced roms?

merge:
> @merge Lazy here sorry to poke you. But in Skulls instructions, do you FORCE users to unlock IFD? Could a simple instruction be pushed to current users that if they intend to flash x230-external-flash, t430-external-flash, x220-external-flash t420-external-flash and all other future coming sandy / ivy bridge based upraders have to flash with a manual command from command line not only flashing BIOS region but whole flash?

no the skulls "external" install scripts don't force IFD unlocking. They have the `-u` switch for that.

tlaurion:
@Thrilleratplay last rebase requested, #703 was merged!

Thrilleratplay:
@tlaurion  Rebased.

boards/t420-maximized/t420-maximized.config
```diff
@@ -0,0 +1,67 @@
...
+#SSH server (requires ethernet drivers, eg: CONFIG_LINUX_E1000E)
+CONFIG_DROPBEAR=y
+
+export CONFIG_BOOTSCRIPT=/bin/gui-init
+export CONFIG_BOOT_REQ_HASH=n
+export CONFIG_BOOT_REQ_ROLLBACK=n
+export CONFIG_BOOT_KERNEL_ADD="intel_iommu=on intel_iommu=igfx_off"
+export CONFIG_BOOT_KERNEL_REMOVE="quiet"
+export CONFIG_BOOT_DEV="/dev/sda1"
+export CONFIG_BOARD_NAME="ThinkPad T420"
```
tlaurion, line 62: T420-maximized

boards/x220-maximized/x220-maximized.config
```diff
@@ -0,0 +1,67 @@
...
+#SSH server (requires ethernet drivers, eg: CONFIG_LINUX_E1000E)
+CONFIG_DROPBEAR=y
+
+export CONFIG_BOOTSCRIPT=/bin/gui-init
+export CONFIG_BOOT_REQ_HASH=n
+export CONFIG_BOOT_REQ_ROLLBACK=n
+export CONFIG_BOOT_KERNEL_ADD="intel_iommu=on intel_iommu=igfx_off"
+export CONFIG_BOOT_KERNEL_REMOVE="quiet"
+export CONFIG_BOOT_DEV="/dev/sda1"
+export CONFIG_BOARD_NAME="ThinkPad X220"
```
tlaurion, line 62: X220-maximized

tlaurion (changes requested):

boards/t420-maximized/t420-maximized.config
Thrilleratplay: Updated.

boards/x220-maximized/x220-maximized.config
Thrilleratplay: Updated.

tlaurion:
LGTM (while commits could be squashed with a bit more detail in each commit, basically on your script and what it does!)
We would be ready to merge :)

Thrilleratplay:
@tlaurion Squashed.

PR #913, opened 2020-11-28
Remove x220 t420 boards ( replaced with their xx20-external-flash equivalents (needing external flash or unlocked IFD to flash internally )

Thrilleratplay:
Removes X220 and T420 boards.

Wait until #912 has been merged.


tlaurion:
Community members have refused removal. We will (hopefully) see xx20/xx30-minimal boards emerge when xx20 and xx30 boards don't produce roms anynore, and guide users to move to their maximized counterparts when the time comes. As a personal statement, I won't maintain xx20 and xx30 boards. New blood needs to make the changes and propse changes. My path is to use freeable space and I invite everyone else to do the same.

paulmenzel:
I only now saw the long merge/pull request description. It’d be great if you added a summary of the goal of the merge/pull request to the description. Also, as an outsider, it’s unclear to me, why such a consolidation would be bad. What are the problems?

flammit:
@tlaurion Please don't misrepresent my position.  My objection to deprecating the x230 board which is a useful configuration as it stands.  I didn't make statements about x220/t420.

tlaurion:
@flammit : I don't really understand your position to be franck. The x220 board has around 3.5mb of BIOS space in the 8m SPI flash chip. For the xx20 boards to survive, scripts were made to reduce de facto ME. So if I understand your point correctly, your opposition comes solely on unlocking the IFD. But what is proposed for xx20 and xx30 boards, as opposed to newer boards, is ME with only ROMP and BUP modules. I would love, considering this, to understand why you oppose on having an unlocked IFD if ROM is measured and ME is kinda immuable, since the laptop will refuse to boot if those are modified and doesn't match Intel signature. All other modules, are removed.

PR #923, opened 2020-12-09
WIP: xx30-*-maximized: Add MMC sdcard support, bump kernel to 4.19.162 LTS.

tlaurion:
This PoC includes MMC kernel support (sdcard slot) support to cover Realtek and Ricoh card reader (Yes. They changed from Ricoh to Realtek card reader somewhere in 2012.)

Other PRs, depending on this one, will follow to use those functionalities in the goal of [attacking Heads GPG keygen and keytocard, with backup on encrypted storage](https://github.com/osresearch/heads/issues/771#issuecomment-740850208).

- Kernel is bumped to latest LTS version because Realtek+Ricoh was malfunctioning/not avail in 4.14. 
- `CONFIG_LINUX_MMC` [board config option was added](https://github.com/osresearch/heads/compare/master...tlaurion:xx30-maximized-linux-4.19.162_MMC_support?expand=1#diff-216178cbb68b280411057b6686d755a1b2b3ca356e395daf99e723ae987d5ebbR20) (will be extended to other boards upon review and successful tests)
- modules/linux contains [conditional statements of inclusion of the added MMC modules](https://github.com/osresearch/heads/compare/master...tlaurion:xx30-maximized-linux-4.19.162_MMC_support?expand=1#diff-2b51dc9d1602ab31c054d57804a441059a68dcfdf7f3c407f83a85148bcecfc6R108-R120)

Changes in compressed payload size is:
`7438884 - 7192516 = 246368` increase in size.

[master 671522e commit build](https://app.circleci.com/pipelines/github/osresearch/heads/260/workflows/1912855b-2150-4986-82fe-498be1700e80/jobs/275/parallel-runs/0/steps/0-199)
```
Built lenovo/x230 (ThinkPad X230)
make[1]: Leaving directory '/root/project/build/coreboot-4.8.1'
2020-12-03 22:16:20+00:00 DONE coreboot
touch /root/project/build/coreboot-4.8.1/x230-hotp-maximized/.build 
"/root/project/build/coreboot-4.8.1/x230-hotp-maximized/cbfstool" "/root/project/build/coreboot-4.8.1/x230-hotp-maximized/coreboot.rom" print
Name                           Offset     Type           Size   Comp
cbfs master header             0x0        cbfs header        32 none
fallback/romstage              0x80       stage           84708 none
cpu_microcode_blob.bin         0x14c00    microcode       25600 none
fallback/ramstage              0x1b080    stage           81118 none
config                         0x2edc0    raw               768 none
[...]

PR #924, opened 2020-12-09
xx30-*-maximized linux 4.19.162 mmc support: add external storage support

tlaurion:
- Based on #923 (Which adds MMC kernel support and board configure option, bumps kernel to LTS and add modules inclusion)
- Add support into mount-usb to be able to deal with mmc device, verifying that MMC support is present prior of checking for external storage device

@MrChromebox : waiting for your approving into renaming mount-usb to mount-external-storage and add changes in all codebase.


tlaurion:
Added LUKS support @MrChromebox . If you want me to split this up in another PR, no problem.

MrChromebox:
@tlaurion not an issue right now, but eventually we might end up with a device that uses internal eMMC and need to distinguish between internal/fixed and external/removable media. I'll do a more thorough review tomorrow

tlaurion:
@MrChromebox ping!

MrChromebox:
@tlaurion would it not make sense to have a 3rd commit in this PR that renames mount-usb to mount-external (and adjusts usages accordingly)?

tlaurion:
@MrChromebox last commit moves mount-usb to mount-external-storage, and changes all references in codebase to point to it.

Issue #789, opened 2020-08-01
QubesOS 4.1 installer doesn't boot on sandy/ivy bridge (xx20/xx30)


tlaurion:
Booting 
https://openqa.qubes-os.org/tests/10816/asset/iso/Qubes-4.1-20200726-x86_64.iso results in this even though https://github.com/osresearch/heads/commit/116787fd03436909f2fe81032895f1d70098a60a was merged in : 

![signal-2020-08-01-154642](https://user-images.githubusercontent.com/827570/89109344-593caa00-d40e-11ea-97e0-162148819837.jpeg)

@marmarek any idea?

_Originally posted by @tlaurion in https://github.com/osresearch/heads/pull/788#issuecomment-667547990_

tlaurion:
Trying to find relevant information and will edit this comment (@marmarek)
- [no-real-mode disappeared](https://github.com/QubesOS/qubes-issues/issues/5529#issuecomment-568888958)?

tlaurion:
@marmarek: `no-real-mode` [gone](https://github.com/QubesOS/qubes-issues/issues/5529#issuecomment-568889327) under xen?

osresearch:
I just downloaded the Qubes 4.1 ISO and it appears to be using Xen 4.13.  Grepping through the 4.13 source it appears that `no-real-mode` should still be valid in the early command line parser, although the main parser doesn't seem to have it:

```
static bool skip_realmode(const char *cmdline)
{
    return find_opt(cmdline, "no-real-mode", false) || find_opt(cmdline, "tboot=", true);
}
```

which is used here:

```
void __stdcall cmdline_parse_early(const char *cmdline, early_boot_opts_t *ebo)
{
    if ( !cmdline )
        return;

    ebo->skip_realmode = skip_realmode(cmdline);
    ebo->opt_edd = edd_parse(cmdline);
    ebo->opt_edid = edid_parse(cmdline);

#ifdef CONFIG_VIDEO
    vga_parse(cmdline, ebo);
#endif
}
```

This parser has moved from `arch/x86/boot/cmdline.S` in Xen 4.8 to `cmdline.c`, although it is invoked in a funny way.

The Xen code in 4.13 that does the trampoline setup has changed, so I wonder if we can retry with the ancient patch that prevents the EBDA from being consulted.  From my 2016 xen-devel post:

```
diff --recursive -u /home/hudson/build/clean/xen-4.6.3/xen/arch/x86/boot/head.S 
[...]

PR #928, opened 2020-12-12
(WiP) xx20 boards: add xx20-hotp-maximized & remove HOTP and NITROCLI support from xxx0-maximized boards

tlaurion:
- xx20-maximized boards were supporting HOTP by error. This adds the xx20-hotp-maximized boards and removes HOTP support from xx20-maximized boards for coherence with the xx30-maximized boards.
- Add those boards to CircleCI.

@Thrilleratplay should be that originally I guess, where you can't test hotp right?

Also unsure if NKSTORAGECLI should be here, since they needed to be removed under https://github.com/tlaurion/heads/commit/00c2ffffade493e17edf6a93b1f6970c01d23719 for https://app.circleci.com/pipelines/github/tlaurion/heads/648/workflows/337f0829-e3e9-47aa-85b1-1e10a9199e54/jobs/696 to build (cryptsetup2 support for future #789 support since LUKS2 at default install)

~~Doesn't build as of right now, since boards configs declare cryptsetup2 dependency.~~ EDIT: removed dependency to cryptsetup2. Should not be linked issues.


Thrilleratplay:
@tlaurion I do not have a Nitrokey/LibremKey so I was not able to test this.  I only included them to match xx30 boards.

tlaurion:
@Thrilleratplay which is why I seperated this into xx30 as well. Now xx20-maximized builds do not contain HOTP (which you can test) and xx20-htop which you can't. Merge?

tlaurion:
As a side note, I would go forward and remote NKSTORECLI inside of those board configs also, reducing future maintenance of those boards (nobody wants NKSTORECLI by default). I will go forward and we wil lbe able to use those two builds to compare size difference.

That would close https://github.com/osresearch/heads/issues/926#issuecomment-743793731 with one stone.

Thrilleratplay:
@tlaurion I understand that but `CONFIG_HOTPKEY=y` was taken from [x230-maximized](https://github.com/osresearch/heads/blob/master/boards/x230-maximized/x230-maximized.config#L39). If it is being removed from the xx20-maximized boards, they should also removed from the xx30-maximized boards.  I assume would be the same for `CONFIG_NKSTORECLI=y` in xx30-maximized and xx20-maximixed boards.

If doing so allows for enough room for cryptsetup2 to be installed in the xx20-maximized boards and does not take away any functionality, then I would suggest doing so for the xx30-maximized boards too.  This will keep them functionally the same for now.

tlaurion:
Agreed. pushing. Will create a x230-PrivacyBeast for my own needs adding this in I guess instead of using the xx30-htop-maximized.

PR #943, opened 2020-12-30
Kgpe d16 flashrom fix

Tonux599:
Allows KGPE-D16 users to flash the BMC internally.

@MrChromebox please review also as makes modifications to your work. As laid out in the commit messages:

> * Flashrom was being fetched with git and was always using `master`
> * No patches were being applied (i.e. `0100-enable-kgpe-d16.patch` was being ignored).

### DIFF
--- a/modules/flashrom
+++ b/modules/flashrom
@@ -2,14 +2,11 @@ modules-$(CONFIG_FLASHROM) += flashrom
 
 flashrom_depends := pciutils $(musl_dep)
 
-flashrom_version := git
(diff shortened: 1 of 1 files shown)


MrChromebox:
would want to CP/test, but LGTM at a glance

tlaurion:
@MrChromebox any regression on your side?

MrChromebox:
@tlaurion will test tomorrow and approve once done

MrChromebox (approved):

PR #946, opened 2021-01-01
(WIP) (KGPE-D16) Bump to Linux 5.4.69

Tonux599:
**Note: Not currently functioning as intended**

This bumps Linux to 5.4.69 for the KGPE-D16 which should generally bring better graphics support.

However there appears to be an issue in that when coreboot loads the latest fam15 microcode, Linux panics. The symptoms are outlined here also: [https://lkml.org/lkml/2019/1/3/540](https://lkml.org/lkml/2019/1/3/540) and after a brief discussion with @paulmenzel the possible solutions appear to be:
1. Ensure that the user has both types of coloured memory slots populated. _I prefer to avoid this as it alienates users_
2. Somehow pin an older, working microcode update to be include in the CBFS. Users can then have newer microcode loaded by Linux later if desired.
3. Include no microcode in the CBFS. _My concern with this is if the users would have enough stability in the heads environment given that [6300 series CPUs have buggy microcode built-in](https://libreboot.org/docs/hardware/kgpe-d16.html#processor-system-bus)_
4. Include the microcode update in initramfs and be loaded by Linux.
5. Reach out to developers on mailing lists to see why when identical microcode is loaded by coreboot, kernel panic occurs as opposed to when it is loaded by Linux and if there is any resolution to this. (@miczyg1 @pietrushnic ?)


Tonux599:
Currently modelling option 4. Messy as hell. Haven't tested yet. If builds and runs consider POC. Personal preferred option is 5.

tlaurion:
Needs https://github.com/osresearch/heads/pull/954 related changes

Tonux599:
> Needs #954 related changes

@tlaurion I wouldn't bother doing it immediately, when #954 is merged I can rebase which will implement it anyway. Lemme know your opinion of the OP and post above!

tlaurion:
Agreed with option #3

tlaurion:
@Tonux599 If I understand correctly, this PR will change removing ucode state in Makefile and basically upgrade kernel.
If that is, then I would select the latest LTS version including upgrade graphical support. This is basically it under Heads. Not really a threat vector per se, since going into recovery console should not be trusted and should invalidate measurements (Disk Unlock Key released by TPM for the win, typed by BMC or on local PS/2 keyboard for better security, where USB keyboard users should be aware of added risk...)

tlaurion:
@Tonux599 : maybe adding a note in the board configs about microcode upgrade being left ot the responsibility of the OS?

Tonux599:
> Needs #954 related changes

Rebased and done.
> @Tonux599 : maybe adding a note in the board configs about microcode upgrade being left ot the responsibility of the OS?

Done.

So now modelling option 3, build is happening and will point to test roms soon.

Tonux599:
Superseded by #957

PR #947, opened 2021-01-01
Makefile: prefetch crossgcc tarballs (ftpmirror.gnu.org failing downloads on coreboot builds)

Tonux599:
Ever look at your CircleCI build for about 2 hours and then the whole thing fails because it failed to fetch some tarballs for a coreboot crossgcc version that hasn't been build yet? Hopefully this will help that. Fetches the tarballs required for coreboot's crossgcc for version 4.8.1, 4.11 and 4.13 (these appear to be the ones currently in use) early on.

Possible extra feature that I haven't implemented yet is checking the checksums early too but as it stands this should help a bit.

PR #950, opened 2021-01-03
CircleCI: Add coreboot+musl-cross cache

tlaurion:
The idea here is a cache to restore from (musl-cross from coreboot version bound crosscomipler, from which coreboot is built)

1- Reuse existing cache for all modules and patches created digest's hash. (If a single module or patch changes, we have cache miss)
2- Reuse existing coreboot and musl-cross-make created digest's hash (If a patch was added to current coreboot, or new coreboot version was added in coreboot module definition, we have a cache miss. Otherwise, we reuse the whole musl-cross-make and coreboot cache for current used versions.)
3- Reuse existing musl-cross-make created digest's hash matching cache (If musl-cross-make module didn't change, we don't rebuild it)

Per https://github.com/osresearch/heads/pull/947#issuecomment-753507412 proposition
@Tonux599 please challenge me!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,23 +17,32 @@ jobs:
             git reset --hard "$CIRCLE_SHA1" \
 
       - run:
-          name: Creating all modules and patches digest
+          name: Creating all modules and patches digest (All modules cache digest)
           command: |
             find ./patches/ ./modules/ -type f | sort -h |xargs sha256sum > /tmp/all_modules_and_patches.sha256sums \
 
       - run:
-          name: Creating musl-cross-make and musl-cross-make patches digest
+          name: Creating coreboot (and associated patches) and musl-cross-make modules digest (musl-cross-make and coreboot cache digest)
           command: |
-            find modules/musl-cross* -type f | sort -h | xargs sha256sum > /tmp/musl-cross_module_and_patches.sha256sums \
+            find ./modules/coreboot ./modules/musl-cross* ./patches/coreboot* -type f | sort -h | xargs sha256sum > /tmp/coreboot_musl-cross.sha256sums \
+
+      - run:
+          name: Creating musl-cross-make and musl-cross-make patches digest (musl-cross-make cache digest)
+          command: |
+            find modules/musl-cross* -type f | sort -h | xargs sha256sum > /tmp/musl-cross.sha256sums \
 
 
       - restore_cache:
           keys:
-            #Restore existing cache for modules checksums validated to be exactly the same as in github current commit
+            #Restore existing cache for matching modules digest, validated to be exactly the same as in github current commit.
+            #This cache was made on top of below caches, if previously existing. If no module definition changed, we reuse this one. Otherwise...
             - heads-modules-and-patches-{{ checksum "/tmp/all_modules_and_patches.sha256sums" }}{{ .Environment.CACHE_VERSION }}
-            #If precedent fails. Restore cache for musl-cross module checksum validated to be exactly the same as in github current commit
-            - heads-cross-musl-{{ checksum "/tmp/musl-cross_module_and_patches.sha256sums" }}{{ .Environment.CACHE_VERSION }}
-
+            #If precedent cache not found, restore cache for coreboot module (and patches) and musl-cross-make digests (bi-yearly modified)
+            #Otehrwise....
+            - heads-coreboot-musl-cross-{{ checksum "/tmp/coreboot_musl-cross.sha256sums" }}{{ .Environment.CACHE_VERSION }}
+            #If precedent cache not found. Restore cache for musl-cross-make module digest (rarely modified).
+            #Otherwise, we build cleanly.
+            - heads-musl-cross-{{ checksum "/tmp/musl-cross.sha256sums" }}{{ .Environment.CACHE_VERSION }}
 # linuxboot steps need something to pass in the kernel header path
 # skipping for now
 #      - run:
@@ -578,11 +587,17 @@ jobs:
 
       - save_cache:
           #Generate cache for the same musl-cross module definition if hash is not previously existing
-          key: heads-cross-musl-{{ checksum "/tmp/musl-cross_module_and_patches.sha256sums" }}{{ .Environment.CACHE_VERSION }}
+          key: heads-musl-cross-{{ checksum "/tmp/musl-cross.sha256sums" }}{{ .Environment.CACHE_VERSION }}
           paths:
             - crossgcc
             - build/musl-cross-*
-
+      - save_cache:
+          #Generate cache for the same coreboot mnd musl-cross-make modules definition if hash is not previously existing
+          key: heads-coreboot-musl-cross-{{ checksum "/tmp/coreboot_musl-cross.sha256sums" }}{{ .Environment.CACHE_VERSION }}
+          paths:
+            - build/coreboot-*
+            - crossgcc
+            - build/musl-cross-*
       - save_cache:
           #Generate cache for the exact same modules definitions if hash is not previously existing
           key: heads-modules-and-patches-{{ checksum "/tmp/all_modules_and_patches.sha256sums" }}{{ .Environment.CACHE_VERSION }}
@@ -590,6 +605,7 @@ jobs:
             - packages
             - crossgcc
             - build
+            - install
 
 workflows:
   version: 2
--- a/modules/coreboot
+++ b/modules/coreboot
@@ -72,7 +72,7 @@ coreboot_depend += linux initrd $(musl_dep)
 
 COREBOOT_TOOLCHAIN="$(build)/$(coreboot_base_dir)/.xcompile"
 $(COREBOOT_TOOLCHAIN): $(build)/$(coreboot_base_dir)/.canary
-	$(MAKE) -C "$(build)/$(coreboot_base_dir)" CPUS=$$CPUS crossgcc-i386
+	$(MAKE) -C "$(build)/$(coreboot_base_dir)" CPUS=$(CPUS) crossgcc-i386
 
 $(build)/$(coreboot_dir)/.configured: $(COREBOOT_IASL) $(COREBOOT_TOOLCHAIN)
 $(COREBOOT_IASL): $(build)/$(coreboot_base_dir)/.canary
--- a/patches/coreboot-4.11/0080-util-crossgcc-builgcc_temp_fix_from_https_to_http.patch
+++ b/patches/coreboot-4.11/0080-util-crossgcc-builgcc_temp_fix_from_https_to_http.patch
@@ -1,0 +1,42 @@
+--- a/util/crossgcc/buildgcc.orig	2021-01-03 19:58:54.084000000 -0500
++++ b/util/crossgcc/buildgcc	2021-01-03 19:58:56.985000000 -0500
+@@ -67,23 +67,23 @@
+ # These are sanitized by the jenkins toolchain test builder, so if
+ # a completely new URL is added here, it probably needs to be added
+ # to the jenkins build as well, or the builder won't download it.
+-GMP_ARCHIVE="https://ftpmirror.gnu.org/gmp/gmp-${GMP_VERSION}.tar.xz"
+-MPFR_ARCHIVE="https://ftpmirror.gnu.org/mpfr/mpfr-${MPFR_VERSION}.tar.xz"
+-MPC_ARCHIVE="https://ftpmirror.gnu.org/mpc/mpc-${MPC_VERSION}.tar.gz"
+-GCC_ARCHIVE="https://ftpmirror.gnu.org/gcc/gcc-${GCC_VERSION}/gcc-${GCC_VERSION}.tar.xz"
+-BINUTILS_ARCHIVE="https://ftpmirror.gnu.org/binutils/binutils-${BINUTILS_VERSION}.tar.xz"
+-GDB_ARCHIVE="https://ftpmirror.gnu.org/gdb/gdb-${GDB_VERSION}.tar.xz"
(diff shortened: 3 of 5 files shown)


Tonux599:
@tlaurion I think this is a more concise implementation and am happy for it to replace #947.

I don't know how I feel about caching the whole of `build/coreboot-*`, personally I would prefer just to cache `build/coreboot-*/util/crossgcc` but only because its relatively unchanging.

tlaurion:
@Tonux599 : Well, building coreboot (each coreboot version), with 4 cores, takes around 1.5 hours each per past CircleCI runs.
Having it in cache permits to economize a lot of it, if not all.

Note the all modules cache actually caches `build` altogether, including coreboot, musl-cross and basically everything else.
In this PR, I go even farther, and cache `install`.... While the current gnu mirror issue is stopping us from doing a test out of it.

Interesting fact, since having reverted coreboot to be built on coreboot's provided muslt-cross (by reverting patch to build against Heads' musl-cross-make), each board build is attemtping to redownload all tarballs, and if found, rebuilds the cache, which is speeded up because binaries are already built. This is because .xcompile is deleted upon each rebuilt. I do not quite know how to bypass this.

Tonux599:
> Note the all modules cache actually caches `build` altogether, including coreboot, musl-cross and basically everything else.
> In this PR, I go even farther, and cache `install`.... While the current gnu mirror issue is stopping us from doing a test out of it.

I don't think it is though? It's only downloading the tarballs for each coreboot version. For example the tarballs are downloaded for `kgpe-d16_workstation` (coreboot 4.11) then used for for the other kgpe-d16 boards. Then `librem_mini` (coreboot 4.13) then for `x220-hotp-maximized` (coreboot 4.8.1) which is used for the rest.

It takes maybe 1.5 hours to build crossgcc, but far less once that is build. We would hopefully be looking for at a few minutes each if the whole of crossgcc was cached.

Tonux599:
Again maybe we are thinking too big just for the moment. The immediate problem is not being able to connect to gnu.org and that wasting a lot of time in CI. We could cache only `build/coreboot-*/util/crossgcc/tarballs` and fix that problem at least.

tlaurion:
> > Interesting fact, since having reverted coreboot to be built on coreboot's provided muslt-cross (by reverting patch to build against Heads' musl-cross-make), each board build is attemtping to redownload all tarballs, and if found, rebuilds the cache, which is speeded up because binaries are already built. This is because .xcompile is deleted upon each rebuilt. I do not quite know how to bypass this.
> 
> I don't think it is though? It's only downloading the tarballs for each coreboot version. For example the tarballs are downloaded for `kgpe-d16_workstation` (coreboot 4.11) then used for for the other kgpe-d16 boards. Then `librem_mini` (coreboot 4.13) then for `x220-hotp-maximized` (coreboot 4.8.1) which is used for the rest.
> 
> EDIT: wrong quote

@Tonux599 sorry. Was not clear.
Each coreboot board built _tries_ to download (wget) tarballs if not present (`build/coreboot-*/util/crossgcc/tarballs/`) else verifies hashes per build/coreboot-*/util/crossgcc/buildgcc script. This is why caching is interesting, with additional build cache present.
As you pointed out, kgpe-d16 is 4.11, after which librem_mini with its coreboot version etc etc



> Again maybe we are thinking too big just for the moment. The immediate problem is not being able to connect to gnu.org and that wasting a lot of time in CI. We could cache only `build/coreboot-*/util/crossgcc/tarballs` and fix that problem at least.

unfortunately, this approach, as opposed as you past one, will fail when switching coreboot version per board builds, and will fail after having lost precious build time, without caching successful builds and directories.

This does not fix the issue of gnu mirror at all, but would prevent mirror problems, in the future, to impact the CI once a full build is successful. For you original solution, I just didn't approve because it was requiring outside of tree maintennce of gatehring filenames and such, where I pointed out that calling `buildgcc -u` for each coreboot version build dependencies and looping until successfull could be a fix for current gnu mirror problem. I just do not want to add additional maintainership overhead on coreboot version release.

tlaurion:
> @tlaurion I think this is a more concise implementation and am happy for it to replace #947.
> 
> I don't know how I feel about caching the whole of `build/coreboot-*`, personally I would prefer just to cache `build/coreboot-*/util/crossgcc` but only because its relatively unchanging.

@Tonux599 those changes only per coreboot release, per coreboot definition. So if bundled in digest with coreboot patches, we insure that that cache life is extended to its maximum to happen only when 1- a new coreboot version is added 2-when patches are added in heads because patches were made against upstream coreboot. This should be relatively stable, while each time those happens, we have a risk that mirrors fail and loose build time.

So #947 is still desirable. Added insight there, where I have no concrete implementation to fix current mirror issue. This PR here is more for a long term objective in reaction of that. We do not want to redowwload and rebuild coreboot if other Heads modules changed. Which is why I splitted the cache system in 3 parts instead of 2.

tlaurion:
@Tonux599 manual tests showed that some of the mirrors are failing because of bad certs.
Example, when landing on https://gnu.mirror.constant.com/mpfr/mpfr-4.0.2.tar.xz

https://github.com/osresearch/heads/pull/950/commits/98cc8984b8f90db38c7b666d752c1fbd50ec37ef patches all active coreboot versions to fetch through http instead of https. Hopefully, the build will succeed. But unfortunately, the digest hash will be different once we remove those patches. So hopefully, GNU will react and nuke those bad mirrors until they comply. Meahwile, this is what I've got, where hashes are still validated prior of decompression, let it be http or https.... So this seems good enough for me. Better to have something that works then something that doesnt.

tlaurion:
@Tonux599 @MrChromebox : 
1- Cache layer was added to be middle ground between musl-cross-make cache being reused all the time when no other cache can be extracted since a module/patch changed/was added, changing the digest hash. Now coreboot cache is being reused until coreboot patches are added/new versions are introduced into coreboot module, while all modules/patches is being made on top of previous caches and reused for script changes under Heads (no module was changed)
2- A hack was introduced in coreboot patches to change download URLs of coreboot tarballs over http, since GNU mirrors are redirecting to some bad https websites with invalid certs. That is the best I came with.
3- A quick fix was added into coreboot module for consistency $$CPUS -> $(CPUS) to correct the change so that the coreboot cache layer can be rebuilt and reused asap in master for all other commits.

tlaurion:
Successful initial build: https://app.circleci.com/pipelines/github/tlaurion/heads/682/workflows/b8bb7fe7-214f-4313-987c-6ac8e50ec375/jobs/750
Rebuild to see rebuild time: https://app.circleci.com/pipelines/github/tlaurion/heads/682/workflows/01c917bf-dab4-4ebf-a1e5-656564e053ec/jobs/752

tlaurion:
Nevermind, the gnu problem is still a thing and builds are still randomly failing: https://app.circleci.com/pipelines/github/tlaurion/heads/676/workflows/7f47ad53-3ef0-4baa-b9e6-5ee99bc41905/jobs/754

@MrChromebox : if you give your go, I merge.

Tonux599:
@tlaurion I am a little confused to why in that CI build, under Restoring Cache it does not do so for `./build/coreboot-*`?

tlaurion:
@Tonux599 for https://app.circleci.com/pipelines/github/tlaurion/heads/682/workflows/01c917bf-dab4-4ebf-a1e5-656564e053ec/jobs/752 ? 

It [restored](https://app.circleci.com/pipelines/github/tlaurion/heads/682/workflows/01c917bf-dab4-4ebf-a1e5-656564e053ec/jobs/752/parallel-runs/0/steps/0-107) the first restorable cache in line, which is the all modules and all patch digest matching cache.

Let me explain again:
1- [Digests](https://github.com/osresearch/heads/blob/7d2ba3d0b82030cf90bb3a897c0ef311903324f6/.circleci/config.yml#L19-L32) are created based on files hashes
```
      - run:
          name: Creating all modules and patches digest (All modules cache digest)
          command: |
            find ./patches/ ./modules/ -type f | sort -h |xargs sha256sum > /tmp/all_modules_and_patches.sha256sums \
      - run:
          name: Creating coreboot (and associated patches) and musl-cross-make modules digest (musl-cross-make and coreboot cache digest)
          command: |
            find ./modules/coreboot ./modules/musl-cross* ./patches/coreboot* -type f | sort -h | xargs sha256sum > /tmp/coreboot_musl-cross.sha256sums \
      - run:
          name: Creating musl-cross-make and musl-cross-make patches digest (musl-cross-make cache digest)
          command: |
            find modules/musl-cross* -type f | sort -h | xargs sha256sum > /tmp/musl-cross.sha256sums \
```
2- Prior of starting build, [proper cache is restored, in order of declaration (first match)](https://github.com/osresearch/heads/blob/7d2ba3d0b82030cf90bb3a897c0ef311903324f6/.circleci/config.yml#L35-L45). Here we want the most complete cache to be restored if it exists. The cache hash is caculated on the created digest at step 1 above:
```
      - restore_cache:
          keys:
            #Restore existing cache for matching modules digest, validated to be exactly the same as in github current commit.
            #This cache was made on top of below caches, if previously existing. If no module definition changed, we reuse this one. Otherwise...
            - heads-modules-and-patches-{{ checksum "/tmp/all_modules_and_patches.sha256sums" }}{{ .Environment.CACHE_VERSION }}
            #If precedent cache not found, restore cache for coreboot module (and patches) and musl-cross-make digests (bi-yearly modified)
            #Otehrwise....
            - heads-coreboot-musl-cross-{{ checksum "/tmp/coreboot_musl-cross.sha256sums" }}{{ .Environment.CACHE_VERSION }}
            #If precedent cache not found. Restore cache for musl-cross-make module digest (rarely modified).
            #Otherwise, we build cleanly.
            - heads-musl-cross-{{ checksum "/tmp/musl-cross.sha256sums" }}{{ .Environment.CACHE_VERSION }}
```
3- At the end of a CI run, the [caches are created](https://github.com/osresearch/heads/blob/7d2ba3d0b82030cf90bb3a897c0ef311903324f6/.circleci/config.yml#L588-L608) only if non-existing for a digest hash (unique, create once, cache):
```
      - save_cache:
          #Generate cache for the same musl-cross module definition if hash is not previously existing
          key: heads-musl-cross-{{ checksum "/tmp/musl-cross.sha256sums" }}{{ .Environment.CACHE_VERSION }}
          paths:
            - crossgcc
            - build/musl-cross-*
      - save_cache:
          #Generate cache for the same coreboot mnd musl-cross-make modules definition if hash is not previously existing
          key: heads-coreboot-musl-cross-{{ checksum "/tmp/coreboot_musl-cross.sha256sums" }}{{ .Environment.CACHE_VERSION }}
          paths:
            - build/coreboot-*
            - crossgcc
            - build/musl-cross-*
      - save_cache:
          #Generate cache for the exact same modules definitions if hash is not previously existing
          key: heads-modules-and-patches-{{ checksum "/tmp/all_modules_and_patches.sha256sums" }}{{ .Environment.CACHE_VERSION }}
          paths:
            - packages
            - crossgcc
            - build
            - install
```

> I am a little confused to why in that CI build, under Restoring Cache it does not do so for ./build/coreboot-*?

@Tonux599 : Since the most complete cache matching declaration on restore was  `heads-modules-and-patches-{{ checksum "/tmp/all_modules_and_patches.sha256sums" }}{{ .Environment.CACHE_VERSION }}`, that is the cache that was extracted prior of build, simply because it was the first found in the `restore_cache` declared order. Only one cache is extracted.

The only case in which `heads-coreboot-musl-cross-{{ checksum "/tmp/coreboot_musl-cross.sha256sums" }}{{ .Environment.CACHE_VERSION }}` would be restored is if other `modules/*` then coreboot was modified. That additional cache was created because of this: coreboot should be touched bi-annually (long term cache) where musl-cross-make cache should rarely change, but where `heads-modules-and-patches-{{ checksum "/tmp/all_modules_and_patches.sha256sums" }}{{ .Environment.CACHE_VERSION }}` will be valid only until another module content is changed (which happens lot more often).

We actually want to always use  
1- `heads-modules-and-patches-{{ checksum "/tmp/all_modules_and_patches.sha256sums" }}{{ .Environment.CACHE_VERSION }}` if it exists (shorter build: ~1h)
2- `heads-coreboot-musl-cross-{{ checksum "/tmp/coreboot_musl-cross.sha256sums" }}{{ .Environment.CACHE_VERSION }}` build (longer)
3- `heads-musl-cross-{{ checksum "/tmp/musl-cross.sha256sums" }}{{ .Environment.CACHE_VERSION }}` ~4 hours build.

Tonux599:
@tlaurion Ahhh I get it, so the condition of caches being restored is based on whether or not is is required to restore a different cache (i.e. not all caches are required to be restored). I was under the assumption that all three caches needed to be restored. Thanks for taking the time to explain I get it now :slightly_smiling_face:

tlaurion:
@Tonux599 
1- If no cache exist, we rebuild from scratch. At the end of build, all three caches will be created.
2- If no other cache signature matches but musl-cross-make, musl-cross-make cache will be restored (economy of one hour on that build). That means, with this PR, that coreboot or its patches were modifies, or that another module signature or their patches, changed.
3- If all_modules cache doesnt exist (digest of a module different then coreboot or its patches changed) then coreboot cache will be restored. Economy of 2 hours in CI run
4- if all_modules cache exit, we have a CI run of an hour prior of having all actual boards roms produced.

Glad I made it clear this time! Will refer here in case of questions!

tlaurion:
@Tonux599 @Thrilleratplay @MrChromebox : now, if you have other recommendation, you should be able to review/approve that PR! CIs are important but a pain in the arse to deal with and understand. Had to wrap my head around it for necessity, now it would be nice that a lot understands it so that tweaks can be proposed!

Next step: tackle upload to github release (and then LVFS for fwupd support....)

Thrilleratplay:
@tlaurion @Tonux599 

> There are only two hard things in Computer Science: cache invalidation and naming things.
> -- Phil Karlton

PR #955, opened 2021-01-05
Improve Boot Device Detection Filtering

MrChromebox:
Grepping on just 'Disk' can lead to disk UUID identifier strings
being added to /tmp/disklist, which then fail to parse later on.

Avoid this by grepping on 'Disk /dev' instead.

### DIFF
--- a/initrd/etc/functions
+++ b/initrd/etc/functions
@@ -312,7 +312,7 @@ detect_boot_device()
 	fi
 
 	# generate list of possible boot devices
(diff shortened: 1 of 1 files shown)


tlaurion:
Don't have a test case to prove it wrong. But I thought this /sys/ was a better approach to deal with it? https://github.com/osresearch/heads/pull/924/files#diff-a7f6ce6d049ded4cf944a5c38c4ccc213c6a04602ab39eda57d2e46efaeddb38R26

MrChromebox:
@tlaurion I'm not sure how that would be a better way of identifying *internal* storage devices (SATA, NVME) but if someone has a better approach I'm open

PR #960, opened 2021-01-11
help

jtmoree-github-com:
add help and usage (-h) to boot utilities kexec-boot and kexec-select-boot

### DIFF
--- a/initrd/bin/kexec-boot
+++ b/initrd/bin/kexec-boot
@@ -1,13 +1,89 @@
 #!/bin/sh
 # Launches kexec from saved configuration entries
-set -e -o pipefail
-. /tmp/config
-. /etc/functions
+
+usage() {
+	cat <<End-of-message
+Usage
+=====
+
+Launches kexec from saved configuration entries
+	
+$0 -b /boot -e 'kexec params|...|...' [<options>]
+	
+Options
+-------
+
+### -a <additional confiurations>
+
+Anything provided to the -a argument is appended to the kernel command line (usually comes from parsing the grub, syslinux, BLS files). Ex.
+
+	-a 'export CONFIG_BOOT_KERNEL_ADD="intel_iommu=on intel_iommu=igfx_off"'
+
+### -b <boot path>
+
+path to boot files.  e.g.  -b /boot
+
+### -e <kexec params>
+
+used to prepare parameters passed to kexec system call for booting operating systems on disk.  Fields are pipe separated. 
+
+	"<description>|[elf,xen,multiboot]|kernel <path>|initrd <path>|<parameters>"
+
+First field is ignored.  The second field is the type of booting.  See examples below.
+
+The other fields are kernel, initrd, and parameters to pass to the linux kernel.
+
+EXAMPLE 1: ELF
+
+boot a linux kernel and specify the root partition
+
+	-e "my custom boot|elf|kernel /vmlinuz|initrd /initrd.img|append root=/dev/sda1"
+
+EXAMPLE 2: XEN
+
+boot Qubes, with Xen hypervisor
+
+	-e "xen boot|xen|module /vmlinuz|module /vmlinuz-4.4.67-13.x86_64 placeholder root=/dev/mapper/root ro|module /initramfs-4.4.67-13.x86_64.img"
+
+EXAMPLE 3: MULTIBOOT
+
+boot Qubes, with Xen hypervisor and LUKS encrypted root partition
+
+	-e "Qubes, with Xen hypervisor|multiboot|kernel /xen-4.6.5.gz placeholder |module /vmlinuz-4.4.67-13.pvops.qubes.x86_64 placeholder root=/dev/mapper/luks-UUID ro rd.qubes.hide_all_usb|module /initramfs-4.4.67-13.pvops.qubes.x86_64.img"
+
+Currently, any multiboot entry is interpreted as a Xen-variant and kexec-boot overrides the arguments to the multiboot kernel with custom arguments. 
+
+### -f
+
+some sort of debug mode which prints information about files and sets dryrun on
+
+### -h
+
+show help
+
+### -i
+
+some sort of debug mode which prints information about initrd and sets dryrun on
+
+### -o <initrd>
+
+use the specified initrd for booting
+
+### -r <configurations to remove>
+
+Anything provided to the -r argument is removed from the kernel command line. Ex.
+
+	-r 'export CONFIG_BOOT_KERNEL_REMOVE="quiet"'
+
+The 'quiet' option will hide needed information from the user.  In this case 'quiet' will be removd from the kernel configuration before it is booted.
+
+End-of-message
+}
 
 dryrun="n"
 printfiles="n"
 printinitrd="n"
-while getopts "b:e:r:a:o:fi" arg; do
+while getopts "b:e:r:a:o:fih" arg; do
 	case $arg in
 		b) bootdir="$OPTARG" ;;
 		e) entry="$OPTARG" ;;
@@ -16,11 +92,21 @@ while getopts "b:e:r:a:o:fi" arg; do
 		o) override_initrd="$OPTARG" ;;
 		f) dryrun="y"; printfiles="y" ;;
 		i) dryrun="y"; printinitrd="y" ;;
+		h) usage; exit 1;;
 	esac
 done
 
-if [ -z "$bootdir" -o -z "$entry" ]; then
-	die "Usage: $0 -b /boot -e 'kexec params|...|...'"
+set -e -o pipefail
+. /tmp/config
+. /etc/functions
+
+if [ -z "$bootdir" ]; then
+	usage
+	die "-b <boot> is required"
+fi
+if [ -z "$entry" ]; then
+	usage
+	die "-e <boot entry> is required"
 fi
 
 bootdir="${bootdir%%/}"
--- a/initrd/bin/kexec-select-boot
+++ b/initrd/bin/kexec-select-boot
@@ -1,8 +1,82 @@
 #!/bin/sh
 # Generic configurable boot script via kexec
-set -e -o pipefail
-. /tmp/config
-. /etc/functions
+
+usage() {
+	cat <<End-of-message
+Usage
+=====
+
+Generic configurable boot script via kexec
+	
+$0 -b /boot [options]
+	
+Options
+-------
+
+### -a
+
+Anything provided to the -a argument is appended to the kernel command line (usually comes from parsing the grub, syslinux, BLS files). Ex.
+
+	-a 'export CONFIG_BOOT_KERNEL_ADD="intel_iommu=on intel_iommu=igfx_off"'
+
+### -b
+
+path to boot files.  e.g.  -b /boot
+
+### -c = '*.cfg'
+
+specify the grub config file(s)
+
+### -d
+
+set the params device.  defaults to same as -b
+
+### -f
+
+force boot by considering hash and rollback checks as good
+
+### -g
+
+use graphical menus.  default is text only.
+
+### -h
+
+show help
+
+### -i
+
+ignore bad hash and/or bad rollback
+
+### -m
+
+show a boot menu
+
+### -o <initrd>
+
+use the specified initrd for booting
+
+### -p
+
+set the params folder.  defaults to same as -b
+
+### -r <configurations to remove>
+
+Anything provided to the -r argument is removed from the kernel command line. Ex.
+
+	-r 'export CONFIG_BOOT_KERNEL_REMOVE="quiet"'
+
+The 'quiet' option will hide needed information from the user.  In this case 'quiet' will be removd from the kernel configuration before it is booted.
+
+### -s
+
+Skip confirmation prompts.
+
+### -u
+
+specifies that the boot entries should be filtered to remove duplicates
+
+End-of-message
+}
 
 add=""
 remove=""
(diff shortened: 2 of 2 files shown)


MrChromebox:
perhaps edit the PR with a more descriptive title?

initrd/bin/kexec-select-boot
```diff
@@ -1,8 +1,82 @@
...
+$0 -b /boot [options]
+	
+Options
+-------
+
+### -a
+
+Anything provided to the -a argument is appended to the kernel command line (usually comes from parsing the grub, syslinux, BLS files). Ex.
+
+	-a 'export CONFIG_BOOT_KERNEL_ADD="intel_iommu=on intel_iommu=igfx_off"'
```
tlaurion, line 20: Same. Since board configs exports variable, those variables content are available.
On the following lines, we expend cmdadd and cmdremove with the content of the variables. If present, content is added, otherwise the string is appended with an empty string, so not expended.

https://github.com/osresearch/heads/blob/0d5b3d75e7b84ca85ff252fe5fe23aea990da435/initrd/bin/kexec-boot#L32-L33

initrd/bin/kexec-select-boot
```diff
@@ -1,8 +1,82 @@
...
+
+use graphical menus.  default is text only.
+
+### -h
+
+show help
+
+### -i
+
+ignore bad hash and/or bad rollback
```
tlaurion, line 48: Those are implicit in code when required to boot insecurely, from boot options menu.
The script starts by setting valid_hashes and valid_rollback to n.

Otherwise, same thing here, the config options are coming from BOARD config, exported and evaluated here:
https://github.com/osresearch/heads/blob/0d5b3d75e7b84ca85ff252fe5fe23aea990da435/initrd/bin/kexec-select-boot#L289-L294

if the board config they are required but not present, booting will fail.

initrd/bin/kexec-select-boot
```diff
@@ -1,8 +1,82 @@
...
+
+### -p
+
+set the params folder.  defaults to same as -b
+
+### -r <configurations to remove>
+
+Anything provided to the -r argument is removed from the kernel command line. Ex.
+
+	-r 'export CONFIG_BOOT_KERNEL_REMOVE="quiet"'
```
tlaurion, line 66: same as above

tlaurion (changes requested):

initrd/bin/kexec-boot
```diff
@@ -1,13 +1,89 @@
...
+$0 -b /boot -e 'kexec params|...|...' [<options>]
+	
+Options
+-------
+
+### -a <additional confiurations>
+
+Anything provided to the -a argument is appended to the kernel command line (usually comes from parsing the grub, syslinux, BLS files). Ex.
+
+	-a 'export CONFIG_BOOT_KERNEL_ADD="intel_iommu=on intel_iommu=igfx_off"'
```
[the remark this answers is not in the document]
jtmoree-github-com: Happy to fix the description but I am not sure what needs to be changed.  

Lines 57 and 122 use these settings to adjust $cmdadd and then in lines 98 and 114 the $cmdline var is added to $kexeccmd -- which is a call to 'kexec'.   As I understand it, the options are passed to kexec to become inputs to the kernel in the boot process.  

Do you mean that the options from -a might be processed in heads in addition to being passed to the booting kernel?

initrd/bin/kexec-select-boot
jtmoree-github-com: How about
```
### -i

Use this parameter to ignore the hash and rollback security settings and boot a potentially compromised system.
```

MrChromebox:
> Use this parameter to ignore the hash and rollback security settings and boot a potentially compromised system.

this reads like the parameter will roll back the security settings, rather than ignore the validity of the rollback counter

I'd suggest:

Use this parameter to skip the verification of boot hashes and rollback counter, and boot a potentially compromised system.

initrd/bin/kexec-boot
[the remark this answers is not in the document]
jtmoree-github-com: @tlaurion  line 124 (in my code) even says "Failed to add required kernel commands: $cmdadd".   Is the current description in the PR misleading?  If not, this adds a lot of information that does not exist currently which is helpful for troubleshooting in the heads environment.

jtmoree-github-com:
> I'd suggest:
> 
> Use this parameter to skip the verification of boot hashes and rollback counter, and boot a potentially compromised system.

checked in:
Use this parameter to skip the verification of boot hashes and rollback counter.  This will boot a potentially compromised system.

jtmoree-github-com:
I noticed that heads is struggling with size limitations.  Should I look for alternatives to adding all of this text to every utility inside of the heads env?  Instead I could build out the docs online in the wiki and reference those in 1 line of the help section.  I would still add '-h' to each utility to conform to *nix conventions.

initrd/bin/kexec-boot
```diff
@@ -1,13 +1,89 @@
...
+
+### -o <initrd>
+
+use the specified initrd for booting
+
+### -r <configurations to remove>
+
+Anything provided to the -r argument is removed from the kernel command line. Ex.
+
+	-r 'export CONFIG_BOOT_KERNEL_REMOVE="quiet"'
```
tlaurion, line 76: @jtmoree-github-com What I meant to say here (REMOVE and ADD variables, loaded where I pointed to their origin before.) is that they normally come from board configuration.

If you want to explain this, please do not point to export, it makes the user think that this is how they can call those.

in present exemple:
 -r "quiet"
 Would remove "quiet" from the default grub specified options, prior of passing resulting modified boot options to kexec'ed kernel.

initrd/bin/kexec-select-boot
tlaurion: @jtmoree-github-com The origin of add and remove is from here.

initrd/bin/kexec-select-boot
```diff
@@ -1,8 +1,82 @@
...
+
+use graphical menus.  default is text only.
+
+### -h
+
+show help
+
+### -i
+
+Use this parameter to skip the verification of boot hashes and rollback counter.  This will boot a potentially compromised system.
```
tlaurion, line 48: @jtmoree-github-com : -i for ignore would be more intuitive.

initrd/bin/kexec-boot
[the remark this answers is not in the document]
tlaurion: @jtmoree-github-com 

> Lines 57 and 122 use these settings to adjust $cmdadd and then in lines 98 and 114 the $cmdline var is added to $kexeccmd -- which is a call to 'kexec'. As I understand it, the options are passed to kexec to become inputs to the kernel in the boot process.

Exactly. They are modifiers. Those commands are called recursively inside of normal Heads codepaths, taking what is in BOARD config, resulting into /etc/config file, that can be overrided by /etc/config.user (and flashed back into the ROM, measured in PCR7, so the user can modify Heads configuration options on actual ROM and where /etc/config/user is merge with precedence to overrides over /etc/config


> Do you mean that the options from -a might be processed in heads in addition to being passed to the booting kernel?

if you call it directly from command line, -a and -r will remove just what you pass on command line. In Heads codepath, those removal and additions come from board config, /etc/config.user overrides by defaut. Doc should make it clear if doc is added for users wanting to read doc instead of code, while this documentation will need to be maintained if code change.

tlaurion:
> I noticed that heads is struggling with size limitations. Should I look for alternatives to adding all of this text to every utility inside of the heads env? Instead I could build out the docs online in the wiki and reference those in 1 line of the help section. I would still add '-h' to each utility to conform to *nix conventions.

@jtmoree-github-com no need to optimize shell script size. Shell scripts are compressed massively (text) in Heads' build process, as opposed to binaries. Please do not mind shell script size.

initrd/bin/kexec-boot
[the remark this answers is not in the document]
jtmoree-github-com: I hear your concerns as:

1. The help I have written for -a and -r does not include the context in which kexec-boot exists--namely board configs.  It is only discussing what happens between kexec-boot and kexec
2. Help documentation has to be kept in sync with working code so we should scope help to relevant information only

Is this correct?

initrd/bin/kexec-boot
jtmoree-github-com: updated help to use the correct format for -a, -r

initrd/bin/kexec-boot
```diff
@@ -1,13 +1,89 @@
...
+	
+Options
+-------
+
+### -a <additional confiurations>
+
+Anything provided to the -a argument is appended to the kernel command line (usually comes from parsing the grub, syslinux, BLS files). Ex.
+
+	-a 'intel_iommu=on intel_iommu=igfx_off'
+
```
tlaurion, line 21: @jtmoree-github-com 
Those normally come from [board configuration statements](https://github.com/osresearch/heads/blob/0d5b3d75e7b84ca85ff252fe5fe23aea990da435/initrd/bin/kexec-boot#L32-L33)

initrd/bin/kexec-boot
```diff
@@ -1,13 +1,89 @@
...
+### -o <initrd>
+
+use the specified initrd for booting
+
+### -r <configurations to remove>
+
+Anything provided to the -r argument is removed from the kernel command line. Ex.
+
+	-r 'quiet'
+
```
tlaurion, line 77: @jtmoree-github-com 
[kernel option remove statements normally come from board configuration](https://github.com/osresearch/heads/blob/0d5b3d75e7b84ca85ff252fe5fe23aea990da435/initrd/bin/kexec-boot#L32-L33) to pass board specific tweaks, needed for that board from OS perspective.

@jtmoree-github-com Something like that. :)

initrd/bin/kexec-select-boot
```diff
@@ -1,8 +1,82 @@
...
+
+set the params folder.  defaults to same as -b
+
+### -r <configurations to remove>
+
+Anything provided to the -r argument is removed from the kernel command line. Ex.
+
+	-r 'export CONFIG_BOOT_KERNEL_REMOVE="quiet"'
+
+The 'quiet' option will hide needed information from the user.  In this case 'quiet' will be removd from the kernel configuration before it is booted.
```
tlaurion, line 68: Same clarifications needs to be specified here.
Leaving -r 'export CONFIG_BOOT_KERNEL_REMOVE='quiet"' is misleading. User needs to understand where such parameter is coming from, and how to pass it directly to command line in case of testing and debugging. Otherwise, this documentatio nis aimed at users trying to understand how it works, and reading this line with export is confusing to me, so not sure it would be helpful, but more confusing for other users.

As a general guideline, in code documentation should state where command arguments are coming from with a really minimalist perspective. Otherwise we open a can of worm, as in here, where trying to cover more leads to more maintenance.

We have to remember that calling those directly is rarely ever needed, and called generally from gui-init, where the output of the transformation is actually given on screen as a result of the manipulations.

So if that part needs more documentation, in my opinion, there might be a bigger need to augment output on console where echo are talking from themselves for people who will actually debug the code and output of it, where general, high level documentation should state the higher level functionning of Heads without going in detail, where code should speak more about itself.

tlaurion (changes requested):

initrd/bin/kexec-select-boot
tlaurion, line 20: Same as remove comments!
I restate that this, showing export statement itself instead of its board origin, is more misleading then clarifying!

tlaurion (changes requested):
Please remove misleading statements. And replace with origin of the export instead of inlining them. Otherwise, more questions will come. We want less, not more :)

initrd/bin/kexec-select-boot
jtmoree-github-com: I didn't touch kexec-select-boot bc we were still working on kexec-boot.

initrd/bin/kexec-boot
jtmoree-github-com: I have reworded this to follow the suggestion below

initrd/bin/kexec-boot
jtmoree-github-com: I have reworded this to follow the suggestion

initrd/bin/kexec-boot
```diff
@@ -71,11 +73,13 @@ use the specified initrd for booting
...
 
-Anything provided to the -r argument is removed from the kernel command line. Ex.
+kernel option remove statements normally come from board configuration to pass board specific tweaks at the OS level.  Anything provided to the -r argument is removed from the kernel command line. 
+
+For EXAMPLE:
 
 	-r 'quiet'
 
-In this case 'quiet' will be removd from the kernel configuration before it is booted because the 'quiet' option will hide needed information from the user.  
+In this case 'quiet' will be removd from the kernel command line before it is booted.
```
tlaurion, line 82: @jtmoree-github-com 
"will be removd" -> "will be removed"

initrd/bin/kexec-select-boot
```diff
@@ -13,11 +13,13 @@ $0 -b /boot [options]
 Options
 -------
 
-### -a
+### -a <additional confiurations>
```
tlaurion, line 16: @jtmoree-github-com  : <additional confiurations>  -> <additional configurations>

tlaurion (changes requested):
Please correct typos! Sorry for the delay of responses. Life has been crazy.

initrd/bin/kexec-boot
jtmoree-github-com: fixed

initrd/bin/kexec-select-boot
jtmoree-github-com: fixed

jtmoree-github-com:
> Please correct typos! Sorry for the delay of responses. Life has been crazy.

no problem.  i have only been working on the wiki changes lately.

PR #993, opened 2021-05-09
Fix exit codes for ME download scripts

Tonux599:
Incorrect parentheses brackets used in those scripts meant that the script as a whole did not return the correct exit code. The use of `( )` brackets created a sub-shell to which the exit code applied to that sub-shell. Changing to `{ }` does not create a sub-shell and as such, the script will return its true return code.

This is evident by changing a character of `FINAL_ME_BIN_SHA256SUM` in `./blobs/xx20/download_parse_me.sh` and running `./download_parse_me.sh && echo "Okay" || echo "Failed"`.

It is important that the exit codes are correct as we want to catch errors as early as possible in CI builds.

### DIFF
--- a/blobs/xx30/download_clean_me.sh
+++ b/blobs/xx30/download_clean_me.sh
@@ -36,19 +36,19 @@ extractdir=$(mktemp -d)
 cd "$extractdir"
 
 echo "### Downloading https://download.lenovo.com/pccbbs/mobiles/g1rg24ww.exe..."
-wget  https://download.lenovo.com/pccbbs/mobiles/g1rg24ww.exe || ( echo "ERROR: wget not found" && exit 1 ) 
+wget  https://download.lenovo.com/pccbbs/mobiles/g1rg24ww.exe || { echo "ERROR: wget not found" && exit 1; }
 echo "### Verifying expected hash of g1rg24ww.exe"
-echo "$ME_EXE_SHA256SUM" | sha256sum --check || ( echo "Failed sha256sum verification on downloaded binary..." && exit 1 )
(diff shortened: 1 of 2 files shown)


blobs/xx30/download_clean_me.sh
```diff
@@ -36,19 +36,19 @@ extractdir=$(mktemp -d)
...
 
 echo "### Downloading https://download.lenovo.com/pccbbs/mobiles/g1rg24ww.exe..."
-wget  https://download.lenovo.com/pccbbs/mobiles/g1rg24ww.exe || ( echo "ERROR: wget not found" && exit 1 ) 
+wget  https://download.lenovo.com/pccbbs/mobiles/g1rg24ww.exe || { echo "ERROR: wget not found" && exit 1; }
 echo "### Verifying expected hash of g1rg24ww.exe"
-echo "$ME_EXE_SHA256SUM" | sha256sum --check || ( echo "Failed sha256sum verification on downloaded binary..." && exit 1 )
+echo "$ME_EXE_SHA256SUM" | sha256sum --check || { echo "Failed sha256sum verification on downloaded binary..." && exit 1; }
 
 echo "### Extracting g1rg24ww.exe..."
 innoextract ./g1rg24ww.exe || exit 1 "Failed calling innoextract. Tool installed on host?"
```
tlaurion, line 44: @Tonux599 same would apply here then!

tlaurion (changes requested):
Thanks for spotting this!

blobs/xx30/download_clean_me.sh
Tonux599: Missed that, good catch. Done :)

PR #1002, opened 2021-06-30
Add Talos 2 board (OpenPower)

SergiiDmytruk:
This provides a configuration for OpenPower that builds (wasn't fully tested yet). It seems to be the first non-x86 target for Heads and might be the time to think how you want to support multiple targets.

This isn't a finished PR and the point is to request comments. Below are some suggestions which you might want to consider.

***

I haven't tried to enable all modules. I think this is a typical config with some things disabled. I'm not sure which are necessary. Might enable and fix build of other things. At the moment at least 31 out of 48 available modules build fine.

Building for power produces `zImage` instead of `bzImage`.  The latter doesn't work (`no rule to make bzImage`). Image name is configurable now.

For coreboot, I specifed `-b` in `coreboot_repo` to switch to the branch. Might want to improve this.

For now I hoisted everything to board configuration via `CONFIG_<MODULE>_TARGET` variables and I don't really like how it looks (naming is very inconsistent). Maybe you want to have a generic `CONFIG_TARGET`, which will then be translated within each module and fail for unknown targets?  This might ease adding new targets in the future thanks to more obvious errors.


modules/musl-cross
```diff
@@ -32,22 +33,28 @@ musl-cross_libraries := \
 else
 
 # Force a full build of the cross compiler
-# have to build both x86_64 and i386 versions for coreboot
 
+ifeq "$(CONFIG_MUSL_TARGET)" "x86-64"
+# have to build both x86_64 and i386 versions for coreboot
 musl-cross_configure := \
-	/bin/echo -e >> Makefile 'both:' ; \
```
tlaurion, line 38: @SergiiDmytruk the disappearance of this statement seems to be the cause of the failing in CI around line 1833 https://app.circleci.com/pipelines/github/tlaurion/heads/734/workflows/978d2ee9-07b5-4014-b570-b9f69a787d53/jobs/1106

SergiiDmytruk: @tlaurion You mean this error?
```
cc1: error: no include path in which to search for stdc-predef.h
```
Not sure why renaming a target would cause this, but maybe something else does.

I haven't tested x86 build with my changes, that's a good point. For now I'm trying to solve issues with `/init` not being execute by the kernel (`error -2`). I'll fix x86 build once I get it to work on power to not fix it more than once.

tlaurion: No, I was referring to the difference in output from last successful full build (not from cache) where weirdly
> cc1: error: no include path in which to search for stdc-predef.h

Is also shown, but where [it is not a problem](https://circleci.com/api/v1.1/project/github/tlaurion/heads/1058/output/103/0?file=true&allocation-id=60d24e2c9f1dbc44db03306d-0-build%2F2B2E18CF) for both i386 and x64
```

echo | /root/project/build/musl-cross-38e52db8358c043ae82b346a2e6e66bc86a53bc1/build/local/x86_64-linux-musl/obj_gcc/./gcc/xgcc -B/root/project/build/musl-cross-38e52db8358c043ae82b346a2e6e66bc86a53bc1/build/local/x86_64-linux-musl/obj_gcc/./gcc/ -E -dM - | \
  sed -n -e 's/^#define \([^_][a-zA-Z0-9_]*\).*/\1/p' \
	 -e 's/^#define \(_[^_A-Z][a-zA-Z0-9_]*\).*/\1/p' | \
  sort -u > tmp-macro_list
cc1: error: no include path in which to search for stdc-predef.h
/bin/bash ../../src_gcc/gcc/../move-if-change tmp-macro_list macro_list
echo timestamp > s-macro_list
```

```
echo | /root/project/build/musl-cross-38e52db8358c043ae82b346a2e6e66bc86a53bc1/build/local/x86_64-linux-musl/obj_gcc/./gcc/xgcc -B/root/project/build/musl-cross-38e52db8358c043ae82b346a2e6e66bc86a53bc1/build/local/x86_64-linux-musl/obj_gcc/./gcc/ -E -dM - | \
  sed -n -e 's/^#define \([^_][a-zA-Z0-9_]*\).*/\1/p' \
	 -e 's/^#define \(_[^_A-Z][a-zA-Z0-9_]*\).*/\1/p' | \
  sort -u > tmp-macro_list
cc1: error: no include path in which to search for stdc-predef.h
/bin/bash ../../src_gcc/gcc/../move-if-change tmp-macro_list macro_list
echo timestamp > s-macro_list
```

But where the path searched for compiled musl-cross is not found [here](https://circleci.com/api/v1.1/project/github/tlaurion/heads/1106/output/103/0?file=true&allocation-id=60e2325708b5c9541ae6c52d-0-build%2F507C5687):
```
../scripts/gcc-version.sh: line 26: /root/project/crossgcc/bin/x86-64-linux-musl-gcc: No such file or directory
../scripts/gcc-version.sh: line 27: /root/project/crossgcc/bin/x86-64-linux-musl-gcc: No such file or directory
```
Where that doesn't occur in x86 builds. will have to dig.

SergiiDmytruk: Right, thanks for pointing it out. `x86-64` should be `x86_64` (`-` -> `_`).

tlaurion:
@SergiiDmytruk did a regression [build](https://app.circleci.com/pipelines/github/tlaurion/heads/736/workflows/bfc6b6a5-9c45-4560-9e54-3d6f0b8c86cd)

Everything x86 builds as expected from last commit. I would recommend following your branch from CircleCI so that each commit pushed triggers a CircleCI build for that commit.

tlaurion:
@SergiiDmytruk 

You could [change](https://github.com/osresearch/heads/blob/master/.circleci/config.yml#L164-L312)
for something similar to the following and tune to have your boards built first with save_cache. (build_and_persist is costy since saving workstpace and restoring workspaces). You might want to only build talos_2 and talos_2_lite and wipe the rest to speed up testing also if regression testing is not needed in all builds (I think changes that are coming next should probably only happen in linux and coreboot configs and trees, so touching other x86 and testing for regression might not be needed.

```
# Below, sequentially build one board for each coreboot
# version. The last board in the sequence is the dependency
# for the parallel boards built at the end, and also save_cache.
      #Coreboot Talos
      - build_and_persist:
          name: talos_2
          target: talos_2
          requires:
            - prep_env

      # Coreboot 4.8.1
      - build_and_persist:
          name: qemu-coreboot
          target: qemu-coreboot
          requires:
            - prep_env

      # Coreboot 4.11
      - build_and_persist:
          name: kgpe-d16_workstation
          target: kgpe-d16_workstation
          requires:
            - qemu-coreboot

      # Coreboot 4.13
      - build_and_persist:
          name: librem_mini
          target: librem_mini
          requires:
            - kgpe-d16_workstation


      - save_cache:
          requires:
            - librem_mini

#
#
# Those onboarding new boards should add their entries below.
#
#
      - build:
          name: talos_2_le
          target: talos_2_le
          requires:
            - talos_2

      - build:
          name: x220-hotp-maximized
          target: x220-hotp-maximized
          requires:
            - librem_mini

      - build:
          name: x220-maximized
          target: x220-maximized
          requires:
            - librem_mini

      - build:
          name: t420-hotp-maximized
          target: t420-hotp-maximized
          requires:
            - librem_mini

      - build:
          name: t420-maximized
          target: t420-maximized
          requires:
            - librem_mini

      - build:
          name: x230-flash
          target: x230-flash
          requires:
            - librem_mini

      - build:
          name: t430-flash
          target: t430-flash
          requires:
            - librem_mini

      - build:
          name: t430
          target: t430
          requires:
            - librem_mini

      - build:
          name: x230
          target: x230
          requires:
            - librem_mini

      - build:
          name: x230-hotp-verification
          target: x230-hotp-verification
          requires:
            - librem_mini

      - build:
          name: x230-maximized
          target: x230-maximized
          requires:
            - librem_mini

      - build:
          name: t430-hotp-maximized
          target: t430-hotp-maximized
          requires:
            - librem_mini

      - build:
          name: x230-hotp-maximized
          target: x230-hotp-maximized
          requires:
            - librem_mini

      - build:
          name: t430-maximized
          target: t430-maximized
          requires:
            - librem_mini

      - build:
          name: qemu-coreboot-fbwhiptail
          target: qemu-coreboot-fbwhiptail
          requires:
            - librem_mini

      - build:
          name: kgpe-d16_workstation-usb_keyboard
          target: kgpe-d16_workstation-usb_keyboard
          requires:
            - librem_mini

      - build:
          name: kgpe-d16_server
          target: kgpe-d16_server
          requires:
            - librem_mini

      - build:
          name: kgpe-d16_server-whiptail
          target: kgpe-d16_server-whiptail
          requires:
            - librem_mini

      - build:
          name: librem_l1um
          target: librem_l1um
          requires:
            - librem_mini

      - build:
          name: librem_mini_v2
          target: librem_mini_v2
          requires:
            - librem_mini
```

tlaurion:
Or without x86 regression testing this simpler version:

```
# Below, sequentially build one board for each coreboot
# version. The last board in the sequence is the dependency
# for the parallel boards built at the end, and also save_cache.
      #Coreboot Talos
      - build_and_persist:
          name: talos_2
          target: talos_2
          requires:
            - prep_env

      - save_cache:
          requires:
            - talos_2

#
#
# Those onboarding new boards should add their entries below.
#
#
      - build:
          name: talos_2_le
          target: talos_2_le
          requires:
            - talos_2
```

SergiiDmytruk:
@tlaurion, thanks. I left one x86 target to see if CI works in this case. I'm not sure because it's not enough to just replace `build/` to switch targets, need to also remove `install/` and `crossgcc/` or they will mess up the build (libraries and includes will be used from wrong toolchain).

tlaurion:
> @tlaurion, thanks. I left one x86 target to see if CI works in this case. I'm not sure because it's not enough to just replace `build/` to switch targets, need to also remove `install/` and `crossgcc/` or they will mess up the build (libraries and includes will be used from wrong toolchain).

@SergiiDmytruk 

I see.... x86 centric and completely understandable until now :/

Probably better to fix [INSTALL](https://github.com/osresearch/heads/blob/master/Makefile#L29) by appending TARGET/arch suffix to that directory ([while musl-cross path should deal with x86 and power from inclusion of the module def](https://github.com/osresearch/heads/blob/master/Makefile#L163-L189)) from now on. 

Fixating caches to include those new directories into them would also be required to be reused:
- Include Makefile so that caches digests (tag) are not reused when Makefile changes ([that didn't change for a while and wasn't included](https://github.com/osresearch/heads/blob/master/.circleci/config.yml#L57-L69))
  - Names can be reused without change (above will change digest, so caches will not be restored if digest changed)
- Where cache directories backuped/restored should be changed to [match new directories created having TARGET (or arch) appended](https://github.com/osresearch/heads/blob/master/.circleci/config.yml#L130-L156)

tlaurion:
@SergiiDmytruk your board dir and board names are matching to "[talos-2"](https://github.com/osresearch/heads/blob/61002eb088c5819867e12dcdb0d5591bc33a9aca/boards/talos-2/talos-2.config) and "[talos-2-le](https://github.com/osresearch/heads/blob/61002eb088c5819867e12dcdb0d5591bc33a9aca/boards/talos-2-le/talos-2-le.config)" not "talos_2" and "talos_2_le" which you fixed under https://github.com/osresearch/heads/blob/61002eb088c5819867e12dcdb0d5591bc33a9aca/.circleci/config.yml

The last build was manually cancelled, but should have started a [build](https://app.circleci.com/pipelines/github/SergiiDmytruk/heads/4/workflows/f774d735-4ef5-45ef-9e87-254428c31a57/jobs/6)!

SergiiDmytruk:
@tlaurion It wasn't me:

    We have detected an anomaly that violates the CircleCI Terms of Service.
    We did not test this push because the project has been suspended. Please contact us if you think this is a mistake.

tlaurion:
@SergiiDmytruk weird. This is an error.
Building [here](https://app.circleci.com/pipelines/github/tlaurion/heads/737/workflows/f7068e37-a156-4e37-93b1-f1d4288f9378) meanwhile, but yes, this is open source project so should not be a violation of their terms of services.

Let's see if that build is also suspended.

tlaurion:
@SergiiDmytruk note that old cache is being reused for last successful build with same modules and patches created digests as explained in previous post. Those are saved into workspace now and will be passed to board builds to be reused, so it will probably fail since ./install and ./crossgcc (and build/musl-cross) will be shared between x86 and power.

SergiiDmytruk:
@tlaurion CI works again for me after contacting support. Haven't updated how caching is done yet.

SergiiDmytruk:
@tlaurion Currently Heads produces kernel and initrd separately. Was embedding of initrd into the kernel ever considered? We probably won't need this for coreboot (skiroot needs it), but I thought I'll bring it up just in case there were attempts and known issues. At the moment I only see this way:
1. Create empty `initrd.cpio`
2. Build everything including real initrd
3. (Re)link kernel

tlaurion:
> @tlaurion Currently Heads produces kernel and initrd separately. Was embedding of initrd into the kernel ever considered? 
@SergiiDmytruk never thought about it....

initrd is actually what Heads is (outside of the patches applied to linux kernel and patches applied to coreboot (less and less, but still.) Just to make sure, you can see those compressed artifacts under /build/$BOARD dir and under CI. [Example artifacts for a built board](https://app.circleci.com/pipelines/github/tlaurion/heads/736/workflows/bfc6b6a5-9c45-4560-9e54-3d6f0b8c86cd/jobs/1119/artifacts)
- bzImage (kernel)
- hashes.txt (contains hashes of all compiled and bundled steps to be able to compare between builds)
- final rom for board (coreboot+ kernel + initrd per coreboot last step of putting everything together)
- heads.cpio ([scripts under /initrd](https://github.com/osresearch/heads/blob/883f4958f8c8a000d979c6f3a63b581eaff858a1/Makefile#L575-L576))
- modules.cpio ([linux kernel modules per linux heads's module](https://github.com/osresearch/heads/blob/c3b0bd6ffbe816430dd41ef54e649af52ed1ff3b/modules/linux#L145-L164))
- tools.cpio ([modules bin, modules libs](https://github.com/osresearch/heads/blob/883f4958f8c8a000d979c6f3a63b581eaff858a1/Makefile#L583-L611))
- initrd.cpio.xz emcompassing [dev.cpio](https://github.com/osresearch/heads/blob/883f4958f8c8a000d979c6f3a63b581eaff858a1/Makefile#L546), then [modules.cpio, tools.cpio and heads.cpio](https://github.com/osresearch/heads/blob/883f4958f8c8a000d979c6f3a63b581eaff858a1/Makefile#L547-L569) above.

Some reverse understanding is currently possible since initrd is built separately:
[tools.cpio, modules.cpio **initrd.cpio** and heads.cpio are bundled under initrd.cpio.xz](https://github.com/osresearch/heads-wiki/issues/70#issuecomment-848279135). The more we bundle things the more difficult it is for newcomers to reverse the bundling inspect and diagnose reproducibility problems, but that might be only a documentation guide away. I do not have a preference over this, if you think the same outcome can be achieved, while understanding what is happening here.

> We probably won't need this for coreboot (skiroot needs it), but I thought I'll bring it up just in case there were attempts and known issues. At the moment I only see this way:
> 
>     1. Create empty `initrd.cpio`
> 
>     2. Build everything including real initrd
> 
>     3. (Re)link kernel

Let me know if this idea is still applicable following details provided above.

SergiiDmytruk:
@tlaurion, thanks for the detailed explanation.

> Let me know if this idea is still applicable following details provided above.

I think everything should be OK as is once coreboot will be able to use Heads as payload.

macpijan:
> I think everything should be OK as is once coreboot will be able to use Heads as payload.

I do not think it will be in the near future. The goal for now is to replace `skirtoot` with `heads` (while still using `skiboot` as the only available `coreboot` payload). As defined here: https://github.com/osresearch/heads/issues/720#issuecomment-867522437

> We won't be able to start heads kernel as a coreboot Linux payload as it is currently done for other platforms. At least for now, we only can start skiboot from coreboot. We need it for OPAL. Then, the skiboot can load Linux - in this case it could load heads kernel and initramfs.

SergiiDmytruk:
@tlaurion I've re-committed changes and marking this PR for review (but will see if CI passes with all the boards). Below is a summary of the changes since the previous comment.

The build eventually started to fail due to mixing of files of different architectures under `build/`, I looked into updating makefiles of modules, but it seemed unfeasible (very specific to each project and would add many more patches), so now it's `build/<arch>/<module|board>` with gawk being build in `build/` once. Hopefully, I updated configs and CircleCI configuration right.

It's just one board again with BE coreboot and LE everywhere else, which seems to be typical for PowerPC.

Because we need skiboot, Heads can now optionally build bundled Linux kernel and the new board asks for it.

Patches for coreboot can be dropped after https://github.com/3mdeb/coreboot/pull/79 is merged.

There is now `CONFIG_TARGET` (`x86` by default) and modules which had build issues branch on it to determine host triplet instead of hard-coding this module-specific knowledge in board definition.

SergiiDmytruk:
Finally, CI is green.

tlaurion:
> I think everything should be OK as is once coreboot will be able to use Heads as payload.
@SergiiDmytruk Not sure about the next steps here. Do we wait to merge once [Heads can be used as a payload](https://github.com/osresearch/heads/pull/1002#issuecomment-877445321)?

SergiiDmytruk:
@tlaurion

> Do we wait to merge once Heads can be used as a payload?

It's not clear when that will be the case (i.e., when coreboot will support OPAL). A much closer target is the ability to start `skiboot`, although I don't know its exact status (might work already, not much changes on the branch) and when it will make its way upstream.

> Where coreboot image doesn't seem to bundle much. Can you detail the current state of the build?

It's small because its payload is `skiboot`. `zImage.bundled` should be inserted into flash separately and `skiboot` will pick it up from there. `coreboot` needs to be installed the same way. Actually we need `.signed.ecc` ROM as well as `bootblock.signed.ecc` in this case. I'll change that and also test all build artifacts together, thanks.

> Not sure about the next steps here.

My concern about this PR is mostly about changes to the build system, which is the bulk of it. Do they look OK or something needs to be done differently?

.circleci/config.yml
```diff
@@ -201,115 +219,115 @@ workflows:
...
           requires:
-            - librem_mini
+            - talos-2
 
       - build:
           name: librem_mini_v2
           target: librem_mini_v2
           requires:
-            - librem_mini
+            - talos-2
```
tlaurion, line 330: Not sure why requires librem_mini was changed here for talos-2 (not same arch, not same set of dependencies) where from CircleCI perspective, what is important is [the dependency of save_cache to happen only after talos-2 has been built](https://github.com/osresearch/heads/pull/1002/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R208-R210)?

tlaurion: Otherwise, we have to wait for another 1:58 minutes to build talos-2 prior of building all other boards, as the mapping shows [here](https://app.circleci.com/pipelines/github/SergiiDmytruk/heads/30/workflows/0e3e4ba6-79ab-4b36-867a-fc02520cc2ae)

tlaurion:
As a basic regression test, I flashed https://113-380596387-gh.circle-artifacts.com/0/build/x86/x230-hotp-maximized/heads-x230-hotp-maximized-v0.2.0-1047-g586ddd0.rom internally and booted Tails with detached signature file verified successfully.

I will review the changes done on the build system, but I do not see anything problematic.
@MrChromebox ?

.circleci/config.yml
SergiiDmytruk: Initially it wasn't like this, but then I noticed that other `build` jobs are run after *all* `build_and_persist` and changed it in case it's important (e.g., if it's not OK to save cache while other jobs are running). I'll undo it.

SergiiDmytruk:
@tlaurion Switched the board to use `gui-init` with `whiptail` in console mode. Tested GUI mode and it works fine, but it's much less convenient to use remotely.

Also flashed coreboot and kernel built by Heads, but `skiboot` hanged on CPU initialization. Something is probably not finished yet.

tlaurion:
> @tlaurion Switched the board to use `gui-init` with `whiptail` in console mode. Tested GUI mode and it works fine, but it's much less convenient to use remotely.

@SergiiDmytruk Can you elaborate on "less convenient"? Not scriptable for sure unless doing screen captures and automatic input for automatic tests. :/

But gui-init is what is currently used by Purism, Nitrokey and Insurgo. generic-init is kind of Heads legacy base, which was used prior of HOTP support (joint effort between Heads, Nitrokey and Purism) and is not much used anymore compared to gui-init codebase.

SergiiDmytruk:
@tlaurion

> Can you elaborate on "less convenient"?

I use it either via SSH or through a Web-interface. And OpenBMC doesn't seem to accept input where it shows the screen in the browser, so it has to be provided in a separate window where it shows textual console. That's all I meant, text is more universal. That said, how it will look finally doesn't matter too much, because it's configurable, so I can enable `CAIRO` and `FBWHIPTAIL` if you like it more.

> But gui-init is what is currently used by Purism, Nitrokey and Insurgo.

I think you're trying to justify switching to `gui-init`, which I already did (works fine). Just not in a graphical mode.

tlaurion:
> I think you're trying to justify switching to gui-init, which I already did (works fine). Just not in a graphical mode.

Graphical mode (FBWHIPTAIL, cairo and other dependencies) can be switched on for workstation builds later on. 
My advice here was to create other board configs for workstation (FBWHIPTAIL)/server (WHIPTAIL) with different subsets of dependencies tested so that building those is also confirmed functional.

I understand what you are saying. qemu-coreboot-fbwhiptail is slow. kgpe-d16 server is using whiptail for BMC, not fbwhiptail. Sorry for the noise.

MrChromebox:
@tlaurion the only thing I see that's less than ideal is adding another subdir (arch) that needs to be accounted for in the coreboot configs, adding another `../` in the payload and initrd paths.

I think the git repo handling could be done better, perhaps specifying a commit hash vs a branch, since the former is immutable and the latter is not. I have a patch in Purism's tree to do that, but can't be used with `--depth=1` in the git clone so there's room for improvement in my current implementation

SergiiDmytruk:
@MrChromebox

> the only thing I see that's less than ideal is adding another subdir (arch) that needs to be accounted for in the coreboot configs, adding another ../ in the payload and initrd paths.

I didn't like that editing either. Have you considered doing `automake`-like expansion in configs?  Something like
```
CONFIG_PAYLOAD_FILE="../../../build/x86/librem_l1um/bzImage"
```
would become
```
CONFIG_PAYLOAD_FILE="@BOARD_DIR@/bzImage"
```
`sed` should be enough to do such simple replacements during copying of a config.

MrChromebox:
@SergiiDmytruk yes something along those lines would be ideal, have @BOARD_BUILD_DIR@ and maybe @BLOB_DIR@ as well

MrChromebox:
@SergiiDmytruk I like the @VAR@ expansion commit, but given the size of this PR, it might be better to do separately

tlaurion:
@MrChromebox I am still unsure how to properly manage big changes under Heads like this one which requires Makefile, modules and boards changes altogether.

How would you seperate this in smaller PRs?

SergiiDmytruk:
@MrChromebox @tlaurion Changes fall into the following categories:
* changing build layout
* adding power arch
* adding new boards
* smaller changes like building musl in parallel

I can make a separate PR with non-Talos changes and this one will be rebased on top of it. Or two new PRs: build layout change in the first and everything else but new boards in the second.

MrChromebox:
@tlaurion I'm not seeing why they can't be broken down as @SergiiDmytruk mentioned above. I'd be ok with 3 PRs - one with the build changes / adding power arch, one for adding the new boards, and then the non-Talos changes as a third.

SergiiDmytruk:
@MrChromebox I removed commit that adds `--depth 1` to not interfere with checking out specific revision. Now the only change that isn't required for new boards is building musl with `-jN`. Didn't create a PR for a trivial change like that or should it also include things like supporting bundled kernel, etc.? In that case third PR makes sense.

tlaurion:
> Didn't create a PR for a trivial change like that or should it also include things like supporting bundled kernel, etc.? In that case third PR makes sense.

@SergiiDmytruk @MrChromebox I think this makes sense (and for Heads CI, a nice improvement for initial build stage of musl-cross-make)

macpijan:
> It's not clear when that will be the case (i.e., when coreboot will support OPAL).

Not planned for a near future. We must assume `skiboot` would load heads.

> A much closer target is the ability to start skiboot, although I don't know its exact status (might work already, not much changes on the branch) 

The `skiboot` already can be started by coreboot. Although, the `skiboot` cannot fully correctly boot Linux kernel yet, because of missing bits in coreboot. We would expect to boot `coreboot -> skiboot -> Linux` next month.

> and when it will make its way upstream.

That is another story

SergiiDmytruk:
> I think the git repo handling could be done better, perhaps specifying a commit hash vs a branch, since the former is immutable and the latter is not. I have a patch in Purism's tree to do that, but can't be used with `--depth=1` in the git clone so there's room for improvement in my current implementation

@MrChromebox Couldn't find your commit, added a different one to get rid of the `-b` option in the wrong place.

MrChromebox:
@SergiiDmytruk these two:
https://source.puri.sm/firmware/pureboot/-/commit/4ebc1e493eac096c84f420c107a28fde4b0e82bf
https://source.puri.sm/firmware/pureboot/-/commit/1acb636167164aa8c7eac133c1f8783ee972dee8

SergiiDmytruk:
@MrChromebox Wasn't looking there, thanks. Took your version.

modules/coreboot
```diff
@@ -24,6 +34,11 @@ else ifeq "$(CONFIG_COREBOOT_VERSION)" "4.14"
 	coreboot_hash := d907379b727561d7ddd1d80b2fabaa373db00c9805719116f591cbc948173c6e
 	coreboot-blobs_hash := 9ee2fe5ba37d0e214000b8655acf8922de6df792adb75790559db4c160847921
 	coreboot_depends := $(if $(CONFIG_PURISM_BLOBS), purism-blobs)
+else ifeq "$(CONFIG_COREBOOT_VERSION)" "talos_2_support_payload"
+	coreboot_version = git
+	coreboot_patch_version := talos_2_support_payload
+	coreboot_commit_hash = ce64c091228ed8878263263fd7fe700e64557148
```
macpijan, line 40: @SergiiDmytruk Can we use dasharo and latest commit from release branch?  https://github.com/Dasharo/coreboot/tree/raptor-cs_talos-2/release/

patches/coreboot-talos_2_support_payload/0001-Makefile.inc-fix-depending-on-SBSIGNTOOLS.patch
```diff
@@ -0,0 +1,49 @@
+From 8a8f12059cc4dc4fd5c3719dece457bc8e442e32 Mon Sep 17 00:00:00 2001
+From: Sergii Dmytruk <sergii.dmytruk@3mdeb.com>
+Date: Thu, 1 Jul 2021 18:20:10 +0300
+Subject: [PATCH 1/6] Makefile.inc: fix depending on $(SBSIGNTOOLS)
+
+The variable points to a directory, which might have its timestamp newer
```
macpijan, line 6: @SergiiDmytruk Can it be fixed in main repo, rather than as a patch?

patches/coreboot-talos_2_support_payload/0002-external-skiboot-do-not-hard-code-value-of-CROSS.patch
```diff
@@ -0,0 +1,45 @@
+From 9e4102c27e73c9943d07f419451a39bfa098c02d Mon Sep 17 00:00:00 2001
+From: Sergii Dmytruk <sergii.dmytruk@3mdeb.com>
+Date: Thu, 1 Jul 2021 18:24:02 +0300
+Subject: [PATCH 2/6] external/skiboot: do not hard-code value of CROSS
+
+Currently building with `powerpc64-linux-musl-` fails because of this.
```
macpijan, line 6: @SergiiDmytruk Let's remove this by picking this to Dasharo: https://github.com/3mdeb/coreboot/pull/79#issuecomment-962015740

patches/coreboot-talos_2_support_payload/0004-Makefile.inc-directly-call-create-container.patch
```diff
@@ -0,0 +1,77 @@
+From 514ad8f2a486d5efcb28d8e6f487a4afc7735e9d Mon Sep 17 00:00:00 2001
+From: Sergii Dmytruk <sergii.dmytruk@3mdeb.com>
+Date: Tue, 6 Jul 2021 18:22:33 +0300
+Subject: [PATCH 4/6] Makefile.inc: directly call create-container
+
```
macpijan, line 5: I believe we could use explanation for this patch. And again: fixable in main repo, rather than patch?

modules/coreboot
macpijan: Or maybe the develop branch. As we want to pick some commits to reduce number of patches. And it will happen quicker on develop, rather than on release.

SergiiDmytruk: Sure, I'll update references and remove all patches when changes land in develop branch of dasharo repo.

`talos_2_support_payload` won't suit anymore, make it `talos_2` or do you have a better naming idea?

tlaurion: talos_2 is fine!

tlaurion:
LGTM. @MrChromebox ?

MrChromebox (approved):

tlaurion:
@SergiiDmytruk : this would need a rebase so that next release is built from heads:
- docs: https://github.com/Dasharo/docs/pull/264/files
- code: https://github.com/Dasharo/coreboot/tree/raptor-cs_talos-2_v0.6.0

tlaurion:
(I also need to check if no-tpm code has been merged, or otherwise check with Purism codebase to do PR. Basically, since they had a board (mini if I recall correctly) that newly had TPM support, they changed their fork to have TPMTOTP/NO-TPM support and not fail close, but warn user at GUI that no-tpm was detected (in case of hardware failure foing TPM->No TPM, where TPM was active in both coreboot and defined in board config).

Also, since usb keyboard is now optional (and should be in under Talos because no PS2), having HID built as module (and toggle USB keyboard option in board config) would also permit to use Librem Key to measure ROM from flashrom in case there is no TPM. Untested as of now.

tlaurion:
@SergiiDmytruk build fails.

Will need more time reviewing why all modules need change instead of modifying  global make file instead with arch specific tweaks.

SergiiDmytruk:
> build fails.

Looks like GCC is newer than 9 on CircleCI with its `-fno-common` by default.

> Will need more time reviewing why all modules need change instead of modifying global make file instead with arch specific tweaks.

Defining targets in one place didn't work because modules use different syntax for target strings, so I made each module handle architectures explicitly.

tlaurion:
> > build fails.
> 
> Looks like GCC is newer than 9 on CircleCI with its `-fno-common` by default.

Then something broke with using musl-cross-make, which should be used to build every module (but coreboot).
The host gcc i ssupposed to be only used to build locally used tools, including musl-cross-make and coreboot's build stack, which are then used to build, not debian-11's GCC.

For x86 prior of PR:
```
user@heads-tests:~/heads$ /home/user/heads/crossgcc/bin/x86_64-linux-musl-gcc -v
Using built-in specs.
COLLECT_GCC=/home/user/heads/crossgcc/bin/x86_64-linux-musl-gcc
COLLECT_LTO_WRAPPER=/home/user/heads/crossgcc/bin/../libexec/gcc/x86_64-linux-musl/8.3.0/lto-wrapper
Target: x86_64-linux-musl
Configured with: ../src_gcc/configure --enable-languages=c,c++ --disable-werror --target=x86_64-linux-musl --prefix= --libdir=/lib --disable-multilib --with-sysroot=/x86_64-linux-musl --enable-tls --disable-libmudflap --disable-libsanitizer --disable-gnu-indirect-function --disable-libmpx --enable-libstdcxx-time --with-build-sysroot=/home/user/heads/build/musl-cross-38e52db8358c043ae82b346a2e6e66bc86a53bc1/build/local/x86_64-linux-musl/obj_sysroot AR_FOR_TARGET=/home/user/heads/build/musl-cross-38e52db8358c043ae82b346a2e6e66bc86a53bc1/build/local/x86_64-linux-musl/obj_binutils/binutils/ar AS_FOR_TARGET=/home/user/heads/build/musl-cross-38e52db8358c043ae82b346a2e6e66bc86a53bc1/build/local/x86_64-linux-musl/obj_binutils/gas/as-new LD_FOR_TARGET=/home/user/heads/build/musl-cross-38e52db8358c043ae82b346a2e6e66bc86a53bc1/build/local/x86_64-linux-musl/obj_binutils/ld/ld-new NM_FOR_TARGET=/home/user/heads/build/musl-cross-38e52db8358c043ae82b346a2e6e66bc86a53bc1/build/local/x86_64-linux-musl/obj_binutils/binutils/nm-new OBJCOPY_FOR_TARGET=/home/user/heads/build/musl-cross-38e52db8358c043ae82b346a2e6e66bc86a53bc1/build/local/x86_64-linux-musl/obj_binutils/binutils/objcopy OBJDUMP_FOR_TARGET=/home/user/heads/build/musl-cross-38e52db8358c043ae82b346a2e6e66bc86a53bc1/build/local/x86_64-linux-musl/obj_binutils/binutils/objdump RANLIB_FOR_TARGET=/home/user/heads/build/musl-cross-38e52db8358c043ae82b346a2e6e66bc86a53bc1/build/local/x86_64-linux-musl/obj_binutils/binutils/ranlib READELF_FOR_TARGET=/home/user/heads/build/musl-cross-38e52db8358c043ae82b346a2e6e66bc86a53bc1/build/local/x86_64-linux-musl/obj_binutils/binutils/readelf STRIP_FOR_TARGET=/home/user/heads/build/musl-cross-38e52db8358c043ae82b346a2e6e66bc86a53bc1/build/local/x86_64-linux-musl/obj_binutils/binutils/strip-new
Thread model: posix
gcc version 8.3.0 (GCC) 
```

SergiiDmytruk:
> Then something broke with using musl-cross-make, which should be used to build every module (but coreboot).

No, it failed for `dtc` tool built for host.

.circleci/config.yml
```diff
@@ -159,20 +169,17 @@ jobs:
...
           paths:
             - crossgcc
-            - build/musl-cross-38e52db8358c043ae82b346a2e6e66bc86a53bc1
+            - build/*/musl-cross-*
             - packages
-            - build/coreboot-4.11
-            - build/coreboot-4.13
-            - build/coreboot-4.14
-            - build/coreboot-4.15
+            - build/*/coreboot-*
```
tlaurion, line 182: same

tlaurion (commented):
Unfortunately, highly doubt CircleCI fixed their wildcard limitation for cache save/restore. Once cache is saved and build is done, please trigger a rebuild for same commit on your CircleCI instance.

.circleci/config.yml
```diff
@@ -159,20 +169,17 @@ jobs:
           key: heads-musl-cross-{{ checksum "./tmpDir/musl-cross.sha256sums" }}{{ .Environment.CACHE_VERSION }}
           paths:
             - crossgcc
-            - build/musl-cross-38e52db8358c043ae82b346a2e6e66bc86a53bc1
+            - build/*/musl-cross-*
```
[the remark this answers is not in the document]
tlaurion: @SergiiDmytruk :

Confirmed on the saving cache steps that are happening now under https://app.circleci.com/pipelines/github/SergiiDmytruk/heads/57/workflows/dda551e0-812a-4c6a-94d6-bd59a640f0d5/jobs/780

musl-cross level 1 cache:
https://app.circleci.com/pipelines/github/SergiiDmytruk/heads/57/workflows/dda551e0-812a-4c6a-94d6-bd59a640f0d5/jobs/780/parallel-runs/0/steps/0-102:
`Warning: could not archive /root/project/build/*/musl-cross-* - Not found`

coreboot level 2 cache:
https://app.circleci.com/pipelines/github/SergiiDmytruk/heads/57/workflows/dda551e0-812a-4c6a-94d6-bd59a640f0d5/jobs/780/parallel-runs/0/steps/0-103
```
Warning: could not archive /root/project/build/*/musl-cross-* - Not found
Warning: could not archive /root/project/build/*/coreboot-* - Not found
```

level 3 cache will most probably be successful since it encompasses everything `build/` (unless it is too big... we will see):
https://app.circleci.com/pipelines/github/SergiiDmytruk/heads/57/workflows/dda551e0-812a-4c6a-94d6-bd59a640f0d5/jobs/780/parallel-runs/0/steps/0-104

tlaurion:
> (I also need to check if no-tpm code has been merged, or otherwise check with Purism codebase to do PR.)
> 
> Basically, since they had a board (mini if I recall correctly) that newly had TPM support and a float of boards without TPM, they changed their fork to have TPMTOTP/NO-TPM support and not fail close, but warn user at GUI that `TOTP: NO TPM` was detected (in case of hardware failure foing TPM->No TPM, where TPM was active in both coreboot and defined in board config).
> 
> Also, since usb keyboard is now optional (and should present under Talos because no PS2), having HID built as module (and toggle USB keyboard option in board config) would also permit to use Librem Key to measure ROM from flashrom in case there is no TPM. Untested as of now.

As expected, Heads code to switch dynamically board defined CONFIG_TPM is not merged currently: https://github.com/osresearch/heads/pull/836/files#diff-63d474ce25ab525343c0d061b286b1cd45199cd14fdf9ea535f396fd8be5de45R47

This could be the time to restart discussions in that PR for merge, or have 4 flavors for talos (workstation/server)x(with/without TPM). This is also related to kgpe-d16 boards as well.

Any thoughts?

SergiiDmytruk:
I guess a separate PR for detecting presence of `/dev/tpm0` will look better than adding two more boards. I'd also run that code only if `CONFIG_TPM=o` ("o" for optional), so that boards would have to opt in for this. If that's all there is to it, I can make a PR.

By the way, is there a reason for `librem_14` depending on `x230-hotp-maximized` in CI config? Changing dependencies like this should make build noticeably faster unless it will fail for some reason that I don't see:
```yaml
      - build_and_persist:
          name: x230-hotp-maximized
          requires:
            - prep_env

      - build_and_persist:
          name: librem_14
          requires:
            - prep_env

      - build_and_persist:
          name: talos-2_workstation
          requires:
            - prep_env

      - save_cache:
          requires:
            - librem_14
            - x230-hotp-maximized
            - talos-2_workstation
```

tlaurion:
> I guess a separate PR for detecting presence of `/dev/tpm0` will look better than adding two more boards. I'd also run that code only if `CONFIG_TPM=o` ("o" for optional), so that boards would have to opt in for this. If that's all there is to it, I can make a PR.

I like the idea, but as justified in OP by @MrChromebox, I think it is sufficient for the user to see there is no TOTP code and that TOTP code is actually: NO TPM to notify him that the TPM was not involved. Also, for boards that have TPM disk encryption key setuped, not having a TPM will fail booting.

I have no strong option on this implementation, you approach being opt-in is definitely better and would have the boards define that a TPM is optional, which is maybe better.



> By the way, is there a reason for `librem_14` depending on `x230-hotp-maximized` in CI config? Changing dependencies like this should make build noticeably faster unless it will fail for some reason that I don't see:
> 
> ```yaml
>       - build_and_persist:
>           name: x230-hotp-maximized
>           requires:
>             - prep_env
> 
>       - build_and_persist:
>           name: librem_14
>           requires:
>             - prep_env
> 
>       - build_and_persist:
>           name: talos-2_workstation
>           requires:
>             - prep_env
> 
>       - save_cache:
>           requires:
>             - librem_14
>             - x230-hotp-maximized
>             - talos-2_workstation
> ```

I admit that I don't remember if that would work or not, CircleCI was being picky. 

I understand from your question the desire of parallelizing all the builds, where the choice of current implementation was not related to saving cache but chaining workspaces between builds so that modules are not recompiled). At some point passing workspaces consumes time (downloading and decompressing multiple workspaces just overwrites the files, which is permitted by CircleCI) while if I remember correctly, caches are combined prior of being really saved, and if some same files are present, the cache will fail to be saved. Of course, testing that requires a clean build, or overriding your CircleCI project configuration's environment variable's CACHE_VERSION to something unique that was not used in the past 30 days (caches are kept only 30 days, as artifacts).

My insight is that 
```
      - save_cache:
          requires:
            - librem_14
            - x230-hotp-maximized
            - talos-2_workstation
```

Will fail for aforementioned reasons, which if I remember correctly is the reason why @Tonux599 and I went for workspace cache being passed to the latest common board reusing the workspace cache prior of using that single workspace to be used in save_cache.

Who knows if CircleCI fixed that and now permit merging multiple workspaces containing same paths of existing files that might be different. I guess their implementation just dodged the complicated choice of deciding which cache should be applied to be saved, but that bug might be a feature for them, who knows. Try it!

SergiiDmytruk:
You were right, `save_cache` [fails](https://app.circleci.com/pipelines/github/SergiiDmytruk/heads/60/workflows/2a14bd98-8a80-49f8-8bdf-66eb97d2972e/jobs/897):
```
Error applying workspace layer for job 4293e77d-61cd-41cc-972e-0d56c872aa94: Concurrent upstream jobs persisted the same file(s)
```
Build probably needs to happen in three separate directories in order to work around this.

tlaurion:
> You were right, `save_cache` [fails](https://app.circleci.com/pipelines/github/SergiiDmytruk/heads/60/workflows/2a14bd98-8a80-49f8-8bdf-66eb97d2972e/jobs/897):
> 
> ```
> Error applying workspace layer for job 4293e77d-61cd-41cc-972e-0d56c872aa94: Concurrent upstream jobs persisted the same file(s)
> ```
> 
> Build probably needs to happen in three separate directories in order to work around this.

But doing so would increase the cache size without benefit. When building, specifically the coreboot module part, one can see that everything is decompressed again, and that patches are reapplied on top of extracted archive. This is good for our use case, since the same files are basically overwritten but the compiled stuff is not if not needed, which speeds the build.

For modules cache (build/module), there is currently no need to have them cached multiple times either per arch (new here). They should be reused per cache.

One optimization that could be done here is to try to see if there would be a way to not try to combine multiple caches having the same path. But as of now, even with the addition of a new arch, the only step that increased in time exponentially is the save_cache part, since all workspaces need to be downloaded again to create the caches to be saved based on the paths defined to be included in different caches.

Any advice?

SergiiDmytruk:
I thought trading memory for time will work here. There would be 5 jobs: prep, x230-hotp-maximized and derived, librem_14 and derived, talos and save_cache. Only save_cache would download all four workspaces, the rest would need to download only prep workspace and produce their own. That said I don't really understand these multitude of caches, so you might be right that it's without benefit.

tlaurion:
> I thought trading memory for time will work here. There would be 5 jobs: prep, x230-hotp-maximized and derived, librem_14 and derived, talos and save_cache. Only save_cache would download all four workspaces, the rest would need to download only prep workspace and produce their own. That said I don't really understand these multitude of caches, so you might be right that it's without benefit.

yeah. Can try to put more thoughts in the multiple layers, but if I try to put in in simple terms of the current cache system, there is a mix a workspace caches, one for each coreboot version used per board config definitions, which now, could also be for a different arch, and might compile its own musl-cross and its own modules. So that gives us 4 workspace cache prior of Talos II inclusion (prep_env to pass previously valid save_cache for whatever biggest cache available, 4.13(x230-hotp-maximized: coreboot and built modules, 4.15: librem_14 coreboot and built modules)




Talos II piggy-backs on librem-14 cache as if it was reusing cache of coreboot 4.15 and built modules for the same arch. As of now, I do not see any hurt here.


The problem with save_cache, as you witnessed, is that it needs to have non-conflicting files between dependencies of what it should depend on. 

So if we wanted to have save_cache to be created upon both talos_II workspace (coreboot, musl-cross and built modules) and the biggest x86 workspace (currently librem14), we would have to make sure nothing conflicts between those two workspace caches.



Note that currently (but not a problem) some boards take more time then others to build. This is explainable since the boards do not build the same modules. For example, qemu-coreboot depends on whiptail, as opposed to fbwhiptail on a clean build (without cache of already built modules, as it is the cache here, and if those built modules are part of save_Cache to be restored on next build)

We could try having Talos_II_workstation depend on prep_env, and have save cache depend on both librem_14 and talos_II_workstation to speed the build. The problem there would be to remove files duplicates from what is being copied from one workspace to another being just compiled stuff (persist_to_workspace statement is home=~ and then caching `.`. This basically mean to cache everything.

Definitely, more thinking needs to be put into that. save_cache will fail if a single file is coming from two workspaces downloaded and extracted when trying to create the cache that is used to create the 3 different cache layers we keep to speed things up: 1-everything built, 2-coreboot build dirs (including coreboot's version's buildstack), 3- musl-cross

Note that each workspace layer (being a different coreboot version) takes longer build time because of coreboot buildstack being built (This is the reason you see x230-htop-maximized taking 30 minutes (building musl-cross, then 4.13 coreboot buildstack and then modules) while librem_14 takes 29 minutes (reusing musl-cross, building coreboot buildstack but reusing built modules cache), where boards taking longer then others are the boards building other modules that were not built in the passed workspace (generally whiptail vs fbwhiptail).

Each time we move from the left to the right in the images above, we pass an additional workspace cache. 
prep_env is nearly empty and only contains blobs passed along to x230-hotp-maximized ( 0second spent in "Attaching workspace")
When building librem_14 (and all other boards in that column) "Attaching workspace" takes around 1 minute attaching prep_env (12mb) and x230-hotp-maximized workspace (2.3gb)
When building librem_15v3 (and all other boards in that column) "Attaching workspace" takes around 3 minutes, attaching prep_env (12mb),  x230-hotp-maximized workspace (2.3gb) and librem_14 (3.7gb)
When saving cache, "Attaching workspace" takes around XX minutes, attaching prep_env (12mb),  x230-hotp-maximized workspace (2.3gb) and librem_14 (3.7gb)

An intuition that comes to mind would be to add a step on the workspace layers that are aimed to be cached by save_cache, to make sure that nothing conflicting files exists between those layers. In what concerns us right here right now is that adding Talos adds more then 1h30 (and counting) of build time on a clean build, but yet again, that should not happen that often; it would happen when coreboot modules definition change, when new patches are added under ./patches and when the Makefile changes. As long as a step doesn't take more then one hour, we do not have a real problem, but being patient when no cache (everything. coreboot, musl-cross) can be reused because hash digests mismatches on coreboot module/patches/Makefile.

tlaurion:
@SergiiDmytruk : can you trigger a same commit rebuild from CircleCI interface? I would love to see how much time takes a rebuild on cache for all platforms, to prioritize CircleCI revamp. Prior of this PR, building from "Everything" cache (which will be the case here) took 30 minutes. If that delays is increased a lot, CircleCI workspace content being cached will need some investigation prior of merging and affecting all builds.

tlaurion:
@SergiiDmytruk :


I pushed your PR to build on my CircleCI instance which built without cache (2h46 build time):
https://app.circleci.com/pipelines/github/tlaurion/heads/1149/workflows/6317c668-b11b-4bcc-9657-c6c8842508f2

Triggered a rebuild to build from cache (everything's cache: 1h48):
https://app.circleci.com/pipelines/github/tlaurion/heads/1149/workflows/69aa6837-69e4-4d14-9969-9621f17e052c

Latest rebuild on master using "everything" cache took (51m).

Not the end of the world if we were to merge as is, but the above intuitions were confirmed, which most of the time difference between clean build and cache based rebuild is due to passing the massive cache between board dependencies (workspace cache being passed being past layer + actual layer) up to the last step of the build speaks the most:
https://app.circleci.com/pipelines/github/tlaurion/heads/1149/workflows/69aa6837-69e4-4d14-9969-9621f17e052c/jobs/9262


The same cache is downloaded 3 times (3x6.8gb for a total of 19g downloaded and decompressed under 9 minutes).
Build time of the board on cache: 6 minutes (expected)
Then another 14 minutes persisting to workspace that gigantic layer to be passed to save_cache which spends 13 minutes downloading the same layers for nothing: a cache existing for all 3 layers with the same key (matching modules Makefile and patches).


Basically, at this point, caching trades time building to time downloading/unpacking. CircleCI doesn't care and see time as time.
So we could probably parallelize the builds more, at this point and maybe give up on saving caches if we had a docker image including musl-cross and coreboot musl-cross'ed buildstack per version of what we use, but that is another story.

Some additional introspection/intuition notes on caches:
- not sure if workspace layers cache need to obey the same as save_cache with known exact paths.
for example, we could maybe workspace cache:
/build/linux-version/$BOARD
/build/coreboot-version/$BOARD

-There is also configurable options to have different CI configs used per CircleCI config
- It doesn't really make any sense to have a single cache used by a coreboot 4.15 board to also have the cache layer of talos II, not older coreboot version caches....
- It doesn't really make sense to have x86 have the ppc64 cache.
But all of those would make sense to be in a docker, since a docker that is reused and has not changed is cached by CircleCI and not redownloaded every time. Again, that would require a Makefile coreboot.musl target that could called when building the docker image.

Again if you have any insights on any small optimization that could be applied here, let me know. Otherwise I have hopes that having a docker image and or working around directory structure in the build process could permit us to use caches more efficiently (./build/x86, /build/ppc64 could easily be hardcoded in CI/CD and passed as non-conflicting files from the last target into the cache, where first worker of a cache could simply remove the arch/coreboot versions unreleated to them to be passed along without conflicting, and then we could have a dummy common task (merge_workspaces) downloading all workspace caches and combining them into a final cache to be downloaded once per prep_env, passed to workers (which could delete anything unrelated to them. That is the best I came up with as of now.

SergiiDmytruk:
My rebuild with cache also took 1h 48m, I thought you saw it. I pushed an attempt similar to the branching I had before, which fails at `save_cache`, but only moved Talos to a separate branch and constrained workspace to arch-specific directories ([here](https://app.circleci.com/pipelines/github/SergiiDmytruk/heads/65/workflows/887d7d77-1711-464c-be7f-52fafe4195e3)). If it works, CI running time should be similar to values without Talos.

I now see that my naive idea with 5 jobs will hit limits. Docker image sounds promising, because overhead of workspaces is way too large. I was also wondering whether `ccache` could be helpful, it's installed, but doesn't seem to be used. Files in `~/.cache/ccache` that have matching names should be identical, but if CircleCI fails to merge identical files the idea to cache cache rather than output files won't work.

tlaurion:
@SergiiDmytruk seems like I missed this glob documentation https://circleci.com/docs/parallelism-faster-jobs#globbing-test-files

Will point commits to here for individual tests of your faster-ci branch

SergiiDmytruk:
My last attempt didn't have globs and still failed. Judging from the logs the paths should start with `project/...`, so it might work this time, then can return to experiments with globs.

tlaurion:
> My last attempt didn't have globs and still failed. Judging from the logs the paths should start with `project/...`, so it might work this time, then can return to experiments with globs.

Exactly what my intuition gave as well seeing https://app.circleci.com/pipelines/github/tlaurion/heads/1155/workflows/16ea4be7-b7da-49a7-bf18-8c744a2a0687/jobs/9392/parallel-runs/0/steps/0-108 output:
```
Found a cache from build 9230 at heads-modules-and-patches-HPrq963VYPLJeLxLr181U2MT1a3QYHAudaIgR4N9beg=<no value>
Size: 6.4 GiB
Cached paths:
  * /root/project/crossgcc
  * /root/project/build
  * /root/project/packages
  * /root/project/install

Downloading cache archive...
Unarchiving cache...
```

I think this might work!

tlaurion:
Rebuilding with clean cache my branch (which is the same as yours) changing CACHE_VERSION under CircleCI project settings to talos.

Intuition is that packages will be downloaded at each board, since no ./build/arch/package_name/.canary will be found unless second layer workspace have that configured.
That might be problematic or not.

This works!
@SergiiDmytruk : you should probably have save_cache depend on Talos II server, so that your ppc64 cache contains both whiptail and fbwhiptail built dependencies to have less long builds on builds based on cache. Of course, no cache is saved in your latest test, since cache keys have not changed.

tlaurion:
@SergiiDmytruk some notes on your last build https://app.circleci.com/pipelines/github/SergiiDmytruk/heads/71/workflows/9b7ca741-8921-4c71-8200-525d6a151448 for https://github.com/SergiiDmytruk/heads/commit/087b754e226c1e2501dd3424c58291ca72cb1342 which included two patches as opposed to last full build that successfully created a cache.

https://app.circleci.com/pipelines/github/SergiiDmytruk/heads/71/workflows/9b7ca741-8921-4c71-8200-525d6a151448/jobs/1111/parallel-runs/0/steps/0-108
```
No cache is found for key: heads-modules-and-patches-_HRCdgnWzy9K7C1h2vPMrcofQlTuTqUhlZCb+2+6dC4=ci-opt-build
Found a cache from build 1052 at heads-coreboot-musl-cross-vb2jPQCaIG8cDtAF01mvN1JR0NxouC8UzROMPf+3mWo=ci-opt-build
Size: 4.7 GiB
Cached paths:
  * /root/project/crossgcc
  * /root/project/build/x86/musl-cross-38e52db8358c043ae82b346a2e6e66bc86a53bc1
  * /root/project/build/ppc64/musl-cross-38e52db8358c043ae82b346a2e6e66bc86a53bc1
  * /root/project/packages
  * /root/project/build/x86/coreboot-4.11
  * /root/project/build/x86/coreboot-4.13
  * /root/project/build/x86/coreboot-4.14
  * /root/project/build/x86/coreboot-4.15
  * /root/project/build/ppc64/coreboot-git

Downloading cache archive...
Unarchiving cache...
```

This is why builds took a little longer, they were only taking coreboot cache from built modules. So a rebuild should take less time then what we see here, which is good news.

tlaurion:
As noted under https://github.com/SergiiDmytruk/heads/commit/9879e5853e898da2414ca041986e32f647260272

I doubt save_cache https://app.circleci.com/pipelines/github/SergiiDmytruk/heads/72/workflows/a1515fdc-e92b-4276-b35b-ee114ba66de0 will work, since packages dir will contain conflicting files :/

tlaurion:
> Rebuilding from clean commit ( Changed CACHE_VERSION under CircleCI project) with save_cache depending on Talos II server so that cache contains variants of whiptail and fbwhiptail https://app.circleci.com/pipelines/github/tlaurion/heads/1158/workflows/ee43e8e4-d487-43a9-b705-2caa833c7948
> 
> This is really good progress for CircleCI caching! Thanks @SergiiDmytruk !

Hmmm. We would need a way to skip already applied patches if packages cannot be cached, otherwise patches creating files will fail, just like it is happening under https://app.circleci.com/pipelines/github/tlaurion/heads/1158/workflows/ee43e8e4-d487-43a9-b705-2caa833c7948/jobs/9467?invite=true#step-103-3342 for the already created `src/syscfg/lock-obj-pub.powerpc64le-unknown-linux-musl.h` 

The actual logic is that decompressed archives have ./packages/package-version_verify file created after file integrity validation, which tells archives to not be decompressed again nor patched on top of it if it was already downloaded, verified and extracted.

tlaurion:
Not sure if the following would resolve the issue, but we could:
- skip the strong dependency between build/arch/package_name-version/.canary and ./packages/package_name-version_verify, so that if canary exists, package is not re-extracted and re-patched if package is downloaded while build dir exists (so that patches creating files are not making builds fail)
- have the packages downloaded under ./arch/packages, but that would be a tweak oriented to CI/CD environments for caching,

> No cache is found for key: heads-modules-and-patches-_HRCdgnWzy9K7C1h2vPMrcofQlTuTqUhlZCb+2+6dC4=ci-opt-build
> Found a cache from build 1052 at heads-coreboot-musl-cross-vb2jPQCaIG8cDtAF01mvN1JR0NxouC8UzROMPf+3mWo=ci-opt-build
> Size: 4.7 GiB
> Cached paths:
>   * /root/project/crossgcc
>   * /root/project/build/x86/musl-cross-38e52db8358c043ae82b346a2e6e66bc86a53bc1
>   * /root/project/build/ppc64/musl-cross-38e52db8358c043ae82b346a2e6e66bc86a53bc1
>   * /root/project/packages
>   * /root/project/build/x86/coreboot-4.11
>   * /root/project/build/x86/coreboot-4.13
>   * /root/project/build/x86/coreboot-4.14
>   * /root/project/build/x86/coreboot-4.15
>   * /root/project/build/ppc64/coreboot-git
> 
> Downloading cache archive...
> Unarchiving cache...

@SergiiDmytruk : It is to note that the builds succeeded because ./packages was part of the cache here too.

tlaurion:
#1188 was merged, so two additional boards+coreboot config files need #1009 relative fixes

tlaurion:
> > My last attempt didn't have globs and still failed. Judging from the logs the paths should start with `project/...`, so it might work this time, then can return to experiments with globs.
> 
> Exactly what my intuition gave as well seeing https://app.circleci.com/pipelines/github/tlaurion/heads/1155/workflows/16ea4be7-b7da-49a7-bf18-8c744a2a0687/jobs/9392/parallel-runs/0/steps/0-108 output:
> 
> ```
> Found a cache from build 9230 at heads-modules-and-patches-HPrq963VYPLJeLxLr181U2MT1a3QYHAudaIgR4N9beg=<no value>
> Size: 6.4 GiB
> Cached paths:
>   * /root/project/crossgcc
>   * /root/project/build
>   * /root/project/packages
>   * /root/project/install
> 
> Downloading cache archive...
> Unarchiving cache...
> ```
> 
> I think this might work!

As can be seen here, the old cache for `heads-modules-and-patches-HPrq963VYPLJeLxLr181U2MT1a3QYHAudaIgR4N9beg` (biggest cache available) included packages as well, which explains why builds were successful (_verify and .canary files being present)

SergiiDmytruk:
> Wondering if I should merge #1201 first prior of your master rebasing?

Rebased. Order shouldn't matter. I don't think the two PRs conflict. This one doesn't add any new `patch` invocations.

tlaurion:
@SergiiDmytruk reverted #1201 Makefile changes since it broke all builds. Not sure I understand what happens, dropped you messages under Matrix.

tlaurion:
@SergiiDmytruk please take a look at https://github.com/tlaurion/heads/tree/replace_util-linux_agetty_busybox_getty which is built on top of this branch.

Basically, those two changes:
- https://github.com/tlaurion/heads/commit/d020609ced719cc6f5698767d75daac9b4255302#diff-7906e5e771d8da6406d4081c2f8e0b51443c1dd41b0300f55f50c016034e9077R528
- https://github.com/tlaurion/heads/commit/d020609ced719cc6f5698767d75daac9b4255302#diff-63d474ce25ab525343c0d061b286b1cd45199cd14fdf9ea535f396fd8be5de45R140-R142

As a result, we would be not asking all board roms to have a 70kb agetty binary that most do not need to use, where adding getty from busybox costs an additional 1k.

```
user@heads-tests:/tmp/talos-without-getty-in-busybox$ wget https://output.circle-artifacts.com/output/job/c9af1bd2-335e-4712-bbc9-34d3946ca4ee/artifacts/0/build/ppc64/talos-2_server/initrd.cpio.xz
user@heads-tests:/tmp/talos-without-getty-in-busybox$ xz -d initrd.cpio.xz ; for i in initrd.cpio modules.cpio tools.cpio heads.cpio; do cpio -i < $i; done
user@heads-tests:/tmp/talos-without-getty-in-busybox$ ls bin/busybox -al
-rwx------ 1 user user 724952 Aug 31 12:23 bin/busybox
user@heads-tests:/tmp/talos-without-getty-in-busybox$ ls -al bin/agetty 
-rwx------ 1 user user 68376 Aug 31 12:23 bin/agetty
```

```
user@heads-tests:/tmp/testing-getty-in-busybox$ wget https://output.circle-artifacts.com/output/job/64aaf2b4-bf77-4679-8c85-1eff5a92fa3b/artifacts/0/build/ppc64/talos-2_workstation/initrd.cpio.xz
user@heads-tests:/tmp/testing-getty-in-busybox$ xz -d initrd.cpio.xz ; for i in initrd.cpio modules.cpio tools.cpio heads.cpio; do cpio -i < $i; done
user@heads-tests:/tmp/testing-getty-in-busybox$ ls -al bin/busybox 
-rwx------ 1 user user 725040 Aug 31 12:48 bin/busybox
```

tlaurion:
@SergiiDmytruk feel free to borrow code from https://github.com/tlaurion/heads/commit/2e2a6c04f8daf6c00be907c65afae7b9e1763e4e if it is doing its job. I was just a bit against adding util-linux's agetty when busybox could provide the same functionality with near to none footprint.

SergiiDmytruk:
`getty` works from recovery shell (how I tested it), but doesn't seem to work from `/init`. Doesn't error, just does nothing there. `/dev/tty0` should be in the same state there, don't see why it should fail.

How would you include `agetty` binary only for Talos boards? Something like `CONFIG_USE_AGETTY=y` in config and then like this in `modules/util-linux`?
```
util-linux_agetty-$(CONFIG_USE_AGETTY) := agetty
util-linux_output := $(util-linux_agetty-y)
```

tlaurion:
> `getty` works from a recovery shell (how I tested it), but doesn't seem to work from `/init`. Doesn't error, just does nothing there. `/dev/tty0` should be in the same state there, don't see why it should fail.
> 
> How would you include `agetty` binary only for Talos boards? Something like `CONFIG_USE_AGETTY=y` in config and then like this in `modules/util-linux`?
> 
> ```
> util-linux_agetty-$(CONFIG_USE_AGETTY) := agetty
> util-linux_output := $(util-linux_agetty-y)
> ```

@SergiiDmytruk yes, exactly. 
Best example for this is how board config options are encompassing kernel modules to be added by kernel under module/linux.

```
user@heads-tests:~/heads$ grep -R CONFIG_LINUX_E1000E boards/x230-hotp-maximized modules/ Makefile
boards/x230-hotp-maximized/x230-hotp-maximized.config:CONFIG_LINUX_E1000E=y
boards/x230-hotp-maximized/x230-hotp-maximized.config:#SSH server (requires ethernet drivers, eg: CONFIG_LINUX_E1000E)
modules/linux:linux_modules-$(CONFIG_LINUX_E1000E)	+= drivers/net/ethernet/intel/e1000e/e1000e.ko
```

tlaurion:
> `getty` works from a recovery shell (how I tested it), but doesn't seem to work from `/init`. Doesn't error, just does nothing there. `/dev/tty0` should be in the same state there, don't see why it should fail.
> 
> How would you include `agetty` binary only for Talos boards? Something like `CONFIG_USE_AGETTY=y` in config and then like this in `modules/util-linux`?
> 
> ```
> util-linux_agetty-$(CONFIG_USE_AGETTY) := agetty
> util-linux_output := $(util-linux_agetty-y)
> ```

@SergiiDmytruk yes, exactly. 
Best example for this is how board config options are encompassing kernel modules to be added by kernel under module/linux.

```
user@heads-tests:~/heads$ grep -R CONFIG_LINUX_E1000E boards/x230-hotp-maximized modules/ Makefile
boards/x230-hotp-maximized/x230-hotp-maximized.config:CONFIG_LINUX_E1000E=y
boards/x230-hotp-maximized/x230-hotp-maximized.config:#SSH server (requires ethernet drivers, eg: CONFIG_LINUX_E1000E)
modules/linux:linux_modules-$(CONFIG_LINUX_E1000E)	+= drivers/net/ethernet/intel/e1000e/e1000e.ko
```

tlaurion:
A more omplicated example was for ifdtool under coreboot module, needed to be crosscompiled where ifdtool also needed to be natively compiled for coreboot to continue to be able to pack firmware.

Giving example because this one was relative to module binary, not kernel module as under linux.
Otherwise, everything under Heads is include/exclude as of now, we are now starting to create exceptions :)

https://github.com/osresearch/heads/pull/1195/commits/a4b5381e65f0f2da5fbbe742696ad6b15f1b9bc2#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R464-R479

PR #1016, opened 2021-08-07
Display TOTP secret as text

elliotvirzi:
Enable use of TOTP devices without a camera

### DIFF
--- a/initrd/bin/seal-totp
+++ b/initrd/bin/seal-totp
@@ -87,4 +87,4 @@ url="otpauth://totp/$HOST?secret=$secret"
 secret=""
 
 qrenc "$url"
-#echo "$url"
+echo "$url"


tlaurion:
Not against. A lot of users actually ask for this since they do not want to rely on smartphones for TPMTOTP remote attestation. 

Doing so here would permit to have the Qr Code AND secret outputted on screen one after the other.

Showing the Qr code on screen is the equivalent  of the url as the encoded form, where the Qr code removes the possibility of user input error of manually transmitting through keystroke. The only concern here, equal to both secret output on screen, is that the user should do so only in safe environment where no eavesdropping can occur through external video surveillance.

Maybe adding a warning (confirmation: Y/n, Y being the default) prior of showing Qr Code + url should be added prior of outputting that secret on screen so users are warned that such secret will be outputted and may decide to do so in a later time when they are confortable with such secret being showed on their screen?

Users desiring to have to input manually url in their second computer KeepassXC or other passphrase managers can be considered already aware of the lesser convenience of needing that second computer to do the remote attestation of TOTP there vs using a smartphone which they normally have with them all the time (necessity of having the second computer to validate TOTP code). 

@elliotvirzi thoughts?

elliotvirzi:
My use case for displaying the URL is to use a programmable TOTP device (such as the [Token 2 Molto-1-i](https://www.token2.com/shop/product/token2-molto-1-i-multi-profile-totp-hardware-token)).  For me, this is advantageous since (1) it doesn't require a smartphone and (2) since I can set a PIN, I feel comfortable storing the device alongside my laptop since it would be difficult to tamper with, unlike a device such as the Nitrokey Pro v2 or the Librem Key.

Regarding adding a confirmation: I could see that feature being useful to some users, but I personally would not find it useful.  Regardless, I believe that is independent of this PR and might be better suited in a new issue or PR.

tlaurion:
> it would be difficult to tamper with, unlike a device such as the Nitrokey Pro v2 or the Librem Key.

@elliotvirzi Could you elaborate on this?

tlaurion:
@elliotvirzi updated https://github.com/osresearch/heads/pull/1016#issuecomment-894832695

tlaurion:
@MrChromebox : not against.

MrChromebox (approved):

PR #1025, opened 2021-09-15
etc/functions: Filter boot device options with '/dev/'

MrChromebox:
Grepping on just 'Disk' can lead to disk UUID identifier strings
being added to /tmp/disklist, which then fail to parse later on.
Avoid this by grepping on 'Disk /dev' instead.


MrChromebox:
whoops, already pushed this as #955

PR #1032, opened 2021-09-23
gui-init overhaul

MrChromebox:
* Refactor gui-init so that all operations are contained with shell functions, leading to much clearer program flow between menus and sub-menus
* Ensure when user selects 'skip/continue to main menu' that is what happens (no further prompts)
* Improve consistency of background color use on error/warning screens, and persist background color / error state from initial checks to the main menu to sub-menus
* Improve error handling when no fixed disks used/found on system


tlaurion:
Wow😃
Will review more in depth but this is awesome!

Expect merge day middle of next week

MrChromebox:
I tested all these with qemu-fbwhiptail, so only really scratched the surface. I'll be testing on the L14 tomorrow/next week and may have a few additional fixes/commits based on how that goes

tlaurion:
@MrChromebox any plan on moving gui functions to /etc/gui_functions ?

MrChromebox:
I suppose that's an option. Where would we draw the line between what goes in gui_functions and gui_init?

tlaurion:
Well, the way you are heading would permit calling

. /etc/gui_functions if whiptail is present and permit to do a lot without leaving GUI and needing To go into authenticated  recovery shell.

Then maybe even unifying them with /etc/functions later.

This would remove redundancy (mount_boot, verify_global_checkums and others ) while permitting other scripts to reuse what is there and cleaner code in the long run?

If the main scripts (policies) were calling only functions which are doing what they are supposed to be doing,  Heads would become way more flexible,  which seems to be the desire of the community with forks like @hardenedvault with its own autoboot script now, tpm2 included and system76 systems interest from the community that seems to be present out there.

More reusability means more collaboration and endorsement which is somehow a really good thing for a project like this to move faster.

MrChromebox:
> More reusability means more collaboration and endorsement which is somehow a really good thing for a project like this to move faster.

of course, that's certainly one of the goals of this effort. Moving things into /etc/gui-functions only makes sense IMO if we're going to split things up such that the same function names can be used with or without whiptail, and the versions in /etc/gui_functions mirror those in /etc/functions just with whiptail. We'd need to split the common stuff out so it could be included in both cases.

Primarily though, my goal here is to improve the reliability and UX of the gui-based boards, since that's what's of concern to Purism's users. Refactoring to improve readability, reuse, and ease of future enhancements is a close 2nd though :)

MrChromebox:
@tlaurion ok, done adding to this, should be ready for review/testing

tlaurion:
Oupsies everyone. This is merged as per #955 having been merged first.

PR #1033, opened 2021-09-29
fix bug with e.g. nosplash parameter in kernel command line

chri2:
Strings from $cmdremove should only be removed from $cmdline if they are enclosed by spaces of if they are at the beginning of $cmdline followed by a space or if they are at the end of $cmdline prepended by a space.

For more information about the bug fixed by this commit please look at https://source.puri.sm/firmware/pureboot/-/issues/4 .

If the kernel command line configured contains eg. 'nosplash' and 'splash' should have been removed from the kernel command line this left over 'no' (without the 'splash'). This led to a situation with pureboot that disabled key entry when the Nitrokey is used for decryption of a luks keyslot.

### DIFF
--- a/initrd/bin/kexec-boot
+++ b/initrd/bin/kexec-boot
@@ -49,7 +49,9 @@ adjusted_cmd_line="n"
 adjust_cmd_line() {
 	if [ -n "$cmdremove" ]; then
 		for i in $cmdremove; do
-			cmdline="${cmdline//$i/}"
+			cmdline="${cmdline#$i }"
+			cmdline="${cmdline/ $i / }"
+			cmdline="${cmdline% $i}"
 		done
 	fi
 


tlaurion:
@MrChromebox the patch here is taking [3 cases into consideration](https://source.puri.sm/firmware/pureboot/-/merge_requests/12)
- prepended by a space, enclosed by spaces or beginning of cmdline.

@chri2 I am not sure here what is the real use case of this change, since KERNEL_ADD and KERNEL_REMOVE defined under board config should remove only default boot configuration that are unwise for a specific board, and/or for development purposes, where the real changes, if important, should be applied into OS's grub config?

MrChromebox:
@tlaurion yes I understand, but that seems to be the "long" way to do an exact word match. 

Something like: `cmdline=$(echo $cmdline | sed 's/\b$1\b//g)`

PR #1034, opened 2021-09-30
Bump coreboot and Linux for qemu-coreboot

bluecmd:
The older versions do not build cleanly under latest
OSes like Ubuntu 21.04. Bump the versions to the
latest ones that are already defined.


tlaurion:
Just a quick note. The reason this hasn't been merged is that the qemu boards are meant to fit the lower end machines supported. By increasing CBFS_SIZE to 8mb, users trying to make a configuratio nhat would fit other devices limited by 7.5Mb would fail to see a problem. 5.x kernels are way bigger then 4.14 counterpart, even if the extended  4.14 .config is passed to newer kernel.

It seems that make tinyconfig and specifying requirements might be a guide to do to ease maintainership in the future, otherwise just passing to new kernel version just explodes in size and is not workable.

Another point going into #590 

Some attempts were made recently to pass all modules from -O3 and -O2 to -Os, which actually liberated a lot of space, while not enough to have legacy x220 boards (non-maximized, having 7.5mb of space) have cryptsetup2 module compiled in.

So bumping to coreboot 4.13 was done, while upgrading to newer kernel is not done.

MrChromebox:
coreboot 4.16 was just released FYI :)

PR #1035, opened 2021-09-30
Fix make 4.2.1 to build on newer systems

bluecmd:
This patch makes it possible to build make 4.2.1 using newer systems
that otherwise would complain about a missing definition of __stat.

Issue #983, opened 2021-03-03
Fedora 33 IASL build failure


osresearch:
```
/build/make-4.2.1/make 
/home/hudson/heads/./build/make-4.2.1/make -C "/home/hudson/heads/build/coreboot-4.8.1" CPUS=128 iasl
make[1]: Entering directory '/home/hudson/heads/build/coreboot-4.8.1'
Welcome to the coreboot cross toolchain builder v1.52 (June 11th, 2018)

Building toolchain using 128 thread(s).

Downloading and verifing tarballs ...
 * acpica-unix2-20180531.tar.gz (cached)... hash verified ("17717140438d506533b4a56e34350749d7b84d6c")
Downloaded tarballs ... ok
Unpacking and patching ...
Unpacked and patched ... ok
Building packages ...
Building IASL v20180531 for host ... failed. Check 'build-IASL/build.log'.
make[2]: *** [Makefile:36: build_iasl] Error 1
make[1]: *** [util/crossgcc/Makefile.inc:52: iasl] Error 2
make[1]: Leaving directory '/home/hudson/heads/build/coreboot-4.8.1'
make: *** [modules/coreboot:79: "/home/hudson/heads/build/coreboot-4.8.1/util/crossgcc/xgcc/bin/iasl"] Error 2
```

and the IASL build log:

```
make[1]: Entering directory '/home/hudson/heads/build/coreboot-4.8.1/util/crossgcc/acpica-unix2-20180531/generate/unix/iasl'
- Link obj/iasl
/usr/bin/ld: obj/aslcompilerparse.o:(.bss+0x8): multiple definition of `AslCompilerlval'; obj/aslcompilerlex.o:(.bss+0x0): first defined here
/usr/bin/ld: obj/prparserlex.o:(.bss+0x0): multiple definition of `LexBuffer'; obj/dtparserlex.o:(.bss+0x0): first defined here
collect2: error: ld returned 1 exit status
make[1]: *** [../Makefile.rules:15: obj/iasl] Error 1
make[1]: Leaving directory '/home/hudson/heads/build/coreboot-4.8.1/util/crossgcc/acpica-unix2-20180531/generate/unix/iasl'
make: *** [Makefile.common:7: iasl] Error 2
cp: cannot stat 'bin/iasl': No such file or directory
```

osresearch:
Appears to have been fixed in a more recent version (26 March 2020): https://acpica.org/sites/acpica/files/changes_50.txt

[...]

Issue #1036, opened 2021-10-06
Build error on Ubuntu 21.04


copyvar:
I need some help to build heads. I installed all packages mentioned in line 13 https://github.com/osresearch/heads/blob/master/.circleci/config.yml

```
...
make[3]: Entering directory '/home/test/heads/build/make-4.2.1/doc'
make[3]: Nothing to be done for 'all'.
make[3]: Leaving directory '/home/test/heads/build/make-4.2.1/doc'
make[3]: Entering directory '/home/test/heads/build/make-4.2.1'
gcc  -g -O2 -Wl,--export-dynamic  -o make ar.o arscan.o commands.o default.o dir.o expand.o file.o function.o getopt.o getopt1.o guile.o implicit.o job.o load.o loadapi.o main.o misc.o posixos.o output.o read.o remake.o rule.o signame.o strcache.o variable.o version.o vpath.o hash.o remote-stub.o glob/libglob.a   -ldl 
/usr/bin/ld: glob/libglob.a(glob.o): in function `glob_in_dir':
/home/test/heads/build/make-4.2.1/glob/glob.c:1269: undefined reference to `__stat'
/usr/bin/ld: glob/libglob.a(glob.o): in function `glob':
/home/test/heads/build/make-4.2.1/glob/glob.c:1036: undefined reference to `__stat'
/usr/bin/ld: /home/test/heads/build/make-4.2.1/glob/glob.c:814: undefined reference to `__stat'
/usr/bin/ld: /home/test/heads/build/make-4.2.1/glob/glob.c:960: undefined reference to `__stat'
collect2: error: ld returned 1 exit status
make[3]: *** [Makefile:656: make] Error 1
make[3]: Leaving directory '/home/test/heads/build/make-4.2.1'
make[2]: *** [Makefile:798: all-recursive] Error 1
make[2]: Leaving directory '/home/test/heads/build/make-4.2.1'
make[1]: *** [Makefile:534: all] Error 2
make[1]: Leaving directory '/home/test/heads/build/make-4.2.1'
LANG=C MAKE=/home/test/heads/build/make-4.2.1/make /home/test/heads/build/make-4.2.1/make  all
/bin/sh: 1: /home/test/heads/build/make-4.2.1/make: not found
make: *** [Makefile:668: all] Error 127
```

copyvar:
I am stuck after hours of research in the web. Anyone got an idea?

paulmenzel:
[...]

PR #1038, opened 2021-10-11
CircleCi on ubuntu 2104

tlaurion:
CircleCI was stuck to debian 10 for a while.
Thanks @bluecmd and @itworksfinebro for individual proposed fixes, that merged together in this PR, should permit building of older boards on newer OSes.

This PR:
- Attempts inclusion of #1035 
- Automates manual fix proposed by @itworksfinebro #983 (comment) in form of coreboot 4.8.1 and 4.11 patches
- Removes manual https to http patch that was pushed in emergency when upstream mirrors were not upgrading their certs while I was doing build tests back in time.

Should fix :
#983
#1036

Close #1035

Waiting for the result of the build here. My CircleCI is rate limiting threads since i'm overusing my instance. Pushing it as PR should use @osresearch's and give faster result.


MrChromebox:
@tlaurion why are we not bumping iasl for coreboot 4.11 the same as we are for 4.81?

On the Librem 13/15 with Heads built using coreboot 4.9+, Qubes 4.0.x freezes right after kexec with no useful output. I'll retest with coreboot 4.14 and Qubes 4.1 RC and report back

tlaurion:
> @tlaurion why are we not bumping iasl for coreboot 4.11 the same as we are for 4.81?
> 

Could definitely give it a try for 4.11.... But for the moment, I have no clue on how to make 4.8.1 support forward. Could swap boards builds on CI so that a 4.11 board is built first bumping IASL. Patches welcome.

tlaurion:
Before merge:

-[] 4.8.1 boards needs to compile or boards need to be upgraded to 4.13+

-[] 4.11 kgpe-16 CiecleCI artifact needs to be tested

MrChromebox:
> @MrChromebox have your confirmed https://1959-103208611-gh.circle-artifacts.com/0/build/librem_l1um/heads-librem_l1um-v5.0.1-53-ga57e41b.rom works as expected too?

yes, tested working

tlaurion:
@MrChromebox 4.8.1 coreboot builds on newer OSes is stuck here with current patchset https://github.com/osresearch/heads/pull/1038#issuecomment-943721373

tlaurion:
Next

tlaurion:
Was fixed a while ago, circleci being based on debian-11 and the only supported path as of now.

PR #1041, opened 2021-10-19
System info: add battery status (health and charge)

tlaurion:
Actual full charge capacity vs design full charge capacity is an important thing to know. For OEMs QA as for users.

This adds 
- a calc function under /etc/functions
- a print_battery_health function under /etc/functions which returns current battery capacity/factory capacity. (percent integer)
- a print_battery_charge function under /etc/functions which returns current battery capacity/factory capacity. (percent integer)
- an additional battery status block under System Info that is present only if BAT0 exists for the moment. Otherwise the output is unchanged.
- Changed header of etc/gui_functions since gui-init is usable without fbwhiptail (whiptail: without fbwhiptail nor cairo).

Why: 
- To be able to test for battery health in custom init scripts.
- For users to know if their battery not lasting long is because really used and quickly ask for replacement.
- For OEMs to be able to test this in custom scripts if desired prior of order fulfillment...

Edited after discussion and force push.

### DIFF
--- a/initrd/bin/gui-init
+++ b/initrd/bin/gui-init
@@ -474,12 +474,19 @@ reset_tpm()
 
 show_system_info()
 {
+  battery_charge="$(print_battery_charge)"
+  battery_health="$(print_battery_health)"
+  if [ -n $battery_charge -a -n $battery_health ];then
+	  battery_status="\nBattery charge: $battery_charge%\nBattery health: $battery_health%\n"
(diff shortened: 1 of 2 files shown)


tlaurion:
@MrChromebox This is health status of the battery.

MrChromebox:
@tlaurion tested working on a Librem 14. LGTM

PR #1042, opened 2021-10-19
gui-init: seperate gui functions from gui init

tlaurion:
Includes #1041

This is a proposition to clarify what needs to be into etc/functions, /etc/gui_functions and gui-init in the goal of having reusable functions and more specialized init scripts while reusing at the maximum shared functions.

In this simple variation from master:
- etc/functions is imported by etc/gui_functions
- tmp/config is imported in etc/gui_functions
- gui-init is left in a really bare condition

@MrChromebox : you would want to clean things more then this? From basic testing, everything works as it is. I came to realize that no etc/gui_functions were even imported buy gui-init as of now. We could remove duplication between etc/functions and etc/gui_functions, where etc/gui_functions could simply be screen prompts/gui selectors, and where etc/functions be the actual functions doing the work, which could then be called in an unattended way if some variables are not populated (empty).

The reason I think this is important is that other init scripts could be created to automate some processes, importing required input from outside sources where/when needed.

Ping me if your want to discuss this more dynamically.


tlaurion:
Done.

PR #1043, opened 2021-10-21
oem-factory-reset: fix custom comment entry

MrChromebox:
copy/paste error resulted in user-entered comment never
being set/checked/used, and email address being overwritten.
Fix variable usage so comment and email are set correctly.


tlaurion:
Oups :/

PR #1046, opened 2021-10-27
Update Librem boards to use coreboot 4.15

MrChromebox:
- Update coreboot version to 4.15
- Remove blob download scripts
- Update purism-blobs module
- Update coreboot config for Qubes compatibility
- Update/Unify IOMMU params across Librem boards
- Remove extraneous config param from Librem 14

IMPORTANT: Updating the Librem 13/15 boards to use coreboot 4.15 breaks compatibility with Qubes 4.0.x; Qubes 4.1-RC1 or later must be used.


tlaurion:
CircleCI context deadline exceeded @MrChromebox ?

tlaurion:
We might want to merge #1002 first and apply the changes needed in each boards after?

MrChromebox:
@tlaurion no opinion either way on merging #1002 first.

Also not sure how to mitigate the context deadline issue, given I didn't change anything related to building for coreboot-qemu

tlaurion:
> Also not sure how to mitigate the context deadline issue, given I didn't change anything related to building for coreboot-qemu

So confused about this 
https://circleci.com/blog/building-open-source-projects-on-circleci/

modules/coreboot
```diff
@@ -42,7 +42,7 @@ CONFIG_COREBOOT_CONFIG ?= config/coreboot-$(BOARD).config
 # Ensure that touching the config file will force a rebuild
 $(build)/$(coreboot_dir)/.configured: $(CONFIG_COREBOOT_CONFIG)
 
-EXTRA_FLAGS := -fdebug-prefix-map=$(pwd)=heads -gno-record-gcc-switches -Wno-error=packed-not-aligned
+EXTRA_FLAGS := -fdebug-prefix-map=$(pwd)=heads -gno-record-gcc-switches -Wno-error=packed-not-aligned -Wno-error=address-of-packed-member
```
tlaurion, line 45: @MrChromebox This breaks builds for non coreboot 4.15 boards.

Better approach is to define per coreboot version and leave the generic line if not specified
```

diff --git a/modules/coreboot b/modules/coreboot
index 10cc9088..806ee608 100644
--- a/modules/coreboot
+++ b/modules/coreboot
@@ -29,6 +29,7 @@ else ifeq "$(CONFIG_COREBOOT_VERSION)" "4.15"
         coreboot_hash := 20e6aaa6dd0eaec7753441c799711d1b4630e3ca709536386f2242ac2c8a1ec5
         coreboot-blobs_hash := c0e2d8006da226208ba274a44895d102cb2879cf139cc67bba5f62e67b871f6d
         coreboot_depends := $(if $(CONFIG_PURISM_BLOBS), purism-blobs)
+       EXTRA_FLAGS := -fdebug-prefix-map=$(pwd)=heads -gno-record-gcc-switches -Wno-error=packed-not-aligned -Wno-error=address-of-packed-member
 else
        $(error "$(BOARD): does not specify coreboot version under CONFIG_COREBOOT_VERSION")
 endif 
@@ -47,7 +48,7 @@ CONFIG_COREBOOT_CONFIG ?= config/coreboot-$(BOARD).config
 # Ensure that touching the config file will force a rebuild
 $(build)/$(coreboot_dir)/.configured: $(CONFIG_COREBOOT_CONFIG)
 
-EXTRA_FLAGS := -fdebug-prefix-map=$(pwd)=heads -gno-record-gcc-switches -Wno-error=packed-not-aligned -Wno-error=address-of-packed-member
+EXTRA_FLAGS :? -fdebug-prefix-map=$(pwd)=heads -gno-record-gcc-switches -Wno-error=packed-not-aligned
 
 coreboot_configure := \
        mkdir -p "$(build)/$(coreboot_dir)" \

```

MrChromebox: @tlaurion makes sense to me, I'll adjust / re-push when back at my desk next week

tlaurion:
@MrChromebox saving cache failed, first time I see an attempt to saving cache of two different workspaces.
Normally workspaces are passed between steps, in serial way, where the latest results in the cache being used for cache.

In this case, librem14 should depend on x230-hotp-maximized, and save_cache, depend on librem14.
This will increase build time because musl-cross-make, then x230-hotp-maximized then librem14, but this should be true only when we change modules definitions. Otherwise, Heads have a commit history of less then 30 days between them, so hopefully, we will have builds base on create cache.

tlaurion:
@MrChromebox So it worked.
You do not want to add the other boards into CircleCI at the same time?

That would be to add boards depending on librem14 and that would be it. The build process should be parallelized consequently.

Otherwise, LGTM.

MrChromebox:
@tlaurion yes I'll do that and re-push

PR #1061, opened 2021-11-23
bump common boards to coreboot 415 and  kernel 510

tlaurion:
WiP


tlaurion:
Last tests shows that a different kernel config will be necessary for xx30-flash legacy and maximized boards, where only legacy and maximized currently exists.

As far as we are concerned, and since no gpg nor encryption nor network operations are happening under xx30-flash boards, the kernel could be axed more and have network support and crypto algos completely removed under kernel configurations. That is next step. I will push local changes in a hit so collaboration can happen.

No, a simple kernel version is not possible under 4mb CBFS available total space for xx30-flash boards.

Another solution is to keep 4.14 kernel there, but as far as build time is important, this would be less desirable both for local and CI builds. The more shared the happier will be CircleCI free tier at the end of the month

tlaurion:
The simplest solution as of now is to keep legacy xx30-flash boards on kernel 4.14 and bump others to newer kernel.
Will test that, but as we know, that means longer build times on CI because multiple kernel versions will need to be built.

tlaurion:
Replaced by #1381

PR #1062, opened 2021-11-23
oem-factory-reset: fix whiptail height/width order

icequbes1:


tlaurion:
@icequbes1 nice catch. Thanks!

PR #1069, opened 2021-12-04
Use issue templates

eganonoa:
In its current form, Heads has so many variables it appears very difficult to determine the cause of issues being raised. There is a risk in development being stalled or taken off track by a rise of very specific bugs reported that are not readily reproducible or only appears in a very specific set of circumstances. In addition, its increasingly difficult to identify separate feature requests from bug reports. 

These three templates provide templates for: (a) feature requests; (b) bug reports related to build errors for Heads; and (c) bug reports that come from the flashing or operation of Heads. They are designed to better help the identification of the best people to respond to any given issue, as well as to help triage issues.


tlaurion:
Awesome idea!

We try to have a one commit/change being as self explanatory as possible.

I could squash them, but doing so I would become the author which you should get the credit for.

Please (from memory since on phone):
git checkout osresearch/master
git fetch osreseach
git reset --force

Then
git cherry-pick Commit-ids-desired
git checkout -b issue_templates
git push YourRepoConfig

tlaurion:
I would love to comment directly but will wait for you to do a clean push first @eganonoa

tlaurion:
Some comments meanwhile.

User flashes a compiled/downloaded  ROM and has a brick:
- He knows filename. In filename is the commit id in last portion of filename after the first g character.

User finds a bug while system boots
- git commit and other useful information is found under /etc/config. 
- We should push the user into doing a screen capture of that. 
- That would give us the commit id and the board config with a lot of other useful information.

Thoughts @eganonoa ?

eganonoa:
> Awesome idea!
> 
> We try to have a one commit/change being as self explanatory as possible.
> 
> I could squash them, but doing so I would become the author which you should get the credit for.
> 
> Please (from memory since on phone): git checkout osresearch/master git fetch osreseach git reset --force
> 
> Then git cherry-pick Commit-ids-desired git checkout -b issue_templates git push YourRepoConfig

Thanks. Yes, sorry. I was being lazy trying to avoid going through the process for the cherry pick as I made things complicated by putting the 15" screen board changes into my master as opposed to a branch.  Will fix and resubmit.

eganonoa:
> Some comments meanwhile.
> 
> User flashes a compiled/downloaded ROM and has a brick:
> 
>     * He knows filename. In filename is the commit id in last portion of filename after the first g character.
> 
> 
> User finds a bug while system boots
> 
>     * git commit and other useful information is found under /etc/config.
> 
>     * We should push the user into doing a screen capture of that. ( `cat /etc/config` )
> 
>     * That would give us the commit id and the board config with a lot of other useful information.
> 
> 
> Thoughts @eganonoa ?

Getting that additional info is a great idea. I'll put that into the template.

Will add also the filename convention, though there will be times where folks won't have them, particularly if they've gotten the rom from another source.

tlaurion:
@eganonoa well, i am bot aware of such other "source". Heads is still a "geek" project where most board owners build the firmware themselves and where the goal would be to have the CircleCI builds being uploaded in releases in some fixed timeframe. 

If a user bought hardware from vendor, don't think those users are not using support channels to get support.

The goal here is mainly for bug reports of rolling releases. At least that is my understanding, where bug reports to vendor should make their way faster upstream then from community reports by vendor support meanings. @eganonoa

eganonoa:
> @eganonoa well, i am not aware of such other "source" of firmware. There are flashed ROMs, CircleCI and vendor provided ones. Flashing some other source kinda break Heads contract, no?
> 
> Heads is still a "geek" project where most board owners build the firmware themselves and where the goal would be to have the CircleCI built ROMs being uploaded in releases in some fixed timeframe in the future, where commit ids are public and known. (With reproducibility issues still being opened and needing love, which kinda means that CircleCI is the best case we have for known build system and buidlstack troubleshooting as of now...)
> 
> If a user bought hardware from vendor, don't think those users are not using support channels to get support if needed.
> 
> The goal here is mainly for bug reports of rolling releases.
> 
> At least that is my understanding, where bug reports to vendor should make their way faster upstream then from community reports, by vendor support meanings wanting less support requests.

"flashed ROMs, CircleCI and vendor provided", plus roms built by people themselves (as well as probably roms built by one of the many coreboot flashing services you can find on ebay, roms shared by some person or another, and roms renamed, etc).  I think you might find it rather useful for triage to have basic info regarding the provenance of the rom, and will find that quite a number of the issues being raised related to sui generis rom builds.

tlaurion:
@eganonoa they should all share the same commit ids. And as a consequence, content should match hashes.txt produced locally, exception made for modules known to not be reproducible.

If a third party modifies content of a ROM for a specific commit Id, we expect the ROM as being tagged dirty.

This is why commit Id and content of roms should match at the very least on the script levels, where other modules known to be not reproducible should be made reproducible with some community help :)

tlaurion:
Otherwise : what are we testing?

eganonoa:
> @eganonoa they should all share the same commit ids. And as a consequence, content should match hashes.txt produced locally, exception made for modules known to not be reproducible.
> 
> If a third party modifies content of a ROM for a specific commit Id, we expect the ROM as being tagged dirty, or not matching any commit id.
> 
> This is why commit Id and content of roms should match at the very least on the script levels, where other modules known to be not reproducible should be made reproducible with some community help. Otherwise, how can we provide support if code running doesn't match what was ever upstream?

@tlaurion I agree. I'm not suggesting **_not_** asking for that data. However, I'm a firm believer in the value of plain language. And I think you will gain significantly from the added transparency of having those questions regarding the provenance of the rom. If it's a downloaded rom it is simple: where did you get it, what is its name. But for roms being self-built it is useful to have additional information about what was done with the repository used (e.g. change version of coreboot). And generally you will be better able to get others to give input into bug reports without requiring them to be as cognizant as you are about what a specific build ID equals. Well set out bug reports that help signpost these things will make it easier for others to respond to bug reports, to test and validate changes, and, more generally, to **avoid** raising duplicative reports.

eganonoa:
Closing to clean up. Will reissue.

Issue #1065, opened 2021-11-29
Heads on w530 (dGPU wanted?)


eganonoa:
I have successfully built and flashed a w530 with coreboot 4.13 and 4.14 using the maximized boards and flash (see #1071). As with the T530 this was tested with the dGPU (I believe all W530's must have the dGPU), building off of master with minimal changes. The issue that appears to impact t430 boards with the dGPU (#1057) does not appear to impact the T530 or W530. However, I have built acting as if the dGPU is not present on these boards at all, and as such they are not initialized. I see some advantages to this, specifically:

1. Lower power consumption means that you do not need the **giant** 135w charger to power the device and charge the battery and get a bit better battery life generally;
2. You can make full use of the additional super-chunky heat sinks that the T530 and W530 models with the dGPU have keeping things incredibly cool

As the dGPU in the T530 is old and not really useful today, I had decided not to bother working to build coreboot in order to activate it. However, as I understand it the dGPU's in the w530's are a bit more useful. I would be happy to play around with figuring out how to get it working if other owners of w530's would find it interesting and could help with some testing.

tlaurion:
The dGPU is an interesting feature for Qubes passthrough and gaming, while I see this being pertinent more in platforms like the kgpe-d16 where a performant GPU is chosen and initialized, let it be for mining and or a gaming workstation.

Since I am not a gamer myself and the dGPU of those machines is quite old, I would leave for comment from the community to decide or not if this beast would be interesting or not to be enabled properly. 

That would be the resort of the OS though. Coreboot has little to play here, if not already done, setting ACPI tables correctly so that the PCI device is made available to the OS with proper VBT blobs, which I think were pushed with corehoot 4.13

eganonoa:
@tlaurion: It turns out that the w530 cannot use an external display (either in the dock or directly) if you do not have the dGPU initialized. For some this won't be an issue, but for others (like me) it will be necessary. So I have done that and have everything working.  I have a working board for this. However, I am wondering whether or not this is something that should be added in as a new board (e.g. w530-dgpu) or whether it is something better just to have documented for people to do on their own. It is a multi-stage process, and one that might be a challenge to automate. The process is as follows:

1. Extract the integrated and discrete VGA roms. A tool such as VBiosFinder (https://github.com/coderobe/VBiosFinder) can automate this. But it is not included in Heads and isn't the simplest thing to build for any given system.
[...]

PR #1071, opened 2021-12-05
Add w530 boards

eganonoa:
This is a cleaned up PR for adding w530 boards (from #1059) separating the w530 from t520 and t530 for clarity and ease of merging. Boards for flash, maximized and hotp-maximized have been added. I am not intending to support non-maximized boards. Tested and working.  All w530 have dGPUs. Coreboot does not initialize the dGPU using these configs. I do not intend to add support for initializing the dGPU at this stage (it will require some changes to the config file and addition of blobs in the blob directory) but will if there is community demand (see #1065). 

Boards currently point to Coreboot 4.13, but 4.14 also tested and working. Support for the w530 was not added in coreboot until version 4.9 so it cannot be built using the current heads:master, which points to 4.8.1 for thinkpad boards.

Note: it can be tricky to properly read the 8MB chip on the T530 and W530 (see #996). This appears to be because the two chips share a power source on these boards (see generally). This is true for both the Rasbperry Pi and the CH431a. There appears to be two workarounds: (1) the Wake on Lan method (which has the one downside of only working once, because it is not supported by coreboot and is, therefore, not available after flashing); or (2) keeping the CMOS battery plugged in, which works also post-flashing. Both methods provide a little extra power to the board and appear to then be sufficient to get enough power to the 8MB chip to get a proper read of it. The Skulls team generally suggest using an external power source with the CH341a, and it's possible that this might help, but I have not tested


tlaurion:
Merged per #1074

PR #1078, opened 2021-12-12
Generic framebuffer and ps2 keyboard

eganonoa:
Have finally gotten test boards and tested the changes made in other lenovo boards re the framebuffer and ps2 keyboard. Tested and working on a w530 with K1000M dgpu (no non-dgpu models in the w530 range). So see no reason not to change to be consistent with the other boards.


tlaurion:
@eganonoa following your other recent posts, this seems to be a good enhancement but vbios still seems to be required to have a fully working (external displays working) w530 board.

I would personally advocate into keeping things separated and push the boards in more clearer forms but not sure how to do properly.

W530-igpu and variants replacing w530? Document more clearly in board xonfig and docs that external displays won't work?

eganonoa:
@tlaurion Yes, agreed. I'm not seeing a genuine benefit with this change alone, as you know. So if being aligned isn't worth it, then this would be better bundled when I do a whole dGPU thing for w530 and t530.

PR #1079, opened 2021-12-13
Add dGPU support (t530 & w530)

eganonoa:
This PR adds support for dGPU's for the t530 and w530. In doing so, it includes scripts that do for the vbios was the Intel ME script does, installing the [VBiosFinder](https://github.com/coderobe/VBiosFinder) tool and extracting the relevant roms from the latest lenovo bios update for each board. This allows the process to be automated (with one exception, see below). 

Adding the dGPU support is necessary to:

- Have an external display via the DisplayPort for the T530 dGPU (VGA works fine irrespective);
- Have an external display via either the VGA or DisplayPort for the W530.

Beyond this, enabling the dGPU for those boards is not needed, and indeed might be preferable depending on use-case as you get better thermals and battery life with the dGPU not initialized.

After building, the dGPU is not completely initialized. It is still necessary to take the resulting 12MB rom and run nvramtool on it to change the default option from Integrated Graphics to Dual Graphic or Discrete Graphics.  This is documented in the README_vbios file in the blobs directory, with the board config files pointing towards this file.

Because the roms are different for each board, it isn't possible to use just one bios update as with the ME.  You can use any update from the same series to get the integrated rom for all boards in the same series. But different boards have different NVIDIA cards, and thus different roms. Nonetheless, the script is simple and the process of adding additional boards will be fairly trivial. 

These boards have been tested on T530 (no-dGPU), T530 with dGPU and W530 with K1000m dGPU and are working.  The scripts have been successfully run on Debian 10 and Ubuntu 21.04.


eganonoa:
Have messed up this PR with subsequent actions. Cleared CircleCI tests yesterday but now its a mess. Closing and will reopen a new cleaner one.

PR #1080, opened 2021-12-14
Add dGPU support (t530 & w530)

eganonoa:
Republishing #1079 with a cleaner PR.

This PR adds support for dGPU's for the t530 and w530. In doing so, it includes scripts that do for the vbios was the Intel ME script does, installing the VBiosFinder tool and extracting the relevant roms from the latest lenovo bios update for each board. This allows the process to be automated (with one exception, see below).

Adding the dGPU support is necessary to:

    Have an external display via the DisplayPort for the T530 dGPU (VGA works fine irrespective);
    Have an external display via either the VGA or DisplayPort for the W530.

Beyond this, enabling the dGPU for those boards is not needed, and indeed might be preferable depending on use-case as you get better thermals and battery life with the dGPU not initialized.

After building, the dGPU is not completely initialized. It is still necessary to take the resulting 12MB rom and run nvramtool on it to change the default option from Integrated Graphics to Dual Graphic or Discrete Graphics. This is documented in the README_vbios file in the blobs directory, with the board config files pointing towards this file.

Because the roms are different for each board, it isn't possible to use just one bios update as with the ME. You can use any update from the same series to get the integrated rom for all boards in the same series. But different boards have different NVIDIA cards, and thus different roms. Nonetheless, the script is simple and the process of adding additional boards will be fairly trivial.

These boards have been tested on T530 (no-dGPU), T530 with dGPU and W530 with K1000m dGPU and are working. The scripts have been successfully run on Debian 10 and Ubuntu 21.04.


blobs/xx30/vbios_t530.sh
```diff
@@ -0,0 +1,79 @@
...
+echo "### Creating temp dir"
+extractdir=$(mktemp -d)
+cd "$extractdir"
+
+echo "### Installing basic dependencies"
+sudo apt update
+sudo apt install -y wget ruby ruby-dev ruby-bundler p7zip-full upx-ucl 
+
+echo "### Downloading rom-parser dependency"
+wget https://github.com/awilliam/rom-parser/archive/refs/heads/master.zip
```
tlaurion, line 19: As discussed in other channel, that should be https://github.com/coderobe/VBiosFinder/archive/c2d764975115de466fdb4963d7773b5bc8468a06.zip

This is valid commit ID related direct link as of today. Unless they public interesting fixes for our use case, the code in current state matches our need. Pointing to master will break validation tests in the short term. This is to be dodged since 
`
VBIOS_FINDER_SHA256SUM="bbffe3c9c8a64c31485618fae2a2892fe2f4da7b45adaaba136fd45056bce9cd master.zip"`

Will be invalid one point in time. We want to not have to deal with that, which is why we want to point to a commit id specific archive, not master.

blobs/xx30/vbios_t530.sh
```diff
@@ -0,0 +1,79 @@
...
+unzip master.zip
+rm master.zip
+cd rom-parser-master
+make
+sudo cp rom-parser /usr/sbin/
+cd ..
+rm -r rom-parser-master
+
+echo "### Downloading UEFIExtract dependency"
+wget https://github.com/LongSoft/UEFITool/releases/download/A58/UEFIExtract_NE_A58_linux_x86_64.zip
```
tlaurion, line 34: That should be probably measured too since it is a fixed version.

blobs/xx30/vbios_t530.sh
```diff
@@ -0,0 +1,79 @@
...
+echo "### Verifying expected hash of UEFIExtract"
+echo "$UEFI_EXTRACT_SHA256SUM" | sha256sum --check || { echo "Failed sha256sum verification..." && exit 1; }
+
+echo "### Installing UEFIExtract"
+unzip UEFIExtract_NE_A58_linux_x86_64.zip
+sudo mv UEFIExtract /usr/sbin/
+rm UEFIExtract_NE_A58_linux_x86_64.zip
+
+echo "### Downloading VBiosFinder"
+wget https://github.com/coderobe/VBiosFinder/archive/refs/heads/master.zip
```
tlaurion, line 45: Same here. We do not want to point to master.
Clicking on lastest commit https://github.com/coderobe/VBiosFinder/commit/c2d764975115de466fdb4963d7773b5bc8468a06 then browse file and then "Code" button shows the URL https://github.com/coderobe/VBiosFinder/archive/c2d764975115de466fdb4963d7773b5bc8468a06.zip under download button. This is the archive, for that commit ID, that you can force a sha256sum verification hash inside of the code that will work forever. That is, whoever will download that archive will download tree until that commit. This should be measurable and the same in the future. If this is not, then Huston we have a problem.

blobs/xx30/vbios_w530.sh
```diff
@@ -0,0 +1,83 @@
...
+echo "### Creating temp dir"
+extractdir=$(mktemp -d)
+cd "$extractdir"
+
+echo "### Installing basic dependencies"
+sudo apt update
+sudo apt install -y wget ruby ruby-dev ruby-bundler p7zip-full upx-ucl 
+
+echo "### Downloading rom-parser dependency"
+wget https://github.com/awilliam/rom-parser/archive/refs/heads/master.zip
```
tlaurion, line 21: same

blobs/xx30/vbios_w530.sh
```diff
@@ -0,0 +1,83 @@
...
+unzip master.zip
+rm master.zip
+cd rom-parser-master
+make
+sudo cp rom-parser /usr/sbin/
+cd ..
+rm -r rom-parser-master
+
+echo "### Downloading UEFIExtract dependency"
+wget https://github.com/LongSoft/UEFITool/releases/download/A58/UEFIExtract_NE_A58_linux_x86_64.zip
```
tlaurion, line 36: same

blobs/xx30/vbios_w530.sh
```diff
@@ -0,0 +1,83 @@
...
+echo "### Verifying expected hash of UEFIExtract"
+echo "$UEFI_EXTRACT_SHA256SUM" | sha256sum --check || { echo "Failed sha256sum verification..." && exit 1; }
+
+echo "### Installing UEFIExtract"
+unzip UEFIExtract_NE_A58_linux_x86_64.zip
+sudo mv UEFIExtract /usr/sbin/
+rm UEFIExtract_NE_A58_linux_x86_64.zip
+
+echo "### Downloading VBiosFinder"
+wget https://github.com/coderobe/VBiosFinder/archive/refs/heads/master.zip
```
tlaurion, line 47: same

tlaurion (changes requested):
@eganonoa please fix!

eganonoa:
@tlaurion Awesome. Thanks. A cool trick.  See the following changes in my repo:

- [vbios_w530.sh](https://github.com/eganonoa/heads/commit/06c6e776ed6c22636f933b27afcdf11df8d4bc74)
- [vbios_t530.sh](https://github.com/eganonoa/heads/commit/5f36bbfe349f965fb9255518345b2c5ae2391d10)

Note the fixed release UEFIExtract_NE_A58_linux_x86_64.zip was already being measured in the existing scripts.

All tested and working. Would like to add into this PR and to my dGPU board but am afraid of triggering yet another round on this CircleCI drama.  So let me know.

eganonoa:
Scripts updated as requested, including edit provided on the scripts on eganonoa:master.  Everything has been run on CircleCI on my fork as well as successfully in Debian 10 and Ubuntu 21.04 docker containers.

eganonoa:
@tlaurion As discussed offline, scripts have been updated to remove duplication and in doing so make easier to maintain and use as the basis for other boards with dgpu. Thanks for the guidance. I think these scripts are now massively improved over the ones I initially submitted with sha256sum verification, downloads from fixed releases, and a more efficient structure.

PR #1090, opened 2022-01-18
qubes-measure-luks: Change implementation from luksDump to luksHeaderBackup

tlaurion:
PoC with minimal changes.

Fixes #1089

PR #1091, opened 2022-01-19
Use luksHeaderBackup rather than luksDump to measure luks headers.

root-hardenedvault:


tlaurion:
Tested on x230-hotp-maximized; replaces closed #1090.
Does not implement fix for #1092

Issue #1092, opened 2022-01-19
LUKS header is not verified directly when a TPM Disk Unlock Key setuped/verified on default boot and fails unsealing


tlaurion:
@MrChromebox @kylerankin: what @Tonux599 [proposed has a workflow](https://github.com/osresearch/heads/issues/1089#issuecomment-1015775472) is about having LUKS headers being verified prior of booting an OS from a LUKS container that may have been tampered with and different then measured, as reported by @rustybird on #1089 OP for possible LUKS2 existing vulns and future ones:
>But attacks against LUKS may involve the crafting of an unusual header (CVE-2021-4122, QSB-019), and a hypothetical future attack of this type might not change the luksDump output at all.

#1090 (closed) #1091(merged) just replaces luksDump as requested in OP. But would not inform of what failed in measurements leading to a failed unsealing of TPM disk unlock key.

It would just refuse to unseal TPM LUKS sealed disk unlock key. Without telling why.

While not having LUKS disk unlock sealed file released by TPM leaves the user blunt about anything having happened on the LUKS container at all for the moment, other then #1091.

_Originally posted by @tlaurion in https://github.com/osresearch/heads/issues/1089#issuecomment-1015987124_

tlaurion:
@root-hardenedvault changed the behavior of TPM sealed Disk Unlock Key from luksDump to luksBackupHeader under #1091, which whole binary headers are now hashed. That hash is then measured and sealed/unsealed as part of the TPM released disk unlock key when a default boot is configured to release such key from TPM NVRAM space when the the firmware measurements, loaded kernel modules, user config files (including public key injected in rom) are still consistent and while Heads haven't landed in recovery shell, which scrambles the PCR measurements.

Under #1092, he adds an additioanl verification when TPM Disk Unlock key passphrase is not able to release the Disk Unlock Key since measurements are incoherent. Since PCR 6 is now measuring the actual configured disks luksBackupHeaders, when the passphrase is unable to unlock TPM NVRAM region, an additional check is made to verify luksHeaderBackup hashes saved under /boot against to the one generated on the fly. If those are the same, a message is given to the user saying so. 

This informs the user if the LUKS headers changed against detached signed /boot/kexec* files, every time TPM Disk Unlock Key passphrase is unable to release TPM Disk Unlock Key if the LUKS Headers did not change.

PR #1093, opened 2022-01-20
Make it possible to report headers of which LUKSes to be unlocked via TPM change

root-hardenedvault:
#1092

### DIFF
--- a/initrd/bin/kexec-insert-key
+++ b/initrd/bin/kexec-insert-key
@@ -51,6 +51,8 @@ tpm extend -ix 4 -ic generic \
 
 # Check to continue
 if [ "$unseal_failed" = "y" ]; then
+	diff "$(dirname $INITRD)/kexec_lukshdr_hash.txt" /tmp/luksDump.txt \
+		&& echo "Headers of LUKSes to be unlocked via TPM do not change."
 	confirm_boot="n"
 	read \
 		-n 1 \
--- a/initrd/bin/kexec-seal-key
+++ b/initrd/bin/kexec-seal-key
@@ -152,3 +152,6 @@ fi
 
 shred -n 10 -z -u "$TPM_SEALED" 2> /dev/null \
 || warn "Failed to delete the sealed secret - continuing"
+
+cp /tmp/luksDump.txt "$paramsdir/kexec_lukshdr_hash.txt" \
+|| warn "Failed to have hashes of LUKS header - continuing"


initrd/bin/kexec-insert-key
```diff
@@ -51,6 +51,8 @@ tpm extend -ix 4 -ic generic \
 
 # Check to continue
 if [ "$unseal_failed" = "y" ]; then
+	diff "$(dirname $INITRD)/kexec_lukshdr_hash.txt" /tmp/luksDump.txt \
```
tlaurion, line 54: @root-hardenedvault : wouldn't be preferable to inform the user if the hashes differ? (! diff)

initrd/bin/kexec-insert-key
```diff
@@ -51,6 +51,8 @@ tpm extend -ix 4 -ic generic \
...
+		&& echo "Headers of LUKSes to be unlocked via TPM do not change."
```
tlaurion, line 55: Otherwise, tense of verb seems wrong, since they were previously measured.
"Headers of LUKSes to be unlocked via TPM do not change." -> "Headers of LUKSes to be unlocked via TPM did not change."

But as said in previous comment at line 54, it would be preferable for end user to know that it changed, not that it didn't change.

tlaurion (changes requested):
@root-hardenedvault quick review!
Analysis given [here](https://github.com/osresearch/heads/issues/1092#issuecomment-1018641881)

tlaurion:
@root-hardenedvault  ping?

tlaurion:
@root-hardenedvault https://github.com/osresearch/heads/pull/1093#discussion_r789788266 ?

tlaurion:
Testing header change after encrypted device reencryption (to change header).

Prior of reencryption of device, trying to unseal TPM encryption key with bad TPM disk unlock key passphrase, the code doesn't go through changed code to show "Headers of LUKSes to be unlocked via TPM do not change." there.  

Instead, and as expected, since the header is part of what is measured and sealed, we only get a bad TPM unseal password, which happens under tpm errors at https://github.com/osresearch/heads/blob/14c76d062c199f17ff6369091a80d2885cb54914/initrd/bin/kexec-unseal-key#L34-L38. It outputs: "Error Authentication failed (Incorrect Password) from TPM_Unseal".

On reboot after reencrypting drive (creyptsetup-reencrypt), attempting to boot the default option, prompt to type the TPM disk unlock key passphrase gives: Unable to unseal disk encryption key, also not running through modified code.

tlaurion:
@root-hardenedvault 

The following works in both case: when the user enters a bad TPM disk encryption key passphrase, and when he enters a good one but the LUKS headers changed:

```
diff --git a/initrd/bin/kexec-unseal-key b/initrd/bin/kexec-unseal-key
index e016f5bd..b410f706 100755
--- a/initrd/bin/kexec-unseal-key
+++ b/initrd/bin/kexec-unseal-key
@@ -44,6 +44,9 @@ for tries in 1 2 3; do
 
        pcrs
        warn "Unable to unseal disk encryption key"
+       if ! diff /boot/kexec_lukshdr_hash.txt /tmp/luksDump.txt > /dev/null 2>&1; then
+               warn "Encrypted LUKS(es) headers changed since they were measured and sealed in TPM when you configured a disk unlock key. You might want to investigate."
+       fi
 done
 
 die "Retry count exceeded..."
```

I was never able to trigger the message in your PR. Please update.

tlaurion:
Ok got it.
So without code addition [above](https://github.com/osresearch/heads/pull/1093#issuecomment-1065958459), with ed1c23aaa3f8d5264c5b6e5f66ad359406bed166 the user is actually hitting your code if he types his TPM disk unlock unlock code 3 times, and receives a raw diff output, without a warning in the case the LUKS Header changed.

Following  ed1c23aaa3f8d5264c5b6e5f66ad359406bed166, current PR shows to the user, only after 3 bad TPM disk unlock passphrase:
- uncontextualized diff output if mismatch
- No message if mismatch (message requires a positive diff exit (no change) to show the message (&&) that there was no change (there was change).

See picture: 
signal-2022-03-12-165352

tlaurion:
Superseeded by #1625

PR #1023, opened 2021-09-03
Update Tails public signing distro key (expires on 2022-01-19)

tlaurion:

PR #1094, opened 2022-01-21
tails.key : merging of new long-term signing key with old one so old ISOs can still be booting from Heads

tlaurion:
Tails 4.26 won't boot from ISO with Heads currently included public key.

Reproducibility notes at https://github.com/osresearch/heads/pull/1023#issuecomment-1018735659 to combine old key with newly available one, while minimizing key footprint.

Fixes #1023


tlaurion:
Tested working with 4.26 ans 4.21.

PR #1099, opened 2022-01-28
modules/coreboot: pass EXTRA_FLAGS correctly to non coreboot 4.15 coreboot builds

tlaurion:
`:?` is invalid. `?=` defines if not previously defined (was not passed to 4.11 nor 4.13 coreboot builds)


tlaurion:
@mrchromebox

MrChromebox (approved):

PR #1101, opened 2022-01-28
modules/coreboot: don't build IASL separate from toolchain

MrChromebox:
Older coreboot versions (pre-4.11) required IASL to be built separate
from the main toolchain (crossgcc), but that is no longer the case,
and doing so causes random failures from trying to build IASL as
part of the toolchain and separately, in parallel, each using
-j$(CPUS) threads.

Test: build any board using coreboot 4.15 under Debian 11, observe
no random failures from building the toolchain due to false positives
for a missing depencency .


tlaurion:
Will test this branch over debain-11 prior of merging

MrChromebox:
@tlaurion I created PR #1102 with the coreboot 4.11 patches

PR #1112, opened 2022-02-16
Update x230-flash.config

lynxred:


tlaurion:
@lynxred I appreciate your attempt but something seems to have changed in coreboot positioning of binaries in the 12 image we extract the 4mb image from.

Still struggling on my side.

@flammit this is the time where xx30-flash are dying, see pinned issue if you have some insight on reviving xx30-flash boards (t430-flash and x230-flash are not booting anymore. Same kernel, ~same coreboot config, no joy anymore.

PR #1116, opened 2022-02-18
Linux x230 cleanup

tlaurion:
Tested on x230 i5:

- External flash of https://4367-103208611-gh.circle-artifacts.com/0/build/x230-flash/heads-x230-flash-v5.0.1-99-g065cbfd-top.rom
- Internal flash of local built x230-hotp-verification and x230-hotp-maximized which share this linux-x230.config file
- GPG OEM factory reset presented on boot.
- Qubes 4.0.4 ISO boot with detached signature. Got to the screen where installation options are given (past previous error given for virtualization extensions not being present which was https://github.com/osresearch/heads/issues/1107
- Qubes 4.1 ISO boot with detached signature file.

Some gain stats: (Wille be edited


tlaurion:
@MrChromebox not quite yet. I see a lot of discrepencies between x230-flash and x230. Will remove draft when done.

tlaurion:
I think its time that linux config are stored in their full oldconfig output instead of makesavedefconfig.
Its the main reason why kernel version bumps are not happening.

By storing into oldconfig format (output of make nconfig/menuconfig) the files stored would be bigger, but could be used as a base for the next version, where comparing output of old kernel version vs new kernel version would should mostly directly what are the new default to deactivate.

paulmenzel:
> I think its time that linux config are stored in their full oldconfig output instead of makesavedefconfig.
> Its the main reason why kernel version bumps are not happening.

Reading all the comments in the thread, I do not understand the motivation for your comment. Can you please elaborate?

tlaurion:
> > I think its time that linux config are stored in their full oldconfig output instead of makesavedefconfig.
> > Its the main reason why kernel version bumps are not happening.
> 
> Reading all the comments in the thread, I do not understand the motivation for your comment. Can you please elaborate?

@paulmenzel Thanks for following up!

This was a note for me to take into account when i'm going to rebase this PR on master later on.

As stated above, if ones takes a look into `config/linux*` files kept under Heads repo (same for coreboot config files), they were historically requiring the maintainer to do a `make savedefconfig` into `build/kernel-version` directory prior of moving the resulting defconfig back into `config/linux-board.config`.

That `config/linux-board.config` being in makesavedefconfig format doesn't make it clear to the end user nor maintainer of what modules were built-in, activated as modules or deactivated, unless that config was manually expended to audit. When it comes to version bumps, the same logic applies, requiring maintainer to expend that config by copying `config/linux-board.config` into `build/kernel-version/.config` and doing a `make menuconfig` to save that config into .config, and then copying that config into the `build/kernel-new-version/.config` to see what new kernel defaults are actually activated and start tuning of that kernel config to remove new defaults.

Someone just taking `config/linux-board.config` and copying it under `build/kernel-new-version/.config` would have a lot of config options enabled by default which should not be there.

The short version of this is that if the idea of not storing full configurations under github was to optimize for size of what is contained under tree, the maintenance cost to do version bumps is not justifying it.

As seen here, by hiding the defaults and saving only deviations from the default config, we are hiding a lot of things that should be removed, including, as pointed in OP here, uneeded CPU governors, AMD support on Intel and a lot of other device drivers which are becoming activating on newer kernel version and for which a defconfig being used as a base for kernel version bumpwould just use those features by default and consume precious SPI storage because not removed. 

I think its more logical to have full linux configs in tree, while conserving coreboot in savedefconfig format (and here, this is not my opinion for coreboot, but the agreed direction amongst Heads maintainers). In the goal of tackling #590, I think this is the logical first step.

@paulmenzel : any guidance on that level?

PR #1118, opened 2022-02-20
Initial SW raid support

hardestkit:
On many compatible laptop types the CD/DVD ROM device can be swapped out for a caddy for a 2nd SATA block device. When configuring QubesOS to use this 2nd device in a RAID configuration, Heads is unable to see the MD devices to either boot, nor seal a disk encryption key to the TPM 

These changes allow Heads to support booting QubesOS from Software RAID as well as sealing the disk key to TPM when using LUKS over software RAID. The changes expect the devices to be created using the default labelling when using 'custom' disk partitioning in QubesOS Anaconda installer to create /boot as RAID-1 and LUKS on a RAID-1 called 'pv00'


hardestkit:
To cover some areas of why..

On some compatible laptops, the CD/DVD caddy can be removed and a 2nd SATA device added. This opens the opportunity to run QubesOS using RAID-0 striping to offer additional storage capacity, or to use RAID-1 for mirroring which will provide data integrity (via mirroring) to kernel, hashes and signatures (/boot) and the OS/Storage (LUKS pv00). 

Without these changes, Heads cannot detect the software RAID-0 or RAID-1 devices and cannot boot the OS, nor can it seal a LUKS disk key to the TPM. 

These changes should also work for any other device which is or may be ported to heads that has multi drive, software RAID capability. The user must configure their partitioning via QubesOS anaconda installers 'custom' partitioning mode and ensure both /boot is set up as RAID-1 (or RAID-0) as per;


[https://raw.githubusercontent.com/hardestkit/howtos/main/boot-configure.png](https://raw.githubusercontent.com/hardestkit/howtos/main/boot-configure.png)


The Thin pool LVM should be set up over LUKS on RAID-1 or RAID-0 as per 
[https://raw.githubusercontent.com/hardestkit/howtos/main/thinlvm-configure.png](https://raw.githubusercontent.com/hardestkit/howtos/main/thinlvm-configure.png)

Heads should see the RAID devices as /dev/md/localhost:boot (/boot) and /dev/md/localhost:pv00 (LUKS disk), which you can see when starting up. 

[https://raw.githubusercontent.com/hardestkit/howtos/main/heads-raid-boot.png](https://raw.githubusercontent.com/hardestkit/howtos/main/heads-raid-boot.png)

Please ensure to use /dev/md/localhost:pv00 

A video howto will be made available in the near future.

hardestkit:
@tlaurion i was torn between modifying the CONFIG_BOOT_DEV in the board config vs overriding it in initrd/init when CONFIG_SWRAID is present. The reason I have set it in initrd is because sw raid may merge it to other boards in future as a simple bool config option (CONFIG_SWRAID) without the need to maintain separate board config files. My changes would probably have lent this was from the start, were it not for the fact xx30 devices which can have a 2nd drive are reliant upon the x230 kernel config and there is little point in adding bloat to that when those devices have no physical space. I dont think there would be much kernel bloat and if the community thinks its easier to simply have RAID0/1 included in the shared linux-x230.config and just be able to add CONFIG_SWRAID as an option to any board/xyz/xyz.config then I can rework the pull request to run it that way. 

tl/dr: if we accept that linux-x230.config can have raid0 and raid1 modules included, then any xx30 board simply needs CONFIG_SWRAID as a bool to enable.

tlaurion:
@hardestkit 

> tl/dr: if we accept that linux-x230.config can have raid0 and raid1 modules included, then any xx30 board simply needs CONFIG_SWRAID as a bool to enable, rather than have long term support implications for having a 2nd board 'raid' config for compatible devices and linux-xx30-raid kernel config.

Definitely a better option considering linux-x230.conf is shared amongst all xx20 and xx30 boards and modules get included only if defined under board config.

Would also facilitate maintainership in the future.

tlaurion:
I just kicked CircleCI buildof your last commit. It seems that you aren't following your own branch from CircleCI, hence the PR is being built on osresearch's CircleCI, not your own.

hardestkit:
@tlaurion 

> Definitely a better option considering linux-x230.conf is shared amongst all xx20 and xx30 boards and modules get included only if defined under board config (with Makefile magic, Linux module config magic and board config magic).

OK I will close this PR and rework changes to include RAID modules in linux-x230.config and make the support a simple boolean. I will re-test and then submit a new PR once validated.

PR #1127, opened 2022-03-01
Move enable_usb earlier

marmarek:
It is going to be enabled later anyway (if CONFIG_HOTPKEY=y), so it can
also be simplified by enabling it at the very beginning.

This enables USB keyboard consistently during all boot menus, including
the "No Bootable OS Found" prompt. It isn't a big deal for "normal"
laptop usage, but it is important for automatic tests and also
non-laptop systems.

### DIFF
--- a/initrd/bin/gui-init
+++ b/initrd/bin/gui-init
@@ -202,7 +202,6 @@ update_hotp()
 {
   if [ -x /bin/hotp_verification ]; then
     HOTP=`unseal-hotp`
-    enable_usb
     if ! hotp_verification info ; then
       if ! whiptail $BG_COLOR_WARNING --clear \
           --title "WARNING: Please Insert Your $HOTPKEY_BRANDING" \
@@ -251,7 +250,6 @@ clean_boot_check()
 
    # check for USB security token
   if [ "$CONFIG_HOTPKEY" = "y" ]; then
-    enable_usb
     if ! gpg --card-status > /dev/null ; then
       return
     fi
@@ -541,6 +539,10 @@ if [ -r /boot/kexec_hotp_key ]; then
 	HOTPKEY_BRANDING="$(cat /boot/kexec_hotp_key)"
 else
 	HOTPKEY_BRANDING="HOTP USB Security Dongle"
+fi
+
+if [ "$CONFIG_HOTPKEY" = "y" ]; then
+  enable_usb
 fi
 
 if detect_boot_device ; then


tlaurion:
Will give short reasoning on why USB_KEYBOARD is a different board option altogether that was added before, which can be seen in file `initird/init`:
```
# load USB modules for boards using a USB keyboard
if [ "$CONFIG_USB_KEYBOARD" = "y" ]; then
        enable_usb
fi

# If the user has been holding down r, enter a recovery shell
# otherwise immediately start the configured boot script.
# We don't print a prompt, since this is a near instant timeout.
read \
        -t 0.1 \
        -n 1 \
        boot_option
echo

if [ "$boot_option" = "r" ]; then
        # Start an interactive shell
        recovery 'User requested recovery shell'
        # just in case...
        if [ "$CONFIG_TPM" = "y" ]; then
                tpm extend -ix 4 -ic recovery
        fi
        exec /bin/ash
        exit
fi

```

As you can see here, which was subject of debates in the past, having USB keyboard support permits rubber ducky attacks to act as a keyboard and storage device and would permit to go into Heads recovery shell, act, and go back to the GUI, sed'ing policies scripts and even report consistent bypassed measurements altogether. 

So @marmarek : your suggestion of bringing USB initialization earlier under gui-init is a different implementation of the above under init,  not being early enough to permit to use a USB keyboard to go into recovery shell prior of the GUI being fired up.

Consequently, I am not against accepting this PR. We just need to be aware here that USB KEYBOARD support is already in, and activated for kgpe-d16_workstation-usb-keyboard, librem_minis and librem_l1um through board config addition: `export CONFIG_USB_KEYBOARD=y `, which permits a rubber ducky to act soon enough to go into recovery shell prior of launching gui-init and be able to mount-usb and then run a script from there that could show no output at all, and even launch a modified gui-init that could fake it all. Thanks to early USB keyboard initialization.

If that fits your use case, we could add a debug board with CONFIG_USB_KEYBOARD in, or just merge this PR knowing the difference between the two options.

From a user paranoid user perspective, seeing the GUI flicker (its already initialized in your proposition), showing that something automatized is playing with the GUI to go into recovery shell and back should be enough to distrust Heads IMHO, which your PR would prevent better and where CONFIG_USB_KEYBOARD would be less obvious.

Let's remember that going into recovery shell, as the snippet shows above, will invalidate measurements, prior of going into the recovery shell.

marmarek:
> We just need to be aware here that USB KEYBOARD support is already in

Ah, I missed that. I think is a better solution - enable it if you need it, but keep disabled otherwise. I can do my build for tests, no problem. It's just a bit unexpected to see USB keyboard not working in only this very screen, but otherwise functional. Maybe USB keyboard should be disabled everywhere by default? I'm not sure if that's possible, without breaking Nitrokey/LibremKey support (which also uses USB HID)...

> Let's remember that going into recovery shell, as the snippet shows above, will invalidate measurements, prior of going into the recovery shell.

Exactly, so if you verity the HOTP (either against mobile app, or Nitrokey/LibremKey), the attack would be detected.

tlaurion:
> > We just need to be aware here that USB KEYBOARD support is already in
> 
> Ah, I missed that. I think is a better solution - enable it if you need it, but keep disabled otherwise. I can do my build for tests, no problem. It's just a bit unexpected to see USB keyboard not working in only this very screen, but otherwise functional. Maybe USB keyboard should be disabled everywhere by default? I'm not sure if that's possible, without breaking Nitrokey/LibremKey support (which also uses USB HID)...
> 
> > Let's remember that going into recovery shell, as the snippet shows above, will invalidate measurements, prior of going into the recovery shell.
> 
> Exactly, so if you verity the HOTP (either against mobile app, or Nitrokey/LibremKey), the attack would be detected.

Yesssish. Like I said, non-paranoid users trusting blindly could have a dongle attached between their USB keyboard and laptop, even less visible behind workstations and servers. If that is made, that init code enabling usb prior of going into gui-init might go unnoticed and HOTP could report on screen that good `HOTP:Success`, where a more advanced and possible attack involving backuping ROM, extracting measurements, replaying the right measurements could lead into TOTP/HOTP value being properly faked, making TOTP code good and HOTP dongle flashing green.

@marmarek : what I aimed to say prior is that your PR is not breaking that contract; CONFIG_USB_KEYBOARD is. Your code just brings usb init prior of disk detection. So a rubber ducky attack could not, as in the CONFIG_USB_KEYBOARD option, go into recovery shell and play there, without the end user seeing a lot of flicking.... The rubber ducky here would need to replay actions in the menu, coing into recovery shell to play there, and then call back gui-init from the recovery shell after having tampered with the scripts (sed) or calling scripts over mounted-storage. 

Short version: I do not have a problem with your PR, which is in my opinion, better then the USB_KEYBOARD_CONFIG, while again. You might not use HOTP and still have a USB keyboard. So different use cases. And in your case, debugging/playing around might require USB Keyboard, and maybe not HOTP in some future use cases.

@MrChromebox @kylerankin @jans23 : I am not aware of being able to disable HID functions, which was talked about a while back. From linux kernel perspective, my checks have proven (shortly) that disabling HID would remove USB Keyboard support altogether with some problems with Nitrokey/Librem Keys with some downsides I do not recall. @szszszsz?

MrChromebox:
> Short version: I do not have a problem with your PR, which is in my opinion, better then the USB_KEYBOARD_CONFIG.

USB_KEYBOARD_CONFIG allows one to boot directly to a recovery shell with a USB keyboard; this PR does not

>  disabling HID would remove USB Keyboard support altogether with some problems with Nitrokey/Librem Keys with some downsides I do not recall.

I'd need to re-test, but I also recall this being problematic, which is why it is not implemented

tlaurion:
> > Short version: I do not have a problem with your PR, which is in my opinion, better then the USB_KEYBOARD_CONFIG.
> 
> USB_KEYBOARD_CONFIG allows one to boot directly to a recovery shell with a USB keyboard; this PR does not
>

Exactly. No USB Keyboard could be used otherwise to enter the recovery shell typing `r` keyboard key frenetically on early boot.
 
> > disabling HID would remove USB Keyboard support altogether with some problems with Nitrokey/Librem Keys with some downsides I do not recall.

MrChromebox:
> Exactly. No USB Keyboard could be used otherwise to enter the recovery shell typing r keyboard key frenetically on early boot.

that's not a good thing on boards which need it

MrChromebox (approved):

tlaurion:
> > Exactly. No USB Keyboard could be used otherwise to enter the recovery shell typing r keyboard key frenetically on early boot.
> 
> that's not a good thing on boards which need it

We spoke off channel. We agree to merge. The check enables USB earlier in case of HOTP being enabled.
@marmarek that would permit you to be able to use roms built by CI without having to compile yourself, while the change is not reducing security.

This is not a replacement for CONFIG_USB_KEYBOARD for users relying on USB keyboard.

tlaurion:
(BTW sorry for CircleCI which now requires to be kicked at least once to complete for now 2 weeks. racing conditions on their dockers. No clue!)

marmarek:
I've noticed some issue with this: `/etc/config` doesn't have `CONFIG_HOTPKEY`, unless I change relevant config in `boards/` to have `export CONFIG_HOTPKEY=y`. Am I doing something wrong?

marmarek:
Sure, will change that. I was confused, because `CONFIG_HOTPKEY` was use in `gui-init` before already.

tlaurion:
@MrChromebox previous comment relates to most probable issues.

Everything should refer to binary presence which is consequential of module being deployed.

marmarek:
https://github.com/osresearch/heads/pull/1130

PR #1128, opened 2022-03-02
Makefile: have HEADS_GAWK depend on locally built HEADS_MAKE if make …

### LINKED ISSUE #1126: Error 1 on Build
Hello everyone,
I am trying to build x230-maximized Board on Debian 10.4.0.
I have installed all the extra required packages.
Right at the start of building I get error 1, The certificate of coreboot.org cannot be trusted, the certificate is invalid.
I have already failed building on Debian 9 (had the wrong gawk and make versions), but this is a new error for me.
I have even tried and downloaded the coreboot-blob file with --no-certificate-check and put it in heads/packages
but the installer still gives out the same error.
I have also updated ca-certificates package on debian, but still no luck.
Any ideas of what I'm doing wrong?

Thanks in advance for your help!

```
root:xxxxxx....~/heads# make BOARD=x230-maximized
2022-02-25 15:00:13+01:00 Local built HEADS_GAWK only if different then provided by OS:
2022-02-25 15:00:13+01:00 Local built HEADS_MAKE only if different then provided by OS:
2022-02-25 15:00:13+01:00 Heads build system will call make from now on as: MAKE: make.
2022-02-25 15:00:13+01:00 WGET https://www.coreboot.org/releases/coreboot-blobs-4.13.tar.xz
--2022-02-25 15:00:13--  https://www.coreboot.org/releases/coreboot-blobs-4.13.tar.xz
Auflösen des Hostnamens www.coreboot.org (www.coreboot.org)… 2a01:4f8:121:1254::2, 78.46.105.101
Verbindungsaufbau zu www.coreboot.org (www.coreboot.org)|2a01:4f8:121:1254::2|:443 … verbunden.
FEHLER: Dem Zertifikat von »www.coreboot.org« wird nicht vertraut.
FEHLER: Das Zertifikat von »www.coreboot.org« ist abgelaufen.
make: *** [Makefile:505: /root/heads/packages/coreboot-blobs-4.13.tar.xz] Fehler 1
root:xxxxxx....~/heads#
```

tlaurion:
Makefile: have HEADS_GAWK depend on locally built HEADS_MAKE if make and gawk are not found in expected 4.2.1 versions locally.

This will be the case on debian-11 and other newer OSes.

- Fixes a bug where "make BOARD=xyz bootstrap" was needed prior of being able to build anything else otherwise HEADS_GAWK was expecting HEADS_MAKE to be built already, which was not the case. (fixes #1126)

- As of now, I do not see any OS make/gawk being used anymore. All modules except coreboot (which uses coreboot buildstack), including musl-cross-make, are being built with local make and local gawk. Improves https://github.com/osresearch/heads/issues/936#issuecomment-746693146 which showed that local make (/usb/bin/make, /usr/local/bin/make, /usr/sbin/make, /usr/local/sbin/make and /usr/bin/gawk were called locally. To restest)


tlaurion:
@SergiiDmytruk @MrChromebox : I didn't find how to have "all" depend on "bootstrap" which would have been the cleaner option.
Before, we were having locally built gawk sha256sum in hashes.txt. Now we have make.
We still have lost sha256sum of the coreboot files in hashes.txt, which I have no clue why up until now.

Any ideas?

Makefile
```diff
@@ -91,9 +91,9 @@ $(HEADS_MAKE): $(build)/$(make_dir)/.configured
 
 # Once we have a proper Make, we can just pass arguments into it
 all linux cpio run: $(HEADS_MAKE)
-        LANG=C MAKE=$(HEADS_MAKE) $(HEADS_MAKE) $(MAKE_JOBS) $@
+        LANG=C MAKE=$(HEADS_MAKE) $(MAKE_JOBS) $@
```
SergiiDmytruk, line 94: Does this work? I think you might have removed wrong part of the string (`$(HEADS_MAKE)` instead of `MAKE=$(HEADS_MAKE)`).

SergiiDmytruk:
@tlaurion

> I didn't find how to have "all" depend on "bootstrap" which would have been the cleaner option.

Probably like this:
```make
bootstrap: ;
all: bootstrap
```

> Before, we were having locally built gawk sha256sum in hashes.txt. Now we have make.

Don't see anything obvious, would need to reduce Makefile to debug this.

Makefile
tlaurion: @SergiiDmytruk the line 94 shows `LANG=C MAKE=$(HEADS_MAKE) $(MAKE_JOBS) $@`
I basically changed:         
`LANG=C MAKE=$(HEADS_MAKE) $(HEADS_MAKE) $(MAKE_JOBS) $@` 
for 
`LANG=C MAKE=$(HEADS_MAKE) $(MAKE_JOBS) $@`

Since originally there was two `$(HEADS_MAKE) $(HEADS_MAKE)` passed. No harm, but make logs under build/logs/* were showing that double `/build/make-4.2.1/make` everywhere.

tlaurion:
> ```makefile
> bootstrap: ;
> all: bootstrap
> ```
@SergiiDmytruk :Some attempts. As said, this is my make fu that is probably the problem. I have to miss a concept.

In master: `bootstrap` is currently defined under `HEADS_GAWK` and `HEADS_MAKE` targets. So as said previously, I tried to have `all` call the `bootstrap` target, since calling make bootstrap works correctly (this is what CircleCI is doing right now):

```
user@heads-tests:~/heads$ rm -rf build/make-4.2.1/ build/gawk-4.2.1/
user@heads-tests:~/heads$ make^C
user@heads-tests:~/heads$ make BOARD=x230-hotp-maximized bootstrap
2022-03-02 14:35:01-05:00 Wrong OS deployed make detected: 4.3. Building and using 4.2.1...
2022-03-02 14:35:01-05:00 Wrong OS deployed gawk detected: 5.1.0. Building and using 4.2.1...
2022-03-02 14:35:01-05:00 Local built HEADS_GAWK only if different then provided by OS: /home/user/heads/build/gawk-4.2.1/gawk
2022-03-02 14:35:01-05:00 Local built HEADS_MAKE only if different then provided by OS: /home/user/heads/build/make-4.2.1/make
2022-03-02 14:35:01-05:00 Heads build system will call make from now on as: MAKE: /home/user/heads/build/make-4.2.1/make.
Makefile:264: warning: overriding recipe for target 'all'
Makefile:141: warning: ignoring old recipe for target 'all'
tar xf "/home/user/heads/packages/make-4.2.1.tar.bz2" -C "/home/user/heads/build"
touch "/home/user/heads/build/make-4.2.1/.extract"
( cd "/home/user/heads/build/make-4.2.1/" ; patch -p1 ) < "patches/make-4.2.1.patch"
patching file glob/glob.c
Hunk #1 succeeded at 208 with fuzz 2.
touch "/home/user/heads/build/make-4.2.1/.patch"
cd "/home/user/heads/build/make-4.2.1/" ; \
./configure 2>&1 \
| tee "/home/user/heads/build/log/make.configure.log" \
> /dev/null
touch "/home/user/heads/build/make-4.2.1/.configured"
make -C "/home/user/heads/build/make-4.2.1/" -j2 --max-load 16 \
	2>&1 \
	| tee "/home/user/heads/build/log/make.log" \
	> /dev/null
tar xf "/home/user/heads/packages/gawk-4.2.1.tar.xz" -C "/home/user/heads/build"
touch "/home/user/heads/build/gawk-4.2.1/.extract"
touch "/home/user/heads/build/gawk-4.2.1/.patch"
cd "/home/user/heads/build/gawk-4.2.1/" ; \
./configure 2>&1 \
| tee "/home/user/heads/build/log/gawk.configure.log" \
> /dev/null
touch "/home/user/heads/build/gawk-4.2.1/.configured"
/home/user/heads/build/make-4.2.1/make -C "/home/user/heads/build/gawk-4.2.1/" -j2 --max-load 16 \
	2>&1 \
	| tee "/home/user/heads/build/log/gawk.log" \
	> /dev/null
user@heads-tests:~/heads$ 

```




So adding an `all: bootstrap` target early:

```
diff --git a/Makefile b/Makefile
index 6ac87a37..71a8fbd7 100644
--- a/Makefile
+++ b/Makefile
@@ -105,6 +105,8 @@ ifeq "" "$(filter $(LOCAL_GAWK_MAJOR_VERSION)%,$(gawk_version))"
 $(eval $(shell echo >&2 "$(DATE) Wrong OS deployed gawk detected: $(LOCAL_GAWK_VERSION). Building and using $(gawk_version)..."))
 HEADS_GAWK := $(build)/$(gawk_dir)/gawk
 
+all: bootstrap
+
 # How to download and build the correct version of gawk
 $(packages)/$(gawk_tar):
        $(WGET) -O "$@.tmp" "$(gawk_url)"

```

```
user@heads-tests:~/heads$ rm -rf build/make-4.2.1/ build/gawk-4.2.1/
user@heads-tests:~/heads$ make BOARD=x230-hotp-maximized
2022-03-02 14:30:22-05:00 Wrong OS deployed make detected: 4.3. Building and using 4.2.1...
2022-03-02 14:30:22-05:00 Wrong OS deployed gawk detected: 5.1.0. Building and using 4.2.1...
2022-03-02 14:30:22-05:00 Local built HEADS_GAWK only if different then provided by OS: /home/user/heads/build/gawk-4.2.1/gawk
2022-03-02 14:30:22-05:00 Local built HEADS_MAKE only if different then provided by OS: /home/user/heads/build/make-4.2.1/make
2022-03-02 14:30:22-05:00 Heads build system will call make from now on as: MAKE: /home/user/heads/build/make-4.2.1/make.
Makefile:264: warning: overriding recipe for target 'all'
Makefile:141: warning: ignoring old recipe for target 'all'
tar xf "/home/user/heads/packages/gawk-4.2.1.tar.xz" -C "/home/user/heads/build"
touch "/home/user/heads/build/gawk-4.2.1/.extract"
touch "/home/user/heads/build/gawk-4.2.1/.patch"
cd "/home/user/heads/build/gawk-4.2.1/" ; \
./configure 2>&1 \
| tee "/home/user/heads/build/log/gawk.configure.log" \
> /dev/null
touch "/home/user/heads/build/gawk-4.2.1/.configured"
/home/user/heads/build/make-4.2.1/make -C "/home/user/heads/build/gawk-4.2.1/" -j2 --max-load 16 \
	2>&1 \
	| tee "/home/user/heads/build/log/gawk.log" \
	> /dev/null
make: *** [Makefile:134: /home/user/heads/build/gawk-4.2.1/gawk] Error 127
```


Applying `all: bootstrap` target late:
```
user@heads-tests:~/heads$ git diff
diff --git a/Makefile b/Makefile
index 6ac87a37..71aa7f6e 100644
--- a/Makefile
+++ b/Makefile
@@ -705,3 +705,5 @@ real.clean:
                fi; \
        done
        cd install && rm -rf -- *
+
+all: bootstrap
user@heads-tests:~/heads$ rm -rf build/make-4.2.1/ build/gawk-4.2.1/
user@heads-tests:~/heads$ make BOARD=x230-hotp-maximized
2022-03-02 14:38:07-05:00 Wrong OS deployed make detected: 4.3. Building and using 4.2.1...
2022-03-02 14:38:07-05:00 Wrong OS deployed gawk detected: 5.1.0. Building and using 4.2.1...
2022-03-02 14:38:07-05:00 Local built HEADS_GAWK only if different then provided by OS: /home/user/heads/build/gawk-4.2.1/gawk
2022-03-02 14:38:07-05:00 Local built HEADS_MAKE only if different then provided by OS: /home/user/heads/build/make-4.2.1/make
2022-03-02 14:38:07-05:00 Heads build system will call make from now on as: MAKE: /home/user/heads/build/make-4.2.1/make.
Makefile:262: warning: overriding recipe for target 'all'
Makefile:139: warning: ignoring old recipe for target 'all'
tar xf "/home/user/heads/packages/gawk-4.2.1.tar.xz" -C "/home/user/heads/build"
touch "/home/user/heads/build/gawk-4.2.1/.extract"
touch "/home/user/heads/build/gawk-4.2.1/.patch"
cd "/home/user/heads/build/gawk-4.2.1/" ; \
./configure 2>&1 \
| tee "/home/user/heads/build/log/gawk.configure.log" \
> /dev/null
touch "/home/user/heads/build/gawk-4.2.1/.configured"
/home/user/heads/build/make-4.2.1/make -C "/home/user/heads/build/gawk-4.2.1/" -j2 --max-load 16 \
	2>&1 \
	| tee "/home/user/heads/build/log/gawk.log" \
	> /dev/null
make: *** [Makefile:132: /home/user/heads/build/gawk-4.2.1/gawk] Error 127
```

tlaurion:
@SergiiDmytruk : 
Haaaaaaa.

So `all: bootstrap;` it is. Thanks for the lesson! Really appreciated!

user@heads-tests:~/heads$ git diff
```
diff --git a/Makefile b/Makefile
index 6ac87a37..1344adb5 100644
--- a/Makefile
+++ b/Makefile
@@ -143,6 +143,9 @@ all linux cpio run: $(HEADS_GAWK)
 bootstrap: $(HEADS_GAWK)
 endif
 
+#We want to make sure we build bootstrap if needed in all cases.
+all: bootstrap;
+
 BOARD          ?= qemu-coreboot
 CONFIG         := $(pwd)/boards/$(BOARD)/$(BOARD).config
 
user@heads-tests:~/heads$ rm -rf build/make-4.2.1/ build/gawk-4.2.1/
user@heads-tests:~/heads$ make BOARD=x230-hotp-maximized
2022-03-02 14:43:56-05:00 Wrong OS deployed make detected: 4.3. Building and using 4.2.1...
2022-03-02 14:43:56-05:00 Wrong OS deployed gawk detected: 5.1.0. Building and using 4.2.1...
Makefile:147: warning: overriding recipe for target 'all'
Makefile:139: warning: ignoring old recipe for target 'all'
2022-03-02 14:43:56-05:00 Local built HEADS_GAWK only if different then provided by OS: /home/user/heads/build/gawk-4.2.1/gawk
2022-03-02 14:43:56-05:00 Local built HEADS_MAKE only if different then provided by OS: /home/user/heads/build/make-4.2.1/make
2022-03-02 14:43:56-05:00 Heads build system will call make from now on as: MAKE: /home/user/heads/build/make-4.2.1/make.
Makefile:265: warning: overriding recipe for target 'all'
Makefile:147: warning: ignoring old recipe for target 'all'
tar xf "/home/user/heads/packages/make-4.2.1.tar.bz2" -C "/home/user/heads/build"
touch "/home/user/heads/build/make-4.2.1/.extract"
( cd "/home/user/heads/build/make-4.2.1/" ; patch -p1 ) < "patches/make-4.2.1.patch"
patching file glob/glob.c
Hunk #1 succeeded at 208 with fuzz 2.
touch "/home/user/heads/build/make-4.2.1/.patch"
cd "/home/user/heads/build/make-4.2.1/" ; \
./configure 2>&1 \
| tee "/home/user/heads/build/log/make.configure.log" \
> /dev/null
touch "/home/user/heads/build/make-4.2.1/.configured"
make -C "/home/user/heads/build/make-4.2.1/" -j2 --max-load 16 \
	2>&1 \
	| tee "/home/user/heads/build/log/make.log" \
	> /dev/null
tar xf "/home/user/heads/packages/gawk-4.2.1.tar.xz" -C "/home/user/heads/build"
touch "/home/user/heads/build/gawk-4.2.1/.extract"
touch "/home/user/heads/build/gawk-4.2.1/.patch"
cd "/home/user/heads/build/gawk-4.2.1/" ; \
./configure 2>&1 \
| tee "/home/user/heads/build/log/gawk.configure.log" \
> /dev/null
touch "/home/user/heads/build/gawk-4.2.1/.configured"
/home/user/heads/build/make-4.2.1/make -C "/home/user/heads/build/gawk-4.2.1/" -j2 --max-load 16 \
	2>&1 \
	| tee "/home/user/heads/build/log/gawk.log" \
	> /dev/null
/home/user/heads/build/make-4.2.1/make -C "/home/user/heads/build/coreboot-4.13" CPUS=2 crossgcc-i386
make[1]: Entering directory '/home/user/heads/build/coreboot-4.13'
Welcome to the coreboot cross toolchain builder v ()

Building toolchain using 2 thread(s).

Target architecture is i386-elf

Found compatible Ada compiler, enabling Ada support by default.

Downloading and verifying tarballs ...
 * gmp-6.2.0.tar.xz (cached)... hash verified (052a5411dc74054240eec58132d2cf41211d0ff6)
 * mpfr-4.1.0.tar.xz (cached)... hash verified (159c3a58705662bfde4dc93f2617f3660855ead6)
 * mpc-1.2.0.tar.gz (cached)... hash verified (0abdc94acab0c9bfdaa391347cdfd7bbdb1cf017)
 * binutils-2.35.tar.xz (cached)... hash verified (6bdd090ce268b6d6c3442516021c4e4b5019e303)
 * gcc-8.3.0.tar.xz (cached)... hash verified (c27f4499dd263fe4fb01bcc5565917f3698583b2)
Downloaded tarballs ... ok
Unpacking and patching ...
 * gmp-6.2.0.tar.xz
   o gmp-6.2.0_generic-build.patch
 * mpfr-4.1.0.tar.xz
 * mpc-1.2.0.tar.gz
 * binutils-2.35.tar.xz
^CStop
make[3]: *** [Makefile:26: build_gcc] Error 1
make[2]: *** [Makefile:51: build-i386] Interrupt
make[1]: *** [util/crossgcc/Makefile.inc:34: crossgcc-i386] Interrupt
make: *** [modules/coreboot:85: "/home/user/heads/build/coreboot-4.13/.xcompile"] Interrupt
```

tlaurion:
@SergiiDmytruk 
> > Before, we were having locally built gawk sha256sum in hashes.txt. Now we have make.
> 

What i'm looking for is what went wrong under 7ce12fe621fe95a460c2c5b0038ab12ad6b3b7b0  causing rom sha256sum output to not be part of generated hashes.txt anymore. This is now replaced by gawk/make in hashes.txt as of now and since that commit.

> Don't see anything obvious, would need to reduce Makefile to debug this.

This is the change that was applied between parent commit c096a1f54d828022a6b8b845b276345316fda63e (generates rom sha256sum under hashes.txt correctly) and culprit commit 7ce12fe621fe95a460c2c5b0038ab12ad6b3b7b0 (which doesn't and generated gawk sha256sum instead):
```
user@heads-tests:~/heads$ git diff c096a1f54d828022a6b8b845b276345316fda63e 7ce12fe621fe95a460c2c5b0038ab12ad6b3b7b0  > diff
user@heads-tests:~/heads$ cat diff
diff --git a/Makefile b/Makefile
index 10faebf7..fd76b057 100644
--- a/Makefile
+++ b/Makefile
@@ -36,6 +36,58 @@ DATE=`date --rfc-3339=seconds`
 
 # This is the correct version of Make
 
+# Check we have a suitable version of gawk
+# that's at least the same major version
+LOCAL_GAWK_VERSION := $(shell gawk --version 2>/dev/null | head -1 | cut -d' ' -f3 | cut -d, -f1)
+LOCAL_GAWK_MAJOR_VERSION := $(patsubst .%,.,$(LOCAL_GAWK_VERSION))
+include modules/gawk
+
+# Wrong version
+ifeq "" "$(filter $(LOCAL_GAWK_MAJOR_VERSION).%,$(gawk_version))"
+# Wrong gawk version detected -- build our local version
+# and re-invoke the Makefile with it instead.
+$(eval $(shell echo >&2 "$(DATE) Wrong gawk detected: $(LOCAL_GAWK_VERSION)"))
+HEADS_GAWK := $(build)/$(gawk_dir)/gawk
+
+# Once we have a suitable version of gawk, we can rerun make
+all linux cpio run: $(HEADS_GAWK)
+	LANG=C HEADS_GAWK=$(HEADS_GAWK) $(MAKE) $(MAKE_JOBS) $@
+%.clean %.vol %.menuconfig: $(HEADS_GAWK)
+	LANG=C HEADS_GAWK=$(HEADS_GAWK) $(MAKE) $@
+
+bootstrap: $(HEADS_GAWK)
+
+# How to download and build the correct version of gawk
+$(packages)/$(gawk_tar):
+	$(WGET) -O "$@.tmp" "$(gawk_url)"
+	if ! echo "$(gawk_hash)  $@.tmp" | sha256sum --check -; then \
+		exit 1 ; \
+	fi
+	mv "$@.tmp" "$@"
+
+$(build)/$(gawk_dir)/.extract: $(packages)/$(gawk_tar)
+	tar xf "$<" -C "$(build)"
+	touch "$@"
+
+$(build)/$(gawk_dir)/.patch: $(build)/$(gawk_dir)/.extract
+#	( cd "$(dir $@)" ; patch -p1 ) < "patches/gawk-$(gawk_version).patch"
+	touch "$@"
+
+$(build)/$(gawk_dir)/.configured: $(build)/$(gawk_dir)/.patch
+	cd "$(dir $@)" ; \
+	./configure 2>&1 \
+	| tee "$(log_dir)/gawk.configure.log" \
+	$(VERBOSE_REDIRECT)
+	touch "$@"
+
+$(HEADS_GAWK): $(build)/$(gawk_dir)/.configured
+	$(MAKE) -C "$(dir $@)" $(MAKE_JOBS) \
+		2>&1 \
+		| tee "$(log_dir)/gawk.log" \
+		$(VERBOSE_REDIRECT)
+endif
+
+
 BOARD		?= qemu-coreboot
 CONFIG		:= $(pwd)/boards/$(BOARD)/$(BOARD).config
 
@@ -124,6 +176,10 @@ CROSS_TOOLS_NOCC := \
 	PKG_CONFIG_PATH="$(INSTALL)/lib/pkgconfig" \
 	PKG_CONFIG_SYSROOT_DIR="$(INSTALL)" \
 
+ifneq "$(HEADS_GAWK)" ""
+CROSS_TOOLS_NOCC += AWK=$(HEADS_GAWK)
+endif
+
 CROSS_TOOLS := \
 	CC="$(heads_cc)" \
 	$(CROSS_TOOLS_NOCC) \
diff --git a/modules/gawk b/modules/gawk
new file mode 100644
index 00000000..e51b8ab5
--- /dev/null
+++ b/modules/gawk
@@ -0,0 +1,14 @@
+# This is not added to the module list since it is a special case
+# of things that need to be built for the host system, instead of the
+# target platform.
+#modules += gawk
+
+gawk_version := 4.2.1
+gawk_dir := gawk-$(gawk_version)
+gawk_tar := gawk-$(gawk_version).tar.xz
+
+gawk_url := http://gnu.mirror.constant.com/gawk/$(gawk_tar)
+gawk_hash := d1119785e746d46a8209d28b2de404a57f983aa48670f4e225531d3bdc175551
+
+# This is built for the local machine, not the target, so it doesn't have any
+# of the build instructions.
```

@SergiiDmytruk any insight on what happened to:
```
all:
        @sha256sum $< | tee -a "$(HASHES)"
```

Which for some reason doesn't apply to coreboot produced rom since that culprit commit?

tlaurion:
@SergiiDmytruk : unfortunately https://github.com/osresearch/heads/pull/1128/commits/4e57d3f18a773f5bf0ce949e95dd05f90f2ab819, implementing:
`all: bootstrap;`

Also hits the `all: sha256sum` and fails with:
```
sha256sum: bootstrap: No such file or directory
make: *** [Makefile:265: all] Error 1
```

So it seems that https://github.com/osresearch/heads/pull/1128/commits/9f062ae01a36cc727065aa1367915661c65fd454 was a better implementation.

Makefile
SergiiDmytruk: @tlaurion I probably miss something. It's just that `LANG/MAKE=...` sets environment variables and `$(MAKE_JOBS)` expands to `-j...` which is what should be interpreted as a command in this case, I just don't see how it can execute `make` at the end.

SergiiDmytruk:
@tlaurion Try making it an [order-only prerequisite](https://www.gnu.org/software/make/manual/html_node/Prerequisite-Types.html) like this:
```make
all: | bootstrap;
```
This way it won't appear in automatic variables `$<` or `$^`. That trailing `;` isn't necessary in most cases, unless you have whitespace on the next line or the target comes from a rule like `%.o: %.c`, or you want to make absence of recipe explicit.

SergiiDmytruk:
> What i'm looking for is what went wrong under https://github.com/osresearch/heads/commit/7ce12fe621fe95a460c2c5b0038ab12ad6b3b7b0 causing rom sha256sum output to not be part of generated hashes.txt anymore. This is now replaced by gawk/make in hashes.txt as of now and since that commit.

@tlaurion Same problem, probably should be:
```make
all linux cpio run: | $(HEADS_GAWK)
```
Mind that this prevents targets from being rebuilt when gawk is updated. An alternative for both issues would be to use normal prerequisites and specify rom file as a variable rather than abusing prerequisites:
```make
ifeq ($(CONFIG_COREBOOT), y)
all: $(build)/$(BOARD)/$(CB_OUTPUT_FILE)
else ifeq ($(CONFIG_LINUXBOOT), y)
all: $(build)/$(BOARD)/$(LB_OUTPUT_FILE)
```
Then use it instead of `$<` for `sha256sum`, this should prevent this issue reappearing in the future.

tlaurion:
As of now, I can't implement correctly any suggestions of @SergiiDmytruk, ordering not being respected. I get the same result of having only gawk built, unless 9f062ae is implemented. But we still have lost ROM hash under hashes.txt.

My current intuition to improve farther then 9f062ae would be to move make and gawk build instructions to their individual modules files and have their install/ path hardcoded and passed to the rest of the builds from the global Makefile. Modules binaries and libraries are hashed when deployed to their final directories, and for those to happen, then need to be referred in boards, added in Make file inclusion statements from boards specified variables. I think adding gawk (for libgpg-error if I recall well) was added in Makefile because it might have been the easiest thing to do, but it broke other things.

Tinkering with the actual Make file makes my eyes and brain bleed and is not properly understood by anyone who seem to dare looking at it or improving it without having other things breaking. We are, as @SergiiDmytruk said, really abusing prerequisites as if now and understanding what is going on is really not easy to do, making Heads hard to maintain. I separated gawk from make version test condition since having make at version 4.2.1 was not a guarantee that gawk would also be in version 4.2.1. Then after moving away from Debian-10 to debian-11 docker image, I realized that gawk version test was flawed, and passed as good even if gawk was not installed on host. Then recently, a user reported that building from a clean commit was not working, since gawk was built without make having been locally built. This is what 9f062ae is fixing, while not fixing the whole problem. It also seems that LANG C is not filled correctly, @SergiiDmytruk is right, MAKE_OPTIONS is not filled correctly, and the path to make local binary is passed twice. This has no intended good effect since perl is rolling back to default LANG statement...

It would need a rewrite by someone having the right qualifications or at least have the current process properly documented. Otherwise, its not maintained, because its not currently maintainable.

tlaurion:
@SergiiDmytruk @MrChromebox : 
https://github.com/osresearch/heads/pull/1128/commits/790671f1259e288f771233105c973b70c090c3ac 
- This patch on top of master fixes the mess, outputing hashes correctly under hashes.txt for x230-maximized board, taken as an example. 
  - Not sure, here again, why past "all" targets that were defined under board config was not generating the hashes by themselves, which is why I had to add the @sha256sum statement in the past, which I just realized was only outputting on console and not in hashes.txt. Why?

- I am not clear either on what was attempted by attempting to set `LANG = "C MAKE=/home/user/heads/build/make-4.2.1/make" GAWK=/home/user/heads/build/gawk-4.2.1/gawk` and `$(MAKE_JOBS)`, which was not passed correctly, while `$@` is referring to either heads.cpio or tools.cpio which doesn;t make any sense... Which is why I removed them.


Comments? Recommendations?

Makefile
tlaurion: Now I see your point. Actually, I do not even understand why it was put there. Was removed under https://github.com/osresearch/heads/pull/1128/commits/790671f1259e288f771233105c973b70c090c3ac and commented at https://github.com/osresearch/heads/pull/1128#issuecomment-1059491383

SergiiDmytruk:
@tlaurion

I think you want to add
```make
ifneq "$(HEADS_MAKE)" ""
$(HEADS_GAWK): | $(HEADS_MAKE)
endif
```
to make local gawk actually depend on local make, otherwise they can be built in parallel.

`$@` is what passes target name to the nested make command, without it, default target will probably be used every time.

> * From my understanding, https://github.com/osresearch/heads/blob/790671f1259e288f771233105c973b70c090c3ac/Makefile#L259-L260 should produce hashes under hashes.txt for all boards specified `all` targets? Is there anything i'm missing?

`$<` expands to only the first prerequisite.

Another cause of the trouble is that recipes are actually redefined (that `warning: overriding recipe for target 'all'`), which is a problem because earlier definitions are ignored, so those `LANG=C HEADS_GAWK=$(HEADS_GAWK) ...` commands are probably never executed.

I think (but not sure) it might help to split `Makefile` into two files:
1. `Makefile`, which ensures correct environment and then invokes the second one. I think the invocation should look something like the following:
    ```make
    export HEADS_GAWK
    ifneq "$(HEADS_MAKE)" ""
        MAKE=$(HEADS_MAKE)
    endif

    %: bootstrap
        LANG=C $(MAKE) -f Makefile.main $@
    bootstrap: ;
    ```
4. `Makefile.main`, which assumes correct environment and just uses it.

This should help avoid redefining targets.

tlaurion:
@SergiiDmytruk @MrChromebox : 
As of now, I can only have all the rom hashes outputted under hashes.txt OR have make and gawk automatically built as prerequisites. So I now invite the user to build bootstrap first if there is discrepancy between deployed gawk and make from OS and expected local built ones.

My attempts to separate Makefile to Makefile.inc didn't resolve the issue. 

I am not sure I can bring this farther then  c8cf23f.

tlaurion:
https://github.com/osresearch/heads/blob/182bd6bd7517c8e322bc4712db01906b41cce58c/Makefile#L580-L591

Seems to show guideline on having precedence

PR #1130, opened 2022-03-03
Check for /bin/hotp_verification instead of CONFIG_HOTPKEY

marmarek:
CONFIG_HOTPKEY is not exported to the initrd, check for binary presence.


MrChromebox:
so, was the gpg card status call just silently failing before an no one noticed?

tlaurion:
@MrChromebox I guess so since CONFIG_HOTPKEY is never exported.

marmarek:
> so, was the gpg card status call just silently failing before an no one noticed?

It wasn't failing. It was skipped. So, yes, I can imagine how no one noticed.

PR #1136, opened 2022-03-10
oem-factory-reset: reownership with custom PINs/passpwords prompt without changing OEM workflow

tlaurion:
oem-factory-reset: adapt code so that custom passphrases can be provided by user without changing oem factory reset workflow.
oem-factory-reset: output provisioned secrets on screen at the end of of the process for users to review prior of reboot.

fixes: https://github.com/osresearch/heads/issues/1067 (while not reusing https://source.puri.sm/firmware/pureboot/-/commit/c912f256fde7bcd0039e23170260f629521fb819#a6c94f7187de316661024d0ce4f8837609a83516_89_91)

### DIFF
--- a/initrd/bin/oem-factory-reset
+++ b/initrd/bin/oem-factory-reset
@@ -18,7 +18,20 @@ WIDTH="90"
 USER_PIN_DEF=123456
 ADMIN_PIN_DEF=12345678
 TPM_PASS_DEF=12345678
-CUSTOM_PASS=""
+USER_PIN=""
+ADMIN_PIN=""
+TPM_PASS=""
+
+# What are the Security components affected by custom passwords
+CUSTOM_PASS_AFFECTED_COMPONENTS=""
+
+if [ "$CONFIG_TPM" = "y" ]; then
+    CUSTOM_PASS_AFFECTED_COMPONENTS="TPM Ownership password"
+fi
+CUSTOM_PASS_AFFECTED_COMPONENTS="
+$CUSTOM_PASS_AFFECTED_COMPONENTS
+GPG Admin PIN
+GPG User PIN"
 
 RSA_KEY_LENGTH=3072
 
@@ -156,7 +169,7 @@ generate_checksums()
     # create Heads TPM counter
     if [ "$CONFIG_TPM" = "y" ]; then
       tpm counter_create \
-          -pwdo "$TPM_PASS_DEF" \
+          -pwdo "$TPM_PASS" \
           -pwdc '' \
           -la -3135106223 \
           | tee /tmp/counter \
@@ -190,7 +203,7 @@ generate_checksums()
     # sign kexec boot files
     if sha256sum $param_files 2>/dev/null | gpg \
             --pinentry-mode loopback \
-            --passphrase "$USER_PIN_DEF" \
+            --passphrase "$USER_PIN" \
             --digest-algo SHA256 \
             --detach-sign \
             -a \
@@ -266,7 +279,7 @@ set_default_boot_option()
 if [ "$1" != "" ]; then
     title_text=$1
 else
-    title_text="OEM Factory Reset"
+    title_text="OEM Factory Reset / Re-Ownership"
 fi
 if [ "$2" != "" ]; then
     bg_color=$2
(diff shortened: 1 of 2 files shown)


tlaurion:
@jans23 @MrChromebox @daringer please review!

kylerankin:
@tlaurion For the most part all of this looks good to me and I appreciate how you handled defaults. I see that you renamed "OEM Factory Reset" which may confuse some people who are looking at existing documentation on this. At least to me the new name is a bit unclear (OEM Factory Reset is fairly clear) and I suspect it would be to the end-user as well. If you want to rename it maybe you could elaborate on what you'd want the new name to communicate so we can come up with something that's more clear?

tlaurion:
@kylerankin @MrChromebox @jans23 @daringer :

I understand that we are debating proper names to be placed as default window titles:
https://github.com/osresearch/heads/pull/1136/files#diff-bec315d56e315e18809e58210db23903cfa07559d297a1910f2c3151f6e6e97dR270-R272

https://github.com/osresearch/heads/pull/1136/files#diff-bec315d56e315e18809e58210db23903cfa07559d297a1910f2c3151f6e6e97dR568-R572

----
The OEM factory reset mode is triggered when no user key is being present in the rom. So that mode is currently being used by 3 use cases: 
- End users doing initial provisioning of their security components on initial flash
- OEM Factory Reset
- Users receiving replacement USB Security key (lost in the sea use case) and going under Options-> OEM Factory reset from gui-init (name unchanged in this PR.) 

To limit the extent of documentation changes, we could simply combine the naming of current use cases for which oem-factory-reset script is currently used under Heads, having `OEM Factory Reset` being presented first? 

Let's throw out alternatives and vote in seperate posts?

tlaurion:
What about: `OEM Factory Reset / Security components Re-Ownership` ?

kylerankin:
> What about: OEM Factory Reset / Security components Re-Ownership ?

I'm OK with that.

jans23:
I'm ok with it. However, it's a bit lengthy and I don't understand what "Security components Re-Ownership" should mean. Please excuse my ignorance but perhaps there is a more clear alternative to this. What are "security components"? I don't get it from reading the term and instead the concept behind it needs to be explained in any case. Therefore it might be sufficient and more straight forward to simply call it "Re-Ownership" (altogether: "OEM Factory Reset and Re-Ownership"). What do you think?

If not, I would start `Components` with a capital letter.

tlaurion:
@jans23 OEM Factory Reset / Security Components (Re-)Ownership mean reownership of Security Components! 

As of now, they are:
- USB Security dongle's smartcard GPG Admin PIN (GPG key Admin PIN seems to have been pushed)
- USB Security dongle's smartcard GPG User PIN
- TPM Ownership password

But could be extended in case of Re-Ownership (transfer of ownership upon reception) to:
- LUKS Disk Recovery Key passphrase change (LUKS first slot key passphrase, setuped at OS install [slot 0])
- TPM released Disk Unlock Key (DUK) passphrase change (LUKS second slot key passphrase, setuped at default boot setup to seal/unseal TPM nvram memory, released when proper measurements + DUK passphrase [slot 1])

And could later on be extended to:
- Platform locking bypass passphrase  change [recovery shell access passphrase] for Ivy/Sandy bridge with io386 merged, but activated per board config. That needs a way to access recovery shell if needed which a simple hash of the passphrase could be compared against a value stored in CBFS.... or
  - Could be linked as per past discussions only to GPG detached signing verification, but what if dongle is lost.... Needs a fallback
  - Or could be tested against Disk Recovery Key passphrase unlocking LUKS, but would be vulnerable to disk swap or
  - Disk Unlock Key passphrase, but not enforced everywhere or
  - Having GPG keys being generated per Heads, and copied-to-card with backup on encrypted storage, the unlocking of the signing key saved on encrypted storage for the backup keys could be used to detach sign and verify against public key bound to the firmware.... or
  - We simply force the user to boot unsafely until he receives his USB Security dongle replacement.... And only depend on GPG User PIN to detache sign and verify prior of giving access (authentication based on physical ownership of something they should always have.)

So hereby, Security Components (Re-)Ownership would permit to extend those and more, which are linked to initial ownership (not OEM nor factory reset linked) but could be linked to transfer of ownership from OEM->End User (hence (Re-)Ownership).

tlaurion:
@kylerankin @MrChromebox @jans23 @daringer? 
Please provide alternatives that we can thumb up?
`OEM Factory Reset / Security Components (Re-)Ownership`

tlaurion:
`OEM Factory Reset / Re-Ownership`

tlaurion:
OEM Factory Reset and Re-Ownership

kylerankin:
OEM Factory Reset / Reset Secrets

tlaurion:
> OEM Factory Reset / Reset Secrets

Love it. Everybody understands that.

tlaurion:
Seems like it doesn't talk to non-technicals unanimously.

@kylerankin : Ok with `OEM Factory Reset / Re-Ownership`?
I think it related to ownership (initial or) Re-Ownership of what would otherwise be set by OEM.

I will update PR with that and take screenshots

kylerankin:
> Ok with `OEM Factory Reset / Re-Ownership`?

I'm ok with it.

tlaurion:
signal-2022-03-10-153815
signal-2022-03-10-161309
signal-2022-03-10-153828
signal-2022-03-10-161107
signal-2022-03-10-161212

Ready to merge.

initrd/bin/oem-factory-reset
```diff
@@ -276,43 +278,85 @@ fi
...
-echo -e -n "Would you like to set a custom password? [y/N]: "
+echo -e -n "Would you like to set a single custom password applied to physical security components? [y/N]: "
 read -n 1 prompt_output
 echo
 if [ "$prompt_output" == "y" \
   -o "$prompt_output" == "Y" ] \
 ; then
-  echo -e "\nThe custom password will be used for the
-TPM admin and GPG user/admin passwords.
+  echo -e "\nThe chosen custom password will be applied to:
```
MrChromebox, line 306: I think you want to inform the user of this *before* they make the selection, not after

initrd/bin/oem-factory-reset
```diff
@@ -450,7 +496,7 @@ if [ $GPG_EXPORT -ne 0 ]; then
         ERROR=$(tail -n 1 /tmp/error | fold -s)
         whiptail_error_die "Key export error: unable to copy ${GPG_GEN_KEY}.asc to /media:\n\n$ERROR"
     fi
-    umount /media 2>/dev/null
+    mount -o remount,ro /media 2>/dev/null
```
MrChromebox, line 499: why remount here? we're going to reboot immediately after

tlaurion: @MrChromebox the initial script took for granted it was ran from ramfs, which assumption didn't fit my use case.

The reason it was unmounted is because we lack flush support, where everywhere else in code, we make sure changes are flushed by remounting read only.

This is why /media is also mounted only if not mounted already, and if mounted already, remounted in rw to modify content and then here in ro.

initrd/bin/oem-factory-reset
tlaurion: @mrchromebox : "custom password will be applied to all security components" and the user said yes. The other option is to define different pins/passwords for different components.

 I basically separated the output as it was before to be more clear on what was going to be applied and to be able to add more components having the same password if needed before applying that custom password to all components.

This PR was to be able to apply different passwords to different security components and output provisionned passwords before reboot.

You want the concerned security components to be clearer prior of engaging action?

MrChromebox: it just seems odd to enumerate what the password will be applied to after the user has made the choice rather than before. I would just move the echo block up before the read call

tlaurion:
@MrChromebox  c6e39fa:
signal-2022-03-11-105342
signal-2022-03-11-105113
signal-2022-03-11-104739
signal-2022-03-11-104611

initrd/bin/oem-factory-reset
tlaurion: @MrChromebox : better? https://github.com/osresearch/heads/pull/1136#issuecomment-1065250513

initrd/bin/oem-factory-reset
```diff
@@ -276,43 +289,77 @@ fi
...
           * Sign all of the files in /boot with the new GPG key\n\n
         It requires that you already have an OS installed on a\n
-        dedicated /boot partition. Do you wish to continue?\n" \
+        dedicated /boot partition. Do you wish to continue?" \
         $HEIGHT $WIDTH $CONTINUE $CANCEL $CLEAR $bg_color --title "$title_text" ; then
     exit 1
 fi
 
+# Inform user of security components affected for the following prompts
+echo -e "The following security components will be provisionned with defaults or chosen PINs/passwords:
```
MrChromebox, line 311: provisioned

initrd/bin/oem-factory-reset
```diff
@@ -276,43 +289,77 @@ fi
...
+$CUSTOM_PASS_AFFECTED_COMPONENTS\n"
+
 # Prompt to change default passwords
-echo -e -n "Would you like to set a custom password? [y/N]: "
+echo -e -n "Would you like to set a single custom password that will be provisionned to all security components? [y/N]: "
```
MrChromebox, line 315: provisioned

MrChromebox:
does it make sense at this point to offer both a single pin/pass and unique for the user/admin pins and TPM? I'd argue if not going to use defaults, then set a unique for each. drop the single option

kylerankin:
> I'd argue if not going to use defaults, then set a unique for each. drop the single option

At least for how we use the custom PIN feature at Purism, it's to set a user-provided custom PIN for PureBoot to aid in tamper detection. Some of these user-provided PINs are pretty complex and having to type them in that many times (or worse, multiple long PINs) would be pretty burdensome for the team, so I'd prefer we keep the "one PIN to rule them all" option as that is very useful from the OEM perspective.

MrChromebox (approved):

PR #1138, opened 2022-03-12
MDADM Software RAID0/1 support

hardestkit:
Reworked modifications as per discussion [https://github.com/osresearch/heads/pull/1118](https://github.com/osresearch/heads/pull/1118) to improve long term support and reduce fragmentation of board structures. 

On some Lenovo laptops, the CD/DVD device can be removed and a 2nd SATA device added using an inexpensive caddy. This opens the opportunity to install an operating system using RAID-0 striping to offer additional storage capacity, or to use RAID-1 to provide N+1 data mirroring of kernel, heads hashes and signatures (/boot) and LUKS/OS/Storage. Without the changes in this pull request, Heads cannot detect a /boot partition or OS that has been installed using Software RAID and cannot boot the OS, nor can it seal a LUKS disk key to the TPM for a Software RAID device. This PR contains the changes which allow Heads to become compatible with Linux Software RAID, boot a software RAID installation and also seal LUKS key of a software RAID device to the TPM.

**Tested compatible vs expected compatible**

Known Compatible: T430 (Tested: 2nd drive caddy, QubesOS 4.1 with two MD partitions using default QubesOS naming conventions of 'boot' and 'pv00'. In heads these translate to these device mappings: /boot on /dev/md/localhost:boot and LUKS on /dev/md/localhost:pv00)

Expected to work: T420, T520, T530, W530. These machines use the exact same caddy and conventions as the T430, but I do not have these machines to test with. There should be no reason they do not behave exactly the same as the T430 for Software RAID. In addition, these Software RAID compatibility changes should also work for any other device which is (or may be) ported to heads that has capability to use more than one block device.

**Implementation**

To enable simply add the following to the board config file (see boards/t430-hotp-maximized/t430-hotp-maximized.config for example)

`export CONFIG_SWRAID=y`

The user must configure their partitioning via QubesOS installers 'custom' partitioning mode and ensure both /boot is set up as RAID-1 (or RAID-0) as per the following guidelines:

- Select custom partitioning when installing QubesOS 4.1.
- When partitioning in QubesOS installer, the label for /boot can be left blank, but you should ensure to not change the default name of 'boot' as per this image (centre, bottom), as Heads relies on this in init [https://raw.githubusercontent.com/hardestkit/howtos/main/boot-configure.png
](https://raw.githubusercontent.com/hardestkit/howtos/main/boot-configure.png
)

- The Thin pool LVM should be set up over LUKS on RAID-1 or RAID-0 as per [https://raw.githubusercontent.com/hardestkit/howtos/main/thinlvm-configure.png
](https://raw.githubusercontent.com/hardestkit/howtos/main/thinlvm-configure.png
)

- Heads should see the RAID devices as /dev/md/localhost:boot (/boot) and /dev/md/localhost:pv00 (LUKS disk), which you can see when starting up. [https://raw.githubusercontent.com/hardestkit/howtos/main/heads-raid-boot.png
](https://raw.githubusercontent.com/hardestkit/howtos/main/heads-raid-boot.png
)

- Please ensure to use /dev/md/localhost:pv00 when sealing the disk key to TPM, unles syou are sure which /dev/mdXXX device represents your pv00 partition (this can be obtained by checking size in //proc/mdstat or by using mdadm if you like) [https://raw.githubusercontent.com/hardestkit/howtos/main/heads-luks-seal-raid.png
](https://raw.githubusercontent.com/hardestkit/howtos/main/heads-luks-seal-raid.png
)

**IMPORTANT:** CONFIG_BOOT_DEV from the board config file will be overridden by initrd/init to /dev/md/localhost:boot when CONFIG_SWRAID is set. If you use a MD naming conv


shamen123:
@tlaurion appears that having RAID in the x230-linux kernel config isnt something the t420-hotp-maximized board likes.

Perhaps a better approach would be to use a patch that gets applied to the x230 kernel config file in the event that CONFIG_SWRAID is set to y? any other suggestions?

hardestkit:
@tlaurion it appears that having RAID in the x230-linux kernel config isnt something the t420-hotp-maximized board likes re size.

Perhaps a better approach would be to use a patch that gets applied to the x230 kernel config file in the event that CONFIG_SWRAID is set to y? any other suggestions?

config/linux-x230.config
```diff
@@ -99,6 +99,11 @@ CONFIG_BLK_DEV_DM=y
 CONFIG_DM_CRYPT=y
 CONFIG_DM_VERITY=y
 CONFIG_DM_VERITY_FEC=y
+CONFIG_BLK_DEV_MD=y
```
tlaurion, line 102: Those should all be modules, and those modules should be added in modules/linux section at the end, taking e1000e as an example.

Makefile should take those modules and add them in initrd if the board config keyword is added.

And init, load those modules as early as needed if the madm binary is found on to system, loading the modules in the proper order.

Otherwise, the kernel size just grew and it will fail for smaller spi boards like xx20 here.

I am not in front of computer right now, but if all those can be built as modules and not included in kernel for all (linux-x230.config being shared for all boards not being flash) this fails because if space constraints.

Let me know if you need more information!

tlaurion: https://github.com/osresearch/heads/blob/9016ebccc22eabd40245743886488cf6b89b4309/modules/linux#L94

USB controller modules might be a better example to follow: https://github.com/osresearch/heads/blob/9016ebccc22eabd40245743886488cf6b89b4309/modules/linux#L94

hardestkit:
tnx @tlaurion  will get those added and tested then tag again when its time for a review. unsure at this time if i need autodetect compiled into the kernel or not, ill have to test in the coming weeks. definitely raid0 and raid1 can be modules though.

> Makefile should take those modules and add them in initrd if the board config keyword is added.

can you point to an example line? Currently i just have initrd/init check for $CONFIG_SWRAID but i guess having that only added to initrd/init when needed will save a few bytes

tlaurion:
Mmmm I see.

init is not including functions ( . /etc/functions)

So i guess your best option would be to add a new script under initrd/bin, called init-raid as an example, replicating the right insmod order. You can base yourself on how USB modules are loaded here https://github.com/osresearch/heads/blob/9016ebccc22eabd40245743886488cf6b89b4309/initrd/etc/functions#L92, where you can rhrbinclude . /etc/functions and reuse/add helpers there if you consider they could be reused in the future.

And then call that script from init, just before initialising madm the way you are doing it.

Note that the insmod calls there are actually using a wrapper script, which will measure the modules to be loaded https://github.com/osresearch/heads/blob/9016ebccc22eabd40245743886488cf6b89b4309/initrd/sbin/insmod

No clue if autodetect cannot be a module.

You need to have that early in init, otherwise the code dealing with /boot related stuff will fail.

tlaurion:
@hardestkit something like:
https://github.com/osresearch/heads/blob/master/initrd/bin/network-init-recovery#L6-L11

PR #1144, opened 2022-04-01
Fix current builds (zlib 1.2.11 cannot be downloaded, busybox patch not applied)

tlaurion:
- zlib 1.2.12 release is not respecting cross compiling. We save current builds with old zlib.
  - 1.2.11 disappeared from servers: taking another archive link, same hash.
- busybox 1.32.0 was not patched with 1.28.0 patch. Renaming patch so that its applied in fresh builds.


tlaurion:
@MrChromebox @osresearch : please approve.

We would need a release with that.

MrChromebox:
would you mind rewording the commit msg to something like `modules/zlib: fix download URL` to make it more obvious what the patch is doing?

PR #1145, opened 2022-04-01
Make loading of usb-hid kernel module dependent only for USB keyboard support

tlaurion:
@szszszsz @jans23 @kylerankin @MrChromebox : Addresses an old concern about USB HID that shouldn't be part of kernel if no USB keyboard needed: https://github.com/osresearch/heads/pull/836#issuecomment-711176858

(Where USB keyboard is supported in early init code now through enable_usb if CONFIG_USB_KEYBOARD is exported per board config, while usbhid is build as a kernel module in all linux configurations as from this PR.

@DemiMarie : this should do?
@marmarek : new board addition in CI so you have a x230-hotp-maximized_usb-keyboard rom to test at each commit if wanted
@MrChromebox : I touched config/linux-librem_common.config changing usbhid.ko being built as a module, added in initrd if CONFIG_USB_KEYBOARD is added in board config
@Tonux599 : kgpe-d16-workstation_usb-keyboard is also affected here with the same reasoning (so workstation is having that usbhid.ko compiled as a module as well).

Tested: built usbhid as module (not in kernel): HOTP resealing worked after flash.
@szszszsz @kylerankin : where USB HID is needed for Nitrokey/Librem Keys under Heads?

981cb96f253fa8cc1f2b64a1c3997f019e0e503e is added just for builds to work again (zlib 1.2.11 vanished and zlib 1.2.12 doesn't build properly as of now)


initrd/etc/functions
```diff
@@ -122,6 +122,12 @@ enable_usb()
 		|| die "xhci_pci: module load failed"
 		sleep 2
 	fi
+
+	if [ $CONFIG_USB_KEYBOARD" = "y" ]; then
+		insmod /lib/modules/usbhid.ko \
+		|| die "usbhid: module load failed"
```
DemiMarie, line 128:
```suggestion
	if [ "$CONFIG_USB_KEYBOARD" = "y" ]; then
		insmod /lib/modules/usbhid.ko ||
		die "usbhid: module load failed"
```

DemiMarie (commented):
Other than the trivial shell script fix below, would it be possible to check that if the module is not loaded, USB keyboards do not work but USB security dongles do?  That catches both the case where the `usbhid.ko` module is necessary for security dongles and the case where `usbhid.ko` gets (wrongly) autoloaded.  You might want to add `usbhid.ko` to a modprobe blocklist.

config/linux-x230.config
```diff
@@ -184,11 +183,10 @@ CONFIG_MFD_SYSCON=y
 CONFIG_DRM=y
 CONFIG_DRM_I915=y
 CONFIG_FB_VESA=y
-CONFIG_BACKLIGHT_LCD_SUPPORT=y
 # CONFIG_LCD_CLASS_DEVICE is not set
-CONFIG_BACKLIGHT_CLASS_DEVICE=y
```
DemiMarie, line 189: What is the reason for these changes?  I can’t see how they are related to USB devices.

config/linux-x230.config
```diff
@@ -207,9 +205,7 @@ CONFIG_GENERIC_PHY=y
 # CONFIG_DMIID is not set
 CONFIG_GOOGLE_FIRMWARE=y
 CONFIG_GOOGLE_MEMCONSOLE_X86_LEGACY=y
-# CONFIG_EXT2_FS is not set
 CONFIG_EXT4_FS=y
-CONFIG_EXT4_USE_FOR_EXT2=y
```
DemiMarie, line 212: These changes also appear unrelated.

config/linux-x230.config
```diff
@@ -218,7 +214,6 @@ CONFIG_MSDOS_FS=y
 CONFIG_VFAT_FS=y
 # CONFIG_PROC_SYSCTL is not set
 # CONFIG_PROC_PAGE_MONITOR is not set
-CONFIG_TMPFS=y
```
DemiMarie, line 221: This one as well.  While it makes sense to use ramfs (instead of tmpfs) in an environment that will never have swap, this should be a separate PR.

config/linux-x230.config
```diff
@@ -56,7 +56,6 @@ CONFIG_KEXEC_FILE=y
 CONFIG_PHYSICAL_ALIGN=0x1000000
 # CONFIG_MODIFY_LDT_SYSCALL is not set
 # CONFIG_SUSPEND is not set
-CONFIG_ACPI_VIDEO=y
```
DemiMarie, line 59: This one seems unrelated as well.

DemiMarie (commented):
The kernel config changes appear unrelated.

initrd/etc/functions
DemiMarie: Ah, does Heads disable autoloading?

config/linux-x230.config
tlaurion: @DemiMarie : that is just the result of the savedefconfig. I could definitely revert the changes to only that module change.
I am not really ok with the standard of saving in savedefconfig for kernel configurations, this complicates migrating to newer kernel, since defaults change all the time.

DemiMarie: That’s up to you.

initrd/etc/functions
tlaurion: @DemiMarie Its litterally the other way around: there is no autoloading :)
The modules needs to be loaded in the right order to be loaded here.
CONFIG_LINUX_USB_COMPANION_CONTROLLER is old USB1 (only required per KGPE-D16 boards configs)
https://github.com/osresearch/heads/blob/714dbd7ac3790322166df67fe317b170ab4e33ee/initrd/etc/functions#L92-L133

Then when usb storage is needed, the module dependencies (enable_usb) is called just right before. 
https://github.com/osresearch/heads/blob/182bd6bd7517c8e322bc4712db01906b41cce58c/initrd/bin/mount-usb#L11

tlaurion: Thanks for the code review.

DemiMarie: You’re welcome!

tlaurion: @DemiMarie : this is the wrapper to insmod I was talking about earlier: https://github.com/osresearch/heads/blob/51ecbdc8cb7e280744c6a00eb3a8b2a1f116c1be/initrd/sbin/insmod#L1-L46

Where going to the recovery shell (exiting menu guided interactions) scrambles the PCRs, preventing secrets to be unsealed: https://github.com/osresearch/heads/blob/182bd6bd7517c8e322bc4712db01906b41cce58c/initrd/etc/functions#L13

tlaurion:
CircleCI cache was foo bar. Kicking rebuilt will fix this once https://github.com/osresearch/heads/commit/493eb3ebb7820ee943587a9212d4a86855ccd786 )

tlaurion:
> > I'm confused by this PR title, as the PR seems to be doing the exact opposite -- the usb-hid kernel module is being explicitly tied to, not separated from, usb keyboard support. A better title would be 'Make loading of usb-hid kernel module dependent on USB keyboard support'
> > The main commit here really needs to be broken down into smaller chunks:
> > 
> > * changing usb-hid support from Y to M (and being explicit about the change)
> > * adding the usb-hid module to the relevant linux configs
> > * add loading of usb-hid via enable_usb
> > * add new board x230-max-usb-kb
> > * add CI for new board
> > * ~fix: y -> "y"~
> > 
> > Throwing all of this into a single commit makes it very difficult to parse/review/test.
> * There also needs to be tests that USB security keys work _without_ the `usb-hid` driver installed.

@DemiMarie:
Was [tested already](https://github.com/osresearch/heads/pull/1145#issue-1190216499) in initial local PoC, and retested locally:
-  x230-hotp-maximized (USB keyboard doesn't work)
-  x230-maximized_usb-kb (USB keyboard works. That is for your custom QA setup @marmarek, so you can test Ci builds directly when needed.)

@MrChromebox:
- All done. Ping me when reviewed and/or approved.
- librem_l1um librem_mini_v2  and librem_mini are affected

@kylerankin @szszszsz @jans23 @MrChromebox
- Brings https://github.com/osresearch/heads/pull/836 closer to being mergeable.

DemiMarie:
> @DemiMarie: Was [tested already](https://github.com/osresearch/heads/pull/1145#issue-1190216499) in initial local PoC, also [questionning Nitrokey for Nitrokey/Librem Key features requiring USB HID](https://github.com/osresearch/heads/pull/836#issuecomment-1086323589), while retested locally:
> 
> * x230-hotp-maximized (USB keyboard doesn't work, GPG/HOTP functions works )
> * x230-maximized_usb-kb (USB keyboard works, GPG/HOTP functions works.)
> * @marmarek: That is for your custom QA setup,  so you can test Ci builds directly from Heads CircleCI's builds for each commit, when/where needed.)

Thank you, I suspected so, but I wanted to see it written down explicitly.

MrChromebox:
> There also needs to be tests that USB security keys work _without_ the `usb-hid` driver installed.

Looks good here on a Librem 14 as well

tlaurion:
Merging since pending question to Nitrokey is addressed under https://github.com/osresearch/heads/pull/836#issuecomment-1086323589

Issue #1147, opened 2022-04-05
LUKS re-encrypt/change recovery passphrase options don't allow cancellation


MrChromebox:
Selecting the 'abort' option when prompted to select the LUKS container device does not actually abort, but continues with no LUKS device selected. The only way to cancel is via CTRL+ALT_DEL.

select_luks_container() should return non-zero when abort is selected from the file selector.
Functions calling select_luks_container() should check for a non-zero return value and abort accordingly

MrChromebox:
resolved by #1150

PR #1150, opened 2022-04-06
Luks-reencrypt Cleanup

MrChromebox:
* /bin/reencrypt-luks: rename to /etc/luks-functions
* luks_reencrypt(): remove extraneous call to select_luks_container()
* select_luks_container(): return non-zero when no device found/selected
* /etc/luks-functions: exit function when select_luks_container() returns non-zero

Resolves issue #1147


tlaurion:
And what about https://github.com/osresearch/heads/pull/1148/files#diff-8E00308445E86F1EF8E620F39F9B81A3R76

?

MrChromebox:
> And what about https://github.com/osresearch/heads/pull/1148/files#diff-8E00308445E86F1EF8E620F39F9B81A3R76
> 
> ?

what about it? rather than putting everything inside an if block, I simply exit the calling function if the call to select_luks_container() doesn't return zero.

tlaurion:
@MrChromebox Nevermind, I thought that LUKS Recovery Key passphrase would be in the list of security components that would be presented to be provisioned, but it is not.

Replaces #1148 with a cleaner approach. Thanks @MrChromebox !

PR #1151, opened 2022-04-07
Kexec QOL Improvements

MrChromebox:
* kexec-select-boot: Simplify boot menu entries
* kexec-select-boot: Simplify boot selection confirmation, reverse order
* kexec-select-boot: Skip duplicate prompt when setting new default boot entry
* kexec-select-boot: use 'fold' to wrap kernel args at 80 char

igned-off-by: Matt DeVillier <matt.devillier@puri.sm>

### DIFF
--- a/initrd/bin/kexec-select-boot
+++ b/initrd/bin/kexec-select-boot
@@ -185,11 +185,13 @@ scan_options() {
 }
 
 save_default_option() {
-	read \
-		-n 1 \
-		-p "Saving a default will modify the disk. Proceed? (Y/n): " \
-		default_confirm
-	echo
+	if [ "$gui_menu" != "y" ]; then
+		read \
+			-n 1 \
+			-p "Saving a default will modify the disk. Proceed? (Y/n): " \
+			default_confirm
+		echo
+	fi
 
 	[ "$default_confirm" = "" ] && default_confirm="y"
 	if [[ "$default_confirm" = "y" || "$default_confirm" = "Y" ]]; then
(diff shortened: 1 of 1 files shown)


tlaurion:
As discussed off channel, experiments with "fold" command seems to be a tool which we should take advantage off, being part of busybox, so that users are able to confirm that their new default kernel/xen command line is "good" after having inspected it.
Since it will be signed as default after, that would also be a good QOL improvement.

```
diff --git a/initrd/bin/kexec-select-boot b/initrd/bin/kexec-select-boot
index d0266c46..b6b32059 100755
--- a/initrd/bin/kexec-select-boot
+++ b/initrd/bin/kexec-select-boot
@@ -137,7 +137,7 @@ confirm_menu_option() {
                default_text="Make default"
                [[ "$CONFIG_TPM_NO_LUKS_DISK_UNLOCK" = "y" ]] && default_text="${default_text} and boot"
                whiptail $BG_COLOR_WARNING --clear --title "Confirm boot details" \
-                       --menu "Confirm the boot details for $name:\n\n$kernel \n\n" 20 120 8 \
+                       --menu "Confirm the boot details for $name:\n\n$(echo $kernel| fold -s -w 80) \n\n" 20 120 8 \
                        -- 'd' "${default_text}" 'y' "Boot one time" \
                        2>/tmp/whiptail || die "Aborting boot attempt"
```

A wrapper around fold would also help in all whiptail GUI presentation work later on, since personally, this is still my main show stopper trying to make it information all fit on the screen.

PR #1155, opened 2022-04-13
oem-factory-reset: Only set default boot option if no TPM Disk Unlock Key

### LINKED ISSUE #1154: OEM Factory Reset should set a new boot default in all cases?
Reviewing the code and replying to some users concerns, I realized that the OEM Factory Reset/Re-Ownership wizard is selecting putting the grub default boot option as being the default boot option.

Might not be problematic, but requires additional step as of now from users using the OEM Factory Reset for users going into the Re-Ownership way, and users who previously had a default boot enforcing TPM LUKS Disk Unlock Key. 

Since generate_checksums wipes /boot/kexec* files, then calling set_default_boot_option enforces a simpler unattended version of kexec-select-boot, I would recommend:

1. Not calling of set_default_boot_option (Would not fulfill OEM Factory Reset mode)
1. Not calling of set_default_boot_option if kexec_key_devices.txt exists prior of deleting /boot/kexec*

I will propose a PR for 2. In that case, /boot digests would be created, while on reboot, the user selecting default boot would have to enroll a new default boot option, and go through the normal prompts of setting a new TPM released LUKS Disk Unlock Key. 

Otherwise, the user has to go manually through Boot options-> Show OS boot option, and then select Make default boot option after Re-Ownership. 

Can be done through code or documentation.

tlaurion:
This continues to generate checksums and detach sign them per cached GPG User PIN in all OEM Factory Reset/Re-Ownership use cases, but does not set a default boot option when kexec_encrypted_disks.txt is present on /boot (flag left on /boot when a TPM Disk Unlock Key was previously setuped.)

The user hitting Default Boot post Re-Ownership will go through the setup of a new boot default, which will ask him to setup a Disk Unlock Key if desired. Without this, hitting Default Boot goes into asking the user for its Disk Recovery Key passphrase and confuses the user, requiring him to manually setup a default boot option (Boot Options-> Show OS boot Options) to setup a TPM Disk Unlock Key.

Tested: OEM Factory Reset. Manual setup of a default boot option with LUKS Disk Unlock Key (Otherwise default boot boots setuped defaut boot option from Grub config asking for Disk Recovery Key passphrase). Re-Ownership. Default boot asks to setup a default boot option.

Implement second option of #1154. 
Fixes #1154 once merged.

@MrChromebox

### DIFF
--- a/initrd/bin/oem-factory-reset
+++ b/initrd/bin/oem-factory-reset
@@ -156,6 +156,11 @@ generate_checksums()
         mount -o remount,rw /boot || whiptail_error_die "Unable to mount /boot"
     fi
 
+    #Check if previous TPM Disk unlock Key was set
+    if [ -e /boot/kexec_key_devices.txt ]; then
+        TPM_DISK_ENCRYPTION_KEY_SET=1
+    fi
+
     # clear any existing checksums/signatures
     rm /boot/kexec* 2>/dev/null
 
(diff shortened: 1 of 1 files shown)


MrChromebox (approved):

tlaurion:
https://osresearch.net/InstallingOS

Instructions works as is

PR #1157, opened 2022-04-18
Swraid

hardestkit:
_Note: Reworked modifications as per discussion [1118](https://github.com/osresearch/heads/pull/1118 ) and [1138](https://github.com/osresearch/heads/pull/1138) to improve long term support, reduce fragmentation of board structures, to avoid kernel bloat and to use correct KO module loading for PCR measurement._ 

Without the changes in this pull request, Heads cannot detect a /boot partition or OS that has been installed using Software RAID and cannot boot the OS, nor can it seal a LUKS disk key to the TPM for a Software RAID device. 

On some hardware, running software RAID is straightforward. On others, such as compatible Lenovo laptops, the CD/DVD device can be removed and a 2nd SATA device added using an inexpensive caddy. These options present the opportunity to install an operating system using RAID-0 striping to offer additional storage capacity, or to use RAID-1 to provide N+1 data mirroring/redundancy of kernel, heads hashes and signatures (/boot) and LUKS/OS/Storage. 

This PR contains the changes which allow Heads to become compatible with Linux Software RAID, boot a software RAID installation and also seal LUKS key of a software RAID device to the TPM.

Known Compatible: T430 (Tested: 2nd drive caddy, QubesOS 4.1 with two MD partitions using default QubesOS naming conventions of 'boot' and 'pv00'. In heads these translate to these device mappings: /boot on /dev/md/localhost:boot and LUKS on /dev/md/localhost:pv00)

Expected to work: T420, T520, T530, W530. These machines use the exact same caddy and conventions as the T430, but I do not have these machines to test with. There should be no reason they do not behave exactly the same as the T430 for Software RAID. In addition, these Software RAID compatibility changes should also work for any other device which is (or may be) ported to heads that has capability to use more than one block device.

**Implementation**

To enable simply add the following to the board config file (see boards/t430-hotp-maximized/t430-hotp-maximized.config for example)

`export CONFIG_SWRAID=y`

The user must configure their partitioning via QubesOS installers 'custom' partitioning mode and ensure both /boot is set up as RAID-1 (or RAID-0) as per the following guidelines:

- Select custom partitioning when installing QubesOS 4.1.
- When partitioning in QubesOS installer, the label for /boot can be left blank, but you should ensure to not change the default name of 'boot' as per this image (centre, bottom), as Heads relies on this in init [https://raw.githubusercontent.com/hardestkit/howtos/main/boot-configure.png
](https://raw.githubusercontent.com/hardestkit/howtos/main/boot-configure.png
)

- The Thin pool LVM should be set up over LUKS on RAID-1 or RAID-0 as per [https://raw.githubusercontent.com/hardestkit/howtos/main/thinlvm-configure.png
](https://raw.githubusercontent.com/hardestkit/howtos/main/thinlvm-configure.png
)

- Heads should see the RAID devices as /dev/md/localhost:boot (/boot) and /dev/md/localhost:pv00 (LUKS disk), which you can see when starting up. [https://raw.githubusercontent.com/hardestkit/howtos/main/heads-raid-boot.png
](https://raw.githubusercontent.com/hardestkit/howtos/main/heads-raid-boot.png
)

- Please ensure to use /dev/md/localhost:pv00 when sealing the disk key to TPM, unles syou are sure which /dev/mdXXX device represents your pv00 partition (this can be obtained by checking size in //proc/mdstat or by using mdadm if you like) [https://raw.githubusercontent.com/hardestkit/howtos/main/heads-luks-seal-raid.png
](https://raw.githubusercontent.com/hardestkit/howtos/main/heads-luks-seal-raid.png
)

**IMPORTANT:** CONFIG_BOOT_DEV from the board config file will be overridden by initrd/init to /dev/md/localhost:boot when CONFIG_SWRAID is set. If you use a MD naming convention that differs from QubesOS 4.1 standards, please adapt accordingly


hardestkit:
@tlaurion the answer was autodetect was not needed to be in the kernel so there is zero impact on the size of heads/linux kernel/coreboot for any user who does not have export CONFIG_SWRAID=y in the board config.

tlaurion:
@hardestkit :I can't test, but LGTM!
@mrchromebox?

tlaurion:
@hardestkit you do not want to have a separate board config that does build this by CI? (Means duplicating board config with uncommented line, associated coreboot config pointing to proper build path).

Your comment block under board config should also be copied to other expected to work boards?

MrChromebox:
@tlaurion I agree, we want to avoid untested build paths, so having separate board configs which enable SW RAID would be ideal

hardestkit:
@tlaurion  @MrChromebox the agreement already was to not have separate board configs for raid variants to avoid bloat and to make the option a simple configuration variable. [https://github.com/osresearch/heads/pull/1118#issuecomment-1048074253](https://github.com/osresearch/heads/pull/1118#issuecomment-1048074253) . 

By the logic of having build paths for every option in a board config, we would need to include board variants for those with or without SSHd, those without E1000, those with text based gui instead of cairo/whiptail etc. 

However it does make sense to have at least one board with this enabled by default so that circleci can ensure that build work now and in future if the option is turned on by a user. I will add a t430 board variant which contains this option as active so that we have confidence in its operation.

tlaurion:
@hardestkit board configs will depend on linux-x230-legacy.config or linix-x230-maximized.config soon.

Just so you know, your board config addition that should be added to circleci config to be build automatically will need to be debased on master soon following another PR soon to be merged in master.

tlaurion:
Someone was interested by this today in this [matrix channel thread](https://matrix.to/#/!pAlHOfxQNPXOgFGTmo:matrix.org/$zRh6jlneZq_ZuKormmSKM36Ca6nMrk-4osW8Bb6BX44?via=matrix.org&via=nitro.chat&via=tchncs.de)

PR #1163, opened 2022-05-17
Remove fedora public key. They don't detach sign ISOs since before 2020.

### LINKED ISSUE #785: Can't boot Fedora ISOs : CHECKSUM files are not detached signed.
From [Fedora](https://getfedora.org/en/security/):
> curl https://getfedora.org/static/fedora.gpg | gpg --import

- [ ] The public key inside of Heads is still valid

>gpg --verify-files *-CHECKSUM

```
gpg : Good signature from "Fedora (32) .... "
gpg: not a detached signature; file blahblah.iso was NOT verified
```
>sha256sum -c *-CHECKSUM

`sha256sum: WARNING: 21 of 21 computed checksums did NOT match` while doing a sha256sum manually of the iso and comparing with the SHA256SUM file line shows the same hash obtained by validating manually fro command line...

What do we do with that?
- [kexec-iso-init](https://github.com/osresearch/heads/blob/master/initrd/bin/kexec-iso-init#L20) depends on gpgv wrapper to call its [gpg vefification](https://github.com/osresearch/heads/blob/master/initrd/bin/gpgv#L3)
- `gpg --verify-files` doesn't work, probably from gpg changes since [reverted to gpg 2.2.10](https://github.com/osresearch/heads/commit/5cb45bbc9906aeef6e0d1f78eced0b00d31ad8b1)
- `sha256sum -v` doesn't work since we depend on busybox, not linux directly build binaries.

@MrChromebox, any thoughts? 
Not directly linked to #784 but this prohibited me to test directly your PR, since ISO boot was supposed to work. (Fedora, QubesOS and Tails can boot from iso if detached signed file is provided side by side with iso. No idea exactly when Fedora changed but even their archive repos have `-CHECKSUM` now and no `.iso.asc` nor `.iso.sig`.

tlaurion:
Fixes #785


tlaurion:
@MrChromebox

MrChromebox (approved):

PR #1164, opened 2022-05-26
System info improvements

tlaurion:
Thanks @nestire for 06311ff068810839b069039dd2e14eec65bebc84 pushed upon [nitropad heads' fork](https://github.com/Nitrokey/heads/tree/nitropad).

gui-init: Adding Atheros wifi card, another x230 camera id and Synaptic TouchPad detection
linux-x230.config: Add Synaptic TouchPad driver so that it can be detected and reported as present prior of shipping.

@MrChromebox : want to add ids there? Otherwise devices will be shown as "None|Unknown".


MrChromebox:
This seems like a very OEM-specific implementation. Using a VID/DID match to check for components requires keeping updated lists, which for an OEM is one thing, but for the project overall seems like we'll always be behind, until a user complains about having something listed as unknown and/or makes PR.

I'm also not sure of the value of reporting all this in system info - especially the touchpad

tlaurion:
@MrChromebox : The idea here is to make sure that components are properly connected, from an OEM perspective, without having to fork project.

Would you be comfortable if such script would leave System Info alone, but simply have callable script equivalent from command line?

Not fully comfortable with either concepts of forking project to specialize it, but understand it is what is happening right now on Heads, not sure what to do about it or understand from it.

But on my use case, as well, would need to make sure everything is good per model selected prior of re-ownership and shipping, which is why I cherry-picked and extended what was forked.

Ideas on those matters?

tlaurion:
> @tlaurion as this is x230 specific

This is xx30 specific, agreed. (t430/x230/w530/t530 share similar hardware. For the moment, OEMs are selling x230 and t430)

>I think it makes sense to have it be an x230-specific script that you call from the recovery console to validate everything. Otherwise we just clutter up system info with a lot of text that's not useful to a large % of users

Agreed. 
> 
> perhaps we just add an OEM folder where scripts like this can live?
Moved it to initrd/bin/oem-system-info for the moment? Not perfect, but having blink comparison app permits to have in a quick check differences in hardware and quick validation for PCI and USB bus IDs as well. 

I have seperated linux-x230.config into maximized and legacy as well.
@MrChromebox comments?

MrChromebox:
> I have seperated linux-x230.config into maximized and legacy as well.

I'm not sure that belongs in this PR, I don't see how it's related.

I'd also recommend renaming the oem-system-info script to reflect the boards supported, maybe `oem-system-info-xx30`?

tlaurion:
> > I have seperated linux-x230.config into maximized and legacy as well.
> 
> I'm not sure that belongs in this PR, I don't see how it's related.

It is related, since xx30-hotp-verification boards are so tight in space that there is a risk that injection of public key might fail for those boards, depending if user injects public key with picture or additional signatures on his key. Yep, legacy boards are that tight.

> 
> I'd also recommend renaming the oem-system-info script to reflect the boards supported, maybe `oem-system-info-xx30`?

@MrChromebox : Perfect, will change to that

MrChromebox:
@tlaurion any chance of reorganizing/rebasing the commits to clean things up? right now the creation of the oem-system-info-xx30 script is "hidden" in a commit titled 'gui-init: restore master' which isn't ideal. IMO it should look more like:

* split kernel configs
* add ps2mouse modules to maximized kernel cfg
* create oem-system-info-xx30 (w/trackpad info)

the changes to the main system info in gui-init should be squashed away essentially.

tlaurion:
@MrChromebox : should be mergeable. Squashing lost the original commit of https://github.com/Nitrokey/heads/tree/nitropad

tlaurion:
rom and hashes.txt of https://app.circleci.com/pipelines/github/tlaurion/heads/1064/workflows/4b1abbcc-e06e-4d61-8c1e-990cbcbdaf0d/jobs/7088/parallel-runs/0/steps/0-103 :

[17a7d238740593fbd9e1f6af58ef98aa8fedd2dc99ff52d4d36b63031e8996f8.zip](https://github.com/osresearch/heads/files/8882581/17a7d238740593fbd9e1f6af58ef98aa8fedd2dc99ff52d4d36b63031e8996f8.zip)

MrChromebox (approved):
Thanks, these commits are a lot cleaner, much better for readibility/traceability

PR #1168, opened 2022-06-02
bin/oem-factory-reset: prevent users to choose a GPG Admin PIN > 25 chars which would fail HOTP sealing

tlaurion:
Fixes https://github.com/osresearch/heads/issues/1167
Circumvents https://github.com/Nitrokey/nitrokey-app/issues/223
Adds validation so usr cannot enter GPG User PIN > 64 while we are at it.

Note that GPG PINs can be up to 64 characters.
But GPG Admin PIN will fail HOTP sealing with GPG Admin PIN of more then 25 chars.

@MrChromebox: please review.
@jans23: please be aware.


tlaurion:
signal-2022-06-02-134500

Otherwise providing 26 chars GPG Admin PIN would pass provisioning but fails HOTP sealing:
signal-2022-06-02-113225

PR #1172, opened 2022-06-16
WIP xx30: TXT support ?

tlaurion:
I stayed up late last nights, reading the amazing blog posts from @3mdeb, more specifically their [Fobnail](https://blog.3mdeb.com/2021/2021-10-28-fobnail-promotion/) articles, on limitation of Heads without having no hardware RoT. 
Reread coreboot documentation on TXT, specifically [ACM](https://doc.coreboot.org/security/intel/acm.html#retrieving-acms) and [IBB](https://doc.coreboot.org/security/intel/txt_ibb.html)

Reread the blogs post on Dasharo support of [Optiplex](https://blog.3mdeb.com/2022/2022-03-17-optiplex-txt/) when reading that a lot of problems linked to past coreboot of TXT for Ivy bridge (xx30) linked to TXT were resolved from 3mdeb. Clicked the link and saw that they were merged end 2021.... 

Then the lines
>To use the [Intel](https://www.intel.com/content/www/us/en/developer/articles/tool/intel-trusted-execution-technology.html) TXT, you will need the BIOS ACM and the SINIT ACM. The latter is available for download from Intel while the former can be obtained by NDA only. In case you have access to both here is a short guide on how to configure coreboot to use Intel TXT.

I decided to give it a shot again. Last time I digged into this enough (which lead to the Maximized support with ME neutered), it took a lot of digging to realize, at the end of it all, that Lenovo was gentle enough to distribute ME themselves from their website.

I then checked around deeper in the work that they have done on Optiplex, specifically their fwdeploy work, to realize that it might be highly possible that the ACM BIOS blob shared the same [GUID across all Ivy boards](https://github.com/Dasharo/fwdeploy/blob/3e26706ce87a04e054c2659632656037895da902/scripts/extract_image.sh#L3).... Ran UEFIExtract the same way locally on backups, and found the ACM blob.

And then decided to download latest available x230 firmware executable from Lenovo, do the same thing I've done in the past to extract ME.... to realize that Lenovo is distributing the ACM blob themselves and that like ME, it can be directly downloaded and extracted. 

So long story short, blobs/xx30/download_extract_acm.sh in this PR is the result of the beginning of this PoC.
Hopefully, it will take less long then for the Maximized boards to be debugged and understand what was going wrong.

So here I attach the `cbmem -c -1` log I took from an x230-hotp-maximized-acm flashed laptop, which complains that no IBB is defined and that TXT policies are missing, so no PCR0 containing IBB measurements.

[cbmem_logs.txt](https://github.com/osresearch/heads/files/8922610/cbmem_logs.txt)

The documentation is so sparse.... I hope the community will help me make blob based hardware RoT possible. With that, we should have PCR0 and bootblock measured, without the need to lock it down from flashrom, considering that locking the platform just before kexec is possible.

Help needed!
Todos:

- [ ]  txt_bios_policy.bin not present under CBFS (required?)
- [ ] Figure out to have bootblock in IBB (goal of this inclusion if ACM blobs are desired/tolerated)
- [ ] Optional DRTM?!

Thank you to exist @3mdeb!!!!
@miczyg1 @pietrushnic : hints from the logs/link to documentation I missed?  

-------

- Update coreboot target to that it includes coreboot 4.17
- Addition of blobs/xx30/download_extract_acm.sh
  - Downloads SINIT ACM blob from an archive.org copy of the archive
  - Downloads Latest BIOS from Lenovo and extracts the ACM BIOS from there
- Add a x230-hotp-maximized-ac board config based on coreboot 4.17
  - coreboot config includes SINIT and ACM blobs
- CircleCI modifications
  - Add unzip in apt packages requirements
  - Added a step into CircleCI which calls blobs/xx30/download_extract_acm.sh to have blobs
  - Add x230-hotp-maximized-ac builds


tlaurion:
For the moment there is an issue with tls.mbed.org, so builds are failing for the moment.

>Resolving tls.mbed.org (tls.mbed.org)... 79.170.91.36 Connecting to tls.mbed.org (tls.mbed.org)|79.170.91.36|:443... failed: No route to host. make: *** [Makefile:505: /root/project/packages/mbedtls-2.4.2-gpl.tgz] Error 1CircleCI received exit code 0

tlaurion:
Seems like the required tools were not merged inside of coreboot from this discussion 

https://www.mail-archive.com/coreboot@coreboot.org/msg55950.html

Where converged security suite by 9elements require the work to be done eternally on coreboot ROM.

The goal here would be to have bootblock in IBB at build time...

tlaurion:
Seems like this is what I have to chew on https://9esec.io/blog/first-open-source-drtm-implementation/

tlaurion:
Important update: 
Broadwell+ can have IBB measured in PCR0.
Sandy/Ivy bridge can have TXT setuped, which basically was just done with this PR as it is.

Will rebase and rename board to be happened with txt instead of ACM.

miczyg1:
> txt_bios_policy.bin not present under CBFS (required?)

Not required. This is optional if one wants to define some custom policies and deny booting an untrusted OS for example. The advantage of adding a policy to CBFS is that you not need to add it from an external storage, i.e. hard drive, like tboot does.

> Figure out to have bootblock in IBB (goal of this inclusion if ACM blobs are desired/tolerated)

On Sandy/Ivy there is no concept of TXT IBB yet. And so it is not even measured by the ACM during reset. One of the lines says it: `[INFO ]  TXT-STS: IBB not measured`
Also you can notice: `[INFO ]  TXT-STS: BIOS is not trusted` and `S-CRTM Capability rooted in: BIOS` which means the platform does not protect the BIOS/IBB with a hardware ROT like BootGuard (hence "not trusted", because not verified before reset vector). S-CRTM capability can be rooted either in BIOS or processor. If rooted in the processor it means the whole chain of trust beginning with ME bootROM through BootGuard and processor bootROM up to the reset vector. When it says BIOS, either the hardware is from pre-BootGuard era or Boot Guard is not provisioned.

A good comparison of logs with Broadwell for example should reveal these differences.

miczyg1:
One does not need any manifestation unless Boot Guard is enabled or the machine is at least 10th gen Intel Core CPU (which uses CBnT)

pietrushnic:
@miczyg1 I assume this is related with project charter for TrenchBoot as Qubes OS AEM? Can we link issues your created here?

miczyg1:
https://docs.dasharo.com/projects/trenchboot-aem/
https://github.com/TrenchBoot/trenchboot-issues/issues/3
https://github.com/TrenchBoot/trenchboot-issues/issues/4

tlaurion:
Rebased on master.
Cleaned the facts under commit.
Have x230-hotp-maximized-acm board report its rom hashes at build time and under hashes.txt for comparison.

tlaurion:
To be considered under #1282 where ACM + Sinit blobs can enable RoT in CPU!

tlaurion:
Discussion happening under https://forum.qubes-os.org/t/trenchboot-anti-evil-maid-for-qubes-os/16559/10

For D-RTM enablement and QubesOS AEM support next steps

tlaurion:
To investigate and reuse to find ACM blob https://github.com/coreboot/coreboot/blob/master/util/cbfstool/fit.c#L353

tlaurion:
Maybe bad news:
> IBB measurement happens only with server TXT most likely... There is not much documentation about differences between client TXT and server TXT though

Trace of discussion at https://matrix.to/#/!pAlHOfxQNPXOgFGTmo:matrix.org/$vCp3QEGJ_JhrVAtcqrK0oiUlDHYf5xLcLGUzigVWsmU?via=matrix.org&via=nitro.chat&via=talk.puri.sm

ansiwen:
I gave up on TXT and switched to BtG 1.0 for enabling PCR-0 IBB measurements on the ProDrive Hermes board (C248 chipset, Cannonlake/Coffeelake): https://gist.github.com/ansiwen/255174aa96d6dc34680d95957d9d6fe9

miczyg1:
Just to clarify the possibilities of TXT measuring the IBB to PCR0, a few things that have to be known:

1. There are two Intel TXT variants: client and server. Each have a different "feature set" and behavior.
2. There are ACM and SACM. The difference is that the latter is Startup ACM, which means it is intended to be run at reset vector. The former are to be run by the BIOS code. There is a bit in the ACM header which indicates if it should be run at reset vector or not. (https://cdrdv2.intel.com/v1/dl/getContent/315168 Table 4 bits 6-7 to distinguish server/client TXT, and ACM heard in APPENDIX A ModuleSubType, although the description is misleading, because 0 doesn't;t necessarily mean SINIT ACM, the real BIOS ACM vs SINTI ACM info is described by ChipsetACMType in Table 10 - yeah confusing...)
3. Client TXT BIOS ACMs are not SACMs, thus the CPU does not run it, despite it is present in FIT. Server TXT ACMs are SACMs and are run at reset vector and those MEASURE IBB.
4. ProDrive Hermes board (C248 chipset, Cannonlake/Coffeelake) platform is considered as an entry server board based on client Cannon Lake H chipset, so no server TXT SACMs for it... What is considered a genuine server board then? Probably something between Xeon Scallable Processors (or basically anything that can do two sockets and more). A few codenames that are genuine servers: Purley, Whitley (IceLake-SP), Skylake-SP, CascadeLake, CooperLake, Sapphire Rapids, Emerald Rapids.
5. Of course trying to run the Server TXT SACM will fail because of CPU or chipset mismatch.
6. May require setting the TXT bit in the ME and running SPS firmware. When the TXT was set, the status bits in CPU TXT registers did change.


These are the conclusions we got with @ansiwen when researching the TXT IBB measurement possibilities.

miczyg1:
@ansiwen I believe we cannot get over this one:

```
Client TXT BIOS ACMs are not SACMs, thus the CPU does not run it, despite it is present in FIT. Server TXT ACMs are SACMs and are run at reset vector and those MEASURE IBB.
```

The CPU simply won't load the TXT BIOS ACM if it is not SACM

miczyg1:
> A TXT SACM for the C62X chipset can be found in this firmware: https://www.intel.de/content/www/de/de/download/18911/intel-server-board-s2600wf-family-bios-and-firmware-update-package-for-uefi.html

Despite that, ACMs have a Processor ID list (see https://cdrdv2.intel.com/v1/dl/getContent/315168 Table 13 and Table 14), which will prevent loading the ACM on arbitrary processors, like non-server ones in case of the ACM you are talking about.

ansiwen:
> @ansiwen I believe we cannot get over this one:
> 
> > Client TXT BIOS ACMs are not SACMs, thus the CPU does not run it, despite it is present in FIT. Server TXT ACMs are SACMs and are run at reset vector and those MEASURE IBB.
> 
> The CPU simply won't load the TXT BIOS ACM if it is not SACM

@miczyg1  Sure, but the BtG SACM seems to have _some_ TXT support, because it reacts to the TXT bit, and the Gigabyte board officially has TXT support in the BIOS settings. And I wonder what would happen if this TXT support would be successfully enabled.

ansiwen:
> > A TXT SACM for the C62X chipset can be found in this firmware: https://www.intel.de/content/www/de/de/download/18911/intel-server-board-s2600wf-family-bios-and-firmware-update-package-for-uefi.html
> 
> Despite that, ACMs have a Processor ID list (see https://cdrdv2.intel.com/v1/dl/getContent/315168 Table 13 and Table 14), which will prevent loading the ACM on arbitrary processors, like non-server ones in case of the ACM you are talking about.

Sure, I just wanted to share the find, in case someone has a compatible chipset.

miczyg1:
> > > A TXT SACM for the C62X chipset can be found in this firmware: https://www.intel.de/content/www/de/de/download/18911/intel-server-board-s2600wf-family-bios-and-firmware-update-package-for-uefi.html
> > 
> > 
> > Despite that, ACMs have a Processor ID list (see https://cdrdv2.intel.com/v1/dl/getContent/315168 Table 13 and Table 14), which will prevent loading the ACM on arbitrary processors, like non-server ones in case of the ACM you are talking about.
> 
> Sure, I just wanted to share the find, in case someone has a compatible chipset.

Compatible chipset AND CPU :)

tlaurion:
So if I get this all well, the only way around this is to forget about IBB measured bootblock and go trenchboot's way, enabling drtm with acm blobs from ivy bridge and up and go integrate trenchboot so that SLCM will be responsible to verify that the bootblock matches checksum inside of cbfs whenever requested up to whatever is desired to be measured?


From what I'm reading above, SRTM with SACM measuring bootblock as part of IBB is reserved to server platforms only and cannot be generalized.


Per prior work, Haswell and previous board configs under Heads have "platform locking" prepared by coreboot and enabled prior of final kexec call, locking write access to SPI from OS as of today. I'm still wrapping my head around how different drtm would prevent tampering of both bootblock and bootblock hash in cbfs but it seems that energy invested should go in that direction which should continue under an Heads specific trenchboot integration issue.

@miczyg1 thoughts/corrections?

miczyg1:
> go integrate trenchboot so that SLCM will be responsible to verify that the bootblock matches checksum inside of cbfs whenever requested up to whatever is desired to be measured?

Not sure what SLCM is. There is no point in verifying bootblock in CBFS against its checksum in CBFS, both things can be replaced to malicious content.

tlaurion:
> > go integrate trenchboot so that SLCM will be responsible to verify that the bootblock matches checksum inside of cbfs whenever requested up to whatever is desired to be measured?
> 
> Not sure what SLCM is. There is no point in verifying bootblock in CBFS against its checksum in CBFS, both things can be replaced to malicious content.

@miczyg1 I meant SACM, typo.

Another issue was created under https://github.com/system76/firmware-open/issues/396

I asked about this in both general and random dasharo channels and crosslinked the posts.

PR #1178, opened 2022-06-21
Remove local gawk and make builds, add xx30 top and bottom hashes under hashes.txt

tlaurion:
- Makefile: remove local gawk and make references to see if everything builds
- xx30 boards: add proper statements to have roms output hashes under hashes.txt
- ~.circleci/config.yml: remove bootstrap~
- patches/make-4.2.1.patch removal

Since gawk was outputed into hashes.txt in a fashion I was not able to fix, this is the solution on top of debian-11 based distributions (in terms of gawk and make being > 4.2.1 which was problematic before). 

Replaces #1128.

TODO: If works, remove heads-wiki need to build bootstrap prior of boards


tlaurion:
Some historical analysis: 
- gawk locally build happened into commit https://github.com/osresearch/heads/commit/7ce12fe621fe95a460c2c5b0038ab12ad6b3b7b0 and discussed under https://github.com/osresearch/heads/pull/811 while real solution was to upgrade libgpgerror to libgpg-error 1.37. The patch was integrated and then the gpg toolstack was version bumped later on under https://github.com/osresearch/heads/pull/714, while the gawk local build continued to be present under Makefile. Seems like it should have been removed after version bumping gpg toolstack to newer version.
- Locally building make has an even older history which integration into Heads comes from https://github.com/osresearch/heads/issues/353#issuecomment-373039370, this doesn't fit current build systems, where even debian-10 came with 4.2.1.

There seems to be no reason to keep those from now on. On a reproducibility perspective (reproducible builds broken as of now anyway. But building a commit should always follow CircleCI steps, which defines also the docker image and OS used and dependencies installed. Until we move to another building system or integrate layers for better reproducibility, having gawk and make being built right now is not bringing any value and causes issues (hashes containing gawk per $@ call instead of roms hashes since gawk being locally built being the unseen regression for a while now).

tlaurion:
Based on top of #1182
@MrChromebox : please review

MrChromebox (changes requested):
this should really be 2 separate commits, unless there's a dependency I'm not seeing

tlaurion:
It will be the rebased once #1182 is merged.

This was just a test to see and confirm that builds stop failing with #1182 merged. And that test succeeded.

tlaurion:
@MrChromebox unless you mean removing gawk and make modules and Makefile references in one commit

And another commit adding proper statements to get back rom hashes under hashes.txt into a separate commit?

And then another commit to remove CircleCI bootstrap step?

(The goal of this commit is to bring back ROM hashes, which those 3 actions depend to test this commit...)

tlaurion:
Forgot to remove make patch.

@MrChromebox you want 4 seperate commits

tlaurion:
@MrChromebox seperated in two commits.

MrChromebox:
> @MrChromebox seperated in two commits.

perfect, thx. I'll test then approve

tlaurion:
@MrChromebox Can't wait for merge. Sick of CircleCI builds failing for race conditions linked to old make

tlaurion:
@MrChromebox : conclusion?

PR #1196, opened 2022-08-02
dropbear: Use mirror, main host is down

JonathonHall-Purism:
Switch to mirror https://mirror.dropbear.nl/


JonathonHall-Purism:
@tlaurion Come to think of it I got this error earlier this week, grabbed the package from a cache and forgot about it, so the host has been down for a bit.  Not sure if it'll come back but this mirror is up.

PR #1201, opened 2022-08-21
Makefile: replace patch with git apply

tlaurion:
Otherwise binary patches cannot be patched/created

Additional fixes needed
- flashrom patch was invalid and got catched by git apply. Correcting
- gpg2-2.2.21.patch was pointing to bad target. Correcting

----

This commit is included by #967 as of now, otherwise coreboot patchsets (and other patches including binary patches) cannot be applied by patch.

Once this is merged, #967 will be rebased on master.

Without the present change, [x230-fhd-edp patchset, targeting graphical configuration blob (vbt) cannot be applied](https://github.com/osresearch/heads/pull/967/files#diff-4eeb4abb7c19d1c1aaf5c84906b32586eb85230907d6ee12e2bc885178a23dfbR132-R172)

Discussions on removing `--reject`, adding `--3way` etc happened under https://github.com/osresearch/heads/pull/967#issuecomment-1217046809

Switching to git patch with `--reject` is actually giving better insights on patch failing to apply in builds, showing which chunks failed to apply and producing reject files for those chunks which are easier to understand.

As a consequence, two patches needed to be remade, since git apply doesn't apply fuzzing. Those are included in this PR as they were part of the changes needed to pass `patch` to `git apply`.

This was used to build all boards successfully while x230-hotp-maximized and x230-maximized-fhd_edp rom images were tested functional under #967.

Leaving this PR open for review for a couple of days prior of merging.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -242,7 +242,7 @@ define define_module =
 	git clone $($1_repo) "$(build)/$($1_base_dir)"
 	cd $(build)/$($1_base_dir) && git submodule update --init --checkout
 	if [ -r patches/$($1_patch_name).patch ]; then \
-		( cd $(build)/$($1_base_dir) ; patch -p1 ) \
+		( git apply --verbose --reject --binary --directory build/$($1_base_dir) ) \
 			< patches/$($1_patch_name).patch \
 			|| exit 1 ; \
 	fi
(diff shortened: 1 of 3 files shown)


tlaurion:
FYI @JonathonHall-Purism approved here https://github.com/osresearch/heads/pull/967#issuecomment-1217975788 already

tlaurion:
@SergiiDmytruk depending of what is merged first, this should also work in your branches

SergiiDmytruk:
Thanks, looks like I need to update a couple of patches.

PR #1206, opened 2022-08-26
whiptail: no more whiptail reseting console on call (--clear)

tlaurion:
So we have console logs to troubleshoot errors and catch them correctly

---
@JonathonHall-Purism linked to separate discussion.

This is useful to call scripts (eg. gui-init) from recovery shell so that scripts can be ctrl-c to see console output.

Any error should be corrected in caller scripts anyway and provide at least a warning (warn) to console.

### DIFF
--- a/initrd/bin/config-gui.sh
+++ b/initrd/bin/config-gui.sh
@@ -14,7 +14,7 @@ while true; do
     unset param
   else
     unset menu_choice
-    whiptail $BG_COLOR_MAIN_MENU --clear --title "Config Management Menu" \
+    whiptail $BG_COLOR_MAIN_MENU --title "Config Management Menu" \
     --menu "This menu lets you change settings for the current BIOS session.\n\nAll changes will revert after a reboot,\n\nunless you also save them to the running BIOS." 0 80 10 \
     'b' ' Change the /boot device' \
     's' ' Save the current configuration to the running BIOS' \
--- a/initrd/bin/flash-gui.sh
+++ b/initrd/bin/flash-gui.sh
@@ -7,7 +7,7 @@ set -e -o pipefail
 
 while true; do
   unset menu_choice
-  whiptail $BG_COLOR_MAIN_MENU --clear --title "Firmware Management Menu" \
+  whiptail $BG_COLOR_MAIN_MENU --title "Firmware Management Menu" \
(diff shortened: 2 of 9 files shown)


tlaurion:
Further improvements on #1243.

JonathonHall-Purism:
Agree, this is great.  A while back I had to send out a few custom builds just to capture some error output, this is much better.  Did a few quick tests in qemu-fbwhiptail-tpm1-hotp.

tlaurion:
Rebased.

tlaurion:
@JonathonHall-Purism Merge? I think this will bring some corner case errors from the community we might want to join force fixing, but at least a quick video capture+screenshot extracted from it should now capture it all and help us understand what is happening quicker.

Don't hesitate proposing better approach that 2 seconds sleep on die and 1 second on warn. Alternative thought on my side was prompting user, but that is way too invasive. We could go on having a trace log under /tmp at some point as well, and an option from menu to copy logs to USB drive?

JonathonHall-Purism:
I thought about pausing/prompting on die too, but I agree I think there is too much chance for false positives to go all the way there in one change.  IMO let's go with the 2 seconds, get some field experience with that, and if there are no indications of false positives then consider prompting.

"Copy logs" is a good idea too, but I think this improvement already reaches the point of diminishing returns.  With this change we can just ask to hit Ctrl+C and take a picture of the screen, that's a pretty low barrier to get some output.  "Copy logs" would be good too but we'd probably get most of the relevant output just from this.

I did something crude once to add a "show log" to TPM reset errors that displayed the log with a pager: https://source.puri.sm/firmware/pureboot/-/compare/Release-21...support-4574?from_project_id=217  This would really need a centralized log though to be useful in general.

I say ship it! :ship:

tlaurion:
> With this change we can just ask to hit Ctrl+C and take a picture of the screen, that's a pretty low barrier to get some output.

Unfortunately there is no such thing as Ctrl-C'ing gui-init unless you go to recovery shell then back to gui-init from there, which of course will break HOTP/TOTP/Disk unlock key "unsealed" from TPM.

So as of master, user really needs to film and extract pictures since there is no output|tee output redirection into something useful to troubleshoot anything.

But I agree not wiping console output is useful in between whiptail prompts.

Reviewing change one last time to make sure my sed didn't break anything and merging.

tlaurion:
Testing this as part of https://github.com/osresearch/heads/pull/1234#issuecomment-1317318212

Really useful! Now possible to go to recovery shell to see errors and warnings which are now still on the console.

Merging!

PR #1121, opened 2022-02-22
Pass all modules from O2/O3 (optimized for speed) to Os (optimize for space)

tlaurion:
Adresses @aesrentai suggestion to pass modules optimization from O2 (performance) to Os (space) here: https://github.com/osresearch/heads/issues/590#issuecomment-821788703

Will comment on gains after build succeeds.
~Still impossible as of now to build xx30-flash boards on top of linux 5.10.4......~ EDIT: without kernel networking support, this now passes.

PR #1214, opened 2022-09-16
xx30 - pack cryptsetup2 instead of cryptsetup

tlaurion:
Having #1116 merged and/or #1121 is required for xx20 boards to support cryptsetup2 since xx20 only have 8mb of SPI flash.

Those requirements are not valid for xx30 boards, which can be updated now to have TPM disk encryption key setuped for default boot option for Q4.1


tlaurion:
Build happened for same commit prior of making PR, which is why no builds are showing I guess. 
Those happened under https://app.circleci.com/pipelines/github/tlaurion/heads/1201/workflows/274f7676-2468-400f-91b6-a164c406b155

Artifacts can be downloaded as normal under a board's artifact section.

tlaurion:
No need for code review, merging.

PR #1177, opened 2022-06-21
T430 dgpu support

tlaurion:
merges @eganonoa and fixes on top of it by @walliams.

Nothing of my hand here. Just cherry-picked commits without merging master into @walliams branch.

Replaces #1175

PR #1175, opened 2022-06-17
Add support for t430 dgpu-versions

walliams:
original configs by @eganonoa , merged and conflicts resolved by me.

PR #1218, opened 2022-09-23
Add support for t430-dgpu variant

weyounsix:
maximized boards only

thanks to @eganonoa for the original configs


weyounsix:
if all clear, this should replace PR #1177 (which replaced PR #1175 )

PR #1184, opened 2022-06-23
Linux x230 cleanup remix

tlaurion:
Continuing  #1116 on x230 linux config tuning.
- flash (minimal)
- legacy (reduced)
- maximized 

As opposed to currently stored linux config, those are not stored in defconfig but full oldconfig format.

-----
Rationale: 
As explained under https://github.com/osresearch/heads/pull/1116#issuecomment-1163207850 the reasoning is to facilitate linux kernel version bumps and for users to understand what changed between those directly from the configuration files.

----

Addition:
- MMC card reader

Removal:
- AMD related stuff
- Many network cards that were supported in kernel since last kernel version bump.
- CPU governers not linked to PERFORMANCE
- Random number generators not being TPM/Intel HW support
- DEBUG related stuff

---------

This is based on top of #1178 which will hopefully be merged soon, after which this will be rebased. 

The only relevant commit is WiP here, which will change until satisfaction of size comparison and functionality is met, not causing regression. Then those kernel configurations will be used to have kernel version bump, with clear differences this time.

PR #1232, opened 2022-10-23
Librem image shrink

### LINKED ISSUE #1229: Question: Getting rid of unneeded kernel options on Librem 14
**Is your feature request related to a problem? Please describe.**
I'm trying to add BTRFS (will affect #1202) and exFAT filesystems support with the corresponding tools which means we'll need some space on flash to contain those. While digging the partial config (and comparing it to the full one after loading it) I found several options I'm not sure we want to have.

**Describe the solution you'd like**
So, those options found fall in several categories:

Debug/Verbosity:
```
CONFIG_DYNAMIC_DEBUG=y
CONFIG_DEBUG_INFO=y
CONFIG_DEBUG_INFO_DWARF4=y
CONFIG_GDB_SCRIPTS=y
CONFIG_DEBUG_FS=y
CONFIG_STACKTRACE=y
CONFIG_KALLSYMS=y
CONFIG_UNUSED_SYMBOLS=y
CONFIG_TRIM_UNUSED_KSYMS=n - probably should allow this one
CONFIG_PNP_DEBUG_MESSAGES=y
CONFIG_ATA_VERBOSE_ERROR=y
CONFIG_DEBUG_MISC=y
```
Do we really need any of those in non-debug production setup?

Ancient/obsolete syscalls/features:
```
CONFIG_X86_MPPARSE=y - do we have any target without acpi?
CONFIG_X86_VSYSCALL_EMULATION=y
CONFIG_X86_IOPL_IOPERM=y
CONFIG_LEGACY_VSYSCALL_NONE=n - probably should allow this one
CONFIG_ZONE_DMA=y - are there any devices on any target using this?
CONFIG_ISA_DMA_API=y - isn't it ancient?
```
Do we have anything using these?

Unused features:
```
CONFIG_CPU_ISOLATION=y
CONFIG_X86_5LEVEL=y - isn't it only for new systems with hue memory?
CONFIG_PCI_IOV=y - are  there any targets with pci vfs needed during boot?
CONFIG_RSEQ=y
CONFIG_NR_CPUS=64 - probably should limit it to like 16
CONFIG_RD_ZSTD=y - we don't use zstd initramfs anyway
```

Support for systems it will never be run on:
```
CONFIG_CPU_SUP_HYGON
CONFIG_CPU_SUP_ZHAOXIN
XZ_DEC_POWERPC
XZ_DEC_IA64
XZ_DEC_ARMTHUMB
XZ_DEC_SPARC
```

I'm also not sure about these three:
```
CONFIG_PRINTK=y - will make the kernel mute, but do we care?
CONFIG_BUG=y
CONFIG_MULTIUSER=y - the question is are we using multiuser system or everything runs as root now?
```

**Describe alternatives you've considered**
Applied -Os patch from @tlaurion mentioned in #590 

**Additional context**
Would be great to have any input about these parameters. If there is a chance any of them will make the system unusable it would be nice to know...
Also, maybe it's a good idea to enable discussions in project for questions like this? Issues don't really feel appropriate for it)

Unb0rn:
This should resolve #1229
Tried not to break anything for any Librem machine, however I am able to test it only on Librem 14

Still need some input on some options:
```
CONFIG_X86_IOPL_IOPERM - legacy. Is there something in heads still using it?
CONFIG_ZONE_DMA - are there any devices on librem machines still using it?
CONFIG_ISA_DMA_API - isn't it ancient?
CONFIG_RSEQ
CONFIG_X86_REROUTE_FOR_BROKEN_BOOT_IRQS - do we need it? I think its only needed for some pretty old specific HW
CONFIG_SYN_COOKIES=y - the kernel is almost network-less. Is it some kind of dependency?
CONFIG_PRINTK=y - will make the kernel mute, but do we care?
CONFIG_BUG=y
CONFIG_MULTIUSER=y - the question is are we using multiuser system or everything runs as root now?
```
The format is also switched to full config instead of defconfig


config/linux-librem_common_full.config
```diff
@@ -0,0 +1,3265 @@
...
+CONFIG_CC_HAS_INT128=y
+CONFIG_ARCH_SUPPORTS_INT128=y
+# CONFIG_CGROUPS is not set
+# CONFIG_NAMESPACES is not set
+# CONFIG_CHECKPOINT_RESTORE is not set
+# CONFIG_SCHED_AUTOGROUP is not set
+# CONFIG_SYSFS_DEPRECATED is not set
+CONFIG_RELAY=y
+CONFIG_BLK_DEV_INITRD=y
+CONFIG_INITRAMFS_SOURCE="../../../blobs/dev.cpio"
```
tlaurion, line 147: Should be `CONFIG_INITRAMFS_SOURCE="@BLOB_DIR@/dev.cpio"`

tlaurion:
@Unb0rn replied to your questions under https://github.com/osresearch/heads/issues/1229#issuecomment-1289109686

JonathonHall-Purism:
Hi @Unb0rn, thanks again for the work on this!

Could you please split this up into a few commits so it is easier to review?  I'd suggest one commit converting to full-config (with no other changes), and another commit making the config changes we want.  Anything along those lines is more reviewable though, right now it is hard to see what exactly was changed (i.e. I think BTRFS was enabled here too, which isn't directly related).

You can replace the existing config file, and then we'll get CI builds I can help test.  Right now this doesn't affect anything because the Librem boards still reference the existing config.

Regarding the specific configs in the description, I mentioned some of those in #1229 but didn't catch all of them:
```
CONFIG_X86_IOPL_IOPERM - Should be OK AFAIK, we don't need to run any binaries not specifically built for Heads on this kernel
CONFIG_ZONE_DMA - I'm not aware of anything requiring this, but I'm not sure, probably just need to test it.
CONFIG_ISA_DMA_API - Might be needed for LPC devices, which includes the TPM on L1UM IIRC.
CONFIG_RSEQ - Sounds removable just from some reading about it, but I don't think it adds much to bzImage either
CONFIG_X86_REROUTE_FOR_BROKEN_BOOT_IRQS - I don't think this is needed
CONFIG_SYN_COOKIES=y - I don't think this is needed
CONFIG_PRINTK=y - Keep this please for users that might encounter unexpected issues, for testing, etc.
CONFIG_BUG=y - Keep this too please, as above, I think it is risky to silently ignore the conditions BUG tests for
CONFIG_MULTIUSER=y - I'm OK with removing this as long as it works, i.e. nothing we ship in Heads depends on the syscalls removed by this, we can try it
```

config/linux-librem_common_full.config
Unb0rn: Fixed this one. Remained from Purism's Pureboot config

Unb0rn:
@JonathonHall-Purism yeah, you were right. Updated defconfig just to build and test everything, removed BTRFS and exFAT as they're unrelated.
Regarding low-power subsystem - Isn't it disabled in current defconfig? Other than that, leaved printk() and bug() in place
One more question to you and @tlaurion 
Found some more "bloat" parameters:
```
CONFIG_IOMMU_SUPPORT
VIRTIO_MENU
VHOST_MENU
CONFIG_PTP_1588_CLOCK
```
Do we need any of it in our minimal image?

tlaurion:
> One more question to you and @tlaurion Found some more "bloat" parameters:
> 
> ```
> CONFIG_IOMMU_SUPPORT
> VIRTIO_MENU
> VHOST_MENU
> CONFIG_PTP_1588_CLOCK
> ```
> 
> Do we need any of it in our minimal image?
@Unb0rn I do not think so. @JonathonHall-Purism ?

modules/linux
```diff
@@ -202,8 +202,43 @@ $(build)/$(BOARD)/$(LINUX_IMAGE_FILE).bundled: \
 	@touch $@ # force a timestamp update
 	@sha256sum "$@" | tee -a "$(HASHES)"
 
+# modifydefconfig target allows us edit current in tree defconfig config
+# under linux decompressed+patched directory and put it back in git tree 
+# to check changes with git difftool 
+# useful for development cycle of linux kernel version bumps.
+linux.modifydefconfig:
```
JonathonHall-Purism, line 209: This is great, I had been doing this manually!

JonathonHall-Purism:
Agree, I don't think we need any of those.

Unb0rn:
Well, it looks I started celebrating a bit too early)
I shou;d've paid more attention - in fact what I thought to be a success was a failure with flashrom. It looks like it tries to detect if it is being run as root, and as we don't have those syscalls in non-multiuser kernel it fails with "Not implemented" error. So I think I will make several more tests and push the fixed version later

Unb0rn:
It turned out to be even more interesting - it had nothing to do with multiuser support, however flashrom still uses iopl() or ioperm(), so had to re-enable it. Double checked to be working.
VHOST_MENU, VIRTIO_MENU and PTP_1588 are also disabled, and regarding IOMMU aren't we using it at all for isolation?

tlaurion:
No IOMMU device isolation under Heads.

tlaurion:
- EXTFAT_FS support will definitely be a global improvement for users: Thumb drives and ssd drives generally comes formatted like that. (currently missing)
- As noted under #1184, I notice a lot of things built as modules which are currently not instructed to be packed under initrd, resulting in useless build time (no impact on size of kernel per se, but unneeded if packed as modules and not instructed to be added under Heads per heads global Makefile condition for that config entree defined under board config)
  - This is the case for a lot of crypto stuff
  - AES as module? Suprised it is not braking things, for example with cryptesetup-reencrypt where cryptsetup is instructed to be compiled intoto and using kernel crypto backend?
- IOMMU: No isolation from Heads as said. Though from what I recall, this is setuped there and was needed to have i915 working (xx20 and xx30 here, not Librem), but my human memory might be tricking me. Would have to check what happens when disabled.
-

JonathonHall-Purism:
@tlaurion @Unb0rn I tested this thoroughly on L1UM-1X8C and did some basic testing on Mini v2.  Looks great, no problems at all, and saves a little over 500 KB on the payload.  Let's merge it :100: 

Thanks for the patience while this progressed through my priority list too :bow:

tlaurion:
@JonathonHall-Purism points addressed under https://github.com/osresearch/heads/pull/1232#issuecomment-1320339869 to be dealt separately?

tlaurion:
OK!

Note that dynamic debugging kernel options are removed from kernel from this point!

Merging!

Issue #1001, opened 2021-06-26
NixOS does not boot


macpijan:
NixOS does not boot after installation (the installer boots fine). The kexec command line is not created correctly.

The original grub.cfg file:

```
# Automatically generated.  DO NOT EDIT THIS FILE!

    search --set=drive1 --fs-uuid 56D5-4720
    if [ -s $prefix/grubenv ]; then
    load_env
    fi

    # ‘grub-reboot’ sets a one-time saved entry, which we process here and
    # then delete.
    if [ "${next_entry}" ]; then
    set default="${next_entry}"
    set next_entry=
    save_env next_entry
    set timeout=1
    else
    set default=0
    set timeout=5
    fi

    # Setup the graphics stack for bios and efi systems
    if [ "${grub_platform}" = "efi" ]; then
    insmod efi_gop
    insmod efi_uga
    else
    insmod vbe
    fi

        insmod font
        if loadfont ($drive1)//converted-font.pf2; then
        insmod gfxterm
        if [ "${grub_platform}" = "efi" ]; then
        set gfxmode=auto
        set gfxpayload=keep
        else
        set gfxmode=1024x768
        set gfxpayload=text
        fi
        terminal_output gfxterm
        fi

            background_color '#2F302F'

        insmod png
        if background_image --mode 'normal' ($drive1)//background.png; then
        set color_normal=white/black
        set color_highlight=black/white
        else
        set menu_color_normal=cyan/blue
        set menu_color_highlight=white/blue
        fi


menuentry "NixOS - Default" --unrestricted {
search --set=drive1 --fs-uuid 56D5-4720
  linux ($drive1)//kernels/65f9va1ql6x2vy23x4nqw0x18jz3a3x8-linux-5.10.45-bzImage init=/nix/store/0jpc6ilx8xikwrmcgbqsvakl1inbpbxj-nixos-system-nixos-21.05.1187.8112fbe212f/init loglevel=4
  initrd ($drive1)//kernels/2jaca9jz9qcr08vlbvgnjn2kz3cbp1yf-initrd-linux-5.10.45-initrd
}


submenu "NixOS - All configurations" {
menuentry "NixOS - Configuration 1 (2021-06-26 - 21.05.1187.8112fbe212f)"  {
search --set=drive1 --fs-uuid 56D5-4720
[...]

PR #1240, opened 2022-11-10
Ignore grub device specs in boot entry paths

### LINKED ISSUE #1239: USB boot to PureOS installer shows boot options that don't work
When booting to a PureOS installer on USB, Heads shows some options that always fail when selected.

Two of the options have `($root)` in them (a variable).  We could safely ignore these (they're the EFI boot options, and the BIOS boot options are fine), but some code in `kexec-parse-boot` makes me think it is actually trying to expand variables, so I can't just ignore any boot options containing variables:

```
		echo $(eval "echo \"$entry\"")
```

@tlaurion Is there any reason to keep any boot options with variables in them?  I checked a handful of ISOs and have not found any (Debian live, Fedora live, Qubes installer, although there are tons more...).

JonathonHall-Purism:
Fixes #1239, #1001

Ignore device specs in GRUB boot entry paths, assume they point to /boot.  Check that boot kernel and initrd are valid files; ignore options that would not be able to boot.

Tested booting the following ISOs (directly written to USB disk): Debian (live KDE, netinst still has graphics init issues), Fedora KDE, Qubes, PureOS.

Tested booting Qubes ISO from ext4-formatted USB disk (loopback ISO boot).

Tested booting installed PureOS and Qubes.

### DIFF
--- a/initrd/bin/kexec-parse-boot
+++ b/initrd/bin/kexec-parse-boot
@@ -94,11 +94,17 @@ grub_entry() {
 			modules="$modules|module $path"
 			;;
 		linux*)
-			kernel=`echo $trimcmd | cut -d\  -f2`
+			# Some configs have a device specification in the kernel
+			# or initrd path.  Assume this would be /boot and remove
+			# it.  Keep the '/' following the device, since this
+			# path is relative to the device root, not the config
+			# location.
+			kernel=`echo $trimcmd | sed "s/([^)]*)//g" | cut -d\  -f2`
 			append=`echo $trimcmd | cut -d\  -f3-`
 			;;
 		initrd*)
-			initrd="$val"
+			# Trim off device specification as above
+			initrd="$(echo "$val" | sed "s/([^)]*)//g")"
 			;;
 	esac
 }
--- a/initrd/bin/kexec-parse-boot
+++ b/initrd/bin/kexec-parse-boot
@@ -39,41 +39,41 @@ check_path() {
 }
 
 echo_entry() {
-	if [ "$kexectype" = "elf" ]; then
-		if [ -z "$kernel" ]; then return; fi
+	if [ -z "$kernel" ]; then return; fi
 
-		fix_path $kernel
-		# The kernel must exist - if it doesn't, ignore this entry, it
-		# wouldn't work anyway.  This could happen if there was a
-		# GRUB variable in the kernel path, etc.
-		if ! check_path "$path"; then return; fi
-		entry="$name|$kexectype|kernel $path"
-		if [ -n "$initrd" ]; then
-			for init in $(echo $initrd | tr ',' ' '); do
-				fix_path $init
-				# The initrd must also exist
-				if ! check_path "$path"; then return; fi
-				entry="$entry|initrd $path"
-			done
-		fi
-		if [ -n "$append" ]; then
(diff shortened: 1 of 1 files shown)


tlaurion:
@JonathonHall-Purism Will download and test Tails for functional test and regression.
Can also boot Debian installer on Talos.

Something else?

JonathonHall-Purism:
Thanks @tlaurion.  I think that's pretty comprehensive.  Was that the Debian network installer?  It still has graphics init issues on Mini v2 that I haven't had a chance to dive into yet, but I'm glad it works on Talos if so.

tlaurion:
the minimal install

> Thanks @tlaurion. I think that's pretty comprehensive. Was that the Debian network installer? It still has graphics init issues on Mini v2 that I haven't had a chance to dive into yet, but I'm glad it works on Talos if so.

On Talos this is network installer, correct.
Unfortunately, optional variables are preset in the middle of the linux line as follows:
```
function boot_one {
        echo "Loading ..."
        linux   /install/vmlinux $options --- quiet
        initrd  /install/initrd.gz                 
}                                                  
                                                   
menuentry "Default install" {                      
        boot_one                                   
}        
```

So booting from detached signed iso didn't work there and still requires to construct manual kexec line from recovery shell to boot from iso mounted to /boot as mounted per verified detached signature.

For Fedora Silverblue, it booted just right, but the installer doesn't output to ast, which is unrelated to this issue and stays right there. Will need to troubleshoot more why anaconda output is not working on the framebuffer output, where drm and ast drivers are loaded correctly.

Have not tested yet for Tails on x86

tlaurion:
Hmmm. 

Unless there would be additional "function" parsing, aside of menuentry parsing, it is curretly impossible for current kexec-parse-options to deal with ppc64le netinst's grub.cfg:
```
set default=2
set timeout=-1

insmod echo
insmod gzio
insmod minicmd
insmod normal

menuentry "Debian GNU/Linux installer boot menu" {
	true
}

menuentry "" {
	true
}

function boot_one {
	echo "Loading ..."
	linux	/install/vmlinux $options --- quiet
	initrd	/install/initrd.gz
}

menuentry "Default install" {
        boot_one
}

menuentry "Automated install" {
	set options="auto=true priority=critical"
	boot_one
}

menuentry "Expert install" {
	set options="priority=low"
	boot_one
}

menuentry "Rescue mode" {
	set options="rescue/enable=true"
	boot_one
}
```

Which is totally normal since there is no menuentry containing a valid kernel+initrd line. 
All menu_entry rely on grub internal logic to call boot_one function. Hopefully this is a corner case and not a lot of ISOs are using this....

The following workaround succeeds to install debian ppc64le iso after usb-init call fails, not findoing any menuentry:
```
kexec -l --initrd=/boot/install/initrd.gz /boot/install/vmlinux
kexec -e
```

tlaurion:
Tails boots just fine.
Fedora 36 Silverblue as well (after anaconda install from vnc since ast driver problem is still not figured out. 

Where boot menu options would benefit being alonged since ostree0 is common across different boot options

But where setting boot option shows ostree difference:



So: All good for the relevant change! Could not test Archlinux/NixOS as of now, But I guess we can expect issue to be reopened if not fixed (should be fixed by code review)

PR #1241, opened 2022-11-11
Add CONFIG_BOOT_RECOVERY_SERIAL to qemu board configs to interact with host through serial

tlaurion:
As a result, you will get expected pause_recovery (etc/functions) output "Hit enter to proceed to recovery shell:" prompt on ttyS0 (stdio: where qemu is launched) while whiptail is shown only once on host (qemu/kvm). 

This is important in case of TPM, since entering recovery shell will invalidate measurements and prevent TPM disk encryption key to be unsealed if going to recovery shell.

By removing the CONFIG_BOOT_RECOVERY_SERIAL board configuration export, only dmesg is redirected to ttyS0 (helpful to debug)


tlaurion:
@JonathonHall-Purism please review!

JonathonHall-Purism:
Works for me, did a quick test of qemu-fbwhiptail-tpm1-hotp under Debian Qemu/KVM.  Looks great :ship:

tlaurion:
We can improve on that on #1245

PR #1243, opened 2022-11-14
init: add 2 seconds pause on die, 1 second on warning so users see console messages

tlaurion:
@JonathonHall-Purism please review.


JonathonHall-Purism:
Looks good to me :ship:

PR #1245, opened 2022-11-14
WiP: Digging console problems (help welcome)

tlaurion:
Somewhere in the past, a kernel upgrade/busybox upgrade required from Heads to change console fixate `GPG_TTY=/dev/console`

All was good, until we needed to have multiple console outputs to multiple serial/console output.

This is WiP with too many commits, aimed as a trace to find the best way of detecting what is the current tty from scripts, in the goal of having different behaviors if we are under a fb console or serial-console and be able to refuse to spawn fbwhiptail if we are over serial.

This is working-ish.

----
busybox adds readlink
fbwhiptail patch uses readlink to see what is reported from /procx/self/fd/0 as console
We try to use the RECOVERY_CONSOLE_SERIAL for qemu and Talos workstation's "bmc" terminals

Any hints there on how to do things better?


initrd/init
```diff
@@ -1,6 +1,6 @@
 #!/bin/ash
 mknod /dev/ttyprintk c 5 3
-echo "hello world" > /dev/ttyprintk
+echo "ttyprintk hello world on tty: $(readlink /proc/self/fd/0)" > /dev/ttyprintk
```
tlaurion, line 3: empty as of now.
Replacing this with $(tty) gives "not a tty".

patches/fbwhiptail-0f14a409735b71c219e0b9b3ee63cdae709ba143.patch
```diff
@@ -0,0 +1,9 @@
+--- fbwhiptail-0f14a409735b71c219e0b9b3ee63cdae709ba143/whiptail.orig	2022-11-14 15:21:23.799000000 -0500
++++ fbwhiptail-0f14a409735b71c219e0b9b3ee63cdae709ba143/whiptail	2022-11-14 15:22:17.498000000 -0500
+@@ -1,2 +1,6 @@
+ #!/bin/sh
++if [[ $(echo $(readlink /proc/self/fd/0)) == "/dev/hvc0" || $(echo $(readlink /proc/self/fd/0)) == "/dev/ttyS0" ]]; then
++	. /etc/functions
++	die "fbwhiptail non-usable on serial/non-fb enabled consoles. Run commands manually."
```
tlaurion, line 7: This "die" only shows under direct all of "whiptail".
Otherwise whiptail calls from gpg-gui or gui-init fails through scripts failing to call whiptail, not whiptail wrapper script which is used when fbwhiptail is defined under board configs.

tlaurion:
Interestingly enough, pts under Heads is mounted successfully from 
https://github.com/osresearch/heads/blob/a9bee2885b784417e541cedfe57f2fc213a673c9/initrd/init#L23-L25

And required `CONFIG_UNIX98_PTYS=y` linux config is present under related kernel configurations, as required.
We would expect consequently /dev/pts/* to exist, but they don't.

We have /dev/pts/ptmx and that's it.

Under talos, we have plentory of /dev/pty* and /dev/tty* 
Same under qemu.

The cause of the issue not permitting us to have `GPG_TTY=$(tty)` seems to be linked to not being able to detect what is our console from scripts.

Any input welcome

tlaurion:
Interestingly enough, pts under Heads is mounted successfully from 
https://github.com/osresearch/heads/blob/a9bee2885b784417e541cedfe57f2fc213a673c9/initrd/init#L23-L25

And required `CONFIG_UNIX98_PTYS=y` linux config is present under related kernel configurations, as required.
We would expect consequently /dev/pts/* to exist, but they don't.

We have /dev/pts/ptmx and that's it.

Under talos, we have plenty of /dev/pty* and /dev/tty* though.
Same under qemu-tpm boards.

The cause of the issue not permitting us to have `GPG_TTY=$(tty)` seems to be linked to not being able to detect what is our current running console from scripts.

Any input welcome.

tlaurion:
Working for qemu boards, where another implementation is used to have dual output on Talos board which can be used as reference for this.

Note that qemu boards enable this by having a recovery waiting console from host tty (waiting for enter) where Talos whiptail board outputs whiptail (console base output) to both BMC and VGA console.

PR #1200, opened 2022-08-19
init: assign CONFIG_TPM depending on /dev/tpm0 presence

SergiiDmytruk:
A small piece of https://github.com/osresearch/heads/pull/836 ([this file](https://github.com/osresearch/heads/pull/836/files#diff-63d474ce25ab525343c0d061b286b1cd45199cd14fdf9ea535f396fd8be5de45)) to avoid adding `-tmp`/`-no-tpm` board variants in https://github.com/osresearch/heads/pull/1002 and other cases.

Is it OK that after `gui-init` prints `Unable to locate /boot files on any mounted disk` message `$skip_to_menu` equals `true`, `$TOTP` isn't set and main menu doesn't display `NO TPM` until you press <kbd>r</kbd>?

Also, with these changes all `CONFIG_TPM` in board files is ignored, so maybe it needs to be removed from them or dynamic behaviour should be used only if `CONFIG_TPM=y` in config.

PR #1246, opened 2022-11-23
gui-init: Always go to main menu when selecting "continue to main menu"

JonathonHall-Purism:
If the user selects "continue to main menu" from an error, do not show any more error prompts until reaching the main menu.

We still try to initialize everything (GPG, TOTP, HOTP) so that the main menu can still show TOTP/HOTP if GPG is not configured, etc., but no more prompts are shown after selecting "continue to main menu".

### DIFF
--- a/initrd/bin/gui-init
+++ b/initrd/bin/gui-init
@@ -9,6 +9,12 @@ export BG_COLOR_MAIN_MENU=""
 . /etc/luks-functions
 . /tmp/config
 
+# skip_to_menu is set if the user selects "continue to the main menu" from any
+# error, so we will indeed go to the main menu even if other errors occur.  It's
+# reset when we reach the main menu so the user can retry from the main menu and
+# # see errors again.
+skip_to_menu="false"
+
 mount_boot()
 {
   
@@ -45,7 +51,7 @@ mount_boot()
         exec /bin/usb-init
         ;;
       m )
-        skip_gpg_check="true"
+        skip_to_menu="true"
         break
         ;;
       * )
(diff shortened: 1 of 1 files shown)


JonathonHall-Purism:
This was a side effect of #1200, it was possible that after selecting "continue to main menu" you could now get more error prompts instead of the main menu.  It now skips prompts if "continue to main menu" was already selected so you always get the main menu next.  I didn't notice this when initially testing #1200 but I saw it now while doing release tests.

It still tries to initialize all of GPG, TOTP, and HOTP, which was the intent behind the original change - you can still get a valid TOTP code on the main menu after a GPG error, etc., if TOTP was indeed valid.

One specific way to produce this is to do a clean flash, then cancel the OEM reset prompt.  (Or, on HOTP boards, boot without an HOTP token, since that suppresses the OEM reset prompt.)  In the past you'd get the GPG error and then could continue to the main menu, but now you get both GPG and TOTP errors before reaching the main menu.

initrd/bin/gui-init
```diff
@@ -214,13 +227,13 @@ update_hotp()
...
     # Don't output HOTP codes to screen, so as to make replay attacks harder
-    hotp_verification check $HOTP
+    hotp_verification check "$HOTP"
     case "$?" in
       0 )
         HOTP="Success"
         BG_COLOR_MAIN_MENU=""
       ;;
-      4 )
+      4|7 )	# 4: code was incorrect, 7: code was not a valid HOTP code at all
```
JonathonHall-Purism, line 236: 7 happens here if `HOTP` contained some "Unable to ...." error from unseal-hotp.  It doesn't mean that the token is not connected, it means we don't have a code, so I added it here.  Also quoted "$HOTP" above, otherwise the error was word-split and triggered an invalid-arguments error from hotp_verification.

tlaurion:
Tested on x230 with hotp variant. 
Restesting master to see what you mean

tlaurion:
Right. 
Refusing OEM factory reset, choosing main menu option resulted into another HOTP/TOTP sealing prompt.

Merging

PR #1256, opened 2022-12-14
Revert deleted race condition patch remove, coreboot413 http download patch

tlaurion:
@JonathonHall-Purism Sorry about that. I deleted a coreboot 4.11 patch by error.
I though the http transcient fix was for 4.11 where it was for 4.13. This fixes that.


JonathonHall-Purism:
OK, LGTM :+1:

PR #1259, opened 2022-12-19
gui-init: show boot option was not returning when verify_global_hash failed

### LINKED ISSUE #1257: GPG verification errors in OS boot menu silently make the user return to previous menu
**To reproduce:**

1. Make sure that GPG verification fails by modifying some file.
2. Reboot & select "Show OS boot menu" & select some boot option.

**Expected behaviour:**

Error is shown.

**Actual behaviour:**

User is returned to the previous menu, no error is shown.

tlaurion:
~Fixes #1257~
Nope

### DIFF
--- a/initrd/bin/gui-init
+++ b/initrd/bin/gui-init
@@ -535,9 +535,12 @@ show_system_info()
 select_os_boot_option()
 {
   mount_boot
-  if verify_global_hashes ; then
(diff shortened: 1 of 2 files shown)


tlaurion:
Hypothesis is that /etc/function's check_config should return 1 when no kexec.sig or no kexec*.txt files are found, or should probably die instead

PR #1261, opened 2022-12-28
Makefile: Use path agnostic bash location

Baitinq:
This patch changes the bash location in the makefile from /bin/bash to /usr/bin/env bash. The latter is a more reproducible location as it is common to more *nix systems which don't contain the former, such as NixOS.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -110,7 +110,7 @@ $(shell mkdir -p "$(initrd_lib_dir)" "$(initrd_bin_dir)")
 # proceed with the build.
 
 # Force pipelines to fail if any of the commands in the pipe fail
-SHELL := /bin/bash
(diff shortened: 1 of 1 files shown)


tlaurion:
@JonathonHall-Purism This implements first step into making Heads current buildsystem being able to transition to being build on top of *Nix as per https://github.com/osresearch/heads/issues/936#issuecomment-1351750448 discussions.

(This will need all blobs/ scripts also being modified, which have path to python hardcoded as well as other troubleshooting along the way)

@JonathonHall-Purism : I have nothing against merging this: that would be an improvement.

JonathonHall-Purism:
Agree, let's merge it.  Thanks!

PR #1264, opened 2023-01-04
coreboot-4.11: Fix remaining patch to work with git apply

JonathonHall-Purism:
The last PR fixing these patches missed this one because it was mistakenly deleted and restored later.  Resulted in 'error: patch with only garbage at line 22'.

Built librem_l1um on Debian 11 host to verify.


tlaurion:
So sorry about that, I thought that file was deleted and unneeded

Issue #1266, opened 2023-01-04
HOTP : do not increment counter when HOTP USB Security dongle is not connected ?


tlaurion:
/boot/kexec_htop_counter will increment even when then USB dongle is not connected when time to remote attest at boot when a HOTP board config is used.

If the machine is booted multiple times without HOTP USB Security dongle connected at boot, the counter on USB dongle and what is stored under boot falls out of sync, generating a `HOTP: Invalid code` error which is reported to the user.

In that case, TOTP code is still generated and can still be verified on smartphone, but HOTP needs to be regenerated to correct the situation.

TODO: 
- confirm what is the tolerance threshold of difference between /boot HOTP counter and counter on dongle (10 expected)
- Properly warn user that booting multiple times without the USB Security dongle connected at boot will eventually result in error in the future and not take habit of not booting without dongle?

----

Reported by user in slack/matrix channel. @kylerankin answered with insights at https://app.element.io/#/room/#OSFW-Heads:matrix.org/$nFGyGk-ZBgHuDIlqn6GSDIxY91TqElIgQd733u85xj0

tlaurion:
@JonathonHall-Purism : after discussion and after evaluation security considerations, removing Heads HOTP counter increase might cause side effects that are undesired.

The current 5 increments tolerance difference justifies 5 boot without visual attestation of the firmware. Considering that a user if verifying TOTP code in this case might be prejudicial. At the end of the day, HOTP being desynchronized, if TOTP is verified instead, can be skipped by a user that has lost his HOTP USB Security dongle and waits for a replacement. 

But a user booting a left unattended platform and having a correct TOTP while an invalid HOTP seems to be the only way currently stating that the machine was attemtpted to be booted more then 5 times without the HOTP dongle.

Are we sure we want to remove increments in the absence of HOTP USB Security dongle connected at boot?
It seems that it might be a documentation issue more then a code issue, users being unaware (I wasn't aware) that booting 5 times without dongle connected would result in HOTP counter being unsyned. A way to communicate this properly with the user would be better. 

Here I mean: the next time that the user is booting with his HOTP dongle connected and the counters are unsynced, instead of just throwing an error, it would be better to warn the user that the laptop has been booted more then the threshold so that the user is aware of that fact.

@kylerankin ? @JonathonHall-Purism ?

tlaurion:
@kylerankin said:
[...]

PR #1267, opened 2023-01-04
TOTP code mismatch: better guidelines in code

tlaurion:
Fixes https://github.com/osresearch/heads-wiki/issues/103 

@JonathonHall-Purism : comments?
Also note that we should maybe cover with one rock (two birds) https://github.com/osresearch/heads/issues/1266

### DIFF
--- a/initrd/bin/gui-init
+++ b/initrd/bin/gui-init
@@ -472,11 +472,14 @@ show_tpm_totp_hotp_options_menu()
 prompt_totp_mismatch()
 {
   if (whiptail $BG_COLOR_WARNING --title "TOTP/HOTP code mismatched" \
-    --yesno "TOTP/HOTP code mismatches could indicate either TPM tampering or clock drift:\n\nTo correct clock drift: 'date -s HH:MM:SS'\nand save it to the RTC: 'hwclock -w'\nthen reboot and try again.\n\nWould you like to exit to a recovery console?" 0 80) then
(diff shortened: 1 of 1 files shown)


tlaurion:
GUI Results in:


Console output results in:



Then some issues resulting in "Alarm clock" as reported under https://dev.archive.openwrt.org/ticket/14541

Which if we removed -q option for ntpd would go forever:


Let's remember that network-init-recovery tried to do ntp sync with router, based on DNS provider (router) first, and then tries to talk to ntp.pool.org first resolving address through DNS, then talking to it through NTP. Through my setup (qemu under Qubes appvm) ntp (UDP) doesn't succeed).

@JonathonHall-Purism : we could document further on command line the process if needed for network-init-recovery:
- First: address is attempted to be obtained dynamically through DHCP locally, etting DNS server from response
- Second: NTP sync is attempted on DNS server provided by DCP response.
- Third: if NTP against DNS server fails, ntp.pool.org is attempted to be resolved and then NTP sync is attempted on IP address (UDP). Might be firewalled.

Thoughts?

JonathonHall-Purism:
Regarding the immediate change, it is certainly an improvement, I've had to help people a handful of times to deal with RTC resets.

I don't think we need to spell out everything `network-init-recovery` does.  I think it'd be pretty verbose and would probably distract from the more important information.

Re: #1266, per discussion in Matrix channel, if we can avoid the problem entirely (not increment the counter when the HOTP dongle isn't present), I think that'd be much better than more gotchas and warnings.

tlaurion:
#1266 needs more thought. To be dealt individually.

@JonathonHall-Purism merge this one?

JonathonHall-Purism:
Yep merge it, thanks :ship:

PR #1272, opened 2023-01-11
/boot: make sure operations requiring to be happening under /boot are under subshells

### LINKED ISSUE #1270: Initial install - new file /boot/.auditing-0
> ### Initial install - new file .auditing-0
> 
> I did a clean setup from scratch in qemu (with PureOS but I don't think the distro matters). Everything went smoothly until it was about to boot - somehow a file `.auditing-0` was created between signing the contents /boot and actually booting, so the boot failed due to a new file. I'm not sure what this file is or what created it, but I re-ran the whole process and it was definitely created by something in Heads at that point (it was not there just after OS install, or even after the OEM reset and TOTP/HOTP reset).
> 
> @tlaurion Any idea what creates `/boot/.auditing-0`? I could not find an obvious reference and searching came up empty so far. Maybe GPG or something?

@JonathonHall-Purism : actually, this is a bug.
Depending of when tpm operations are done, sometime currentdir is `/boot` and sometimes its `/`
So when tpm operations are done when currdir is /boot, that file is created.

That should be another issue which solution is to make sure that each time we do write operations is under a subshell.
As said previously, in my tests, the only time that file was created was when sealing a TPM disk unlock key. Will have to dig code deeper to understand why this case is creating `.auditing-0` under /boot.

_Originally posted by @tlaurion in https://github.com/osresearch/heads/issues/1262#issuecomment-1374018818_

tlaurion:
Fixes #1270

@JonathonHall-Purism that seems to do it.

### DIFF
--- a/initrd/bin/kexec-save-default
+++ b/initrd/bin/kexec-save-default
@@ -122,7 +122,7 @@ if [ ! -d $paramsdir ]; then
 fi
 rm $paramsdir/kexec_default.*.txt 2>/dev/null || true
 echo "$entry" > $ENTRY_FILE
-cd $bootdir && kexec-boot -b "$bootdir" -e "$entry" -f | \
+kexec-boot -b "$bootdir" -e "$entry" -f | \
 	xargs sha256sum > $HASH_FILE \
 || die "Failed to create hashes of boot files"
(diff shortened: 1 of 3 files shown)


JonathonHall-Purism:
LGTM :+1:   Re-tested the clean signing flow, verified no .auditing-0 file was created.  :ship:

PR #1275, opened 2023-01-12
Last clear removal add reboot after totp on screen

tlaurion:
Recently, fbwhiptail calls removed `--clear` as part of calls clearing console output in the goal of facilitating troubleshooting.

Two last `printf clear` existed in the codebase, preventing to see full previous console output in case of errors.
This corrects this.

A separate commit currently adds a reboot after showing TOTP Qrcode on screen, which is basically the only secret showed on screen under heads. This seperate commit adds a `reboot` after asking the user to type a key.

@JonathonHall-Purism : we could alternatively replace that reboot by a `printf clear` if you prefer: let me know and I will change it.

### DIFF
--- a/initrd/bin/gui-init
+++ b/initrd/bin/gui-init
@@ -157,6 +157,7 @@ generate_totp_htop()
     echo "Once you have scanned the QR code, hit Enter to continue"
     read
   fi
+  reboot
 }
 
 update_totp()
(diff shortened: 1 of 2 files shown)


JonathonHall-Purism:
@tlaurion I would prefer a clear rather than reboot after showing the QR code please.  Otherwise looks good

tlaurion:
Replaced by #1277

PR #1276, opened 2023-01-12
recovery call: since we operate under set -e, make sure that shred failing doesn't exit

tlaurion:
This was exposed through testing #1262 and attempting to call `recovery` instead of `die` under proposed changeset under https://github.com/osresearch/heads/pull/1262#issuecomment-1381079323

This should be merged prior of #1262, which should be rebased on top of master.


tlaurion:
@JonathonHall-Purism : as discussed!

JonathonHall-Purism:
Looks good to me, let's merge :+1:

PR #1277, opened 2023-01-13
Adjust printf clears

tlaurion:
Recently, fbwhiptail calls removed --clear as part of calls clearing console output in the goal of facilitating troubleshooting.

This PR:
- removes last two printf clear that were getting in the way of having traces on console output (functions)
- adds only needed printf clear on console, which is after output and scan of TPMTOTP QR code confirmed to be scanned by user (was a reboot under #1275)

Replaces #1275 

@JonathonHall-Purism : please review


JonathonHall-Purism:
Looks great, thanks :+1:

PR #1278, opened 2023-01-13
kexec-save-default: do /boot operation in subshell

tlaurion:
fixes screwed commit 6f6f37ec3a8d7cab1c963ce4e6614f89132aeecf

@JonathonHall-Purism : otherwise scanning / not /boot
sorry about that.


JonathonHall-Purism:
Looks correct, thanks, let's merge

PR #1280, opened 2023-01-16
Make crypttab overridable + deal with Debian-11 crypttab default location (TPM disk unlock key unsealing) 

tlaurion:
Replaces #894

Debian based OSes (Debian-11-xfce default partitioning tested with encryption) creates 3 partitions (1,2,5) and LUKS container inside of initramfs is parsing only /cryptroot/crypttab (not /etc/crypttab). 

This PR adds the capacity to pass a custom crypttab filepath inside of board configuration, but deals correctly with Debian based OS (to be confirmed at least also with PureOS) which initramfs's `/scripts/local-top/cryptroot` script is only taking into consideration.

Details at https://github.com/osresearch/heads/pull/894#issuecomment-1384506525


tlaurion:
Replaced by #1284

PR #1281, opened 2023-01-16
/etc/functions: fix order when loading usb modules to prevent warning

tlaurion:
Master:

With this fix:


JonathonHall-Purism:
Makes sense, looks good, let's merge :+1:

PR #1284, opened 2023-01-17
WiP : TPM disk unlock key setup (kexec-save-key) reuses OS initrd's crypttab files as base for /secret.key override (kexec-insert-key)

### LINKED ISSUE #1283: Setting a TPM disk encryption key prevents OS discards down to the LUKS and actual disk layer
Revisiting #894 codepaths, testing #1188 with debian-11-xfce (unencrypted /boot and optional LUKS container for OS), while trying to fix TPM disk unlock key to work for something else then Qubes OS under #1280 made me realize that as of today, the cpio that is created as part of unsealing TPM disk encryption key contains a crypttab and the secret that is applied on top of initramfs.

That codepath was created prior of Qubes 4.1 changing the defaults to have its installer apply discards (TRIM) properly from the FS layers (EXT4 in Templates/qubes), down to the LVM pool layer to setting LUKS correctly to have discard applied there as well, so that TRIM operations can be applied correctly per SSD firmware (which then varies per vendor implementation).

Current Heads implementation
------
When a user's decide to setup a boot default (highly recommended), the user is offered to setup a TPM disk unlock key, which will take not only[ firmware measurements](https://osresearch.net/Keys/#tpm-pcrs) (coreboot's reported measured boot under PCR2) until Heads payload is loaded, which then extend measurements of ifs persistently stored components into PCR7 (GPG key material and config.user override kept under coreboot's CBFS storage area), but also measurements of kernel modules at time of sealing in PCR5 and the bootmode in PCR4, while making sure that booting the default boot option won't unseal the secret if the LUKS header hash changed from what was measured at sealing under PCR6. All of the boot options (kernel options, xen options, kernel and xen binaries hashes) are measured (hashed) and signed/authentified by the user after an OS core component upgrade (dom0 boot components upgrades) prior of being discovered and changed.

All is good until now. A user booting its default boot option will only be permitted if the same detached signed hashes are verified against he public key fused inside of the ROM, and if the sealed TPM nvram region can be unsealed with above PCRs+defined TPM disk enryption key passphrase used to seal the disk unlock key "passphrase" (256 characters key) which is unsealed, packed into an additional cpio to be unpacked by the OS after having unpacked its own initrd to construct its initramfs, which is the the pre-os environement containing OS drivers and pre-boot environement to boot into the end OS.

**The actual problem comes with the fact that that additional cpio that Heads constructs, which contains the secret.key released by the TPM after authentication, also contains a crypttab override that changes expected OS' crypttab behavior from initrd, and deployed in OS' paths.** We will over that.

This happens here on master: https://github.com/osresearch/heads/blob/c3b0bd6ffbe816430dd41ef54e649af52ed1ff3b/initrd/bin/kexec-insert-key#L78

What was proposed under #894: https://github.com/osresearch/heads/blob/3b3ce56bafa674c9a2af76e23c6be1a2d01c01aa/initrd/bin/kexec-insert-key#L98

And now, some examples of what is expected/deployed on some OSes.

debian-11-xfce:

Q4.1:
```
[user@dom0 ~]$ sudo cat /etc/crypttab 
luks-a4ba89e4-65a3-48f9-a045-d128c2a422cc UUID=a4ba89e4-65a3-48f9-a045-d128c2a422cc none discard
```

Conclusion:
Heads currently overwrites the discard/TRIM behavior **to not be applied**:

Controversial and probably desired implementation
----
Since the OS is responsible to deal with the upper layer discard/trim configuration on filesystem level (EXT4,etc) and pool level (LVM etc), it is thought that having Heads force a LUKS discard override as part of its deployed cpio to be applied on top of initramfs would be the best approach.

So #1280's proposed implementation (to override multiple crypttab deployed files to have debian based OSes `/cryptroot/crypttab` initramfs files overwritten, which neglects /etc/crypttab...):
https://github.com/osresearch/heads/blob/990150703bc6b8da4c5f6790838cf48b4972984b/initrd/bin/kexec-insert-key#L102
Look more like:
` echo "luks-$uuid UUID=$uuid /secret.key luks,discard" | tee "$INITRD_DIR/$crypttab_path" "$INITRD_DIR/cryptroot/crypttab" > /dev/null `
**Where `luks,discard` would force discard/trim operations on selected default boot defined LUKS containers**

Possible current impact of lack of discard operations on LUKS container
----
- Impact performance
- Impact anti-forensic
- No TRIM applied down to drives

Litterature
----
- https://patrick-nagel.net/blog/archives/337 (2011)
- https://lwn.net/Articles/787272/ (2019)
- Qubes OS 4.0 enables LUKS discard automatically: https://github.com/QubesOS/openqa-tests-qubesos/commit/2f49b7502e5b94d46db8499febff40c0b8dd5751
- Qubes 4.0 blkdiscard applied on shutdown only so that TRIM is not locking IO https://github.com/QubesOS/qubes-core-admin/commit/16d069289cb01e73272e82f9f2a1c7f8ebdf0403
- Qubes 4.0: enable discard on all supporting filesystem at installer: https://github.com/QubesOS/qubes-anaconda/commit/2d242cb577b835313b67c14aa0fd3380c00bf78d
- Qubes 4.0: apply discard operation on dom0 filesystems by default: https://github.com/QubesOS/qubes-anaconda/commit/6b814910db8db712d82fc0355c54cb878dc0faa0
- Qubes 4.0: add automatic fstrim to trim/discard thin pool volumes down to LUKS layer: https://github.com/QubesOS/qubes-core-admin-linux/commit/bae443dfceffc43c44fd9b3bf06132b8ac28303d

Latest: Q4.1 implements trimming operations automatically to all layers https://github.com/Qubes-Community/Contents/blob/master/docs/configuration/disk-trim.md

> There are some security implications to permitting TRIM (read for example [this article](https://asalor.blogspot.com/2011/08/trim-dm-crypt-problems.html)), but in most cases not exploitable. Conversely, TRIM can improve security against local forensics when using SSDs, because with TRIM enabled deleting data (usually) results in the actual data being erased quickly, rather than remaining in unallocated space indefinitely. However deletion is not guaranteed, and can fail to happen without warning for a variety of reasons.

tlaurion:
This PR takes original work from @root-hardenedvault under #894 and supplements it to automatically parse selected default's boot's initrd file (compressed/uncompressed) to find OS defined crypttab entries.

It keeps track of OS deployed crypttab files and respects OS deployed options, overriding "none" to "/secret.key" and saves 
the overrides under `/boot/kexec_initrd_crypttab_overrides.txt` in the form of `crypttab_file/crypptab config lines` under `kexec-save-default`. 

`kexec-save-default` now informs the user of what was found and overridden, while `kexec-insert-key` applies ithose overrides, while also informing the user of twhat is to be applied through injected additioanl cpio through the final kexec call to the OS when adding a TPM disk unlock key at default boot selection.

If no crypttab file was found in initrd (case for QubesOS), `kexec-insert-key` inform the user of the dynamically created crtypttab entry that is to be injected through cpio override to initramfs' `/etc/crypttab` file and uses previous format:
`luks-$uuid $uuid /secret.key luks,discard`

**BUGFIX: It is to note that prior of this commit, QubesOS (and all OSes not having a crypttab under initrd) _were instructed to not apply discard (TRIM) operations at LUKS layer_:**
`luks-$uuid $uuid /secret.key luks`

------

TODO create additional tasks:
- [ ] Change prompts to user, using blkid + cryptsetup isLuks to present valid "suggested" alternative values, while having default (no user input: Enter to reuse) reusing saved values under kexec_key_lvm.txt and/or kexec_key_devices.txt
- [ ] Save TCPA log/TCPA hashes (cbmem -L output) under /boot/kexec_tcpa_log.txt, so that selecting default boot can validate that current TCPA is consistent. Otherwise, have default boot prompt user to setup a new boot default again since firmware measurements changed/luks changed etc. Point is: if TOTP/HOTP unsealing works but unsealing disk unlocking doesn't, default boot needs to be renewed anyway to pass /secret.key and default boot won't work.

### DIFF
--- a/initrd/bin/kexec-insert-key
+++ b/initrd/bin/kexec-insert-key
@@ -30,12 +30,24 @@ fi
 cat "$TMP_KEY_DEVICES" | cut -d\  -f1 | xargs /bin/qubes-measure-luks \
 	|| die "LUKS measure failed"
 
-# Unpack the initrd and fixup the /etc/crypttab
+# Unpack the initrd and fixup the crypttab
 # this is a hack to split it into two parts since
 # we know that the first 0x3400 bytes are the microcode
 INITRD_DIR=/tmp/secret/initrd
 SECRET_CPIO=/tmp/secret/initrd.cpio
-mkdir -p "$INITRD_DIR/etc"
+
+# Not all distro put crypttab under /etc/ within initramfs, but finding it at
+# runtime needs unpacking, which may be hard to do, so it is made overridable
+# with a file at /boot/kexec_initrd_crypttab_path.txt, whose content could be
+# obtained with $ cpio -t < ${uncompressed_initrd} | grep crypttab .
+
+bootdir=$(dirname "$INITRD")
+if [ -r $bootdir/kexec_initrd_crypttab_path.txt ]; then
+	crypttab_path=$(cat $bootdir/kexec_initrd_crypttab_path.txt)
+else
+	crypttab_path=etc/crypttab
+fi
+mkdir -p "$INITRD_DIR/$(dirname $crypttab_path)"
 
 # Attempt to unseal the disk key from the TPM
 # should we give this some number of tries?
@@ -72,10 +84,18 @@ echo '+++ Building initrd'
 dd if="$INITRD" of="$SECRET_CPIO" bs=512 conv=sync \
 || die "Failed to copy initrd to /tmp"
 
+# The "target" field of the record within the crypttab stored in the root
+# file system for the luks container which is going to be unlocked via
+# kexec-insert-key should be modified into the same "luks-$uuid" format,
+# otherwise the boot sequence will get stuck when OS is trying to unlock them
+# again, in order to map them according to "target" fields written in the
+# crypttab stored in the root fs.
+
 if [ "$unseal_failed" = "n" ]; then
-	# overwrite /etc/crypttab to mirror the behavior for in seal-key
+	# overwrite crypttab to mirror the behavior for in seal-key
 	for uuid in `cat "$TMP_KEY_DEVICES" | cut -d\  -f2`; do
-		echo "luks-$uuid UUID=$uuid /secret.key" >> "$INITRD_DIR/etc/crypttab"
+		# In Debian, the "luks" option at last should not be omitted
+		echo "luks-$uuid UUID=$uuid /secret.key luks" >> "$INITRD_DIR/$crypttab_path"
 	done
 	( cd "$INITRD_DIR" ; find . -type f | cpio -H newc -o ) >> "$SECRET_CPIO"
 fi
--- a/initrd/bin/kexec-insert-key
+++ b/initrd/bin/kexec-insert-key
@@ -35,19 +35,8 @@ cat "$TMP_KEY_DEVICES" | cut -d\  -f1 | xargs /bin/qubes-measure-luks \
 # we know that the first 0x3400 bytes are the microcode
 INITRD_DIR=/tmp/secret/initrd
 SECRET_CPIO=/tmp/secret/initrd.cpio
-
-# Not all distro put crypttab under /etc/ within initramfs, but finding it at
-# runtime needs unpacking, which may be hard to do, so it is made overridable
-# with a file at /boot/kexec_initrd_crypttab_path.txt, whose content could be
-# obtained with $ cpio -t < ${uncompressed_initrd} | grep crypttab .
-
 bootdir=$(dirname "$INITRD")
-if [ -r $bootdir/kexec_initrd_crypttab_path.txt ]; then
-	crypttab_path=$(cat $bootdir/kexec_initrd_crypttab_path.txt)
-else
-	crypttab_path=etc/crypttab
-fi
-mkdir -p "$INITRD_DIR/$(dirname $crypttab_path)"
+mkdir -p "$INITRD_DIR/etc"
 
 # Attempt to unseal the disk key from the TPM
 # should we give this some number of tries?
@@ -84,18 +73,30 @@ echo '+++ Building initrd'
 dd if="$INITRD" of="$SECRET_CPIO" bs=512 conv=sync \
 || die "Failed to copy initrd to /tmp"
 
-# The "target" field of the record within the crypttab stored in the root
-# file system for the luks container which is going to be unlocked via
-# kexec-insert-key should be modified into the same "luks-$uuid" format,
-# otherwise the boot sequence will get stuck when OS is trying to unlock them
-# again, in order to map them according to "target" fields written in the
-# crypttab stored in the root fs.
-
 if [ "$unseal_failed" = "n" ]; then
-	# overwrite crypttab to mirror the behavior for in seal-key
-	for uuid in `cat "$TMP_KEY_DEVICES" | cut -d\  -f2`; do
-		# In Debian, the "luks" option at last should not be omitted
-		echo "luks-$uuid UUID=$uuid /secret.key luks" >> "$INITRD_DIR/$crypttab_path"
-	done
+	# kexec-save-default might have created crypttab overrides to be injected in initramfs through additional cpio 
+	if [ -r "$bootdir/kexec_initrd_crypttab_overrides.txt" ]; then
+		echo "$bootdir/kexec_initrd_crypttab_overrides.txt found..."
+		echo "Preparing initramfs crypttab overrides as defined under $bootdir/kexec_initrd_crypttab_overrides.txt to be injected through cpio at next kexec call..."
+		# kexec-save-default has found crypttab files under initrd and saved them
+		cat "$bootdir/kexec_initrd_crypttab_overrides.txt" | while read line; do
+			crypttab_file=$(echo "$line" | awk -F ':' {'print $1'})
+			crypttab_entry=$(echo "$line" | awk -F ':' {'print $NF'})
+			# Replace each initrd crypttab file with modified entry containing /secret.key path
+			mkdir -p "$INITRD_DIR/$(dirname $crypttab_file)"
+			echo "$crypttab_entry" | tee -a "$INITRD_DIR/$crypttab_file" > /dev/null
+			echo "initramfs's $crypttab_file will be overriden with $crypttab_entry"
+		done
+	else
+		# No crypttab files were found under selected default boot option's initrd file
+		crypttab_file="etc/crypttab"
+		mkdir -p "$INITRD_DIR/$(dirname $crypttab_file)"
+		# overwrite crypttab to mirror behavior of seal-key
+		echo "The following /etc/crypttab lines will be passed through cpio into kexec call for default boot option:"
+		for uuid in `cat "$TMP_KEY_DEVICES" | cut -d\  -f2`; do
+			# NOTE: discard operation (TRIM) is activated by default if no crypptab found in initrd
+			echo "luks-$uuid UUID=$uuid /secret.key luks,discard" | tee -a "$INITRD_DIR/$crypttab_file"
+		done
+	fi
 	( cd "$INITRD_DIR" ; find . -type f | cpio -H newc -o ) >> "$SECRET_CPIO"
 fi
--- a/initrd/bin/kexec-save-default
+++ b/initrd/bin/kexec-save-default
@@ -129,6 +129,52 @@ if [ ! -r $ENTRY_FILE -o ! -r $HASH_FILE ]; then
 	die "Failed to write default config"
 fi
 
+if [ "$save_key" = "y" ]; then
+	# logic to parse OS initrd to extract crypttab, its filepaths and its OS defined options
+	mkdir -p /tmp/initrd_extract
+	cd /tmp/initrd_extract
+	# Get initrd filename selected to be default initrd that OS could be using to configure LUKS on boot by deploying crypttab files
+	current_default_initrd=$(cat /boot/kexec_default_hashes.txt | grep initr | awk -F " " {'print $NF'} | sed 's/\.\//\/boot\//g')
+	
+	# Get crypttab files paths from initrd
+	echo "Checking current selected default boot's $current_default_initrd for existing crypttab files..."
+	# First either decompress or use the original if it's not compressed
+	initrd_decompressed="/tmp/initrd_extract/initrd_decompressed.cpio"
+	zcat < "$current_default_initrd" > "$initrd_decompressed" 2> /dev/null || initrd_decompressed="$current_default_initrd"
+	crypttab_files=$(cpio --list --quiet < "$initrd_decompressed" | grep crypttab 2> /dev/null) || true
+	
+	if [ ! -z "$crypttab_files" ]; then
+		echo "Extracting current selected default boot's $current_default_initrd for found crypttab files analysis..."
+		cpio -id --quiet < $initrd_decompressed $crypttab_files 2> /dev/null
+		rm -f $bootdir/kexec_initrd_crypttab_overrides.txt || true
+		
+		#Parsing each crypttab file found
+		echo "$crypttab_files" | while read filepath; do
+			# Keep only non-commented lines
+			current_filepath_entries=$(cat "$filepath" | grep -v "^#")
+			# Modify each retained crypttab line to contain to be injected /secret.key at next default boots
+			modified_filepath_entries=$(echo "$current_filepath_entries" | sed 's/none/\/secret.key/g')
+			echo "$modified_filepath_entries" | while read single_modified_filepath_entry; do
+				# Append each found filepath:entry into additional kexec_ file that will be part of detached signed digest
+				echo "$filepath:$single_modified_filepath_entry" >> $bootdir/kexec_initrd_crypttab_overrides.txt
+			done
+		done
+
+		cd - > /dev/null
+
+		#insert current default boot's initrd crypttab locations into tracking file to be overwritten into initramfs at kexec-inject-key
+		echo "The following OS crypttab file:entry were modified from default boot's initrd:"
+		cat $bootdir/kexec_initrd_crypttab_overrides.txt
+		echo "Heads added /secret.key in those entries and saved them under $bootdir/kexec_initrd_crypttab_overrides.txt"
+		echo "Those overrides will be part of detached signed digests and used to prepare cpio injected at kexec of selected default boot entry."
+	else
+		echo "No crypttab file found in extracted initrd. Removing $bootdir/kexec_initrd_crypttab_overrides.txt"
+		rm -f "$bootdir/kexec_initrd_crypttab_overrides.txt" || true
+	fi
+	# Cleanup
+	rm -rf /tmp/initrd_extract || true
+fi
+
 # sign and auto-roll config counter
 extparam=
 if [ "$CONFIG_TPM" = "y" ]; then


tlaurion:
Last commit is unclean https://github.com/osresearch/heads/compare/fa53b89044523d541ab49747efcc5b5ca0e2c91a..9f12ad0fcf1d50f35ab3438ec2ac0c900ca9a9dc

But it permits to dynamically extract currently selected boot's default initrd file, get actual OS crypttab path and crypptab content to only change `none` to `/secret.key`, also respecting the OS's discard implemenation (not changing it).

tlaurion:
In current state (with debug traces):

tlaurion:
... Forgot to deal with uncompressed initrd, which is the case with QubesOS (where xz/gz/bz is dealt with zcat):
```
sudo file /boot/initramfs-5.15.81-1.fc32.qubes.x86_64.img 
/boot/initramfs-5.15.81-1.fc32.qubes.x86_64.img: ASCII cpio archive (SVR4 with no CRC)
```

signal-2023-01-19-153413

tlaurion:
Ok, last commit adds dual support for compressed/uncompressed cpio.
Tested on qemu with debian-11-xfce:





Letting build and will test on x230 (QubesOS is the uncompressed cpio case without crypttab, currently failing as reported under https://github.com/osresearch/heads/pull/1284#issuecomment-1397572464

tlaurion:
> 

@JonathonHall-Purism any comment on verbiage presented? What would you add/remove for best UX?

initrd/bin/kexec-save-default
```diff
@@ -129,6 +129,41 @@ if [ ! -r $ENTRY_FILE -o ! -r $HASH_FILE ]; then
...
+if [ ! -z "$crypttab_files" ]; then
+	echo "Extracting current selected default boot's $current_default_initrd for crypttab analysis..."
+	zcat $current_default_initrd | cpio -id --quiet "$crypttab_files" > /dev/null || cpio -id --quiet < $current_default_initrd "$crypttab_files" > /dev/null
+
+	crypttab_entries=$(cat "$crypttab_files"|uniq| sed 's/none/\/secret.key/g')
+	cd - > /dev/null
+
+	#insert current default boot's initrd crypttab location into tracking file to be overwritten into initramfs
+	echo "The following OS crypttab was found in intrd and Heads will overwrite it: $crypttab_files"
+	echo "Adding found crypptab files location under $bootdir/kexec_initrd_crypttab_path.txt"
```
JonathonHall-Purism, line 149: I don't think this is important or useful; it's always `kexec_initrd_crypttab_path.txt` under the boot dir.  IMO, I think we could remove it so the relevant info is more prominent.

I do like the other two traces though, both of those are characteristics of the initrd we found that would be important for troubleshooting.

initrd/bin/kexec-save-default
```diff
@@ -129,6 +129,41 @@ if [ ! -r $ENTRY_FILE -o ! -r $HASH_FILE ]; then
...
+
+	crypttab_entries=$(cat "$crypttab_files"|uniq| sed 's/none/\/secret.key/g')
+	cd - > /dev/null
+
+	#insert current default boot's initrd crypttab location into tracking file to be overwritten into initramfs
+	echo "The following OS crypttab was found in intrd and Heads will overwrite it: $crypttab_files"
+	echo "Adding found crypptab files location under $bootdir/kexec_initrd_crypttab_path.txt"	
+	echo "$crypttab_files" > "$bootdir/kexec_initrd_crypttab_path.txt"
+else
+	echo "No crypttab file found in extracted initrd. Removing $bootdir/kexec_initrd_crypttab_path.txt"
```
JonathonHall-Purism, line 152: I like having this trace and the similar one below, but I'd suggest:
* I don't think we need to say what file we're deleting since it's always the same
* Maybe we should point out that LUKS key sealing will not work since we can't override crypttab

If you had a default boot selected in the past, and are updating it now, is it possible old copies of the files could be hanging around?  Maybe we should `rm -f "$bootdir/kexec_initrd_crypttab_*"` in case an old copy of `..._entries.txt` was afound?

initrd/bin/kexec-save-default
```diff
@@ -129,6 +129,41 @@ if [ ! -r $ENTRY_FILE -o ! -r $HASH_FILE ]; then
...
+fi
+rm -rf /tmp/initrd_extract	
+
+if [ ! -z "$crypttab_entries" ];then
+	#insert current default's boot's initrd crypttab override, only modifying secret location
+	echo "The original crypttab options were overwritten to include secret.key: $crypttab_entries"	
+	echo "Adding modified entries under $bootdir/kexec_initrd_crypttab_entries.txt..."	
+	echo "$crypttab_entries" > "$bootdir/kexec_initrd_crypttab_entries.txt"
+else
+	echo "No crypttab entries found in extracted initrd. Removing $bootdir/kexec_initrd_crypttab_entries.txt"
```
JonathonHall-Purism, line 163: Same verbage suggestions as above, and here it might be more important to `rm -f "$bootdir/kexec_initrd_crypttab_*"` since we are leaving around a `..._path.txt` currently.  Maybe we could refactor this to consolidate the two copies of the failure handling?

JonathonHall-Purism (commented):
Thanks @tlaurion , posted comments and suggestions inline.

initrd/bin/kexec-save-default
```diff
@@ -129,6 +129,41 @@ if [ ! -r $ENTRY_FILE -o ! -r $HASH_FILE ]; then
...
+# User just selected a default boot option. Get initrd filename
+current_default_initrd=$(cat /boot/kexec_default_hashes.txt | grep initr | awk -F " " {'print $NF'} | sed 's/\.\//\/boot\//g')
+# Get crypttab files paths from initrd
+crypttab_files=$(zcat $current_default_initrd | cpio --list --quiet | grep crypttab) > /dev/null || crypttab_files=$(cpio --list --quiet < $current_default_initrd | grep crypttab) > /dev/null
+
+if [ ! -z "$crypttab_files" ]; then
+	echo "Extracting current selected default boot's $current_default_initrd for crypttab analysis..."
+	zcat $current_default_initrd | cpio -id --quiet "$crypttab_files" > /dev/null || cpio -id --quiet < $current_default_initrd "$crypttab_files" > /dev/null
+
+	crypttab_entries=$(cat "$crypttab_files"|uniq| sed 's/none/\/secret.key/g')
```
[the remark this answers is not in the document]
tlaurion: The logic is incorrect here, since multiple crypttab files (paths) should be kept to have heads recreate them in propoer paths, not disturbing OS behavior. 

I think the problem here is `uniq|sed` (so uniq on filename, not on sed output) instead of `sed/uniq`. Will change for review. Otherwise I would have to install PureOS to test (currently testing on real hardware with Qubes (no crypptab in initrd) and debian-11-xfce on qemu.

----

I agree, OS maker should only have one crypptabl file in initrd, but this whole PR goal is to not care about OS choices, just respecting them and to replace `none` with `/secret.key` in OS's initrd expected path (overwriting them in memory) with OS's expected content (hoping OSes have propoer LUKS unlock logic, which they should if they support LUKS partitioning scheme at install).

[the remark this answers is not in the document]
JonathonHall-Purism: The problem is `cat "$crypttab_files"` - that tries to cat a single file whose name is the entire content of that variable, it's not word-split since it's quoted.

If you want to support it, do you want to loop here to apply the replacement to each file?  That would seem to make the fewest assumptions about equivalence, etc., but you'd have to store the new content for each location somewhere.

Maybe you could build the whole "crypttab override initrd" here rather than at boot time?  I.e. perform the replacement individually on each found file, put them all in a cpio archive, and save that single file rather than saving separate files listing the paths of the crypttabs and new contents.  Then boot would just cat the override initrd on, it wouldn't have to build it at all.

initrd/bin/kexec-save-default
tlaurion: @JonathonHall-Purism : Please consider this (as general guideline for user understanding of the minimal technicalities to underatand and therefore trust the process):

- We add two new /boot/kexec_* files.
- We do so otherwise in an obscure place (kexec-save-boot)
  - The actual reason we are doing it there is because this is the place where whatever is modifying /boot gets prompted to be signed without the checksum/dirtree validation to kick in detecting the changes.

Therefore, unless you keep your stance here, I would change the behavior everywhere else in heads to tell the user that some /boot/kexec_* files were added/modified instead of respecting actual Heads behavior which lacks output for user to understand what is happening.

But if you keep your stance here, I have no **strong** opinion. I only remember my own learning path of Heads, and those lines are actually really good entry point to inspect code change and where things are happening.
@JonathonHall-Purism tell me! Otherwise I would really love to have a quick **HEADS_DEBUG=y** possible in board configs and the new code changes implementing those for additional output when developing new features.

initrd/bin/kexec-save-default
tlaurion: @JonathonHall-Purism Same reasoning as before. If for whatever reason a user is keeping backup of /boot into root for backups, I'm sure the user would love to see why there is a new file added, or deleted?

Not sure I understand the reasoning here, I might have missed the point and therefore if that is the case i'm sorry.
Both files are deleted individually in the current PR?

initrd/bin/kexec-save-default
JonathonHall-Purism: Yeah, I understand what you're saying, and I do agree.  We can keep it :+1:   Definitely does help newcomers that want to understand how it works and hack on it, anybody that does look into their own /boot, etc.  Not really a strong opinion on my end either, let's keep it.

initrd/bin/kexec-save-default
JonathonHall-Purism: > @JonathonHall-Purism Same reasoning as before. If for whatever reason a user is keeping backup of /boot into root for backups, I'm sure the user would love to see why there is a new file added, or deleted?

Yeah agree, thanks for spelling out your thought process here.  Let's keep it :+1: 

> Both files are deleted individually in the current PR?

OK, I see in this case it should go on later to also see that `crypttab_entries` is empty and then delete `..._entries.txt` too.  But per above if you want to handle multiple crypttabs properly, maybe it would be simpler and more robust just to create the initrd override archive here, which would also dodge the issue of keeping two files in sync entirely?

tlaurion:
@JonathonHall-Purism In line of past comments, I think the "effective" (to be injected crypttab configuration's option) should be given on screen to the user when cpio is being prepared to tell the user what is the modified runtime behavior applied (kexec-insert-key).

Afterall, this is the reason why master doesn't unlock PureOS right now (luks-uuid doesn't match), and also why QubesOS didn't have discard applied from initramfs (luks option, not luks,discard), and why some other OSes might not have their LUKS container unlocked through TPM disk unlock key in the future (while less possible with this PR, since we will now copy OS behavior+override instead of crafting it). 

If no crypttab found in initrd, inject a generic `luks-uuid UUID=uuid /secret.key luks,discard`  line (QubesOS behavior for now with this PR, just changing luks->luks,discard with injected crafted crypttab), while @root-hardenedvault PR was not working with debian-11 out of the box (requiring double LUKS unlock or CTRL-C at second prompt)

-----
Please comment on sentence changes you would love to be modified from console output below (or whole output from kexec-insert-key to be removed: that is what is shown on default boot):







I gave screenshots here to be reused in other issues I will open (Show boot option for Qubes only shows kernel version. This is far from acceptable)

tlaurion:
As of last commit per qemu testing:
kexec-save-default:

kexec-insert-key:

tlaurion:
@JonathonHall-Purism 

I decided to go OS agnostic to the extend that OSes must provide valid crypttab files into initrd to not fall into Heads default crypttab construction being otherwise `luks-$uuid $uuid /secret.key luks,discard`.

As per previous discussions, I decided to give it another shot and create a single kexec_initrd_crypttab_override.txt containing `cryptab_file:crypttab override` on kexec-save-default (which also takes it as part of detach signed digest per previous code base) and using that file at kexec-insert-key, telling the user what is happening. In case an OS is not compliant, the user will be able to report/modify said kexec_initrd_crypttab_override.txt content, generate+sign digests and boot into modified override until OS fixes behavior themselves.

For PureOS, this means that both /et/crypttab and /cryptroot/crypttab files are reconstructed, where I took liberty to remove comments present in origin files, expecting only actual crypttab related content to be present prior of sed'ing content to populate kexec_initrd_crypttab_override.txt

I haven't tested with PureOS, but this works for both debian-11-xfce (/cryptroot/crypttab override) and QubesOS generated crypttab file (fixing missing and expected discard on luks,discard option)

# qemu+debian-11-xfce 

kexec-save-default:


kexec-insert-key:


debian picking it up successfully:


# Qubes 4.1

kexec-save-default:
signal-2023-01-26-132654

kexec-insert-key:
signal-2023-01-26-132750
 

-------
Possible areas of concerns:
- set +e +o pipefail is added pior of doing the calls to zcat||cpio under kexec-save-default
- set -e -o pipefail is put back after that section as declared initially under kexec-save-default

@JonathonHall-Purism : good enough?
Would be nice for that to be tested with PureOS when you have a second (dual crypttab files under initrd)

Todo:

- [ ] clear/squash commits
- [ ] Thank origin code which talked about doing this automatically (@root-hardenedvault) but somehow irrelevant in current PR since we do not hardcode paths anymore (which required users to know way too much implementation details to do the right thing at each case)
- [ ] Create other issues to simplify TPM disk unlock key creation (reuse LVM/disk encrypted container as defined under kexec_ files, make user only type enter to reuse what is there, make language consistent (TPM disk unlock key is what is kept inside of additioanl TPM nv region, what is asked is TPM disk unlock key passphrase. Verbiage is inconsistent

initrd/bin/kexec-save-default
```diff
@@ -129,6 +129,55 @@ if [ ! -r $ENTRY_FILE -o ! -r $HASH_FILE ]; then
...
+	cd /tmp/initrd_extract
+	# Get initrd filename selected to be default initrd that OS could be using to configure LUKS on boot by deploying crypttab files
+	current_default_initrd=$(cat /boot/kexec_default_hashes.txt | grep initr | awk -F " " {'print $NF'} | sed 's/\.\//\/boot\//g')
+	
+	# Get crypttab files paths from initrd
+	# We expect zcat to fail on uncompressed file and try alternative cpio-only command if failing, redirecting error output to devnull
+	# To do so we disable exiting if things fail here temporarely
+	echo "Checking current selected default boot's $current_default_initrd for existing crypttab files..."
+	set +e +o pipefail
+	crypttab_files=$(zcat $current_default_initrd 2> /dev/null | cpio --list --quiet | grep crypttab 2> /dev/null) || crypttab_files=$(cpio --list --quiet < $current_default_initrd | grep crypttab 2> /dev/null)
```
JonathonHall-Purism, line 144: Saw your note about `set +e +o pipefail`, agree it's a bit complex/fragile :thinking: 

You could break it up into an "optionally decompress" step and then do everything else once, like:

```
# First either decompress or use the original if it's not compressed
initrd_decompressed="/tmp/initrd_decompressed.cpio"
if ! zcat < "$current_default_initrd" > "$initrd_decompressed"; then
    initrd_decompressed="$current_default_initrd" # Use uncompressed image directly
fi

# Then use $initrd_decompressed in the remaining steps - checking for crypttab presence, extracting crypttabs
```

(Used zcat instead of `gzip -d` because the initrd doesn't have a .gz extension.)

JonathonHall-Purism (commented):
@tlaurion Tested this with PureOS, there only fix needed was to unquote $crypttab_files on some cpio invocations, commented inline.

It works with PureOS with that change, tested:
* Setting default boot without sealing a key, booting
* Sealing disk unlock key and booting with sealed key
* Aborting key unseal to boot with recovery passphrase
* Resetting default boot without sealing a key and booting

I noticed that resetting default boot again without sealing a key leaves the existing key devices / crypttab overrides, so booting still prompts to unseal the key and presumably should still work - this is intentional right?  Same in master, not caused by this PR.

Also suggested a couple of spelling nitpicks and an alternative to `set +e +o pipefail`.

initrd/bin/kexec-insert-key
```diff
@@ -73,9 +74,30 @@ dd if="$INITRD" of="$SECRET_CPIO" bs=512 conv=sync \
...
-		echo "luks-$uuid UUID=$uuid /secret.key" >> "$INITRD_DIR/etc/crypttab"
-	done
+	# kexec-save-default might have created crypttab overrides to be injected in initramfs through additional cpio 
+	if [ -r "$bootdir/kexec_initrd_crypttab_overrides.txt" ]; then
+		echo "$bootdir/kexec_initrd_crypttab_overrides.txt found..."
+		echo "Preparing initramfs crypttab overrides as defined under $bootdir/kexec_initrd_crypttab_overrides.txt to be injected through cpio at next kexec call..."
+		# kexec-save-default has found crypttab files under initrd and saved them
+		cat "$bootdir/kexec_initrd_crypttab_overrides.txt" | while read line; do
+			crypttab_file=$(echo "$line" | awk -F ':' {'print $1'})
+			crypttab_entree=$(echo "$line" | awk -F ':' {'print $NF'})
```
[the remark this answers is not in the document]
tlaurion: French native language stikes again

tlaurion:
At this stage, PR contains fixes for above review and didn't cause regression on qemu. Waiting for builds to test on x230+qubesos 4.1, then squashing commits to clean PR.

Will keep https://github.com/osresearch/heads/pull/1284/commits/27f8dff2349390f1f9f922accad253b436b0f63f (#1280) for historic purposes, making it clear that no manual overrides are needed anymore but basing it on that commit still, because it contained the original idea to make it dynamic for desired UX.

tlaurion:
- [ ] Qemu+debian-11-xfce works as prior of code review.
- [ ] x230-maximized+qubesos 4.1 (https://app.circleci.com/pipelines/github/tlaurion/heads/1383/workflows/2ce7ec53-a2e4-4301-9f5f-866d23d72f20/jobs/14977/artifacts)

tlaurion:
@JonathonHall-Purism Cleaned up and ready for final review.

JonathonHall-Purism:
Looks great @tlaurion , tested final with PureOS and all good.  Let's merge!

PR #1289, opened 2023-01-18
/etc/functions: fix detection of virtual flash drive in qemu.

### LINKED ISSUE #1274: Heads can't find default qemu virtual flash drive
Using the qemu configs like `qemu-fbwhiptail-tpm1-hotp`, the automatically created USB flash drive image isn't detected in Heads.

`mount-usb` was recently improved to wait for partitions to appear if the media is partitioned, while still supporting unpartitioned media.  The qemu scripts create an unpartitioned fat32 disk.

Apparently though fat32 filesystem start with a boot sector that has a valid (ish) partition table - so fdisk doesn't produce the expected error, even though the device really isn't partitioned.  We should look and see if the device actually contains one or more partitions instead.

danielp96:
Adds check to detect device formatted as fat32 without partition table.

With fat32 fdisk does not print message about invalid partition table and instead it'll print an empty table with header. In both cases total output has the same length of 5 lines: 3 about device info, 1 empty line and the 5th will be the table header or invalid partition message.

So the amount of lines fdisk outputs is used to detect if it's a formatted device without partition table.

This fixes #1274 

Sample image of the empty table fdisk outputs.
Screenshot from 2023-01-17 12-28-39

Device format cases tested:
- Fat32 without partition table.
- Ext4 without partition table.
- MBR with single fat32 partition.
- GPT with single ext4 partition.
- USB drive with flashed live ISO

### DIFF
--- a/initrd/etc/functions
+++ b/initrd/etc/functions
@@ -155,7 +155,16 @@ list_usb_storage()
 			# never usable directly, and this allows the "wait for
 			# disks" loop in mount-usb to correctly wait for the
 			# partitions.
-			if fdisk -l "$b" | grep -q "doesn't contain a valid partition table"; then
+			# This check: [ $(fdisk -l "$b" | wc -l) -eq 5 ]
+			# covers the case of a device without partition table but
+			# formatted as fat32, which contains a sortof partition table.
+			# this causes fdisk to not print the invalid partition table
+			# message and instead it'll print an empty table with header.
+			# In both cases the output is 5 lines: 3 about device info,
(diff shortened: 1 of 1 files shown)


JonathonHall-Purism:
Thanks @danielp96.  Could you list the specific cases you tested so we can ensure this doesn't regress again?  It looks good to me but I haven't tested myself yet.  Counting lines of fdisk output is a tad fragile as we discussed but busybox tends to be very stable I think, and doing a much more complex awk to look for headers is not worth it IMO.

danielp96:
Updated with list of cases tested.

PR #1296, opened 2023-01-26
/etc/distro/keys/tails.key: Replace with updated and minimized one

tlaurion:
Trace for replication:
```
user@heads-tests:/tmp/tails-signing$ wget https://tails.boum.org/tails-signing.key
--2023-01-26 15:55:52--  https://tails.boum.org/tails-signing.key
Resolving tails.boum.org (tails.boum.org)... 204.13.164.63
Connecting to tails.boum.org (tails.boum.org)|204.13.164.63|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 1379278 (1.3M) [application/octet-stream]
Saving to: ‘tails-signing.key’

tails-signing.key                     100%[=======================================================================>]   1.31M  1.11MB/s    in 1.2s    

2023-01-26 15:55:53 (1.11 MB/s) - ‘tails-signing.key’ saved [1379278/1379278]

user@heads-tests:/tmp/tails-signing$ gpg --home /tmp/tails-signing/ --import tails-signing.key 
gpg: WARNING: unsafe permissions on homedir '/tmp/tails-signing'
gpg: keybox '/tmp/tails-signing/pubring.kbx' created
gpg: key DBB802B258ACD84F: 2172 signatures not checked due to missing keys
gpg: /tmp/tails-signing/trustdb.gpg: trustdb created
gpg: key DBB802B258ACD84F: public key "Tails developers <tails@boum.org>" imported
gpg: Total number processed: 1
gpg:               imported: 1
gpg: no ultimately trusted keys found
user@heads-tests:/tmp/tails-signing$ gpg --home /tmp/tails-signing/ --edit-key tails@boum.org
gpg: WARNING: unsafe permissions on homedir '/tmp/tails-signing'
gpg (GnuPG) 2.2.27; Copyright (C) 2021 Free Software Foundation, Inc.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.


pub  rsa4096/DBB802B258ACD84F
     created: 2015-01-18  expires: 2024-01-04  usage: C   
     trust: unknown       validity: unknown
sub  rsa4096/3C83DCB52F699C56
     created: 2015-01-18  expired: 2018-01-11  usage: S   
sub  rsa4096/98FEC6BC752A3DB6
     created: 2015-01-18  expired: 2018-01-11  usage: S   
The following key was revoked on 2015-10-29 by RSA key DBB802B258ACD84F Tails developers <tails@boum.org>
[...]

PR #1297, opened 2023-01-26
usb-scan->media-scan: usb-init calling media-scan usb

tlaurion:
media-scan accepts direct input of existing blkid and mount that passed device to /media

### DIFF
--- a/initrd/bin/media-scan
+++ b/initrd/bin/media-scan
@@ -11,8 +11,19 @@ if grep -q /boot /proc/mounts ; then
 		|| die "Unable to unmount /boot"
 fi
 
-# Mount the USB boot device
-mount_usb || die "Unable to mount /media"
+available_partitions="$(blkid | while read line; do echo $line | awk -F ":" {'print $1'}; done )"
+
+if [ "$1" == "usb" ]; then
+	# Mount the USB boot device
+	mount_usb || die "Unable to mount /media"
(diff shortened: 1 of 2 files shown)


tlaurion:
Not convinced yet. 

Seems like when usb thumb drive is connected even if unused, Tails boot (with #1296) which won't detect root filesystem if no usb drive connected.

Will review and diagnose prior of putting ready for review.

tlaurion:
Hid last comment since it is due to different options being passed.

@JonathonHall-Purism : please review, since present code change works and permits first steps toward booting from already exposed partitions that could have recovery detached signed ISOs which Heads can boot directly.


-----
The following works:

Recovery shell+ `media-scan /dev/sda1`
signal-2023-01-27-145917

Selecting ISO:
signal-2023-01-27-145928
Results in detached signature validation being successful, while current Heads's Tails ditribution public key is expired (fixed by #1094)

Selection External HD:
signal-2023-01-27-150007

Boots Tails 5.9 (with
signal-2023-01-27-150045

tlaurion:
Also boots Qubes4.1 successfully, meaning that a user could leave a non-boot partition with enough space to keep detached signed ISOs there, as long as he is aware to pass that partition to `media-scan`. I think this is neat.

JonathonHall-Purism:
Looks good, we had a side discussion on possibly factoring out the "how to mount /media" from media-scan but this is a good step.  :+1:

PR #1302, opened 2023-01-31
Testing x220/x230 coreboot configs: deactivate CONFIG_DRIVERS_PS2_KEYBOARD so  that kernel initializes it.

### LINKED ISSUE #958: keyboard backspace bug for x230-hotp-maximized (x230t)
So I was excited to get heads installed and my nitrokey setup. But then realised I have some sort of bug on all three laptops. No matter the OS, if its linux it locks the screen or just doesn't do anything. Same issue on 1 x230 tablet and 2 x230 notebooks.

How can I go about fixing this? I'm not a seasoned developer so please have mercy on me.

Thank you

tlaurion:
Another attempt to fix #958 (this is annoying!)


tlaurion:
@bwachter you can test the artifacts of this PR (green mark, then select your board + variant) and test.
If already running on maximized board, just flash internally.

Issue related topic should continue under #958.

This should be just a a work/not work kind of rejection thing.

Marking as draft to not pollute, and will crosspost in issue to here.

tlaurion:
Not sure why yellow icon is not beside commit id, but builds are happening here https://app.circleci.com/pipelines/github/tlaurion/heads/1428/workflows/97b68324-6455-4900-97bf-bd55476a139b

tlaurion:
Last commit touches all x220 and x230 coreboot configs to set `CONFIG_DRIVERS_PS2_KEYBOARD=n` explicitly.

tlaurion:
This is not fixing anything.
Tested under https://github.com/osresearch/heads/issues/958#issuecomment-1412288038

The only other alternative would be to force i8042 and serio to do the right thing per kernel option under KERNEL_ADD in board configs but this is really weird. Pinging coreboot developers/creating an issue after 4.19 release would also be another possibility.

PR #1305, opened 2023-02-02
Set forcesig if "not forced" after reset

### LINKED ISSUE #1076: oem-factory-reset fails to generate PGP keys with some Yubikeys
**Issue**:
Upon flashing the most recent heads master (commit e492786, Dec 1 2021) on a T430 machine built with `make BOARD=t430` and all settings reset, PGP key generation always failed for at first-boot with a whiptail error displaying "GPG Key automatic keygen failed". The user is then left to generate PGP keys manually and provision heads on their own as detailed on the wiki [1]. Heads works fine after.

This issue describes debugging and identified workarounds and serves to discuss steps forward prior to submission of new Pull Requests.

**Investigation**:
Debugging the failure scenario showed that the prompts supplied by `oem-factory-reset` during the `gpg --card-edit` "generate" command were not what `gpg` was expecting, primarily due to Admin and User PINs.

PR #1063 was opened to fix the issue, assuming it was a regression that had not been caught. Before the Pull Request, Admin PIN followed immediately by User PIN was supplied by `oem-factory-reset` near the beginning of the prompts. The PR change kept User PIN at the beginning, but moved the Admin PIN to the end.

Further analysis showed success depended on which OpenPGP card was being used. In my case, a Yubikey 5 with firmware version 5.1.2 was used. No problems were seen with Nitrokeys before the PR was submitted.

The same discrepancy in gpg prompts were seen outside of the heads environment with the same Yubikey, further hinting at Yubikey being the issue.

In the end, it was found that the Yubikey appears to use an invalid value for an OpenPGP Data Object (PW1 Status), technically violating the OpenPGP spec. This mismatch causes `gpg` to execute different logic when authenticating PINs. 

Until `gpg` performed an explicit action to set PW1 Status to a certain value within the OpenPGP spec, PW1 Status retained an invalid value, that, when evaluated as a boolean is interpreted to mean PW1 (User PIN) is valid for several commands. This is the behavior `gpg` seeks when performing a `generate` command, and once it sees a Nitrokey does not have this value set, it preemptively sets it early in the procedure, which requires an Admin PIN.

**Workarounds, solutions, next steps**:

1. Leave code as-is, document this possibility and encourage use of manual provisioning
1. Leave code as-is, document a "hack" that will make the Yubikey respond similar to a Nitrokey
   - Run `admin`, `forcesig`, `[admin pin`], iif Signature PIN shows as "not forced" in `gpg --card-status` 
1. Make modifications to the code to detect which prompts are asked and respond appropriately

Considering I am the only one who has complained, I would think step 1 is most suitable and step 2 for those who like to tinker.

I don't see the benefit of option 3 if not many people have complained, especially if the identified alternatives in option 1 or 2 are suitable.

Thoughts on simply pursuing the first option? I note that oem-factory-reset is not even documented in the Wiki, so the work here is essentially just documenting oem-factory-reset and caveats (the Yubikey).

------

**Detailed debugging**:

Enabling `--debug ipc` when invoking `gpg` showed the reason for the discrepancy:

- On my Yubikey, after a factory reset, the OpenPGP Data Object "PW1 Status bytes" (tag 0xC4) value was "255". PW1 Status Bytes is a 2-bit field indicating if PW1 (User PIN) is valid for a single command (0b00) or several (0b01). This value is shown as "SCD GETATTR CHV-STATUS" by `gpg`.
- On other implementations, this value was "0" after factory reset.

GnuPG, upon seeing a value of "0" for PW1 Status bytes performs a scdaemon command to set the value to 0b01, so that the user is not continually prompted for the User PIN during the session. However, setting the value to 1 requires PW3 Authentication, or the Admin PIN.

This explains why Admin PIN is requested with the Nitrokey, but is not requested with a Yubikey.

Finally, once `gpg` sends the `SCD GENKEY` command to scdaemon, the Yubikey now requires the Admin PIN as PW3 is required to be verified for the OpenPGP `GENERATE ASYMMETRIC KEY PAIR` command. In the Nitrokey case, the Admin PIN has already been supplied.

It is also observed that peforming a `factory-reset` command within `gpg` or the `gpg-connect-agent`-based reset sending raw APDUs did not affect the value of PW1 Status. My belief is this value is potentially uninitialized before being explicitly set. Once explicitly set, it retains its value, though even after factory resets. It is unknown if the OpenPGP implementation even cares of this value.

The PW1 Status DO can be explicitly set to 0b00 with `gpg --card-edit` by running:

```
gpg/card> admin
gpg/card> forcesig
```

If the output of the `gpg --card-status` indicates:

```
Signature PIN ....: forced
```

Then, the card is in a good state and will now complete `oem-factory-reset` just like a Nitrokey. Signature PIN "forced" equates to NOT PW1_Status.

In the Yubikey, a PW1 Status of 255 (or even a valid value of 0b01) evaluates to `false`, which will show the Signature PIN as "not forced".

In comparison, a Nitrokey indicated "forced", meaning GnuPG will change it to "not forced" or cached, and `oem-factory-reset` will be happy as it is answering prompts under the assumption that the card does not cache PW1 (User PIN).

[1]: https://osresearch.net/Configuring-Keys/

echo-84:
Fixes  #1076 and closes [heads-wiki#102](https://github.com/osresearch/heads-wiki/issues/102)

Code is from https://github.com/osresearch/heads-wiki/issues/102#issuecomment-1255464870 but instead of making a function and call, it's incorporated into the `gpg_key_reset` function after the Nitrokey AES key reset.

**To-Do:**
- [x] Compile
- [x] Flash
- [x] Verify oem-factory-reset
  - [x] Yubikey
  - [x] Nitrokey (Don't have the hardware to verify)

### DIFF
--- a/initrd/bin/oem-factory-reset
+++ b/initrd/bin/oem-factory-reset
@@ -87,6 +87,19 @@ gpg_key_reset()
     if lsusb | grep -q "20a0:4109" &&  [ -x /bin/hotp_verification ] ; then
         /bin/hotp_verification regenerate ${ADMIN_PIN_DEF}
     fi
+    # Toggle forced sig (good security practice, forcing PIN request for each signature request)
+    if gpg --card-status | grep "Signature PIN" | grep -q "not forced"; then
+        {
+            echo admin
+            echo forcesig
+            echo ${ADMIN_PIN_DEF}
+        } | gpg --command-fd=0 --status-fd=1 --pinentry-mode=loopback --card-edit \
+            > /tmp/gpg_card_edit_output 2>/dev/null
+        if [ $? -ne 0 ]; then
+            ERROR=`cat /tmp/gpg_card_edit_output`
+            whiptail_error_die "GPG Key forcesig toggle on failed!\n\n$ERROR"
+        fi
+    fi
     # Set RSA key length
     {
         echo admin


tlaurion:
Will test as part of next test runs on top of #1312

tlaurion:
No regression on my side. LGTM.
@JonathonHall-Purism ?

JonathonHall-Purism:
Agree, this change makes sense.  OEM reset is intended to restore defaults and sign, and "forced" is the default as far as I can tell, at least for the keys I have.

Let's merge it :+1:

PR #1310, opened 2023-02-10
x220 t420: revert non-maximized (legacy) coreboot configurations

tlaurion:
A mishap happened for legacy x220/t420 coreboot configurations.

Legacy here means that x220/t420 boards (non maximized) boards:
- IFD is locked (Intact from Lenovo)
- ME region is locked (Intact from Lenovo)

Consequently, CBFS_SIZE from coreboot config needs to match IFD BIOS region otherwise brick can happen.

I would need one board owner to send me an original OEM bios backup:
- [ ] x220 (xx20): @techge @Thrilleratplay @BlackMaria @srgrint
- [ ] [t420](https://doc.coreboot.org/mainboard/lenovo/t420.html) (xx20): @alexmaloteaux @natterangell (iGPU) @akfhasodh @doob85


tlaurion:
Ok... So even blobs/x220 and blobs/t420 are providing flashmap and scripts to modify local backup's IFD to pbe ready for maximized builds (IFD relocalization of ME space to BIOS region, and IFD unlocking.....)

- https://github.com/osresearch/heads/tree/master/blobs/t420
- https://github.com/osresearch/heads/tree/master/blobs/x220

https://www.calculator.net/hex-calculator.html?number1=007fffff&c2op=-&number2=00018000&calctype=op&x=62&y=17

I'm getting really confused with all of this....
Why are t420 and x220 boards specifying 0x750000
https://github.com/osresearch/heads/pull/1310/commits/0caaafd159cee3a56e8f9443087b90c1cc2cf32c

tlaurion:
Those boards (x220/t420 "non-maximized") boards are then calling `flashrom --ifd --image bios`

https://github.com/osresearch/heads/blob/master/boards/x220/x220.config#L37

Are x220/t420 boards even used by anyone anymore as opposed to x220-maximized/t420-maximized?!

tlaurion:
Really really confused.

So here we go.
If users are expected to run blobs/x220 blobs/t420 extract scripts manually from those directory (as opposed to blobs/xx20 download and extract script) the only difference from t420 and x220 boards is that one might prefer to extract ifd and me from local backup as opposed to using a generic ifd and downloaded ME image from Lenovo.

If that is the case, then basically, x220/t420 and x220-maximized/t420-maximized are equivalent and master is right saying that that CBFS_SIZE under coreboot configurations should be changed 0x750000 -> 7E7FFF to benefit of neutered space of ME inside of ME region.
----

So this is my understanding in short:

- x220/t420 boards are made for users desiring to extract blobs from local taken backups instead of downloading and exracting them from Lenovo as per blobs/xx20 scripts.

tlaurion:
> EDIT:
> - blobs/x220 and bobs/t420 readme and extract scripts are applying layout.txt, unlocks IFD and neuter ME.
> - x220/t420 boards config (flashrom call to limit BIOS region only access through `--ifd --image bios` considers that users might want to relock ifd and ME regions, so those cannot be touched)

I will leave this open, but I think I answered my own question. and master is correct changing bios region to match IFD.

tlaurion:
Those boards should not be renamed maximized boards (documentation says maximized boards flash whole SPI space) and are not legacy boards either (which are documented as not changing IFD/ME).

They COULD be called custom, or user....

But as far as i'm concerned, they are considered equivalents(validated) of their maximized boards in terms of CBFS_SIZE. And if bricks happen, its because neither blobs/xx20 or blobs/t420 blobs/x220 READMEs or their scripts have not be ran correctly. 

So until t420-maximized/x220-maximized boards do not pass CircleCI compilation, those not built boards should be considered as compiling correctly if we add features inside of them (for example as of now, the desire to ditch busybox's ash and to pack bash inside of the rom).

techge:
@tlaurion how would you like to get the images? I have three different versions here, I just could upload it and send it to you. Do you have a preferred way of messaging you?

techge:
Oh, well, what a coincidence. Regarding your message of the last two minutes, I guess you don't need them anymore.

PR #1318, opened 2023-02-21
Makefile: Fix cross compilation variables for pkg-config

JonathonHall-Purism:
pkg-config will still pick up system default directories from PKG_CONFIG_LIBDIR even if PKG_CONFIG_PATH is set.  Per the docs, cross compilation requires clearing PKG_CONFIG_PATH and setting PKG_CONFIG_LIBDIR (which is always searched after PKG_CONFIG_PATH).

Fixes issues observed in tpm2_retry branch picking up packages from host environment.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -125,6 +125,9 @@ heads_cc	:= $(CROSS)gcc \
 	-isystem $(INSTALL)/include \
 	-L$(INSTALL)/lib \
 
+# Cross-compiling with pkg-config requires clearing PKG_CONFIG_PATH and setting
+# both PKG_CONFIG_LIBDIR and PKG_CONFIG_SYSROOT_DIR.
+# https://autotools.info/pkgconfig/cross-compiling.html
 CROSS_TOOLS_NOCC := \
 	AR="$(CROSS)ar" \
 	LD="$(CROSS)ld" \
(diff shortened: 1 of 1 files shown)


tlaurion:
@JonathonHall-Purism This is amazing find! Some undoing of fixating things as a consequence of this will need to be done under modules. Letting this build for all boards and will test rom just to be safe but should be better.

pkg-config was a source of bleeding for local builds for sure.

Nice catch!

JonathonHall-Purism:
t520-maximized is failing to link bootblock:

```
/root/project/build/x86/coreboot-4.13/util/crossgcc/xgcc/bin/i386-elf-ld.bfd: cannot open output file t520-maximized/cbfs/fallback/bootblock.debug: No such file or directory
make[1]: *** [src/arch/x86/Makefile.inc:110: t520-maximized/cbfs/fallback/bootblock.debug] Error 1
```

@tlaurion Not sure why this would be related to pkg-config, have you seen this before?  Maybe something to do with caches since I touched the Makefile in a way affecting many modules but not any modules themselves?

JonathonHall-Purism:
talos-2_server built successfully but failed to persist workspace due to timeout

tlaurion:
@JonathonHall-Purism i had infrastructure errors around the same time. Please "retry from failed" on CircleCI

tlaurion:
(For info, a successful build from clean checkout can take 38 minutes when CircleCI is not overwhelmed.)

tlaurion:
And as far of t520-maximized, it seems like a pretty random race condition. 4.19 is ready to merge in my opinion. We saw what happens with host toolstack getting newer and reproducible builds in the past for now removed 4.11 builds. I would not be surprised if the same thing happened for 4.13 at some point... Until we pin packages/move away to nixos

JonathonHall-Purism:
Build completed after retry :+1:

tlaurion:
No regression under x230-maximized

Issue #1323, opened 2023-02-28
Test T440p with Coreboot NO_GFX_INIT


rbreslow:
From Coreboot's [`src/device/Kconfig`](https://github.com/coreboot/coreboot/blob/9770df1e9dfff99f8e764a56e6f603d4266edfb7/src/device/Kconfig#L117-L123):

> Select this to not perform any graphics initialization in coreboot. This is useful if the payload (e.g. SeaBIOS) can initialize graphics or if pre-boot graphics are not required.

@tlaurion says that the Linux kernel should be able to perform this initialization: https://github.com/osresearch/heads/pull/1282#issuecomment-1445399916. It's possible that allowing Coreboot to initialize graphics as well is redundant/increases boot times.

The pull request should provide before and after `cbmem` timings.

~Waiting on #1282~

srgrint:
I can confirm I have tried with NO_GFX_INIT set with no adverse effects.  For some reason, it still dosen't redraw the screen after the kexec until the deian kernel starts - not quite sure why.

Here it the cbmem log (have built using coreboot 4.19, although 4.17 has the same behaviour)

[coreboot_t440p_no_gfx-init.log](https://github.com/osresearch/heads/files/10854374/coreboot_t440p_no_gfx-init.log)

srgrint:
Also as a side issue regarding fine tuning the t440p kernel config, it is possible to safely remove  CONFIG_DRM_AST=y

(I think this was a leftover for when the kernel config was cloned from the common librem kernel - in order to make it work with their server boards - clearly not needed on the t440p)

tlaurion:
> I can confirm I have tried with NO_GFX_INIT set with no adverse effects. For some reason, it still dosen't redraw the screen after the kexec until the deian kernel starts - not quite sure why.

Thé kexec'ed OS will take over the screen only when and if i915 driver initializes the display

This is where igfx init from coreboot ( or no ifgx_init) *should* create a difference.

tlaurion:
For example, it is known that installing Archlinux will require end users to add i915 into ibts initrd.

Never tested archlinux with igfx on on coreboot. Tha t is actually a good test case for this and will try.

srgrint:
> The kexec'ed OS will take over the screen only when and if i915/another fb driver initializes the display. That can be before/after plymouth.
> 
> This is where igfx init from coreboot (or no ifgx_init) _should_ create a difference.

[...]

PR #1326, opened 2023-03-02
Add CONFIG_NO_GFX_INIT=y to T440p board

srgrint:
This PR adds CONFIG_NO_GFX_INIT=y to the T440p coreboot configuration, leaving graphics init exclusively to the heads kernel.

There are 3 benefits of this change:
- Reduces the size of the ramstage 
- Decreases boot time
- Likely improves security by reducing size of the trusted code base and hence attack surface

Have tested on my intel only T440p and seems to work fine.

Further discussed in #1323


tlaurion:
Merged upstream https://github.com/osresearch/heads/blob/ab1faf5389dbfbc174e3cbf6cdec0be1cba14772/config/coreboot-t440p.config#L16

PR #1321, opened 2023-02-27
patches/util-linux libusb: patch configure scripts for reproducibility

tlaurion:
All hardcode_into_libs=yes -> hardcode_into_libs=no

Also
- modules/libusb: remove --disable-tests (not recognized at configure step)

Adresses two modules that were not reproducible under https://github.com/osresearch/heads-wiki/issues/70

PR #1329, opened 2023-03-07
Fix util-linux reproducibility issue (libblkid.so.1)

tlaurion:
patches/util-linux: patch configure script so that all hardcode_into_libs=yes -> hardcode_into_libs=no

Originally tested and documented under https://github.com/osresearch/heads/pull/1321#issuecomment-1452353251

Replaces #1321


JonathonHall-Purism:
Looks great, per review on #1321 :+1:  Let's merge :ship:

PR #1333, opened 2023-03-10
modules/flashtools : version bump back to osresearch/flashtools

tlaurion:
Changes to support ppc64 were merged upstream


tlaurion:
@JonathonHall-Purism please approve! (we were pointing to same commit from 3mdeb/flashtools fork for which PR is now merged upstream).

JonathonHall-Purism:
Looks good, see we are pointing to the merge commit right over the previously-referenced commit, let's merge :+1:

PR #1292, opened 2023-01-23
TPM2/TPM1 support (testing and bug fixes needed through qemu-(fb)whiptail-tpm[1,2](-hotp) testing boards!

tlaurion:
Integrates and adapts @hardenedvault's vaultboot's TPM2 codebase (https://github.com/hardenedvault/vaultboot/commit/b2800fb008d7434d68376af7f473aa40ee51837f 

-------------

TODOs:
- ~libcurl cannot be built as a tpm2-tools dependency as of now not sure why. curl currently needs to be added in board config to be built~ not needed.
- ~tpm-reset (master and here) needs some review, no handle of no tpm use case. Caller is responsible to not call it otherwise does nothing~ caller scripts does not call (config-gui, gui-init, oem-factory-reset all clean)
- ~init tries to bind fd and fails currently~ 
  - ~Dynamic, depends of bash (required for all boards but can be deactivated (legacy-flash boards)~
- ~Note: Check if whiptail is different of fbwhiptail in clearing screen. As of now every clear seems to be removed, still whiptail clears previous console output~: fbwhiptail doesn't clear screen and is recommended to have more output on console
- When no OS' /boot can be mounted, do not try to TPM reset (will fail)
- ~TPM reset asks for TPM owner password multiple times~
- ~seal-hotpkey is not working properly~ ok
- ~setting disk unlock key asks for TPM ownership passphrase (sealing in NV requires ownership, but text is misleading user as if reowning TPM)~ fixed.
  - ~We should cache input, feed tpm behind the scene and wipe passphrase and state clearly that this is TPM disk unlock kye passphrase.~ tpm reset still required after each commit
- ~primary key from TPM2 is invalid most of the time from kexec-select-boot and verifying global hashes but is setuped correctly at disk unlock key setup~ fixed under kexec-save-default
- ~would be nice to take advantage of bash function tracing to understand where we are for debugging purposes, code takes ash in consideration only~ : would still be nice, but we have TRACE and DEBUG calls now
- ~tpmr says it implements nv calls but actually doesn't. Removing those falsely wrapped functions would help.~
- ~look into tpm2-tss to see precisely what algorithms are required on libcrypto~
- ~REVIEW TODOS IN CODE~
- ~READD CIRCLECI CONFIG~

---- 

Current state:
- TPM seal/unseal TOTP properly
- TPM disk encryption key sealing/unsealing properly
- coreboot support of TPM v2.0 (shared config for TPM2 support across all 4 previous variations) 
[...]

PR #1334, opened 2023-03-10
config/coreboot-*: add ccache to all coreboot config files

tlaurion:
First build will create slightly bigger coreboot build cache. Next builds will be around 50-80% faster builds

This is really useful in both development settings (qemu-coreboot-tpm*) and on CircleCI builds


tlaurion:
@JonathonHall-Purism : included in my last commit for qemu-tpm boards. Will need rebase under #1292 if this is merged first or this one against master.

tlaurion:
changed CircleCI project setting's CACHE_VERSION to 2023-03-10 to build clean.
Goal here is to build clean and then rebuild to see and document gain

tlaurion:
And rebased on master so this cache becomes useful for other builds.

tlaurion:
@JonathonHall-Purism gains in CircieCI are questionable (1h6 vs 56 minutes but yet we added the t440p which is not cached since then.)

I would vouch for merge more then against it. Makes sense to reuse cache (ccache) for coreboot by default for local builds and more globally on CI (where most time we have on rebuild today if because of workspace caches being passed and decompressed and where cache building on clean builds are taking the longest.

For a stupid example, see 10+ minutes into downloading caches.... to not save it here on each builds already having a cache for existing hashes: https://app.circleci.com/pipelines/github/tlaurion/heads/1541/workflows/ee5a6167-fc71-450e-b8b8-ee2a39f63079/jobs/18822

tlaurion:
@saper I thought ccache was used if present/not used if absent kinda thing.
Problem would be if it ccache was present and then it isn't, in which case a user would need to wipe extracted coreboot dir and build again.

Will retry locally and edit this comment

PR #1341, opened 2023-03-14
kexec-save-default: fix case when no crypttab found in initrd

tlaurion:
- /tmp/initrd_extract was attempted to be deleted while under that directory when no crypttab found.
- changing of directory to / is non-conditional prior of deletion: move to cleaning step
- Clarity on message displayed to user when a generic crypttab will be generated in case of no OS override

Otherwise currently showing at setting TPM disk unlock key when setting default boot option on master:

### DIFF
--- a/initrd/bin/kexec-save-default
+++ b/initrd/bin/kexec-save-default
@@ -169,18 +169,21 @@ if [ "$save_key" = "y" ]; then
 			done
 		done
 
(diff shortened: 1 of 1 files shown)


JonathonHall-Purism:
We troubleshooted this together and discussed the message, I like the new wording.  Looks great, let's merge :ship:

PR #1342, opened 2023-03-14
qemu-coreboot-whiptail-tpm2-hotp : add missing HOTP board requirements

tlaurion:
Otherwise, functional equivalent of qemu-coreboot-whiptail-tpm2


JonathonHall-Purism:
Looks good, I had been testing with fbwhiptail so I didn't notice this.  Thanks, let's merge

PR #1345, opened 2023-03-19
Add p8z77-m_pro board

ThePlexus:
**The P8Z77-M Pro** 
This board is a better choice over the [P8H61](https://github.com/osresearch/heads/issues/550) for a cost effective Heads + QubesOS desktop with ME neuter+disable compatibility. The P8H61 ecosystem was complex with multiple variants (some not even having a TPM header, and others having RamInit issues with some memory sticks), while less feature rich than the P8Z77 family. The P8H61s that were compatible still required some ME sections (FCRS,EFFS) to be whitelisted in order to post, which introduced unknowns. The P8H61s also needed a larger flash chip to work with heads than the manufacturer supplied 4M, which added complexity for the average user. 

The P8Z77-M Pro is able to offer more SATA connectors (2x 6Gb, 4x 3Gb, 2x eSATA) as well as more full size expansion ports. The board has a PS/2 keyboard port as recommended for QubesOS. The board comes with 8M flash chip as standard.

The i7-3770 is the best CPU available for the board, with VT-x & VT-d both present

This has been a daily driver Desktop QubesOS machine for me for some time.

**ME & ROM**
The board supports Intel LGA1155, which allow for ME removal (both neuter+disable work), ME region resize/shrinking (aka 'maximized' board), as well as [VSCC table modification.](https://github.com/corna/me_cleaner/issues/80). 

The blob download script uses the manufacturer supplied ME and IFD and performs the necessary hashing. The download script also removes the VSCC table by overwriting a NULL at the VSCC length table and FF bytes at the VSCC identifier table - using a printf with dd. The download script also resizes the rom layout and minimizes ME while maximizing space. 

The P8Z77-M Pro comes as standard with an 8Mb Flash chip, which means that no modification is needed to replace the chip is order to use heads as we shrink ME and 'maximize' this board by default, leaving just 335396 bytes available. 

The P8Z77-M Pro has both TPM1 and TPM2 modules available, though at time of writing only the TPM1 module would be usable with heads until the TPM2 work is completed. All testing was done with a TPM1 module

**Test platform**
BOARD: Asus P8Z77-M Pro 
RAM: 32Gb - 4x TimeTec DDRL3 75TT16NUL2R8-8G
CPU: Intel i7 3770
TPM: Modules tested: Asus branded TPM 1.02H & Foxconn TPM Krypton Rev 1.0


boards/p8z77-m_pro/p8z77-m_pro.config
```diff
@@ -0,0 +1,36 @@
+# Configuration for Asus P8Z77-M Pro
```
tlaurion, line 1: Please add here documentation from PR

boards/p8z77-m_pro/p8z77-m_pro.config
```diff
@@ -0,0 +1,36 @@
...
+CONFIG_FBWHIPTAIL=y
+CONFIG_HOTPKEY=y
+
+CONFIG_LINUX_USB=y
+
+export CONFIG_TPM=y
+export CONFIG_BOOTSCRIPT=/bin/gui-init
+export CONFIG_BOOT_REQ_HASH=n
+export CONFIG_BOOT_REQ_ROLLBACK=n
+export CONFIG_BOOT_KERNEL_ADD="intel_iommu=on intel_iommu=igfx_off nohz=off"
```
tlaurion, line 31: IGPU on that board? i915 Intel supported graphic card?

boards/p8z77-m_pro/p8z77-m_pro.config
```diff
@@ -0,0 +1,36 @@
...
+CONFIG_UTIL_LINUX=y
+CONFIG_LVM2=y
+CONFIG_MBEDTLS=y
+CONFIG_PCIUTILS=y
+CONFIG_POPT=y
+CONFIG_QRENCODE=y
+CONFIG_TPMTOTP=y
+CONFIG_CAIRO=y
+CONFIG_FBWHIPTAIL=y
+CONFIG_HOTPKEY=y
```
tlaurion, line 23: Board should be renamed with -hotp to make clear a Nitrokey pro/storage or librem key is required for remote attestation.

config/linux-p8z77-m_pro.config
```diff
@@ -0,0 +1,334 @@
...
+CONFIG_I2C_MUX_PCA9541=m
+CONFIG_I2C_MUX_REG=m
+# CONFIG_I2C_HELPER_AUTO is not set
+CONFIG_I2C_SLAVE=y
+CONFIG_PTP_1588_CLOCK=y
+# CONFIG_HWMON is not set
+# CONFIG_X86_PKG_TEMP_THERMAL is not set
+CONFIG_MFD_SYSCON=y
+CONFIG_DRM=y
+CONFIG_DRM_I915=y
```
tlaurion, line 195: IGPU on that board? i915 Intel supported graphic card?

config/coreboot-p8z77-m_pro.config
```diff
@@ -0,0 +1,20 @@
...
+CONFIG_GBE_BIN_PATH="@BLOB_DIR@/p8z77-m_pro/gbe.bin"
+CONFIG_NO_GFX_INIT=y
+CONFIG_TPM_MEASURED_BOOT=y
+CONFIG_TPM1=y
+CONFIG_DRIVERS_PS2_KEYBOARD=y
+CONFIG_DEFAULT_CONSOLE_LOGLEVEL_8=y
+CONFIG_PAYLOAD_LINUX=y
+CONFIG_PAYLOAD_FILE="@BOARD_BUILD_DIR@/bzImage"
+CONFIG_LINUX_INITRD="@BOARD_BUILD_DIR@/initrd.cpio.xz"
+CONFIG_LINUX_COMMAND_LINE="intel_iommu=on intel_iommu=igfx_off nohz=off"
```
tlaurion, line 20: IGPU on that board? i915 Intel supported graphic card?

blobs/p8z77-m_pro/download_BIOS_clean.sh
```diff
@@ -0,0 +1,110 @@
...
+dd bs=1024 skip=2 if=P8Z77-V-PRO-ASUS-2104.CAP of=P8Z77-V-PRO-ASUS-2104.ROM || { echo "Failed to de-cap the ROM..." && exit 1; }
+
+echo "### Stock variant AltME & ME Cleaner"
+$IFDTOOL -M 1 $ROMFILENAME
+echo "### extract stock ME"
+$IFDTOOL -x $ROMFILENAME.new
+
+echo "### Applying me_cleaner to neuter and truncate. EFFS,FCRS whitelisted"
+#$MECLEAN -w FCRS,EFFS -t -O "$BLOBDIR/me.bin" flashregion_2_intel_me.bin
+$MECLEAN -t -O "$BLOBDIR/me.bin" flashregion_2_intel_me.bin
```
tlaurion, line 87: Without relocalisation  of modules some space that could be freed and reused in ifd from me to bios region is lost. Me modules cannot be relocalised?

tlaurion:
Circle CI would also need two new entrees, one to call blob download and extraction script, and then another one to build the board on each commit to track regressions.

Can you please look and make sure that ME modules untrimmed cannot be relocalised to give some more free space to reduce me ifd region and then recalculate ifd bios region accordingly?

boards/p8z77-m_pro/p8z77-m_pro.config
ThePlexus: Standard Ivy HD 2500/4000

boards/p8z77-m_pro/p8z77-m_pro.config
ThePlexus: good point I guess i can make two versions

config/linux-p8z77-m_pro.config
ThePlexus: Intel HD, yes

config/coreboot-p8z77-m_pro.config
ThePlexus: Intel HD, yes

tlaurion:
@ThePlexus also, board name would require tpm1 extension since TPM daughter board is replaceable and tpm2 support just landed.

tlaurion:
Glad to see another board being integrated to Heads :)

blobs/p8z77-m_pro/download_BIOS_clean.sh
ThePlexus: good catch. Ive not really had much need for space crunching on this particular test rig, but yes 

```
8192 Mar 19 15:56 gbe.bin
4096 Mar 19 15:56 ifd.bin
98304 Mar 19 15:56 me.bin
```

tlaurion: 98304 is a value common to my eyes and expected after relocalization (shrink)

ThePlexus:
> Can you please look and make sure that ME modules untrimmed cannot be relocalised to give some more free space to reduce me size, the me ifd region and then recalculate ifd bios region accordingly with increased available BIOS region also mirrored into coreboot CBFS_SIZE?
> 
> I'm worried this board might be short lived if that is not figure out today. 

Sure - this is done. Not quite a meg, but not far off. Im just waiting to see if this has any effect on the system abiltiy to stay up, but it posts and gets into heads just fine

TOTP
`(empty)                        0x6e5400   null           933348 none
`
HOTP
`(empty)                        0x6e7800   null           924132 none
`

>As you might have saw, we now pack bash inside of Heads on recent master.

Yes ive been following the TPM2 work closely. Just have not had time to get involved. Ultimately this board is a DIP8 so if we ever get to the stage where 8Mb isnt enough, I already have a procedure that works for migrating to 16M W25q128

blobs/p8z77-m_pro/download_BIOS_clean.sh
tlaurion: > good catch. Ive not really had much need for space crunching on this particular test rig, but yes

That is important. 8mb is really, really limiting and will fail requiring either dropping of platforms at some point or massive collaboration into reducing unneeded features of modules packed into initrd for all boards. That includes massive tweaking of heads kernel and past work to tweak module compilation for space already.

For example, last time I checked, t520 was still using cryptsetup1 and nobody complained. Meaning nobody uses it really and could be dropped.

Goal here being also to lesser maintainer ship burden, which you just accepted to be part of to maintain the life of heads for that platform :)

config/coreboot-p8z77-m_pro.config
```diff
@@ -0,0 +1,19 @@
...
+CONFIG_BOARD_ASUS_P8Z77_M_PRO=y
+CONFIG_HAVE_IFD_BIN=y
+CONFIG_HAVE_ME_BIN=y
+CONFIG_HAVE_GBE_BIN=y
+CONFIG_IFD_BIN_PATH="@BLOB_DIR@/p8z77-m_pro/ifd.bin"
+CONFIG_ME_BIN_PATH="@BLOB_DIR@/p8z77-m_pro/me.bin"
+CONFIG_GBE_BIN_PATH="@BLOB_DIR@/p8z77-m_pro/gbe.bin"
+CONFIG_NO_GFX_INIT=y
+CONFIG_TPM_MEASURED_BOOT=y
+CONFIG_TPM1=y
```
tlaurion, line 13: Sorry for not having been precise enough.

This coreboot config should also be tpm1  suffixed and board coreboot config file named accordingly so that tpm2 board config will just duplicate and name things correctly.

AFK, will be able to review more correctly in week time to compare Linux config gere with others. This should be considered as a first and quick review.

ThePlexus: no problem have fixed. 

the kernel config is currently the same as x230 however i think its best to have a separate kernel config for this board as [running swraid](https://github.com/osresearch/heads/pull/1157) is something I've been doing on desktops - which x230s cant (though my w530 and t430 do with the caddy) . 

The raid kernel configs have been unset though for the purposes of this PR, but its going to cause bloat to x230 and others later to have these options enabled. 

`diff linux-p8z77-m_pro.config linux-x230-maximized.config
102,103d101
< # CONFIG_BLK_DEV_MD is not set
< # CONFIG_MD_RAID1 is not set
`
It would be good to figure out a way for making swraid usable that fits in with the your current board and config logical strategy

tlaurion: > The raid kernel configs have been unset though for the purposes of this PR, but its going to cause bloat to x230 and others later to have these options enabled.
> 
> diff linux-p8z77-m_pro.config linux-x230-maximized.config 102,103d101 < # CONFIG_BLK_DEV_MD is not set < # CONFIG_MD_RAID1 is not set 
> It would be good to figure out a way for making swraid usable that fits in with the your current board and config logical strategy

I'm confused here thought that was discussed and resolved already. If swraid kernel modules are built as modules, and packed under initrd (modules.cpio) on demand per board config and then loaded early under init, this is no issue.

Reason I readdress this here is that if it is changed under xx30 maximized linux config (and reused for w530/t430, and then ultrabay if anyone cares for x230) the board config would be the only place to switch that config option on, which at build time would pack it, and then be able to use it because exported variable through board config.

Looking for Ethernet driver config option name in board config should lead to working example for that, leading you to Makefile and Linux module config required changes to duplicate this without needing that module to be compiled in kernel as opposed to a kernel module. That kernel module compiled as module would be extensively stripped as opposed to being compiled in, and packed with higher compression rate by xz as a bonus. Everything not required to be compiled in kernel should not be compiled in kernel.

And for the sake of clarity, your linux config for that board could be a symlink to  xx30, and changed later on if something shows as needing be be specific. But the more eyes on the same kernel configs and again less maintainership to tune its size across kernel version bump, which is a really heavy task, which explains its not done often enough and when done, leads to some obscure bugs (as seen for t440p based on 5.x which misbehave as compared to 4.x today booting different distros and not showing early console output through i915/DRM drivers as 4.x counterpart).

tlaurion: Other example you might want to look for this is USB keyboard under x230 variant, since USB HID was removed under Linux config as being built in, so that users have to choose a variant if they accept security risk.

This implements all the chain, starting from that board config, the init check, Linux module config change and Makefile changes (if related, I don't remember on top of my head).

You could implement those changes as part of another PR, and have this PR benefit from it simply adding the 
export "SOFT_RAID=y" 
or whatever you see fit.

tlaurion: Yeah, just quickly reviewed https://github.com/osresearch/heads/pull/1157

Unfortunately would require a rebase since those changes should be presend under maximized Linux config to deserve a merge (sorry, the need drove merging of different Linux configs otherwise builds failed).

ThePlexus: Sorry you are right - its been so long i forgot about it.. So basically I could set this in linux-x230-maximized.config and it wont affect other boards, then i just add those modules to my builds, right?

```
CONFIG_MD_RAID1=m
CONFIG_DLK_DEV_MD=m
```

in which case I can just drop the need for the p8z77 kernel config and use x230. I need to push another change anyway as I missed 4096k extra free space in the layout mathematics on the blobs anyway, so ill wait to hear back before pushing that one

tlaurion: Yes, that and other changed from older PR.

No need for separate PR, just make that commit real clear about
-adding kernel modules
-adding mdraid module
- makefile requirement
- Linux module mods

And I will accept the change.
Do a reference to past PR in commit and that would be perfect.


Note that calculations for t440p confirmed that math should be ifd max-min in hex= cbfs_size ( where not sure if +- decimal one. Last time I retried that math for closed PR today on x230, I was 1 DEC byte off, soni think it must be +1 on final result, but yet again unsure, can reverify tomorrow).

ThePlexus: interesting. I had always thought CBFS SIZE was ROM SIZE - (IFD_SIZE+ME_SIZE+GBE_SIZE). Is it really just ROM_SIZE-IFD_SIZE (if so, where does ME and GBE live?)

`FMAP REGION: COREBOOT
cbfs_master_header             0x0        cbfs header        32 none
fallback/romstage              0x80       stage           85168 none
cpu_microcode_blob.bin         0x14dc0    microcode       26624 none
fallback/ramstage              0x1b600    stage          100556 LZMA (223560 decompressed)
config                         0x33f40    raw              1199 none
revision                       0x34440    raw               704 none
build_info                     0x34740    raw                97 none
fallback/dsdt.aml              0x34800    raw              8823 none
vbt.bin                        0x36ac0    raw              1243 LZMA (3902 decompressed)
cmos_layout.bin                0x37000    cmos_layout      1904 none
fallback/postcar               0x377c0    stage           28036 none
fallback/payload               0x3e5c0    simple elf    6975508 none
(empty)                        0x6e5600   null           932836 none
bootblock                      0x7c9200   bootblock       27584 none
`
This is my layout 
`00000000:00000fff fd
0001b000:007fffff bios
00003000:0001afff me
00001000:00002fff gb
`
800000 - 1b000 = 7E5000 (which is what i have set for CONFIG_CBFS_SIZE)



> Do a reference to past PR in commit and that would be perfect.

I recommend we not confuse the two and create dependency. This board being usable is independent of SWRAID and is now using all recommended changes and shares linux-x230-maximized.config so, as this PR stands now we should be able to merge and add this desktop into heads. I can then concentrate on documentation for this board. Once thats done, I will rebase SWRAID under a new PR.

config/coreboot-p8z77-m_pro-tpm1.config
```diff
@@ -0,0 +1,19 @@
...
+CONFIG_GBE_BIN_PATH="@BLOB_DIR@/p8z77-m_pro/gbe.bin"
+CONFIG_NO_GFX_INIT=y
+CONFIG_TPM_MEASURED_BOOT=y
+CONFIG_TPM1=y
+CONFIG_DRIVERS_PS2_KEYBOARD=y
+CONFIG_DEFAULT_CONSOLE_LOGLEVEL_8=y
+CONFIG_PAYLOAD_LINUX=y
+CONFIG_PAYLOAD_FILE="@BOARD_BUILD_DIR@/bzImage"
+CONFIG_LINUX_INITRD="@BOARD_BUILD_DIR@/initrd.cpio.xz"
+CONFIG_LINUX_COMMAND_LINE="intel_iommu=on intel_iommu=igfx_off nohz=off"
```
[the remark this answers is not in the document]
ThePlexus: This disables tickless timer (aka dynamic ticks) as on ASUS boards in the older kernels these throw a bunch of soft IRQ console logs, which could confuse users. Without this option users see repeats of this in console;

`NOHZ: local_softirq_pending XX`

This is down to;

> can_stop_idle_tick() prints "NOHZ: local_softirq_pending HH"
(where "HH" is the hexadecimal softirq vector number) when one or more
non-RCU softirq handlers are still enablded [sic] when checking to stop the
scheduler-tick interrupt. 
https://lore.kernel.org/lkml/20200706164816.GA1299@lca.pw/T/

Disabling the [tickless timer](https://github.com/torvalds/linux/blob/master/kernel/time/tick-sched.c) for our use case should not be an issue as the kernel is short lived in its run time, is not really worrying[ too much about power consumption ](https://stackoverflow.com/questions/9775042/how-nohz-on-affects-do-timer-in-linux-kernel/9776098#9776098)and not performance focused.

blobs/p8z77-m_pro/download_BIOS_clean.sh
```diff
@@ -0,0 +1,109 @@
...
+$IFDTOOL -x $ROMFILENAME.new.new.new || { echo "Failed ifdtool. Tool installed on host?" && exit 1;}
+
+printf '\x00' | dd of=flashregion_0_flashdescriptor.bin bs=1 seek=3837 count=1 conv=notrunc
+printf '\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF' | dd of=flashregion_0_flashdescriptor.bin bs=1 seek=3568 count=32 conv=notrunc
+
+echo "### Verifying expected hash of IFD"
+echo "$FINAL_IFD_SHA256SUM" | sha256sum --check || { echo "Failed sha256sum verification on factory IFD bin..." && exit 1; }
+cp flashregion_0_flashdescriptor.bin $BLOBDIR/ifd.bin || { echo "Failed to copy IFD ..." && exit 1; }
+echo "$FINAL_GBE_SHA256SUM" | sha256sum --check || { echo "Failed sha256sum verification on factory IFD bin..." && exit 1; }
+cp flashregion_3_gbe.bin $BLOBDIR/gbe.bin || { echo "Failed to copy GBE ..." && exit 1; }
```
tlaurion, line 104: Confused here. What is the hardcoded mac address? Any? From my understanding, those needed to be generated otherwise coming from user's taken backups of their own ROM.

I was not aware that the manufacturer would/could provide in their UEFI capsule an IFD and even less a GBE config: which normally includes a MAC address. This is interesting, to say the least.

ThePlexus: ifdtool dependency removed, script updated.

ThePlexus: just took a look for you, I dont see a pre defined MAC in gbe.bin, 0x1000 onwards are 0xFF. Looking at your generated xx30 gbe.bin its clear that what I extract is definitely a gbe.bin file (ex 0xAD 0xBA from 0x40 etc etc)

I don't use networking in coreboot personally and having the GBE partition in this state does not appear to affect operation when booing an linux based OS

ThePlexus:
ifdtool dependency removed, script updated. just used me_cleaner and dd instead of me_cleaner and ifdtool

config/coreboot-p8z77-m_pro-tpm1.config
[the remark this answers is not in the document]
ThePlexus: I realised i had this option in CONFIG_LINUX_COMMAND_LINE in error. The k'execd kernel doesn't need this. Just the coreboot payload. I have push and update

ThePlexus:
> mrc.bin from peppy firmware image

Was it not possible to extract just using offset and dd?

blobs/p8z77-m_pro/download_BIOS_clean.sh
ThePlexus: hold on.. think this may be on me for GBE.bin ... ive had a usb network adapter connected to my test machine. #fail. 

the onboard nic isnt showing up. 

please hold this PR for now till i figure out a solution.

tlaurion:
The same story will happen for me_cleaner here which is found under extracted coreboot build dir from previous build of the board, only difference here is that me_cleaner doesn't need to be built first but just present.

tlaurion:
@ThePlexus I see, you use osresearch's CircleCI instance because you haven't create your own.
Currently changing CACHE_VERSION so that everything builds clean so you will see what i'm talking about.

tlaurion:
@ThePlexus 

Applied logic in question with some minimum visibility improvements and htop board including maximized board, just changing name after incorporating everything maximized under hotp here under branch https://github.com/tlaurion/heads/tree/ThePlexus_p8z77-m_pro_build_test_clean

commit: https://github.com/osresearch/heads/commit/970dad8b962dcd1f908b4ca3a079b6ab2cb95b3e
(patch : https://github.com/osresearch/heads/commit/970dad8b962dcd1f908b4ca3a079b6ab2cb95b3e.patch )


Pipeline building this commit from clean (no cache) at https://app.circleci.com/pipelines/github/tlaurion/heads/1556/workflows/04d36601-09cc-4506-8c4d-b409a4dbc273

ThePlexus:
> osresearch's CircleCI instance because you haven't create your own.

no I dont have cicleci currently. 

> @ThePlexus
> 
> Applied logic in question with some minimum visibility improvements ...

I can see the logic what you mean,I would also think the xx30 approach would have worked too, by adding this to the YAML (as heads ships with a me_cleaner.sh in the codebase already, thus its a not really a dependency on coreboot being extracted)

`-m $(readlink -f ./blobs/xx30/me_cleaner.py)`

However, I dont much mind what the approach is, ive applied your recommendations from your patch and pushed

I also checked out the NIC - the reason it didn't come up for me was because I was using ubuntu without the driver. I have used the gbe.bin generated into heads, booting into QubesOS. the NIC comes up and it uses MAC ff:ff:ff:ff:ff:ff. I assume thats because that is what is at 0x1000 in gbe.bin. I will test tomorrow or in the next few days setting the same MAC as your generated gbe.bin from xx30

Are there any other blockers to pushing this board to master?

blobs/p8z77-m_pro/download_BIOS_clean.sh
ThePlexus: see https://github.com/osresearch/heads/pull/1345#issuecomment-1476777824 

seems fine

tlaurion:
@ThePlexus I think its OK as of now if build from circlrci flashed works for you and is built by ci.

On reusing x230 blobs dir me_cleaner, goal is to get away of it at some point, having in board configs a base for newcomers to replicate instead of building on top of hacks.

Let me know when circleci artifact flashed works for you and I will be liberal for the commits here, since new board and not touching anything else.

tlaurion:
@ThePlexus 

- That would be https://output.circle-artifacts.com/output/job/028d7620-2b40-4377-8056-c58f9bcdc83d/artifacts/0/build/x86/p8z77-m_pro-tpm1-hotp-maximized/heads-p8z77-m_pro-tpm1-hotp-maximized-v0.2.0-1494-gc205c40.rom

ThePlexus:
@tlaurion ill await the next artefact to test. There is no gbe.bin on this board, it was a path copy error on my part when moving in the download bios script. For some PBKAC reason I got my copy of the P8Z77-V not P8Z77-M, the former has gbe.bin and the latter does not. To add to cvonfusing, using the P8Z77-V BIOS for ME, IFD and GBE will allow the P8Z77-M to boot and function normally, it will just have PCI 03:00.0 missing (ethernet card)

The version building now uses the correct vendor BIOS, does not use GBE.bin and gives us 4k back. Not a lot - but was to those of us who use to code ZX81's back in the day ;)

tlaurion:
@ThePlexus 

> The version building now uses the correct vendor BIOS, does not use GBE.bin and gives us 4k back. 
Now i'm confused. Without gbe.bin, and with e1000/e1000e/igb driven ethernet, ethernet will not be useable. I went through that in my first steps creating the first maximized builds to understand this. 

So is not having ethernet functional satisfactory? You set the entry point here for that board support.
You can test this directly from Heads through network-recovery-init script, which ephemerally randomizes mac adress if module loads properly, checking that at least one duplicated section in gbe is valid, otherwise you will see in dmesg the same errors I went through in other issues because gbe.bin was invalid and or not present.

gbe.bin is configuration blob for intel ethernet and if not found, pci device is found, driver will attempt to be loaded but will fail in OS.

ThePlexus:
> @ThePlexus
> 
> > The version building now uses the correct vendor BIOS, does not use GBE.bin and gives us 4k back.
> > Now i'm confused. Without gbe.bin, and with e1000/e1000e/igb driven ethernet, ethernet will not be useable. I went through that in my first steps creating the first maximized builds to understand this.
> 
> So is not having ethernet functional satisfactory? You set the entry point here for that board support. You can test this directly from Heads through network-recovery-init script, which ephemerally randomizes mac adress if module loads properly, checking that at least one duplicated section in gbe is valid, otherwise you will see in dmesg the same errors I went through in other issues because gbe.bin was invalid and or not present.
> 
> gbe.bin is configuration blob for intel ethernet and if not found, pci device is found, driver will attempt to be loaded but will fail in OS.

@tlaurion sorry for confusion, there is * no * gbe.bin in this platform. At all.

The error was i used a download_BIOS_clean.sh script I made for a different test platform and copied that into the heads folder before I made the initial PR. There has been a knock on effect of confusion (including on my part where I couldnt get Ethernet working, then confused the USB adapter I had)

The vendor ROM for this board does also not contain gbe.bin 

```
$ ~/bios/coreboot/util/ifdtool/ifdtool --layout layout.txt P8Z77-M-PRO-ASUS-2203.ROM 
File P8Z77-M-PRO-ASUS-2203.ROM is 8388608 bytes
Wrote layout to layout.txt
$ cat layout.txt 
00000000:00000fff fd
00180000:007fffff bios
00001000:0017ffff me

```

I can confirm with the new script that Ethernet works fine. Herei s heads after POST
Screenshot-2023-03-21-at-13-04-24

and heres the PCI device in heads (10ec:8168 on 03:00.0)
Screenshot-2023-03-21-at-13-04-33

And here it is in a ubuntu OS usb boot, working as intended
Screenshot-2023-03-21-at-14-04-48

I will grab the artefacts and test, of course.

ThePlexus:
Looks good .

Screenshot-2023-03-21-at-14-22-00

Screenshot-2023-03-21-at-14-22-08



> gbe.bin is configuration blob for intel ethernet

This platform uses Realtek RTL8111

ThePlexus:
@tlaurion as to why adding an intel gbe.bin to the bios stops realtek RTL8111 coming up on the PCI bus remains a mystery that I suspect only the ASUS designers would understand.

tlaurion:
@ThePlexus why have you deleted circleci config?

ThePlexus:
@tlaurion i was doing some action in a side branch in git web admin - some how one of the deletes made it to my master branch. Have restored. 

Ideally I want to move to branches for different pull requests in future, which I will move to once the items in this PR are merged to heads master.

Are there any items left to do for this? as I have some videos to finish and I it will be awkward guiding people to download from different repos and/or applying patches to heads.

tlaurion:
Why don't you do beaches and use git on console?

tlaurion:
Otherwise I can co-aurhor your commits cleaning things up

tlaurion:
Kinda late hete

ThePlexus:
@tlaurion to be fair .. this is on me - i do use cli normally and only use branches for pulls where needed. I just didnt expect to have two things 'in flight' for heads at the same time so used master of my fork (https://github.com/osresearch/heads/pull/1357 to fix OEM reownership). So I tried to make a new branch, without CLI while in bed with a partner sleeping next to me - hence clicky on web is not as noisy as mashing keys. 

anyway, lesson learned. Ill tidy this up to one nice neat pull.

PR #1352, opened 2023-03-24
modules/linux: fix linux.*config targets for non-x86 architectures

krystian-hebel:
This patch adds ARCH="$(LINUX_ARCH)" to Linux targets working on config files. Without it, the architecture defaults to that of host, which for cross-compilation isn't right.


tlaurion:
@krystian-hebel temporary issue downloading gmp in pipeline. Please kick a retry on pipeline

PR #1356, opened 2023-03-28
oem-factory-reset: remove duplicates and add proper error redirection to file

tlaurion:
- Trace calls need to happen after sourcing /etc/functions not before
- Move sourcing of external files at beginning of file, remove /etc/functions sourcing duplicate
- gpg error redirection was sent to /dev/null where expected to be added to whiptail in case of error (2>&1 instead and redirection to file)

### DIFF
--- a/initrd/bin/oem-factory-reset
+++ b/initrd/bin/oem-factory-reset
@@ -1,10 +1,14 @@
 #!/bin/bash
 # Automated setup of TPM, GPG keys, and disk
 
-TRACE "Under /bin/oem-factory-reset"
-
 set -o pipefail
+
+## External files sourced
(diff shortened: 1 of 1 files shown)


tlaurion:
@danielp96 if you see this and can test this would be nice, pretty trivial changes but still bugfixes
@JonathonHall-Purism otherwise next week!

danielp96:
@tlaurion Tested OEM factory reset on a Librem 14, everything working fine.
Anything specific you want tested?

tlaurion:
@danielp96 Purely esthetic. Rebasing on master with @ThePlexus fix #1357 and merging

Issue #1203, opened 2022-08-24
Have qemu boards support writeable pflash and internal flashing


tlaurion:
Seems like both qemu and flashrom will need to be patched. Meanwhile, injecting key is made from separate qemu board build statement, and "reflashing" is rebooting qemu/kvm with newer built rom.

------
Current limitations stated under OP under #1188, where the following traces are what needs to be resolved: 

> @JonathonHall-Purism : would like to know what is the issue with nvram you were talking about earlier, and also what avenues possible with pflash since from what I've read from libvirt XML and qemu doc, the flash is not writable. Any input you have there would be useful in other open issues

Looks like the size limit is just a matter of changing this default for max_fw_size: https://gitlab.com/qemu-project/qemu/-/blob/master/hw/i386/pc.c#L1833

Re: writable flash, it seems to exist but the only discussion I can find is about people using it incorrectly: https://bugs.launchpad.net/qemu/+bug/1818367/comments/4

>The <loader> element, with @type='pflash', no other attributes, and then
>no <nvram> sibling element either, happens to be valid, but it is an
>extremely niche use case. It is used when you have a unified, writeable,
>OVMF.fd file that contains both the firmware executable and the live
>variable store.

This does sound like what we want, I have no idea right now how this works with plain qemu, whether flashrom supports it, or what it would take to add.

_Originally posted by @JonathonHall-Purism in https://github.com/osresearch/heads/issues/1188#issuecomment-1218004671_

tlaurion:
Leaving traces of pertinent mailing list sub threads

- spi chips implementation https://www.mail-archive.com/qemu-devel@nongnu.org/msg58003.html

tlaurion:
Track https://github.com/Dasharo/dasharo-issues/issues/828

tlaurion:
Specifically
- https://review.coreboot.org/c/coreboot/+/82555
- https://review.coreboot.org/c/coreboot/+/82639

As referred by https://github.com/Dasharo/dasharo-issues/issues/828

tlaurion:
Question asked under Dasharo OSFV matrix channel at https://matrix.to/#/!MwWOJhMJzlxIdAQxae:matrix.3mdeb.com/$-G0q4GBkLoQwCk7d1geSf-U74XuqTNwtSb6HT072K-s?via=matrix.org&via=nitro.chat&via=matrix.3mdeb.com

tlaurion:
[...]

PR #1357, opened 2023-03-29
fix broken OEM re-ownership process

ThePlexus:
Problem
When using a custom password for TPM, the OEM re-ownership process is broken

Impact 
The OEM re-ownership process breaks for any user setting a custom password and not just using 12345678 

First appeared
https://github.com/osresearch/heads/commit/6923fb5e20cfb662b76c1d1fa48e245ecd8d79fb

Detail
on line 498, if blank, the TPM custom password is overwritten with TPM_PASS_DEF (eg, when no custom password is set by the user installing)

```
if [ "$TPM_PASS" == "" ]; then TPM_PASS=$TPM_PASS_DEF; fi
```
so far so good.  $TPM_PASS should be used for all TPM interaction from this point. $TMP_PASS_DEF is now a disposed of variable.

we see that happens when resetting the TPM on line 712 (generate_checksums) is that $TPM_PASS is used (correctly)

```## reset TPM and set password
if [ "$CONFIG_TPM" = "y" ]; then
  echo -e "\nResetting TPM...\n"
  tpmr reset "$TPM_PASS" >/dev/null 2>/tmp/error
---SNIP
```
The TPM now has either the custom password of the user, or the default of 12345678 depending on user selection.

On line 712, we duck into the generate_checksums sub, which for some reason reverts to TPM_PASS_DEF

```
    # create Heads TPM counter
    if [ "$CONFIG_TPM" = "y" ];then
	    if [ "$CONFIG_IGNORE_ROLLBACK" != "y" ]; then
		    tpmr counter_create \
			 -pwdo "$TPM_PASS_DEF" \
--SNIP
```

This then, rightly, fails due to 
```
Authentication failed (Incorrect Password) (ox1) from TPM_CreateCounter
```

### DIFF
--- a/initrd/bin/oem-factory-reset
+++ b/initrd/bin/oem-factory-reset
@@ -187,7 +187,7 @@ generate_checksums()
     if [ "$CONFIG_TPM" = "y" ];then
 	    if [ "$CONFIG_IGNORE_ROLLBACK" != "y" ]; then
 		    tpmr counter_create \
-			 -pwdo "$TPM_PASS_DEF" \
+			 -pwdo "$TPM_PASS" \
 			 -pwdc '' \
 			 -la -3135106223 \
 			    | tee /tmp/counter \


tlaurion:
Nasty.... And a reminder for testing things outside of qemu :/

Testing now. Not sure how that landed in master (my bad, sorry about that.)

ThePlexus:
> Nasty.... And a reminder for testing things outside of qemu :/ (which fails injecting key in firmware because #1203)
> 
> Testing now. Not sure how that landed in master (my bad, sorry about that.)

As I say to my teams, cant test for every possibility of every combination of every change. we would not progress as we would be doing nothing but testing all the time. The tpm2 changes were a large change set, so small bugs are expected . keep up the good work sir.

PR #1367, opened 2023-04-06
modules/json-c: set cmake build type as minsizerel

danielp96:
By default json-c builds as debug instead of release.

Adding `CMAKE_BUILD_TYPE=minsizerel` ensures it does not add debug info and optimizes for file size.

Fixes https://github.com/osresearch/heads/issues/1328


tlaurion:
@danielp96 : awesome find!!!!! Thanks a lot for that!
Will check for build and redo build locally and compare lib and then merge!

PR #1371, opened 2023-04-12
WIP - Z690 A  DDR5 / Msi ms7d25 TPM2

ThePlexus:
This is my attempt so far to bring the Z690-A DDR5 into Heads. This is a WIP. The board seems to have all regions unlocked for internal programmer. 

**status**
It boots to console only and we see hello world;

Screenshot from 2023-04-12 11-29-59

**connections**
- Flash chip is MX25U25635F WSON8, soldered on
- External flashing is not amazingly easy, but possible [with a WSON8 clip ](https://docs.dasharo.com/variants/msi_z690/recovery)
- External flashing is a bit less fiddly by using the[ SPI bus via the TPM2 header](https://docs.dasharo.com/variants/msi_z690/development/#hardware-connection) and connecting directly to one pin on the ROM chip (I cannibalised a connector from a tpm to make a connector, and used a clamp and probe to attach to the pin on the ROM) 
- Dasharo also show how to get [read only ttyS0 console ](https://docs.dasharo.com/variants/msi_z690/development/#hardware-connection)

**Requirements**
- copied [1251](https://github.com/osresearch/heads/pull/1251) to use Dasharo flashrom for MX25U25635F compatibility 
- Have to use coreboot 4.19 for board support, though i expect 4.20 will be a better choice given[ all the work since 4.19](https://review.coreboot.org/q/+ms7d25)
- Patch made to back port a fix for some Alderlake iGPU https://review.coreboot.org/c/coreboot/+/70101/4#message-0ac2afa1c9d02452721469c837caabb01ccd86b2

**ME**
- Cant me_clean and reduce ME firmware on this generation of cpu. just how it is. 
- We can use ME hap bit to disable ME https://github.com/6d6178667269747a/me_cleaner/commit/d6b313ad75ac4ccda3668042d0a8dacd4fef2f0f 
- I set HAP the bit in download_BIOS_clean.sh by flipping 0x1DE to 0x11 - validated that setting this bit on factory firmware shows ME version 0.0.0.0 in factory BIOS
- as we cant shrink ME, it may be an idea to use VSCC mod by default on this board
- could we make a patch and have coreboot measure the ME region for changes?

**Findings/WIP stuff**
- Board config not using its own kernel, i just stole the librem common as it was the same version and a reasonably nearby CPU family. WIP
- No graphics working yet, console output only. Coreboot uses FSP GOP, so I assume this needs some kernel modules.  
- theres a worrying TPM error message in the console output that needs looking at (see below)
 - Coreboot VBOOT seems to use the onboard fTPM even with a dTPM attached. The above mentioned error still remains. This is going to take a bit more work
 -  network at boot is untested and unlikely. dashro extract a binary blob to use in a UEFI payload, and currently this WIP is CSM mode https://github.com/Dasharo/coreboot/blob/msi_ms7d25/release/build.sh#L16

** TPM error** 
```
[DEBUG]  TPM: CRB Interface is not active. System needs reboot in order to active TPM.
[ERROR]  TPM: Probe failed.
[ERROR]  tlcl_lib_init: tis_init returned error
[ERROR]  TPM: Can't initialize.
```
even though it seems coreboot creates a TPM log and TPM measurements do present.
```
[DEBUG]  ACPI:    * TPM2
[DEBUG]  TPM2 log created at 0x7699b000
-- SNIP
[INFO ]  coreboot TPM log measurements:
[INFO ]   PCR-2 ae47351f1afcf7df124e3b547b8d2a96dc43da705e432bc1ba08942ef6822f89 SHA256 [FMAP: FMAP]
[INFO ]   PCR-2 ea2696180cda0ce19438a62f0a1bdc86007908e825f4dec108147aea1aa58fb5 SHA256 [CBFS: bootblock]
[INFO ]   PCR-2 eeca2517dbe532bab8ceb7c7612338e4e72d9bd7aaa86f518b6ee9feaae52ee9 SHA256 [CBFS: fallback/romstage]
[INFO ]   PCR-2 25ff8837f740d51fe99acf1cb85582df6f97c1425e58b4cd67d14bfb426deec0 SHA256 [CBFS: fspm.bin]
[INFO ]   PCR-2 74eba1e41dc4e310666224177fbefa9fa723952ff058c6948805e35cc7ee2ed4 SHA256 [CBFS: fallback/postcar]
[INFO ]   PCR-2 29de505911e6181cb2564a486fcdb77c9b62d22afb7eba7e75a1adaf20dfbca9 SHA256 [CBFS: fallback/ramstage]
[INFO ]   PCR-2 1dc207481a8570abc93295543da654b24eaceb1b17efaf536a0db07617ff3d15 SHA256 [CBFS: cpu_microcode_blob.bin]
[INFO ]   PCR-2 7ec6adf33596c9051cad0ef295675bdf0b02c78d89675d7a3bfd7e74f677d7d0 SHA256 [CBFS: fsps.bin]
[INFO ]   PCR-2 2502ce2ed0172d622f0eb24aaa7c173689e54719eab258a161422f5409bd3b2e SHA256 [CBFS: vbt.bin]
[INFO ]   PCR-2 5ae7b8ee21400362658243091c1dd30044d5392ca725db4d7837b5a4c32d2ddd SHA256 [CBFS: fallback/dsdt.aml]
[INFO ]   PCR-2 1e0433b680a0deba810e265b0e29680d6c67c3d47203bfeaa52bc332f4087a75 SHA256 [CBFS: fallback/payload]

```


miczyg1:
@ThePlexus just a few moments ago I pushed slightly better approach to flash the board: https://docs.dasharo.com/variants/msi_z690/development/

WSON8 clip is not needed anymore. The BIOS flash CS pin is exposed on JTPM1 pin5 (reserved), so basically any 1.8V level capable programmer should do the work by connecting 2mm to 2.54mm female-female dupont wires to JTPM1 header.

miczyg1395:
Also the board may come with Winbond chip too, not that it is relevant for internal programmer

miczyg1395:
dTPM is not supported for now, because coreboot can work only with one TPM type, either fTPM or dTPM and will need code modifications and fiddling with ME to switch to dTPM.

simonbunny:
Thanks for the knowledge share @miczyg1 . Thats great news to hear - I shall modify the top post here to record the change in header status. 

Re dTPM - thanks for the info I suspected that was the case. I had to do something [similar for the P8H61](https://review.coreboot.org/c/coreboot/+/32080) (though a lot less effort :P )

ThePlexus:
Thanks for the knowledge share @miczyg1  . Thats great news to hear - I shall modify the top post here to record the change in header status.

Re dTPM - thanks for the info I suspected that was the case. I had to do something [similar for the P8H61](https://review.coreboot.org/c/coreboot/+/32080) (though a lot less effort :P )

simonbunny:
@tlaurion see any problems, for example FSP? is this a wasted effort?

boards/msi_ms7d25_ddr5-tpm2/msi_ms7d25_ddr5-tpm2.config
```diff
@@ -0,0 +1,78 @@
...
+CONFIG_COREBOOT_CONFIG=config/coreboot-msi_ms7d25_ddr5-tpm2.config
+
+export CONFIG_COREBOOT=y
+export CONFIG_COREBOOT_VERSION=4.19
+export CONFIG_LINUX_VERSION=5.10.5
+
+#Enable DEBUG output
+export CONFIG_DEBUG_OUTPUT=y
+export CONFIG_ENABLE_FUNCTION_TRACING_OUTPUT=y
+export CONFIG_TPM2_CAPCTURE_PCAP=y
```
tlaurion, line 15: Remove this. Only useful to have pcap trace under /tmp

boards/msi_ms7d25_ddr5-tpm2/msi_ms7d25_ddr5-tpm2.config
```diff
@@ -0,0 +1,78 @@
...
+export CONFIG_TPM2_TOOLS=y
+export CONFIG_PRIMARY_KEY_TYPE=ecc
+CONFIG_TPM2_TSS=y
+CONFIG_OPENSSL=y
+#Set this option to zero out the VSCC table https://github.com/osresearch/heads/pull/1358#discussion_r1153251399 
+export CONFIG_ZERO_IFD_VSCC=y
+
+CONFIG_LINUX_USB=y
+
+export CONFIG_TPM=y
```
tlaurion, line 58: This is tpm1. Remove

boards/msi_ms7d25_ddr5-tpm2/msi_ms7d25_ddr5-tpm2.config
```diff
@@ -0,0 +1,78 @@
...
+export CONFIG_BOOTSCRIPT=/bin/generic-init
+export CONFIG_BOOT_REQ_HASH=n
+export CONFIG_BOOT_REQ_ROLLBACK=n
+export CONFIG_BOOT_KERNEL_ADD="intel_iommu=on intel_iommu=igfx_off debug console=tty0 console=ttyS0,115200 earlyprintk=ttyS0,115200"
```
tlaurion, line 62: console=tty needs to he the last one to tell what is main console (VGA)

config/coreboot-msi_ms7d25_ddr5-tpm2.config
```diff
@@ -0,0 +1,21 @@
...
+CONFIG_HAVE_FSP_GOP=y 
+CONFIG_RUN_FSP_GOP=y 
+CONFIG_TPM_MEASURED_BOOT=y
+CONFIG_TPM2=y
+CONFIG_DRIVERS_PS2_KEYBOARD=y
+CONFIG_DEFAULT_CONSOLE_LOGLEVEL_8=y
+CONFIG_PAYLOAD_LINUX=y
+CONFIG_PAYLOAD_FILE="@BOARD_BUILD_DIR@/bzImage"
+CONFIG_LINUX_INITRD="@BOARD_BUILD_DIR@/initrd.cpio.xz"
+CONFIG_LINUX_COMMAND_LINE="intel_iommu=on intel_iommu=igfx_off debug console=tty0 console=ttyS0,115200 earlyprintk=ttyS0,115200"
```
tlaurion, line 20: Same here. Swap consoles so last is desired main console last

miczyg1:
> * No idea of how graphics are dealt with by FSP GOP, but serial only shows that either kernel i915 is not recent enough, that coreboot kernel command line doest specify right options for Heads to init DRM+GPU to enable framebuffer

FSP GOP works the same way as libgfxinit. The result is a coreboot framebuffer saved in LB tags for use by payload. If heads Linux kernel does not initialize i915 properly, then maybe the Linux version is too old?

modules/flashrom
```diff
@@ -1,34 +1,39 @@
 modules-$(CONFIG_FLASHROM) += flashrom
```
tlaurion, line 1: We might need to separate flashrom module from dasharo-flashrom. Not all older boards will be able to have this new flashrom version which is bigger.

miczyg1: Z690 chipset is available in upstream flashrom already: https://github.com/flashrom/flashrom/blob/master/chipset_enable.c#L2178

tlaurion: @miczyg1 : build of master flashrom happening under https://app.circleci.com/pipelines/github/tlaurion/heads/1604/workflows/f7a74cfc-72f2-4de7-aedc-e8a7c24cc5aa for commit https://github.com/tlaurion/heads/commit/0d4997952cbc63a69ed24f3c7daf31b02143c544 (building Heads flashrom to be as flashrom https://github.com/flashrom/flashrom/commit/f3c21c6439f8bbd2911f120c55f301982ecd35bc)

Will edit this commit for size comparison to explain my point once build is finish

tlaurion: Clearer and direct test is happening here:
- flashrom/flashrom master on top of heads master: https://app.circleci.com/pipelines/github/tlaurion/heads/1607/workflows/23f6c9f3-926f-478f-a709-70be6d87bae4
- dasharo/flashrom (WP+ast1100 support) on top of heads master: https://app.circleci.com/pipelines/github/tlaurion/heads/1605/workflows/26753865-ac5a-4097-8e06-eccc85d748bf

Will edit this once builds are finished

tlaurion: So let's take a clean view now into space differences comparing actual heads master using old flashrom/flashrom and dasharo/flashrom.

Will edit with results.

tlaurion:
@ThePlexus can you provide full serial console log please?

tlaurion:
> Intel_iommu=igfx_off is not applied from coreboot command line down to heads kernel iommu and passed right through to env to be picked up by userland. Edit: it is applied (kernel output confirming it), unknown why init sees it in userland.

@simonbunny I edited previous posts.

ThePlexus:
Closing this out for now, as it will require (at minimum) some significant coreboot modifications and has some less desirable outputs. I will open anew PR with work so far (it does post and gets display) but there are some significant upstream blockers to this 

1. We cannot use HAP bit to disable ME as coreboot on this board. It requires us to use HECI to disable PTT and switch to dTPM. This is method is [using cse_enable_ptt(false);](https://review.coreboot.org/c/coreboot/+/68919), however this function will not operate with HAP bit 0x11 only with HAP bit 0x10.  This would leave only 'soft disable of ME via HECI' as a way to disable ME on this board. 
- @miczyg1 do you think there is any way to use a dTPM without HECI? For example, if I were to commission a coder to work on this upstream in coreboot. Or are we looking at a situation where we cannot use a dTPM with HAP bit enabled, no matter what we do?
- @tlaurion Is soft disable of ME via HECI only a viable option? Im leaning towards no as im not sure how much it would be trusted to not simply be able to be switched back on.

2. When using HECI cse_enable_ptt(false) in the main board silicon init, the system does global reset, and on the next boot it does show PTT as 'already disabled'. However, there is still not dTPM as addressing this is not yet added to the board  upstream in coreboot. 

3. The upstream mainboard Kconfig has TPM_CRB, so it looks like coreboot vboot/measure still just fallbacks to this even with cse_enable_ptt (fTPM PCR2 has measurements in it)

@tlaurion what is the view on using the fTPM in these families, assuming a dead end on dTPM and ME?

Ill do a new PR with the work so far at some point.

tlaurion:
Can you turn that into an issue? @ThePlexus ?

miczyg1:
> * @miczyg1 do you think there is any way to use a dTPM without HECI? For example, if I were to commission a coder to work on this upstream in coreboot. Or are we looking at a situation where we cannot use a dTPM with HAP bit enabled, no matter what we do?

TBH I haven't verified how chipset behaves when ME is disabled with HAP (I have no MSI compliant SPI dTPM to check if TPM can be switched from PTT to SPI dTPM). I know for sure that fTPM stops working with whatever method you use to disable ME.

> 2. When using HECI cse_enable_ptt(false) in the main board silicon init, the system does global reset, and on the next boot it does show PTT as 'already disabled'. However, there is still not dTPM as addressing this is not yet added to the board  upstream in coreboot.

Yes, we need the patches that unify the TPM drivers in coreboot: https://review.coreboot.org/q/topic:%22tpm-unification%22

> 3. The upstream mainboard Kconfig has TPM_CRB, so it looks like coreboot vboot/measure still just fallbacks to this even with cse_enable_ptt (fTPM PCR2 has measurements in it)
> 
> @tlaurion what is the view on using the fTPM in these families, assuming a dead end on dTPM and ME?
> 
> Ill do a new PR with the work so far at some point.

With the patches above, there should be no need for TPM_CRB option anymore.

miczyg1:
Maybe to be more precise, I haven't found a way to use dTPM before disabling fTPM/PTT via HECI or by ME tools.

miczyg1:
@tlaurion I took a quick look at what happens to the PTT state when HAP is enabled aaand... It looks like ME/chipset automatically switches to the next available and enabled/configured TPM interface. In case of MSI, I have checked this register: https://github.com/coreboot/coreboot/blob/master/src/security/intel/txt/txt_register.h#L98 which indicates currently active TPM bus/interface and it indicated SPI TPM. So that's a good news, that we don't have to use HECI to disable PTT explicitly if we use HAP. However, we still need that patches unifying TPM drivers in coreboot to land.

tlaurion:
> However, we still need that patches unifying TPM drivers in coreboot to land.

@miczyg1 something to track?

miczyg1:
> > However, we still need that patches unifying TPM drivers in coreboot to land.
> 
> @miczyg1 something to track?

@tlaurion see my earlier comment, there is a link to gerrit

PR #1372, opened 2023-04-12
Fix benign script errors

JonathonHall-Purism:
A few errors now show up under Bash that Ash had silently ignored.  It happens to behave correctly, but fix the errors.


tlaurion:
Good catch. Merging

PR #1351, opened 2023-03-24
Debugging kernel 5.x kexec issues add tools to ease kernel version bumping

tlaurion:
This is just a tracing test in my attempt to narrow down the problem of why kexec'ing from 4.14 kernel works even though i915 driver is not provided in Tinycore boot of dd'ed iso (simplest test I found) where it just hangs when kexec'ing from 5.x.

As of now, this PR 
- adds i915drmfb into kexec patch to do equivalent of what was done for inteldrmfb (reuse vga)
  -  reusing vga doesn't work. Why, not sure yet. 
- adds build statements into modules/linux to be able to copy to versioned olconfig, and from defconfig to olconfig and back and forth.
- shows a lot of unneeded stuff that was removed in tested configs
- Misses what is needed to work correctly

PR #1380, opened 2023-04-19
coreboot+linux modules: add helpers to edit config, save in oldconfig/defconfig

tlaurion:
Helper tools borrowed from #1351 

Will help me bumping 4.14 kernel configs straight to 5.10 with minimal changes and help other do so in the future.

The way I use them is the following:
`make BOARD=xyz linux.modify_and_save_oldconfig_in_place`: takes the in tree linux board config, modifies it through menuconfig (or not) and saves its oldconfig (full format) back in tree.
`make BOARD=xyz linux.modify_defconfig_in_place`:  takes the in tree linux board config, modifies it through menuconfig (or not) and saves its defconfig (variation from defaults for arch) back in tree.
 
Same applies with `make BOARD=xyz coreboot.modify_and_save_oldconfig_in_place` and `make BOARD=xyz coreboot.modify_defconfig_in_place`

---- 
@JonathonHall-Purism doesn't impact anything else then dev/maintenance cycle

### DIFF
--- a/modules/coreboot
+++ b/modules/coreboot
@@ -147,6 +147,31 @@ coreboot.saveconfig:
 		DEFCONFIG="$(pwd)/$(CONFIG_COREBOOT_CONFIG)" \
 		savedefconfig
 
+coreboot.save_defconfig_in_place:
+	$(MAKE) \
+		-C "$(build)/$(coreboot_base_dir)" \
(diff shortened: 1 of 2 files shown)


JonathonHall-Purism (commented):
Just one tiny nitpick, looks great and super handy :partying_face:

PR #1390, opened 2023-04-27
Bash reproducibility

danielp96:
Removing debug info from bash.

Adding patch to disable bash auto increment of the build number that goes in the version string of the binary.
The increment is triggered when bash/configure is run and bash/.build file already exists.
Patch consists of removing the relevant flag from the call to helper script in makefile template.

Should fix bash reproducibility.


tlaurion:
Good digging and finding!!!! Thanks @danielp96 ! Merging!

PR #1393, opened 2023-05-01
linux 5.10.5: add upstream patch for 5.10.5 (use after free on realloc)

tlaurion:
Thanks to @srgrint for reporting and insisting for backporting this in the meantime we migrate to 5.10.178 for all boards. Address comment https://github.com/osresearch/heads/issues/1392#issuecomment-1530132938 forward.

@JonathonHall-Purism we merge?


JonathonHall-Purism:
Yes thank you @tlaurion, reviewed the patch and looks correct to fix this issue (and it's in a host tool not the kernel proper, so low impact risk anyway).  Let's merge

Issue #1392, opened 2023-05-01
5.10.5 kernel build failure under Debian bookworm


srgrint:
## Context of the Build

**1. What board are you trying to build?**
T440p
**2. What repository:branch are you using to build from?**
- [ X] Heads:Master
- [ ] Other (please specify)

**3. What version of coreboot are you trying to build**
- [ ] 4.13
- [ ] 4.14
- [ ] 4.15
- [X ] 4.17
- [ ] Other (please specify)

**4. In building the rom where did you get the blobs?**
- [ ] No blobs required
- [ ] Provided by the company that installed Heads on the device
- [ ] Extracted from a backup rom taken from this device
- [ ] Extracted from another backup rom taken from another device (please identify the board model)
- [ X] Extracted from the online bios using the automated tools provided in Heads
- [ ] I don't know

**5. If using the automated tools to get the blobs did you run the relevant scripts in the blobs directory**
- [X ] Yes
- [ ] No

**6. What operating system are you using**
Debian bookworm/12

Have tried building the t440p-maximized rom using my test install of Debian 12 (bookworm) which ships with GCC 12

Once I have built the toolchain without gnat support, the build of the 5.10.5 kernel fails due to finding a use after free (due to the new compiler being more aware than what shipped with Debian 11).  The bug has already been fixed upstream last year - if I apply the following patch, heads builds as expected:

https://github.com/torvalds/linux/commit/52a9dab6d892763b2a8334a568bd4e2c1a6fde66

I don't know whether you want to cherry pick this into heads?

tlaurion:
@srgrint linux 5.10.178 (LTS) has the fix backported.

@JonathonHall-Purism you might want to backport referred patch, where my helper to prompt devel (who will bump kernel version) works really sweetly under debugging branch.


Note that bookworm on builder host is not officially supported as of now.

I'm personally tired of supporting multiple hosts buildstacks, where goal would be to have Nix layer on top of any linux distrib to fix this as well as building under docker image, which is more interesting on a maintainership perspective. 

As of now, it is recommended to build on top of debian-10.

srgrint:
> I'm personally tired of supporting multiple hosts buildstacks, where goal would be to have Nix layer on top of any linux distrib to fix this as well as building under docker image, which is more interesting on a maintainership perspective.

Completely agree - Nix layer is definitely the way forward.

The use-after-free kernel bug is real though - just Debian 10 buildstack isn't astute enough to catch it

[...]

PR #1381, opened 2023-04-21
Bump remaining 4.14 boards kernel to 5.10.5 (and coreboot to 4.19)

tlaurion:
This PR brings a lot of cleanup to linux configs after bumping configs from 4.14 to 5.10.5, unifying those to easy compare boards across each other. Detailed changes:
Modules:
- fixed coreboot and linux with helpers used to produce the following changes (save_defconfig, save_oldconfig, modify_oldconfig modify_defconfig etc. Do make BOARD=xyz and abuse of autocomplete with tab)

Linux configs:
- Only the modules possibly packed in modules.cpio under initrd are compiled
- Crypto backend has been brought to the minimal needed for cryptsetup to work (you might want to raise issues if you use twofish or serpent, but it is not clear why those should be added as well, since perf is way beyond with AES-XTS on the platforms, since the CPU has accelerated extensions for AES)
- GENERIC_X86 CPU support is changed to CONFIG_MCORE2 which also brings pentium6+ NOP operations and accelerated checksuming support
- Important to note that CONFIG_RANDOM_TRUST_CPU=y was added in the configs to get rid of reading non-random bytes on early init (entropy pool is filled by TPM and CPU extension here. Should be good enough)
- CONFIG_PREEMPT_NONE=y: Remove preemptiveness as for servers. Under heads, we are single tasking. No point having this big thing in kernel https://lwn.net/Articles/746780/
- IO scheduler: only enable CONFIG_MQ_IOSCHED_DEADLINE=y since we want maximum throughput and do not have concurrent tasks (outside of the gpg-agent working in the background)
- CONFIG_CPU_ISOLATION=y : Enable CPU Isolation across all boards: this permits to make sure that the kernel tasks running on a core are not distrurbed by user tasks. full dynticks are not in.
- CONFIG_MULTIUSER not defined: Removing cluttering since we are single root user under Heads anyway, for now.
- CONFIG_IO_URING=y : limit number of copy operations between kernel and user space from apps
- CONFIG_ZONE_DMA not defined: relevant for older hardware (less then 32bit addressing space)
- CONFIG_X86_MPPARSE not defined: relevant for older smp systems
- CONFIG_CPU_FREQ_DEFAULT_GOV_PERFORMANCE is enabled and SCHEDUTIL is disabled: we want performance with CPU sched with deadline IO. No need for having other schedules compiled in.
[...]

PR #1394, opened 2023-05-02
backport upstream patch for 4.14.62.  Allows building on debian 12

srgrint:
As discussed in #1392 , a use after free bug in the kernel prevents heads kernel being compiled on debian 12.

This has been fixed in #1393 for boards using 5.10.5
The bug also affects boards using kernel 4.14.62.  This PR will fix this and allow building of boards using this kernel on debian 12.

Clearly the future is for all boards to use 5.10 kernels (as per WIP #1381).  Nevertheless, might be sensible to fix in 4.14 branch in the interim until no boards use the old kernel.


tlaurion:
As noted per @JonathonHall-Purism at https://github.com/osresearch/heads/pull/1393#issuecomment-1531759632 for kernel 5.10:

> Yes thank you @tlaurion, reviewed the patch and looks correct to fix this issue (and it's in a host tool not the kernel proper, so low impact risk anyway). Let's merge

PR #1397, opened 2023-05-04
modules/fbwhiptail: Update for reproducibility

danielp96:
Updated to reproducible version of fbwhiptail.
Added flags to remove debug info.
Updated url to current one instead of going through redirect.

This fixes fbwhiptail reproducibility.


tlaurion:
Checked changes merged upstream going down to Make.

PR #1400, opened 2023-05-08
Add P8Z77-V

d-wid:
I used this as a practice board of sorts before moving on to the Z220, hoping I could get away with just internally flashing the HP (unfortunately it wasn't to be and after a mistake I had to flash it externally anyway). It does boot and is able to kexec into something else, but I never tested it with a TPM or HOTP device.


tlaurion:
@d-wid watch for https://github.com/osresearch/heads/pull/1398 changes and rebase/adapt for merge!

tlaurion:
@d-wid wrote 
> Great! Unfortunately I've not been in a position to boot the #1400 Z77 board for some time now and for the foreseeable future. I could "port" the config over to the current Linux and Coreboot versions like for the Z220 but I won't test it (not as much as my now main PC Z220 any time soon in any case), so I guess it'll be abandoned until someone (else) who uses the Z77-V shows up. Would you suggest closing it or leaving it as a draft? (strictly speaking it being merged at some point is not ruled out after all) Will test again if/when #1403 merges, exam schedule permitting.

tlaurion:
Abandoned

PR #1416, opened 2023-06-12
seal-totp: better document inclusion of PCR4 measurements part of sealing op

tlaurion:
commit https://github.com/osresearch/heads/commit/7b949a1a44eced0e90199eb6bd4e01513234f081 would have deserved a PR.

Here is a cbmem -L (tcpa log) on Talos II to explain why we need to include PCR4 measurements which now includes extends with BOOTKERNEL and seperators measurements.

Edit: Also why taking for granted that PCRs are zeroed on boot is not respecting TCG guidelines. Basically, PCRs boot values should be used as a base for tuturecalcpcr operations to determine to be sealed values for next boot unseal operations.

```
~ # cbmem -L
TPM2 log:
	Specification: 2.00
	Platform class: PC Client
TPM2 log entry 1:
	PCR: 2
	Event type: Action
	Digests:
		 SHA1: 27c4f1fa214480c8626397a15981ef3a9323717f
	Event data: 50 FMAP: FMAP
TPM2 log entry 2:
	PCR: 2
	Event type: Action
	Digests:
		 SHA1: 2d5f9d8abb287dc868a1b8505b11989f307d8cce
	Event data: 50 FMAP: BOOTBLOCK
TPM2 log entry 3:
	PCR: 2
	Event type: Action
	Digests:
		 SHA1: cba866ea5956334e5d1c7b58b21c28836cd25c95
	Event data: 50 CBFS: fallback/romstage
TPM2 log entry 4:
	PCR: 2
	Event type: Action
	Digests:
		 SHA1: 0447bd27531fcd0e293fca6f43e7ca6539de874e
	Event data: 50 CBFS: fallback/ramstage
TPM2 log entry 5:
	PCR: 2
	Event type: Action
	Digests:
		 SHA1: 567e5aa96274a685bf6ddda2599013e48bd346e2
	Event data: 50 CBFS: fallback/payload
TPM2 log entry 6:
	PCR: 2
	Event type: Action
	Digests:
		 SHA1: 11b04b24178f9c1f4dbe89a1b6e959dab6caf6c4
	Event data: 50 CBFS: 2-cpus.dtb
TPM2 log entry 7:
	PCR: 3
	Event type: Action
	Digests:
		 SHA256: 6e7b06693452d997ac534e823b1ea79e5bb8ed19ba8a7af878abf10199c3d515
		 SHA1: 6e7b06693452d997ac534e823b1ea79e5bb8ed19
	Event data: 7 VERSION
TPM2 log entry 8:
	PCR: 2
	Event type: Action
	Digests:
		 SHA256: de73053377e1ae5ba5d2b637a4f5bfaeb410137722f11ef135e7a1be524e3092
		 SHA1: de73053377e1ae5ba5d2b637a4f5bfaeb4101377
	Event data: 11 IMA_CATALOG
TPM2 log entry 9:
	PCR: 4
	Event type: Action
	Digests:
		 SHA256: c013f948107aa41f33a44ecdac691cd5d0fe9335584ae2476b2bc0fb7d866e87
		 SHA1: c013f948107aa41f33a44ecdac691cd5d0fe9335
	Event data: 10 BOOTKERNEL
TPM2 log entry 10:
	PCR: 0
	Event type: Separator
	Digests:
		 SHA256: ad95131bc0b799c0b1af477fb14fcf26a6a9f76079e48bf090acb7e8367bfd0e
		 SHA1: d9be6524a5f5047db5866813acf3277892a7a30a
	Event data: 4 ����
TPM2 log entry 11:
	PCR: 1
	Event type: Separator
	Digests:
		 SHA256: ad95131bc0b799c0b1af477fb14fcf26a6a9f76079e48bf090acb7e8367bfd0e
		 SHA1: d9be6524a5f5047db5866813acf3277892a7a30a
	Event data: 4 ����
TPM2 log entry 12:
	PCR: 2
	Event type: Separator
	Digests:
		 SHA256: ad95131bc0b799c0b1af477fb14fcf26a6a9f76079e48bf090acb7e8367bfd0e
		 SHA1: d9be6524a5f5047db5866813acf3277892a7a30a
	Event data: 4 ����
TPM2 log entry 13:
	PCR: 3
	Event type: Separator
	Digests:
		 SHA256: ad95131bc0b799c0b1af477fb14fcf26a6a9f76079e48bf090acb7e8367bfd0e
		 SHA1: d9be6524a5f5047db5866813acf3277892a7a30a
	Event data: 4 ����
TPM2 log entry 14:
	PCR: 4
	Event type: Separator
	Digests:
		 SHA256: ad95131bc0b799c0b1af477fb14fcf26a6a9f76079e48bf090acb7e8367bfd0e
		 SHA1: d9be6524a5f5047db5866813acf3277892a7a30a
	Event data: 4 ����
TPM2 log entry 15:
	PCR: 5
	Event type: Separator
	Digests:
		 SHA256: ad95131bc0b799c0b1af477fb14fcf26a6a9f76079e48bf090acb7e8367bfd0e
		 SHA1: d9be6524a5f5047db5866813acf3277892a7a30a
	Event data: 4 ����
TPM2 log entry 16:
	PCR: 6
	Event type: Separator
	Digests:
		 SHA256: ad95131bc0b799c0b1af477fb14fcf26a6a9f76079e48bf090acb7e8367bfd0e
		 SHA1: d9be6524a5f5047db5866813acf3277892a7a30a
	Event data: 4 ����
TPM2 log entry 17:
	PCR: 7
	Event type: Separator
	Digests:
		 SHA256: ad95131bc0b799c0b1af477fb14fcf26a6a9f76079e48bf090acb7e8367bfd0e
		 SHA1: d9be6524a5f5047db5866813acf3277892a7a30a
	Event data: 4 ����
```

But PCR 5,6,7 are also touched.
Houston, we still have collision https://osresearch.net/Keys/#tpm-pcrs

### DIFF
--- a/initrd/bin/seal-totp
+++ b/initrd/bin/seal-totp
@@ -30,7 +30,7 @@ dd \
 
 secret="`base32 < $TOTP_SECRET`"
 pcrf="/tmp/secret/pcrf.bin"
-DEBUG "Sealing TOTP with actual state of PCR0-4)"
+DEBUG "Sealing TOTP with actual state of PCR0-4"
 tpmr pcrread 0 "$pcrf"
 tpmr pcrread -a 1 "$pcrf"
 tpmr pcrread -a 2 "$pcrf"
@@ -38,7 +38,7 @@ tpmr pcrread -a 3 "$pcrf"
 DEBUG "Sealing TOTP with actual state of PCR4 (Going to recovery shell extends PCR4)"
 # pcr 4 is expected to either: 
 #  zero on bare coreboot+linuxboot on x86 (boot mode: init)
-#  already extended on ppc64 per BOOTKERNEL (skiboot) which boots heads.
+#  already extended on ppc64 per BOOTKERNEL (skiboot) and seperators on PPC64.
 #We expect the PCR4 to be in the right state at unattended unseal operation
 tpmr pcrread -a 4 "$pcrf"
 # pcr 5 (kernel modules loaded) is not measured at sealing/unsealing of totp


tlaurion:
@JonathonHall-Purism @SergiiDmytruk should it be more clear than that in code?

tlaurion:
@SergiiDmytruk :As of now, seperators are extending PCR4,5,6,7. There is no possibility as of now that Heads can seal a disk unlock key into PCR7.  I thought the conclusions were that seperators + bootkernel would be in PCR4, or seperators removed from measurements (what are they anyway) and in last resort have Heads change its sealing/unsealing scheme.

JonathonHall-Purism:
I think the source comments are sufficiently clear for the current behavior.

Nitpicking the actual behavior, IMO it would make the most sense to seal with "the value present at boot" rather than "the value now" (difference being sealing after entering recovery shell).  I don't think it really makes sense to seal with the post-recovery-shell value even if the recovery shell has been entered when sealing.

That said, on PPC64 you'd have to actually grab the value at boot and remember it somewhere, so the behavior now is a reasonable compromise for simplicity IMO.

SergiiDmytruk:
https://github.com/Dasharo/dasharo-issues/issues/415#issuecomment-1563016440:
> https://trustedcomputinggroup.org/wp-content/uploads/TCG_PCClient_PFP_r1p05_v23_pub.pdf shows in table 1 what should be measured to which PCR. Heads depends on coreboot not being compliant with this specification so I'd say that this script should be modified.

In `skiboot` sources:
```
	/*
	 * As defined in the TCG Platform Firmware PWe are done. Extending the digest of 0xFFFFFFFF
	 * in PCR[0-7], and recording an EV_SEPARATOR event in
	 * event log as defined in the TCG Platform Firmware Profile
	 * specification, Revision 00.21
	 */
```

Separators are just hash value with all bits set, it's applied to PCR0-PCR7 to mark the end of boot process.  My conclusion from that Dasharo issue was that Heads shouldn't assume state of the PCRs on boot as that makes it too fragile and non-compliant.   And I took your changes for PCR4 as an agreement with that.  Dropping separators was my idea, but with your changes it's not really necessary.

> There is no possibility as of now that Heads can seal a disk unlock key into PCR7.

I don't see how a different initial value affects sealing.  TPM doesn't care how many times a PCR was extended, only about it having the same value on unsealing.

tlaurion:
Above is true in case of firmwarestate sealing values which happen through seal-totp/unseal-totp (reused by HOTP) for early boot validation. As we know those value are not affected by runtime, those PCRs cannot be reset, just extended. Those are PCR0-7, used by coreboot, skiboot and Heads.

Coreboot, still today, in case of measured boot without ACM, only touches PCR2.
skiboot fires, extends PCR0-7 as @SergiiDmytruk showed above.
Then Heads fires, where current assumptions are invalid.

for totp-seal/totp-unseal, this is not much problematic.
Only problem, fixed on master, is that we cannot take for granted that PCR4 will be 0. Now we take its extended value on sealing. Good.

When comes the time to prepare the sealing of the optional LUKS disk unlock key in TPM, measurements of more PCRs are taken into consideration and current assumptions are simply invalid.
- PCR4 still expected to be 0 (we can take boot value here, no prob)
- PCR5 expected to be 0 if no usb keyboard needed early. This is problematic, since on PPC64, we know its extended and then depending of previous operations, we need to consider its boot value. We could seal with boot value, not so problematic.
- PCR6 is not populated but on LUKS disk unlock key loading LUKS header. On sealing, we futurecalcpcr into PCR16 only the header. But that won't fly, since PCR6 is also extended by skiboot. 
- PCR6 is expected to contain value of cbfs on sealing. Not a problem.

So we have a real problem for PCR6 for the moment on PPC64 and a lesser problem with PCR5.
I will try to play around those, but if one of you has insights on either modifying tpmr and/or kexec-seal-key, let me know.

SergiiDmytruk:
@tlaurion Do I get it right that the problem with PCR6 is to precompute its value?  With a zero value  `pcrreset` helps to make `PCR16 = PCR6` (followed by an extension to PCR16 and reading from it), but non-zero PCR6 value can't be copied to make TPM compute the final value.

Why use TPM for that at all?  Read current PCR value, append hash of a file then hash the result to produce future value of a PCR (which is `future PCR value = hash(<current PCR value bytes><input hash bytes>)`).

JonathonHall-Purism:
Agree with @SergiiDmytruk, we should just calculate the PCR value ourselves without using the TPM.  I'd suggest adding a new futurecalc operation to tpmr that would take the initial state (in this case, the state captured from boot), data to "extend" it with just like the normal extend, and output the new state.  tpmr is a good place for this since the algorithm will differ between TPM1/TPM2.

Then if you capture all PCR states at boot, and futurecalc from those states (or just use as-is to seal against the boot state), it should work for any initial state.

macpijan:
@tlaurion Do we have any conclusions here? Can we proceed with the release on current code, or do you need us to implement some changes? Please let's agree on what's needed and we will push that forward.

tlaurion:
> @tlaurion Do we have any conclusions here? Can we proceed with the release on current code, or do you need us to implement some changes? Please let's agree on what's needed and we will push that forward.

Sorry for the delay. As of now, seal-totp/unseal-totp operations works on both x86 and talos II. The problem is linked to sealing/unsealing disk key, and doesn't work only for talos II.

I thought about this and brought back code from trenchboot for validating pcr values against tcpa log and other sources, while also modifying tpmr code to be able to precalc from 0. Have not yet resolved the issue, not being sure of what is the best way forward into taking boot measurements and then reuse those with precalc to seal disk unlock key. The functions in question are an interaction of kexec-seal-key and qubes-measure-luks, which are called when sealing a disk unlock key when setting a boot default and then when booting that boot default and unsealing the disk unlock key from tpm.

Here is my current patch, implementing the basic changes, but not taking any snapshot of PCRs at early boot, which need to be taken into consideration per kexec-seal-key, not the current PCR values.
```

diff --git a/initrd/bin/kexec-seal-key b/initrd/bin/kexec-seal-key
index e892826e32..3eef8745b3 100755
--- a/initrd/bin/kexec-seal-key
+++ b/initrd/bin/kexec-seal-key
@@ -89,25 +89,26 @@ cat "$KEY_DEVICES" | cut -d\  -f1 | xargs /bin/qubes-measure-luks \
 	|| die "Unable to measure the LUKS headers"
 
 pcrf="/tmp/secret/pcrf.bin"
+#On PPC64, skiboot extends PCR0-7 with seperators, extending digests of 0xFFFFFFFF
+DEBUG "Preparing sealing LUKS disk unlock key with actual state of PCR0-4. Should not be modified at runtime"
 tpmr pcrread 0 "$pcrf"
 tpmr pcrread -a 1 "$pcrf"
 tpmr pcrread -a 2 "$pcrf"
 tpmr pcrread -a 3 "$pcrf"
-# Note that PCR 4 needs to be set with the "normal-boot" path value, which is 0.
-dd if=/dev/zero bs="$(tpmr pcrsize)" count=1 status=none >> "$pcrf"
-if [ "$CONFIG_USB_KEYBOARD" = "y" -o -r /lib/modules/libata.ko -o -x /bin/hotp_verification ]; then
-	DEBUG "Sealing TPM disk unlock key with PCR5 involvement (additional kernel modules are loaded per board config)..."
-	# Here, we take pcr 5 into consideration if modules are expected to be measured+loaded
-	tpmr pcrread -a 5 "$pcrf"
-else
-	DEBUG "Sealing TPM disk unlock key with PCR5=0 (NO additional kernel modules are loaded per board config)..."
-	#no kernel modules are expected to be measured+loaded
-	dd if=/dev/zero bs="$(tpmr pcrsize)" count=1 status=none >> "$pcrf"
-fi
-# Precompute the value for pcr 6
+DEBUG "Preparing sealing TPM disk unlock key with actual state of PCR4 (Going to recovery shell extends PCR4. PPC64 extends with BOOTKERNL digest and seperator)"
+# pcr 4 is expected to either: 
+#  zero on bare coreboot+linuxboot on x86 (boot mode: init)
+#  already extended on ppc64 per BOOTKERNEL (skiboot) and seperators on PPC64.
+#We expect the PCR4 to be in the right state at unattended unseal operation
+tpmr pcrread -a 4 "$pcrf"
+DEBUG "Preparing sealing TPM disk unlock key with PCR5 involvement (additional kernel modules are loaded per board config and seperator on PPC64)..."
+tpmr pcrread -a 5 "$pcrf"
 DEBUG "Precomputing TPM future value for PCR6 sealing/unsealing of TPM disk unlock key..."
-tpmr calcfuturepcr -a "/tmp/luksDump.txt" "$pcrf"
-# We take into consideration user files in cbfs
+#TODO: PCR6 is extended in PPC64. Need to take current value+extend of precalc value for expected PCR6 PPC64 seperator value at unseal
+# NOTE: qubes-measure-luks extends PCR6. Who should do it?
+tpmr pcrread 6 "$pcrf"
+tpmr calcfuturepcr -a -r "/tmp/luksDump.txt" "$pcrf"
+DEBUG "Preparing sealing TPM disk unlock key with PCR7 involvement (user related CBFS files, including config, gpg keyring and trustdb ) and PPC64 seperator value..."
 tpmr pcrread -a 7 "$pcrf"
 
 DO_WITH_DEBUG --mask-position 7 \
diff --git a/initrd/bin/seal-totp b/initrd/bin/seal-totp
index c67a574d91..3db0e1c9c6 100755
--- a/initrd/bin/seal-totp
+++ b/initrd/bin/seal-totp
@@ -30,7 +30,7 @@ dd \
 
 secret="`base32 < $TOTP_SECRET`"
 pcrf="/tmp/secret/pcrf.bin"
-DEBUG "Sealing TOTP with actual state of PCR0-4)"
+DEBUG "Sealing TOTP with actual state of PCR0-4"
 tpmr pcrread 0 "$pcrf"
 tpmr pcrread -a 1 "$pcrf"
 tpmr pcrread -a 2 "$pcrf"
@@ -38,7 +38,7 @@ tpmr pcrread -a 3 "$pcrf"
 DEBUG "Sealing TOTP with actual state of PCR4 (Going to recovery shell extends PCR4)"
 # pcr 4 is expected to either: 
 #  zero on bare coreboot+linuxboot on x86 (boot mode: init)
-#  already extended on ppc64 per BOOTKERNEL (skiboot) which boots heads.
+#  already extended on ppc64 per BOOTKERNEL (skiboot) and seperators on PPC64.
 #We expect the PCR4 to be in the right state at unattended unseal operation
 tpmr pcrread -a 4 "$pcrf"
 # pcr 5 (kernel modules loaded) is not measured at sealing/unsealing of totp
@@ -46,6 +46,7 @@ DEBUG "Sealing TOTP neglecting PCR5 involvement (Dynamically loaded kernel modul
 # pcr 6 (drive luks header) is not measured at sealing/unsealing of totp
 DEBUG "Sealing TOTP without PCR6 involvement (LUKS header consistency is not firmware integrity attestation related)"
 # pcr 7 is containing measurements of user injected stuff in cbfs
+DEBUG "Sealing TOTP with actual state of PCR7 (user injected config.user override, gnupg keyring and trustdb)"
 tpmr pcrread -a 7 "$pcrf"
 tpmr seal "$TOTP_SECRET" "$TPM_NVRAM_SPACE" 0,1,2,3,4,7 "$pcrf" 312 "" "$TPM_PASSWORD" \
 	|| die "Unable to write sealed secret to NVRAM"
diff --git a/initrd/bin/tpmr b/initrd/bin/tpmr
index c392d306aa..1d120fba68 100755
--- a/initrd/bin/tpmr
+++ b/initrd/bin/tpmr
@@ -98,6 +98,11 @@ tpm2_calcfuturepcr() {
 		shift
 	fi
 
+	if [ "$1" = "-r" ]; then
+                RESET=y
+                shift
+        fi
+
 	input_file="$1"
 	output_file="$2"
 
@@ -105,26 +110,47 @@ tpm2_calcfuturepcr() {
 		true >"$output_file"
 	fi
 
-	tpm2 pcrreset -Q 23
+	if [ -n "$RESET" ]; then
+                DO_WITH_DEBUG tpm2 pcrreset -Q 23
+        fi
+
 	DO_WITH_DEBUG tpmr extend -ix 23 -if "$input_file"
 	DO_WITH_DEBUG tpm2 pcrread -Q -o >(cat >>"$output_file") sha256:23
-	tpm2 pcrreset -Q 23
+
+	if [ -n "$RESET" ]; then
+                DO_WITH_DEBUG tpm2 pcrreset -Q 23
+        fi
 }
 tpm1_calcfuturepcr() {
 	TRACE "Under /bin/tpmr:tpm1_calcfuturepcr"
+	#We expect append(-a) and reset(-r) to be passed before input output
 	if [ "$1" = "-a" ]; then
 		APPEND=y
 		shift
 	fi
 
+	if [ "$1" = "-r" ]; then
+                RESET=y
+                shift
+        fi
+
 	input_file="$1"
 	output_file="$2"
 
+	if [ -n "$RESET" ]; then
+		DO_WITH_DEBUG tpm pcrreset -ix 16
+	fi
+
 	if [ -z "$APPEND" ]; then
 		true >"$output_file"
 	fi
 
+	
 	DO_WITH_DEBUG tpm calcfuturepcr -ix 16 -if "$input_file" | hex2bin >>"$output_file"
+
+	if [ -n "$RESET" ]; then
+                DO_WITH_DEBUG tpm pcrreset -ix 16
+        fi
 }
 
 tpm2_extend() {
diff --git a/initrd/etc/functions b/initrd/etc/functions
index 9bb0e665b5..627dea780c 100755
--- a/initrd/etc/functions
+++ b/initrd/etc/functions
@@ -484,6 +484,73 @@ detect_boot_device()
 	return 1
 }
 
+#TPM extend operations simulators
+#GET TCPA logs through cbmem -L, eg for PCR2
+# cbmem -L | grep -A 4 -e "PCR: 2" -e "PCR-2"| grep SHA1 | cut -d ' ' -f3
+#borrowed from https://github.com/TrenchBoot/secure-kernel-loader/blob/master/util.sh
+validate_and_escape_hash () {
+	local TRIM=`echo -n "$1" | sed -r -e "s/ .*//"`
+	if (( ${#TRIM} != 64 && ${#TRIM} != 40 )); then
+		>&2 echo "\"$TRIM\" is not a valid SHA1/SHA256 hash"
+		return
+	fi
+	echo -n $TRIM | sed -r -e "s/([a-f0-9]{2})/\\\x\1/g"
+}
+# extend_sha* HASH file.bin - extend PCR with value=HASH with file.bin's hash
+# extend_sha* HASH1 HASH2 - extend PCR with value=HASH1 with HASH2
+# extend_sha* HASH file file...
+# extend_sha* HASH HASH HASH... - as long as file is not first
+extend_sha1 () {
+	local HASH1
+	local HASH2
+	case $# in
+	[01] )	>&2 echo "extend_sha1 called with not enough arguments provided"
+		return
+		;;
+	2 )	if [ -f "$2" ]; then
+			HASH1="$1"
+			HASH2=`dd if="$2" 2>/dev/null | sha1sum`
+		else
+			HASH1="$1"
+			HASH2="$2"
+		fi
+		;;
+	* )	HASH1=$(extend_sha1 "$1" "$2")
+		shift 2
+		extend_sha1 "$HASH1" $@
+		return
+		;;
+	esac
+	local HASH1_ESC=$(validate_and_escape_hash "$HASH1")
+	local HASH2_ESC=$(validate_and_escape_hash "$HASH2")
+	printf "%b" $HASH1_ESC $HASH2_ESC | sha1sum | sed "s/-/SHA1/"
+}
+extend_sha256 () {
+	local HASH1
+	local HASH2
+	case $# in
+	[01] )	>&2 echo "extend_sha256 called with not enough arguments provided"
+		return
+		;;
+	2 )	if [ -f "$2" ]; then
+			HASH1="$1"
+			HASH2=`dd if="$2" 2>/dev/null | sha256sum`
+		else
+			HASH1="$1"
+			HASH2="$2"
+		fi
+		;;
+	* )	HASH1=$(extend_sha256 "$1" "$2")
+		shift 2
+		extend_sha256 "$HASH1" $@
+		return
+		;;
+	esac
+	local HASH1_ESC=$(validate_and_escape_hash "$HASH1")
+	local HASH2_ESC=$(validate_and_escape_hash "$HASH2")
+	printf "%b" $HASH1_ESC $HASH2_ESC | sha256sum | sed "s/-/SHA256/"
+}
+
 calc()
 { 
 	awk "BEGIN { print "$*" }";
```

tlaurion:
The problem is deeper then just for TPM PCR4.
The problem lies into being able to replay TCPA logs from `cbmem -L`.

@krystian-hebel : here is the bases I was using prior of our discussion yersterday https://github.com/tlaurion/heads/tree/pcr4_better_description_of_change_staging

It seems that helpers from /etc/functions (borrowed from your code) might need to be extended to replace calls currently done under `kexec-seal-key`. 

If I understand the problem correctly:
- We need to store clean boot PCR values (/tmp/secret/boot_pcrs would be a good place, wiped on recovery shell access and on boot)
- We need to have consistent boot measurements as when happening on clean boot when a user attempts to seal a disk unlock key into TPM when setting a new boot default and optionally injecting a LUKS key into header
  - At that point, the user could have done other things, including signing /boot content. Doing so would load kernel modules which would modify PCR5
  - At that point, the user could have went to the recovery shell, extending PCR4.
  - At that point, something else in the bootchain (skiboot, something else) could have extended other PCRs. We cannot take for granted the current PCR values are valid to futurecalc PCR6 with measurement of the LUKS header. We need to be able to replicate clean boot PCR6 value from TCPA log
  - Injecting new user settings into CBFS, measured on boot into PCR7 can be neglicted, since flashing firmware requires a reboot and therefore a resealing.
-  Consequently of the above, it seems that the best approach is to have helpers that would do the same job done per tpmr pcrread operations, but doing so from the measurements being replayed from TCPA logs, and recreate the pcrf file that will be the content of what is sealed into NVRAM with normal tpmr seal operations. Doing so would guarantee that the unseal operations done on clean boot would succeed, without taking for granted that PCRs were not extended prior of sealing a disk unlock key into TPM.

@JonathonHall-Purism Makes sense? Or you had a different approach? Asking since from my experiences in the past days, it doesn't seem possible to simply inject a final expected value into PCR16(tpm1)/23(tpm2) for futurecalc operations. So the easy way expected (dump PCR values early on boot, and simply reuse them as part as sealing op) is a no go. It seeams that we will need to bypass the TPM completely to prepare such expected PCR values, which means we have to either measure the same contents or replay TCPA log externally, through the helpers added through extend_sha256/extend_sha1 under /etc/functions (needing bash).

@macpijan Any challenge on that plan welcome. @krystian-hebel already on it.

For the sake of this PR, I think I will go forward and clean it a bit and merge it, simply so that the DEBUG statements will match what will need to be updated under https://osresearch.net/Keys/#tpm-pcrs to also match what is extended on PPC64(more compliant then coreboot when it comes to TCG)

krystian-hebel:
@tlaurion are init scripts executed by the same shell (ash?) as recovery shell? In other words, if I have something that works in recovery shell, will it behave the same way when executed by those scripts? This environment is somewhat limited, but I'd like to avoid depending on bash if possible.

I've got something that "works", but is too easy to break and is hardcoded in a way that would break x86 at the moment. I'll do some cleaning and add argument checks and it should be ready for review. Here's an example of use:

```
~ # cat /sys/class/tpm/tpm0/pcrs 
PCR-00: 3A 3F 78 0F 11 A4 B4 99 69 FC AA 80 CD 6E 39 57 C3 3B 22 75 
PCR-01: 3A 3F 78 0F 11 A4 B4 99 69 FC AA 80 CD 6E 39 57 C3 3B 22 75 
PCR-02: 53 BE 34 44 24 2A 1C 68 36 06 33 BC DB 26 AB 94 08 16 93 08 
PCR-03: 34 E7 DC C1 16 9B CC 88 ED B0 DD 12 D9 89 A3 FC 9D AD B1 0B 
PCR-04: 89 E7 C1 5B FF 4C C4 B8 DF EB 3C C4 03 FB A4 4B DD 9E 96 F0 
PCR-05: AE BE C1 47 70 40 2B 1D AB E6 B6 84 5E C3 12 50 53 35 9B 79 
PCR-06: 3A 3F 78 0F 11 A4 B4 99 69 FC AA 80 CD 6E 39 57 C3 3B 22 75 
PCR-07: 3A 3F 78 0F 11 A4 B4 99 69 FC AA 80 CD 6E 39 57 C3 3B 22 75 
PCR-08: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
PCR-09: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
PCR-10: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
PCR-11: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
PCR-12: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
PCR-13: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
PCR-14: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
PCR-15: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
PCR-16: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
PCR-17: FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF 
PCR-18: FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF 
PCR-19: FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF 
PCR-20: FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF 
PCR-21: FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF 
PCR-22: FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF 
PCR-23: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
~ # for i in 0 1 2 3 4 5 6 7; do replay_sha1_pcr $i; done
3a3f780f11a4b49969fcaa80cd6e3957c33b2275
3a3f780f11a4b49969fcaa80cd6e3957c33b2275
53be3444242a1c68360633bcdb26ab9408169308
34e7dcc1169bcc88edb0dd12d989a3fc9dadb10b
aac64b69780ac926bf53c5d842389d10c49afdb3
3a3f780f11a4b49969fcaa80cd6e3957c33b2275
3a3f780f11a4b49969fcaa80cd6e3957c33b2275
3a3f780f11a4b49969fcaa80cd6e3957c33b2275
~ # 
```

As you can see, PCRs 4 and 5 have different values. I confirmed that PCR4 is valid (`extend_sha1 aac64b69780ac926bf53c5d842389d10c49afdb3 $(echo -n recovery | sha1sum | cut -d " " -f 1)` gives proper value), haven't checked PCR5 but I don't see why it would be different.

tlaurion:
Discussion happened off channel to clarify above statements.

kexec-seal-key kexec-unseal-key and qubes-measure-luks need to be adapted to not take into consideration pcrs at zeroed in precalc and simulations of pcrread being zeroed

Having logs of extend operations with value extended would be needed so users can replay as well now and in the future.

tlaurion:
Replaced by #1428

Issue #1359, opened 2023-03-31
qemu-coreboot-whiptail-tpm2 fails with openssl error


ThePlexus:
## Context of the Build

**1. What board are you trying to build?**
qemu-coreboot-whiptail-tpm2

**2. What repository:branch are you using to build from?**
- [X] Heads:Master
- [ ] Other (please specify)

**3. What version of coreboot are you trying to build**
- [X] 4.13
- [ ] 4.14
- [ ] 4.15
- [ ] 4.17
- [ ] Other (please specify)

**4. In building the rom where did you get the blobs?**
- [X] No blobs required
- [ ] Provided by the company that installed Heads on the device
- [ ] Extracted from a backup rom taken from this device
- [ ] Extracted from another backup rom taken from another device (please identify the board model)
- [ ] Extracted from the online bios using the automated tools provided in Heads
- [ ] I don't know

**5. If using the automated tools to get the blobs did you run the relevant scripts in the blobs directory**

**6. What operating system are you using**

debian 11 x64 / mint LMDE 5

## Please describe the problem

This board fails to build with an OpenSSL output error
```
$ apt install -y build-essential zlib1g-dev uuid-dev libdigest-sha-perl libelf-dev bc bzip2 bison flex git gnupg gawk iasl m4 nasm patch python python2 python3 wget gnat cpio ccache pkg-config cmake libusb-1.0-0-dev autoconf texinfo ncurses-dev doxygen graphviz udev libudev1 libudev-dev automake libtool rsync innoextract sudo libssl-dev device-tree-compiler u-boot-tools sharutils e2fsprogs parted curl unzip
--SNIP
0 upgraded, 0 newly installed, 0 to remove and 34 not upgraded.
```
```
$ make BOARD=qemu-coreboot-whiptail-tpm2
---SNIP
2023-03-31 15:14:00+01:00 MAKE openssl
tail /home/user/heads/build/x86/log/openssl.log
-----
	apps/openssl-bin-openssl.o apps/openssl-bin-passwd.o \
	apps/openssl-bin-pkcs12.o apps/openssl-bin-pkcs7.o \
	apps/openssl-bin-pkcs8.o apps/openssl-bin-pkey.o \
	apps/openssl-bin-pkeyparam.o apps/openssl-bin-pkeyutl.o \
	apps/openssl-bin-prime.o apps/openssl-bin-progs.o \
	apps/openssl-bin-rand.o apps/openssl-bin-rehash.o \
[...]

Issue #1233, opened 2022-11-01
Future improvements for flash.sh and flash content optimizations


tlaurion:
As noted by @JonathonHall-Purism under https://github.com/osresearch/heads/pull/1230#issuecomment-1291068570

>Additional improvements - the only suggestion I have would be to potentially just flash modified regions when changing settings - i.e. just flash the CBFS region when that's all that's needed, to avoid reading and verifying the rest of flash. Or even better, to move config, GPG keyring, etc., to their own region, so we could separate them from the payload too. There's no need to tie that to this work though

tlaurion:
Also this from OP of #1230

Next steps:

Review changes tested under 76d7ee5 for later work (after upstream flashrom project fixes its flashrom --progress reporting under https://ticket.coreboot.org/issues/390)
Have https://github.com/osresearch/heads/blob/master/patches/flashrom-b1f858f65b2abd276542650d8cb9e382da258967/0100-enable-kgpe-d16.patch upstreamed at least under https://github.com/Dasharo/flashrom/ if not upstream

PR #1417, opened 2023-06-13
add Nitropad NV41/NS50 TPM2 boards / NK3 support

daringer:
This PR introduces two new boards: `nitropad-nv41` + `nitropad-ns50` with TPM2 and overall Nitrokey 3 support (`hotp-verification`). Re-based on the current `master` there is no need for `gpg` patches anymore. 

The current state is that both boards work as expected, tpm2 works nicely, hotp-verification works using Nitrokey 3 (or others). LUKS + TPM is disabled through config, because our tests have not shown a robust behavior here.

Changes:

* kernel 6.1 with patches ported (expect winterfell)
* `.npf` handling (a zip + hash file) during update
* `hotp-verification` with Nitrokey 3 support
* added `CONFIG_GPG_DEFAULT_ALGO` w/o it being set, gpg defaults to create RSA-3072 keys, setting `p256` leads to NIST-P256 keys during key generation on `oem-factory-reset`
* adds modules: `iotools` and `nitrokey-blobs`


tlaurion:
I guess I will revisit this reply with information as I go and we can split this up in smaller discussions as well! 
Will reply inline.

> Changes/Commits which cannot be merged as-is:
> 
>     * [86427bf](https://github.com/osresearch/heads/commit/86427bf8b3315c2cb8fa60465b7908137f805830) changes oem-factory reset to use p256 keys instead of rsa3072. I would suggest to make both variant available through a configuration setting (inside the board) defaulting to rsa3072. Any proposal for the to-be-used config name?
> 
>     * [b428b5b](https://github.com/osresearch/heads/commit/b428b5bcf29c461026281795873dcc08a172ff6e) is an evil one. If the `-V` (verbose flag for flashrom) is not omitted, the oem-factory-reset takes ~2hours as the progress parser seems to fail. This seems to be a flashrom 1.3 issue. What should we do with that? We could introduce a config variable to change the commandline parameters for flashrom here. We have been debugging this, but with no luck yet.

This one is problematic and discussible under https://github.com/osresearch/heads/issues/1233
The short version is that we currently process textual output of -V to produce a progress output ourselves, since depending of hardware, spi read/write/speed, the end user is otherwise provided with an empty prompt, and reading/writing on some platforms actually take forever. The last thing we would want is a user doing a poweroff. So we are stuck waiting that https://ticket.coreboot.org/issues/390 is fixed and flashrom provides a `--progress` output of their own. Quickly checked the commit you picked, but haven't checked upstream for progress on this since forever. Maybe there is commit merged in and a switch to turn that on. If you can check and tell that there is progress bar that we can all use and test, I'm more then willing to test and push for testing for regression on all boards. Let me know maybe under https://github.com/osresearch/heads/issues/1233. When we implemented the changes for progress output under Heads, we knew that most of the processing time was in drawing in the console the progress bar, but we had to do it either way. You can track this under #1233.

 
> Notable other changes, which might need discussion:
> 
>     * [60a9034](https://github.com/osresearch/heads/commit/60a9034aad27190d9e543988a40bafc328d139d1) flashrom bump to 1.3

Same as previously discussed point. Maybe you could check for -V output under /tmp and see if there is some output that could be patched against. There was some output on the dasharo port of flashrom that needed to be removed otherwise it was failing as well. You can check their commit log to see patch commit.

> 
>     * [103d17d](https://github.com/osresearch/heads/commit/103d17d60d530f4438745dd7c1d9dfa404cf660c) kernel 6.1 with patches ported (expect winterfell)
> 
>     * [cf1a217](https://github.com/osresearch/heads/commit/cf1a21745f47a667d10e28d7d475f16c9f75d961) gpg patch backported from https://dev.gnupg.org/rG3c3765405de02b9a57fdc9a3cf901f6e3aca8586 (thus should be save for inclusion)
> 
>     * [dc522d1](https://github.com/osresearch/heads/commit/dc522d115b22b2c7355df493a2c16c0f25ee017f) `.npf` handling (a zip + hash file) during update
> 
>     * [d676bca](https://github.com/osresearch/heads/commit/d676bca7837d2d1d2bf2a9233424043f7c277a4d) `hotp-verification` with Nitrokey 3 support
> 
> 
> A build issue I've found, I can reliably reproduce this error: #1359 ... This happens to me every time (after a fresh checkout) locally and in our CI. To work around this I have to `rm -rf build/x86/openssl-3.8` and re-build to make it work. Could this be a bad dependency ordering? Any ideas?

daringer:
As this looks like this could take some more time - should I pull out the make-the-Nitrokey 3-work into a separate PR ? This would be the `gpg`-patch, `hotp-verification` to v1.4 and a config-var to define the key-generation algorithm ?

saper:
> CONFIG_GPG_DEFAULT_ALGO

We have to remember that current OEM Init/Re-ownership thing completely resets uses' cryptographic token. This would mean one token per device.

Which key needs to be an elliptic curve key? The master public key? The key signing the image?

Maybe it is just enough to have a subkey for those boards, so that the user could maintain their main long term identity keys and one or more device-specific subkeys on the token?

OpenPGP is ugly and allows only 3 keys on the token, but if we have included support for tokens like Nitrokey HSM 2/SmartCard-HSM or other PKCS#11 tokens, it would be easier to manage multiple different keys and key types.

saper:
> * [cf1a217](https://github.com/osresearch/heads/commit/cf1a21745f47a667d10e28d7d475f16c9f75d961) gpg patch backported from https://dev.gnupg.org/rG3c3765405de02b9a57fdc9a3cf901f6e3aca8586 (thus should be save for inclusion)

https://dev.gnupg.org/rGf808012ac2cf67ec563da178d963f300a7f2564d should already be included in https://github.com/osresearch/heads/pull/1350 - can you try if 2.4.0 works for you?

daringer:
> Which key needs to be an elliptic curve key? The master public key? The key signing the image?

Inside `oem-factory-reset` the generation of a new key is currently hard-coded to be rsa3072, the change is simply changing this one to p256. It was also just one algorithm for all keys and this PR does this in a similar fashion, just changing the algorithm. Nevertheless having `CONFIG_GPG_DEFAULT_ALGO` available would already be a first step towards a more flexible approach here.

> OpenPGP is ugly and allows only 3 keys on the token, but if we have included support for tokens like Nitrokey HSM 2/SmartCard-HSM or other PKCS#11 tokens, it would be easier to manage multiple different keys and key types.

Totally second that, the main issue is here that we would also have to replace `gpg` as afaik it does not support PKCS#11 tokens in general.

> https://dev.gnupg.org/rGf808012ac2cf67ec563da178d963f300a7f2564d should already be included in https://github.com/osresearch/heads/pull/1350 - can you try if 2.4.0 works for you?

Didn't test within HEADS, but locally it works as expected

saper:
> the change is simply changing this one to p256

So this is from the "any other changes" category, not required to this board at all? Maybe those should go into separate PRs?

Regarding the use of the elliptic curves in general:

- yes, this is cool, for example I would like to have a possibility to WRITE DOWN the private key on paper and have it on paper "HSM" only. It is possible with EC, since they are small.

However:
- I don't see people using NIST curves with GPG often (which may or may not be a bad thing)
- Most people prefer ed25519 but this is problematic with GPG since there is an ongoing war of standards how to do it in OpenPGP, also there are some pitfalls with it, like having to specify "eddsa" flag with a private key or else it tries ECDSA(!)
- Most importantly, this would mean we would use EDDSA. I am not sure how reliable is random number generation under heads, it might be hard to guarantee the uniqueness of the _k_ semi-private key. libgcrypt [seems to have a "rfc6979" flag](https://www.gnupg.org/documentation/manuals/gcrypt/Cryptographic-Functions.html#Cryptographic-Functions) but I have no idea whether GnuPG uses it/can use it and how.

tlaurion:
@daringer splitting would definitely help things merged faster

tlaurion:
@daringer #1423 is a better approach for flashrom.

tlaurion:
@daringer #1422 might work for you for gnupg?

daringer:
rebased on current master so got rid of any flashrom & gpg fixes - now ready for review & merge ...

initrd/bin/flash-gui.sh
```diff
@@ -27,14 +27,35 @@ while true; do
...
             ROM=$FILE
           fi
 
+          # is a .npf provided?
+          if [ -z "${ROM##*.npf}" ]; then
+            # unzip to /tmp/verified_rom
+            mkdir /tmp/verified_rom
+            unzip $ROM -d /tmp/verified_rom
+            # check file integrity
+						if (cd /tmp/verified_rom/ && sha256sum -cs /tmp/verified_rom/sha256sum.txt) ; then
```
tlaurion, line 44: Indent

initrd/bin/flash-gui.sh
```diff
@@ -53,7 +74,13 @@ while true; do
             whiptail --title 'ROM Flashed Successfully' \
               --msgbox "${ROM#"/media/"}\n\nhas been flashed successfully.\n\nPress Enter to reboot\n" 0 80
             umount /media
-            /bin/reboot
+            
+						
+    				if [ "${CONFIG_BOARD%_*}" = nitropad-nv41 || "${CONFIG_BOARD%_*}" = nitropad-ns51 ]; then
+							/bin/nitropad-shutdown.sh
```
tlaurion, line 80: Indent

config/linux-nitropad-x.config
```diff
@@ -0,0 +1,336 @@
+CONFIG_LOCALVERSION="-heads"
```
tlaurion, line 1: Please use helpers to save in oldefconfig and compare with other boards to tune accordingly if needed

https://github.com/osresearch/heads/blob/3747d58510c7dde30a75cef4aba7977c4cbbfa7b/modules/linux#L233

config/coreboot-nitropad-ns50.config
```diff
@@ -0,0 +1,44 @@
+CONFIG_LOCALVERSION="v1.6.0"
```
tlaurion, line 1: Please use coreboot helper equivalent of linux module to save in oldconfig format. Same for other board if course

tlaurion:
@daringer initial review!

tlaurion:
Your hotp commit would benefit  lillte more detail or a link to changelog?

modules/linux
```diff
@@ -31,6 +31,9 @@ linux_hash := a6fbd4ee903c128367892c2393ee0d9657b6ed3ea90016d4dc6f1f6da20b2330
 else ifeq "$(CONFIG_LINUX_VERSION)" "5.10.5"
 linux_version := 5.10.5
 linux_hash := 3991a9e16a187d78d5f414d89236ae5d7f404a69e60c4c42a9d262ee19612ef4
+else ifeq "$(CONFIG_LINUX_VERSION)" "6.1"
```
tlaurion, line 34: Would be nice if you could reuse version 6.1.2 as of purism otherwise time of builds on circleci will explode for no good reason. Make sure to save your Linux config in oldefconfig format.

https://github.com/osresearch/heads/pull/1419/files#diff-C44C192F63AAB4E550C64696EA703787R34

daringer: ok, can do, this means we'll wait at least until #1419 is upstreamed, then I'll rebase and adapt on top of that ?

daringer: or should I just pull in the 6.1.8 kernel changes and make this PR standalone + 6.1.8-based ?

tlaurion: @daringer you can adapt your code and test on top of that kernel version is what I meant. The commit related to your kernel version bump for the kernel module, it same then Purism PR, will be easy to drop from a dynamic rebase 

`git rebase -i commit_prior_of_that_one` would do and fasten the process.
#1419 is being tweaked with last bits prior of merge, so if you tested it against your own branch, that will be next version of Heads. Most important steps for Nitrokey here is being aware of what Restricted Boot/Basic Boot is and test if their newer fbwhiptail with AVX works as well for you where no regression were observed on x230. Next step there will be to move away of i915drmfb in place of simplefb so we have an easy portability to other platforms in the future not having to hack our way around and tainting the kernel and being a step ahead in legacy boot deprecation that is in the air which would not affect us.

TLDR: please test #1419 @daringer :) If your board depend on that kernel version and that PR works, that will be merged soon.

tlaurion:
Note that coreboot fork affected by this as all coreboot cross buildstack for past version
https://review.coreboot.org/c/coreboot/+/76399

tlaurion:
@daringer #1419 and #1428 merged.

if linuxconfigs saved in oldconfig in tree based on earlier linux version that in this PR, I would be ok merging once this PR is rebased on current master.

config/linux-nitropad-x.config
daringer: done

initrd/bin/flash-gui.sh
daringer: done

config/coreboot-nitropad-ns50.config
daringer: done

modules/linux
daringer: rebased on master, thus using 6.1.8 now

daringer:
okay, thanks for all the inputs, the most recent push includes:
* rebased on top of `master`
* kernel used is now 6.1.8
* coreboot consolidated to one version called `nitrokey` (including a fix for the acpica-unix2 download location)
* fixed indentation fails, ooopsy
* created defconfigs for linux-config and coreboot-configs
* updated `hotp-verification` commit-msg to reflect the changes in version 1.4
* added both nitropads to `.circleci/config.yaml` - there are good chances that this is not correct, only very limited circleci-experience here

tested successfully on Nitropad NV41 + Nitropad NS50

initrd/bin/flash-gui.sh
daringer: done

tlaurion:
> * created defconfigs for linux-config and coreboot-configs

We are changing from defconfig to oldconfig as [said under this comment](https://github.com/osresearch/heads/pull/1417#issuecomment-1634497052)

Please do:
```

make BOARD=nitropad-ns50 coreboot.save_in_oldconfig_format_in_place
make BOARD=nitropad-ns50 linux.save_in_oldconfig_format_in_place
make BOARD=nitropad-nv41 coreboot.save_in_oldconfig_format_in_place
make BOARD=nitropad-nv41 linux.save_in_oldconfig_format_in_place
```

And push result as new commit to pass configs in oldconfig.

initrd/bin/flash-gui.sh
```diff
@@ -59,7 +80,12 @@ while true; do
             whiptail --title 'ROM Flashed Successfully' \
               --msgbox "${ROM#"/media/"}\n\nhas been flashed successfully.\n\nPress Enter to reboot\n" 0 80
             umount /media
-            /bin/reboot
+            
+            if [ "${CONFIG_BOARD%_*}" = nitropad-nv41 || "${CONFIG_BOARD%_*}" = nitropad-ns51 ]; then
+              /bin/nitropad-shutdown.sh
+            else
+              /bin/reboot
+            fi
```
tlaurion, line 88: Hmmmm. Why don't you change /bin/reboot instead? I would prefer to look there instead of having ever growing conditional statements in the code calling reboot for different platforms. You might also look into /bin/poweroff and put this condition after tpm shutdown call?

.circleci/config.yml
```diff
@@ -231,6 +231,14 @@ workflows:
           requires:
             - x230-hotp-maximized
 
+      # Coreboot Nitropads
+      - build_and_persist:
+          name: nitropad-nv41
+          target: nitropad-nv41
+          subcommand: ""
+          requires:
+            - prep_env
```
tlaurion, line 240: Please depend on librem_14 instead. 

Logic here is that we try to only build musl-cross and tools built with musl-cross once and pass workspaces to other boards sharing the most things in common for the same architecture (x86). 

That way, we don't use CircleCI to rebuild musl-cross-make nor modules that were already built (cache is used to rebuild any changes if needed) and only builds coreboot and linux.


By depending on librem_14, you will share the same cache for linux with librems, which now share the same linux kernel version.

tlaurion: @daringer I'm testing your config. For some reason, your fork is not built by circleci. I pushed exact for build test at https://app.circleci.com/pipelines/github/tlaurion/heads/1831/workflows/eab77bd8-f556-4927-ab4d-18e2488f8fd6

Hold on, you might have done the right thing. If save cache works, your approach might speed the builds, but I doubt save_cache can merge multiple directories that are shared across archoitectures (it should fail, we will see). Directories need to be unique, which is why save_cache only had one x86 and one ppc64 cache to merge.

initrd/bin/oem-factory-reset
```diff
@@ -101,24 +101,52 @@ gpg_key_reset()
...
+          echo 1 # RSA
+          echo ${RSA_KEY_LENGTH} #Authentication key size set to RSA_KEY_LENGTH
+          echo ${ADMIN_PIN_DEF}
+      } | gpg --command-fd=0 --status-fd=1 --pinentry-mode=loopback --card-edit \
+          > /tmp/gpg_card_edit_output 2>&1
+      if [ $? -ne 0 ]; then
+          ERROR=`cat /tmp/gpg_card_edit_output`
+          whiptail_error_die "Setting key attributed to RSA ${RSA_KEY_LENGTH} bits in USB security dongle failed."
+      fi
+
```
tlaurion, line 149: unneeded space

initrd/bin/oem-factory-reset
```diff
@@ -101,24 +101,52 @@ gpg_key_reset()
...
-        echo ${ADMIN_PIN_DEF}
-    } | gpg --command-fd=0 --status-fd=1 --pinentry-mode=loopback --card-edit \
-        > /tmp/gpg_card_edit_output 2>&1
-    if [ $? -ne 0 ]; then
-        ERROR=`cat /tmp/gpg_card_edit_output`
-        whiptail_error_die "Setting key attributed to RSA ${RSA_KEY_LENGTH} bits in USB security dongle failed."
+
+    # use p256 for key generation if requested
+    if [ "$CONFIG_GPG_DEFAULT_ALGO" = "p256" ];then
+
```
tlaurion, line 107: unneeded space

initrd/bin/oem-factory-reset
```diff
@@ -101,24 +101,52 @@ gpg_key_reset()
...
+      } | gpg --command-fd=0 --status-fd=1 --pinentry-mode=loopback --card-edit --expert \
+          > /tmp/gpg_card_edit_output 2>&1
+      if [ $? -ne 0 ]; then
+          ERROR=`cat /tmp/gpg_card_edit_output`
+          whiptail_error_die "Setting key to NIST-P256 in USB security dongle failed."
+      fi
+
+    # fallback to RSA key generation by default
+    else
+
```
tlaurion, line 129: unneeded space

tlaurion (commented):
As seen in previous CircleCI buiuld, it fails. making nv depend on librem_14 and having save_cache depend on nv board should work

initrd/bin/oem-factory-reset
daringer: done

initrd/bin/oem-factory-reset
daringer: done

initrd/bin/oem-factory-reset
daringer: done

.circleci/config.yml
daringer: updated based on https://github.com/osresearch/heads/pull/1417#pullrequestreview-1544417462

initrd/bin/flash-gui.sh
daringer: clearly a simpler approach, done

daringer:
* created oldconfigs for all `linux` and `coreboot` configs
* removed some unneeded empty lines
* move `nitropad-shutdown.sh` away from `oem-factory-reset` and into `bin/reboot` and `bin/poweroff`
* updated `.circleci/config.yaml` - is this as expected now? 

Further there is `create_npf.sh` directly within the root dir, is it ok to be there, or is there a better place?

Not tested in detail yet, testing planned for tomorrow (wednesday).

nestire:
tested it on a NS50:
oem-factroy-reset -> works
flashing -> works

tlaurion (changes requested):
@daringer:
Please rebase on master (there are conflicting changes needed under Makefile, circleci etc)

CircleCI:
Make sure save_cache only has only one board per arch cached, and it shoud leb one of the added one here.
Have your boards depend on librem_14, which shares same kernel version as your now and should speed up builds
Please recgister an account on CirlceCI and have it watch and build your branches on github

I stopped review at the point there was conflicts and CircleCI builds were not confirming absence of regressions for other board builds.

daringer:
I am a little confused: at the bottom of this PR it says: "This branch has no conflicts with the base branch", I am also pretty sure and just checked that this PR is based on 06b1b0948d911ab1ee15d08cc4776e41889349cd which is the current master...

In CircleCI the boards also depend on librem_14 ... will look into registering an account at cirecleci....

.circleci/config.yml
```diff
@@ -245,8 +253,8 @@ workflows:
           requires:
             - librem_14
```
tlaurion, line 254: @daringer remove this line. 

One cache per architecture. You are replacing x86 cache that was made from librems by yours.

As of now, this is otherwise failing

tlaurion (changes requested):

.circleci/config.yml
tlaurion: I'm trying to fix the Makefile for coreboot-git to be shared across forks... Not so easy

tlaurion:
@daringer #1462 finally resolves our problems, and under https://github.com/tlaurion/heads/tree/staging_nitrokey_git I pushed what this PR should look like so, reusing #1462 toolcain logic, for which nv41/ns50 builds locally against 4.19 toolchain. So We should be able to piggyback nv41/ns50 to x230-hotp-maximized under CircleCI cache, maining we will reuse musl-cross-make build there and also 4.19 coreboot's buildstack

tlaurion:
Finally! Two testing CircleCI pipelines were successful building without cache and with cache:
- nv41/ns50 coreboot fork, depending on 4.19 coreboot buildstack: https://app.circleci.com/pipelines/github/tlaurion/heads?branch=staging_nitrokey_git3
- nv41/ns50 coreboot fork, depending on its own coreboot's fork buildstack: https://app.circleci.com/pipelines/github/tlaurion/heads?branch=staging_nitrokey_git4

I will update  https://github.com/tlaurion/heads/tree/staging_nitrokey_git to use clevo_release buildstack (second approach above) where there will be need of improving CircleCI/coreboot module's targets to optimize rebuilds when only scripts change. As of now, a full build when modules changes takes nearly 2 hours, where a rebuild from cache takes 50 minutes.

Also to be noted that CircleCI has some instabilities where restoring workspaces and saving workspaces are sometimes failing in the past days, so wil keep an eye on that: it requires CircleCI builds to be restarted from failed step to succeed.

tlaurion:
Superseded by #1469. Please reopen if choice is to rebase this PR and cherry-pick my commit on top of it so that present commits here are not co-authored as under #1469. Closing meanwhile.

tlaurion:
Commits refreshed, reopening!

tlaurion:
Commits refreshed, reopening!

tlaurion:
Reopen

tlaurion:
Commits cleaned, reopening!

tlaurion:
Can't reopen

tlaurion:
Reopen?

tlaurion:
Reopening

tlaurion:
reopen

tlaurion:
reopen and comment doesn't work :(

nestire:
tested ci image on nv41 and ns50:
- oem-factory-reset works
- Booting Qubes 4.1.2 and Ubuntu 22.04 works
- NK3 and NK Pro 2 works

tlaurion:
Thanks for Nlnet for supporting this part of the work https://nlnet.nl/project/HEADS-TPM2.0/#ack

PR #1419, opened 2023-06-21
Upstream PureBoot 27

JonathonHall-Purism:
This is the state of PureBoot 27, except for the branding and build scripts.  This includes all the major features we have added - Basic mode, Restricted Boot, automatic USB boot for basic mode, blob jail, root file hashing, Mini automatic power-on, and improvements to config-gui.sh to control those features.

There are now configs for skipping the pause after the QR code, and to show the "use defaults?" prompt for OEM reset, both are enabled for all Librem boards

I reworked the Heads/PureBoot branding to use CONFIG_BRAND_NAME, so I can carry the PureBoot branding change downstream easily.


JonathonHall-Purism:
There's a patch in here removing the pause after the QR code prompt, I'll revert that for upstream (thanks @tlaurion for finding this)

JonathonHall-Purism:
There are now configs for skipping the pause after the QR code, and to show the "use defaults?" prompt for OEM reset, both are enabled for all Librem boards

tlaurion:
@JonathonHall-Purism first full functional review!

Mainly about the Change configuration settings menu:

- Not working/doing nothing under normal Heads and Qubes 4.2 BRTFS luks install (no readable rootfs)
  - No error given here, just returns to caller menu option
    - "Change the root device for hashing" 
    - "Change the root directories to hash"
    - "Check root hashes at boot"
- Enable Restricted Boot
  - Successfully deactivates recovery shell access
  -  USB BOOT of dd'ed image failed as expected (dd'ed disk here. Message could be better bit good enough)
  - USB boot of Qubes 4.1 detached signed iso was permitted as expected
  - Successfully does not permit unsafe boot
- Size of menus added are not autoexpending to correct width (width fixated? Observed under Disable Restrictive Boot Menu prompt)
- Disable Restrictive Boot stays it will reset TPM. But does it or it wipes TOTP sealed secret only?

- Basic boot mode (no tampering detection. Menaing no TPM involvement)
  - Tamper detection is disabled as expected
  - TPM disk unlock key is also disabled. 
    - If desired, then prompt at setting default boot option should not permit setting a TPM disk unlock key on default boot.
  - Enabling usb autoboot
    - Attemtping to usb autoboot Tails iso (multiple isos on USB thumb drive here)
      - On reboot, I see scanning for "USB storage" prior of fbwhiptail prompt, but i'm showed the Basic boot menu.
        - It seems that when there is multiple ISOs on thumb drive, usb-autoboot fails. 
          - Is that a desired behavior or the list of ISOs should be proposed instead (without validation. Is that why this is skipped? ISO is coupled with notion of detached signature validation and is dropped altogether?) 
    - Attempting to usb autoboot from dd'ed Ubuntu successful.
      -  Seems like the ISO prompts vs USB might be clarified a bit more here?

Understanding:
- Restricted mode is "hardcore mode/paranoia"
- Normal is what Heads enforces normally
- Basic is bypassing TPM/boot detached hash signature validation. Basically, to be ran as a grub/seabios alternative?
 

FBwhiptail:
- A little bit slower on x230: can see initial drawing from top of screen to bottom. Note that x230 (ivy and sandy) do not have AVX2 CPU extension.

initrd/bin/seal-hotpkey
```diff
@@ -73,20 +80,29 @@ else
...
 fi
 
-echo -e ""
-read -s -p "Enter your $HOTPKEY_BRANDING Admin PIN: " admin_pin
-echo -e "\n"
-
-hotp_initialize "$admin_pin" $HOTP_SECRET $counter_value "$HOTPKEY_BRANDING"
+# try using factory default admin PIN
+admin_pin="12345678"
+hotp_initialize "$admin_pin" $HOTP_SECRET $counter_value "$HOTPKEY_BRANDING" >/dev/null 2>&1
```
tlaurion, line 85: Still unconformable with this one.
Could we use `gpg --card-status | grep "Signature counter"` and only have that code run when its under 5 or something (when the USB security dongle was just received?) From my experience, 3 ADMIN PIN attempts is not a lot in the absence of having a Reset code (Reset code can unlock Admin PIN which can unlock User PIN) but as of today, without a private key backup, I had a lot of reports of users locking themselves out. Running this at each HOTP reseal is augmenting the chances a user locks himself out of his USB dongle. 

I would be ok running that code and detecting default Admin PIN being used if the counter signature is low, only.

initrd/etc/functions
```diff
@@ -284,6 +284,19 @@ replace_config() {
...
 }
 
+# Generate secret value using first 20 chars of ROM SHA256 hash
+secret_from_rom_hash() {
+	local ROM_IMAGE="/tmp/coreboot-notpm.rom"
+
+	echo -e "\nTPM not detected; measuring ROM directly\n" 1>&2
+	# use a previously-copied image if it exists
+	if [ -f ${ROM_IMAGE} ]; then
+		sha256sum ${ROM_IMAGE} | cut -f1 -d ' ' | cut -c 1-20 | tr -d '\n'
```
tlaurion, line 360: I thought Nitrokey firmware updated the size of HOTP secret. Why is this still truncated?
Will search for old issue on nitrokey side if that doesn't ring a bell.

initrd/bin/config-gui.sh
```diff
@@ -136,6 +141,97 @@ while true; do
         exit 0
       fi
     ;;
+    "R" )
+      CURRENT_OPTION=`grep 'CONFIG_ROOT_DEV=' /tmp/config | tail -n1 | cut -f2 -d '=' | tr -d '"'`
```
tlaurion, line 190: I think this is where the toggling under Configure Settings menu toggle script fails for non librems since we are `set -e -o pipefail` and the grep fails here?

JonathonHall-Purism:
Thank you @tlaurion !

> @JonathonHall-Purism first full functional review!
> 
> Mainly about the Change configuration settings menu:
> 
>     * Not working/doing nothing under normal Heads and Qubes 4.2 BRTFS luks install (no readable rootfs)
>       
>       * No error given here, just returns to caller menu option
>         
>         * "Change the root device for hashing"
>         * "Change the root directories to hash"
>         * "Check root hashes at boot"

Thanks, will check this out.

>     * Enable Restricted Boot
>       
>       * Successfully deactivates recovery shell access
>       * USB BOOT of dd'ed image failed as expected (dd'ed disk here. Message could be better bit good enough)
>       * USB boot of Qubes 4.1 detached signed iso was permitted as expected
>       * Successfully does not permit unsafe boot

Great, looks like this all works but agree message could be improved, will check the wording.

>     * Size of menus added are not autoexpending to correct width (width fixated? Observed under Disable Restrictive Boot Menu prompt)

Good note, I do not test regular whiptail much, tried to get all the sizes right but I may have missed something, will check.

>     * Disable Restrictive Boot stays it will reset TPM. But does it or it wipes TOTP sealed secret only?

You're right, that was changed, it now only wipes the TOTP secret.  I'll improve the prompt

>     * Basic boot mode (no tampering detection. Menaing no TPM involvement)
>       
>       * Tamper detection is disabled as expected
>       * TPM disk unlock key is also disabled.
>         
>         * If desired, then prompt at setting default boot option should not permit setting a TPM disk unlock key on default boot.

Agree, yes I hadn't noticed that since the PB distribution disables TPM disk unlock key, I'll prevent sealing a key in basic mode.

>       * Enabling usb autoboot
>         
>         * Attemtping to usb autoboot Tails iso (multiple isos on USB thumb drive here)
>           
>           * On reboot, I see scanning for "USB storage" prior of fbwhiptail prompt, but i'm showed the Basic boot menu.
>             
>             * It seems that when there is multiple ISOs on thumb drive, usb-autoboot fails.
>               
>               * Is that a desired behavior or the list of ISOs should be proposed instead (without validation. Is that why this is skipped? ISO is coupled with notion of detached signature validation and is dropped altogether?)
>         * Attempting to usb autoboot from dd'ed Ubuntu successful.
>           
>           * Seems like the ISO prompts vs USB might be clarified a bit more here?

"USB autoboot" was intended for use on headless systems (Librem Minis mainly for an "appliance" use case, e.g. router/NAS/etc.).  That way you can recover the appliance by booting to USB without having to hook up a display/keyboard and with no interaction.  Since it can't prompt anything, it currently only works with plain bootable USB, not ISOs.

Maybe if there was exactly one ISO it could still boot that automatically.  Do you think that's worthwhile?  It can't prompt though, having to prompt defeats the purpose of automatically booting to USB.  So if there's any ambiguity (more than one ISO), just use use the normal boot to USB menus, since you have to select the ISO anyway.

> Understanding:
> 
>     * Restricted mode is "hardcore mode/paranoia"

Yes.  One of the goals here is that booting any unsigned OS creates evidence of tampering (TOTP key is erased, so even disabling RB, then reverting firmware leaves evidence).  It never locks you out of the system (there is no challenge required to disable RB) but it does create evidence if it was disabled.

>     * Normal is what Heads enforces normally
> 
>     * Basic is bypassing TPM/boot detached hash signature validation. Basically, to be ran as a grub/seabios alternative?

Yes exactly.  The power of Linux in firmware without the TOTP/HOTP/GPG/etc. requirements.  It also reduces friction for users to move to the normal Heads flow with TOTP/HOTP - you can switch to Heads in basic mode pretty easily, then once comfortable enable TOTP/HOTP.  Or vice versa, if you decide you don't really need TOTP/HOTP (or temporarily lose your HOTP key, etc.) you can go to basic mode without having to change firmware completely.

> FBwhiptail:
> 
>     * A little bit slower on x230: can see initial drawing from top of screen to bottom. Note that x230 (ivy and sandy) do not have AVX2 CPU extension.

That is what I was worried about - that's the blit from the shadow buffer taking a measurable amount of time.  But if you want to move to simplefb eventually, the shadow buffer method will be a necessity since there is no way to create a GPU backbuffer (no GPU driver, just plain FB).

Sandy and Ivy both have AVX, maybe an AVX copy would improve this enough.  Should be straightforward to put together and we can try it.

initrd/etc/functions
JonathonHall-Purism: Good question.  If it was updated, maybe it is for compatibility with older firmware?  If so, and we can get the firmware version, I could have it only truncate if needed.  If you can find that issue to pinpoint the change let me know, I'll look too.

tlaurion:
> Maybe if there was exactly one ISO it could still boot that automatically. Do you think that's worthwhile? 

Well, maybe in improved version? To me, having verification that flash memory is not corrupted was always a lifesaver, which is the main reason I detach sign my own ISOs that aren't coming with one already. It prevents frustration will increasing a bit friction: It takes a little while to have the ISO verified, but saves hours when installing QubesOS, for example, which otherwise might fail at installing template and then, you have multiple things to blame before knowing what is the cause. I think it would be a nice feature.

initrd/etc/functions
tlaurion: @JonathonHall-Purism https://github.com/osresearch/heads/pull/836#issuecomment-694261654

tlaurion: https://github.com/osresearch/heads/pull/493#pullrequestreview-189586537

tlaurion: @JonathonHall-Purism https://github.com/osresearch/heads/pull/493#issuecomment-487251001

initrd/bin/seal-hotpkey
JonathonHall-Purism: I like the idea of conditionally trying this.  The signature counter idea might work, I think the key there is to smooth out the UX.  If the user neglects to change the PIN (obviously a bad thing), then it'd be even worse if much later they are suddenly asked for a PIN they don't know.  5 signatures later could be a long time after the OEM reset occurred and told them the default PIN.

Maybe we could phase it?
* 0-4 signatures - try default, maybe suggest a bit more strongly to change it (and maybe suggest how since there is no UI for this currently)
* 5-9 signatures - don't try default, but remind user what it was (so at least they get used to entering the default PIN)
* 10+ signatures - don't try default and don't remind, to minimize nuisance reminders in the long run

I wonder if we could also get the number of attempts remaining before making an attempt, to be sure this doesn't consume the last attempt if you had mistyped the pin once or twice and aborted.

initrd/bin/config-gui.sh
JonathonHall-Purism: Yeah you're probably right, good catch! :bow:

initrd/bin/seal-hotpkey
tlaurion: @JonathonHall-Purism 
> I wonder if we could also get the number of attempts remaining before making an attempt, to be sure this doesn't consume the last attempt if you had mistyped the pin once or twice and aborted.

Yes we can! This is `PIN retry counter : 3 3 3`

I love the phasing idea. I would not have any objection into having that upstream, but main concern is really to consume even 1 Admin attempt if the opengpg card was used before to sign other stuff: meaning the user might use the smartcard to encrypt+sign, where we can take assumption that if counter > 5 is definitely used...


Throwing additional idea here: maybe time is a better factor. If the keypair was enerated recently by oem, the key "created" date being recent (check Purism numbers, but is there a delivery that happens after 1 month?)  with a signing counter <=3 should mean we can consume and Admin PIN attempt.

My point here is that it is better sooner then later to have the user reown hist gpg card then later. In the user case a user is actually using his USB Security dongle for something else then detach signing /boot and HOTP, we can think of that user as being one that would run the reownership wizard anyway?

initrd/etc/functions
tlaurion: Seems like it was fixed in firmware in 2019?

tlaurion: > @JonathonHall-Purism [#493 (comment)](https://github.com/osresearch/heads/pull/493#issuecomment-487251001)

Seems like it was fixed in firmware in 2019? https://github.com/Nitrokey/nitrokey-storage-firmware/commit/ec5c0c13076a2064da330ee9bb6206dccaadee71

tlaurion:
@JonathonHall-Purism 
> Sandy and Ivy both have AVX, maybe an AVX copy would improve this enough. Should be straightforward to put together and we can try it.

Tested some more and yeah, this is slow. Moving around in longer menus like Option is slowed down a lot. Weird since AVX and i915drmfb in use here and now.

Can't wait to see this for simplefb

JonathonHall-Purism:
>AVX and i915drmfb in use here and now.

AFAICT the default memcpy is not currently using any CPU extensions.  AVX may be allowed elsewhere but I had to implement a memcpy-ish equivalent using AVX2 (and it is not a perfect replacement due to additional alignment requirements, which probably relates to the fact that memcpy itself is not using these extensions.

The current implementation is fast because it doesn't have to copy from shadow buffer to front buffer at all.  Instead it allocates front and back buffers, and just flips the two (tells the GPU to start displaying the other buffer instead, no copy needed).  But that requires i915 and DRM, which we want to eliminate anyway.

I'm working through all the changes today, will have updates soon :crossed_fingers:

initrd/etc/functions
JonathonHall-Purism: @tlaurion Firmware supports it per your notes, but `hotp` from tpmtotp still requires exactly 20 bytes: https://github.com/osresearch/tpmtotp/blob/master/hotp.c#L26

The secret does not have to be printable, so I can use 20 binary bytes from the hash and get 160 bits of entropy, up from 80 currently (20 hex chars).

tpmtotp can be extended but I feel like this is going down a rabbit hole to go from 160 bits to 256 bits.  Can we track that for an improvement later?

JonathonHall-Purism:
Addressed most of the comments, still to do are the default admin PIN behavior improvements and fbwhiptail AVX copy.  Also will address the merge conflits.

JonathonHall-Purism:
@tlaurion fbwhiptail now uses AVX (not AVX2) for the fast memory copy, this turned out to be a relatively simple change (still 256-bit wide vectors, and performance was the same on Comet Lake so I dropped AVX2 and use AVX everywhere).

Could you try that on Sandy/Ivy Bridge and let me know how performance is?

fbwhiptail MR: https://source.puri.sm/firmware/fbwhiptail/-/merge_requests/3 (would appreciate review of this too)

tlaurion:
Combined both #1419 and #1428 under https://app.circleci.com/pipelines/github/tlaurion/heads for testing.
Will edit here with results.

initrd/bin/seal-hotpkey
JonathonHall-Purism: Got this working, it only tries the default PIN for up to 1 month and only if >= 3 attempts remain (my keys default to 3).  If it doesn't try the default for either reason, it shows a reminder.

I couldn't use `gpg --card-status` to do this as it locked up the keys (https://github.com/Nitrokey/nitrokey-pro-firmware/issues/54) but I was able to do it by getting the key creation time from Heads' copy of the public key, and getting the PIN counter from `hotp_verification info`.

tlaurion:
@JonathonHall-Purism can you debase on master and test with qemu and setup a  TPM unsealed disk encryption key?

tlaurion:
As of now I see no regression. Will need to test ownership a bit more thoroughly.

initrd/bin/seal-hotpkey
```diff
@@ -70,20 +82,63 @@ else
 	HOTPKEY_BRANDING="HOTP USB Security Dongle"
 fi
 
-echo -e ""
-read -s -p "Enter your $HOTPKEY_BRANDING Admin PIN: " admin_pin
-echo -e "\n"
+# Truncate the secret if it is longer than the maximum HOTP secret
+truncate_max_bytes 20 "$HOTP_SECRET"
```
tlaurion, line 86: Why are we still truncating secret?

JonathonHall-Purism: @tlaurion See comments above, will tag you again there

initrd/etc/functions
JonathonHall-Purism: @tlaurion :point_up: Thread on secret truncation

initrd/bin/config-gui.sh
```diff
@@ -141,18 +172,305 @@ while true; do
...
+    ;;
+    "L" )
+      if [ "$CONFIG_RESTRICTED_BOOT" != "y" ]; then
+        if (whiptail --title 'Enable Restricted Boot Mode?' \
+             --yesno "This will disable booting from any unsigned files,
+                    \nincluding kernels that have not yet been signed,
+                    \n.isos without signatures, raw USB disks,
+                    \nand will disable failsafe boot mode.
+                    \n\nThis will also disable the recovery console.
+                    \n\nDo you want to proceed?" 0 80) then
```
tlaurion, line 326: @JonathonHall-Purism an important feature of Restricted boot is not advertised here: with restricted mode enabled, a user won't be able to update firmware unless restricted boot is disabled, as can be seen from flash-gui.sh when an upgrade is attempted. This warning should also be given in the config gui prompt for enablement.

I'm impressed with the effectiveness of restricted boot here and i'm questioning myself on next steps (issues to be opened for implementation):
- Should restricted boot be configured as default under Heads?
- Should theUSB boot option for ISO not having a detached signature be prompted to be detached signed and verified with user's keyring?

JonathonHall-Purism: >This warning should also be given in the config gui prompt for enablement.

Agree :+1:

This is already a somewhat long/wordy prompt, but I think we can capture all of this still.

For comparison, existing wording:
```
This will disable booting from any unsigned files,
including kernels that have not yet been signed,
isos without signatures, raw USB disks,
and will disable failsafe boot mode.

This will also disable the recovery console.

Do you want to proceed?
```

How about:
```
Restricted Boot allows booting:
* Signed installed OS
* Signed ISOs from USB

All other boot methods are blocked.  Recovery console and firmware updates will
be blocked.

Restricted boot can be disabled at any time.  This resets TOTP/HOTP so it is
evident that Restricted Boot was disabled.

Do you want to proceed?
```

It's obviously not short but I think it's more accessible:
* most important information first
* avoids long/complex sentences
* adds that you can still turn it off in case you are unsure this is what you want

I waiver about including the "resets TOTP/HOTP" bit, that's perhaps a bit more information than is really needed, but OTOH there is a really obvious question about "what good is it if anyone can disable it" and that's the answer, disabling it creates evidence.

JonathonHall-Purism: >Should restricted boot be configured as default under Heads?

In principle this is easy to do just by setting `export CONFIG_RESTRICTED_BOOT=y` in a board config if you want a board to have it enabled by default.  Should still work to configure it from the GUI since that overrides board config.

I also offer some "preconfigured" ROMs (where I've just injected a config.user ahead of time) for specific configurations like basic, Mini basic + auto power-on + USB autoboot (headless config), blob jail, etc.  This works really well IMO, coreboot_util.sh offers a choice of configs if you are switching to PB from SeaBIOS.  I don't offer them on PB upgrades because by default it preserves your existing settings, but you actually can flash the preconfigures with the "erasing settings" method and it works.

>Should the USB boot option for ISO not having a detached signature be prompted to be detached signed and verified with user's keyring?

I do think there's a use case here.  IMO, we should think about workflows so we are not bombarding with too many serial questions of "do you want to X", "looks like you are trying to Y".

I think it'd be a good solution to have a separate "signing GUI" to sign an ISO with your key, which would fulfill this.  It'd also be great to be able to configure which OS keys are trusted with RB; maybe you would prefer to only allow booting ISOs you yourself have signed.

tlaurion: I'm an advocate of bullet points detailing what is gonna be lost and active. :/
I like your suggestion, I say let's go with it for simplicity I guess. As you said numerous times, if its too long people don't read. It's sad, but this is what it is.

tlaurion: >I think it'd be a good solution to have a separate "signing GUI" to sign an ISO with your key, which would fulfill this. It'd also be great to be able to configure which OS keys are trusted with RB; maybe you would prefer to only allow booting ISOs you yourself have signed.

Yes. This implies that users will have verified hashes externally, or have the GUI to offer the possibility to generate sha256sum hashes prior of detach signing in case hashes were not already verified on thumb drive. I won't insist enough here, but outside of hashes, so many users are using cheap dongles, and people tend to verify hashes on computer, not on thumb drive. So signing an unverified iso hash is kind of dumb.

Please open issue for tracking on this one. I think it will be beneficial to all and easy to do.

JonathonHall-Purism: I think this is pretty reasonable.  I took out the double spacing as it made the message monstrously tall in fbwhiptail, and the output is already sort of double-spaced.

Reworded the "disable" prompt similarly, let me know if you have any comments on that.

Screenshot_20230711_155809

Screenshot_20230711_155822

tlaurion:
Testing roms from merged #1419 #1428 board images at https://app.circleci.com/pipelines/github/tlaurion/heads/1819/workflows/683cf547-010b-4dea-bb5f-3420931203ea


Tested on qemu boards:
  - [x] Sign /boot
  - [x] Define default boot + setup TPM sealing/unseal of disk unlock key
  - [x] Default boot unseals TPM disk unlock key

- [x] qemu-coreboot-whiptail-tpm1
- [x] qemu-coreboot-whiptail-tpm2
- [x] qemu-coreboot-fbwhiptail-tpm1-hotp
- [x] qemu-coreboot-fbwhiptail-tpm2-hotp
- [x] Talos-2
  - [x] Factory reset (tpm reset, nitrokey pro smartcard gpg in key keygen, flash back key material, default boot partition setup)
  - [x] default boot boots into dedian-11 with TPM released disk unlock key passed through secret.cpio 
- [x] x230-maximized

tlaurion:
> @tlaurion fbwhiptail now uses AVX (not AVX2) for the fast memory copy, this turned out to be a relatively simple change (still 256-bit wide vectors, and performance was the same on Comet Lake so I dropped AVX2 and use AVX everywhere).
> 
> Could you try that on Sandy/Ivy Bridge and let me know how performance is?

So much faster!

> fbwhiptail MR: https://source.puri.sm/firmware/fbwhiptail/-/merge_requests/3 (would appreciate review of this too)

Nothing to report there as per off-channel discussions. From my eyes, changes are proper and gains are amazing.
Will take more time to test https://source.puri.sm/firmware/fbwhiptail/-/commits/simplefb_shadow/ which will have a really big impact for older boards in term of kernel reduction and generalization of kernel and coreboot configs, at least on intel based iGPU, where otherwise whiptail enough is not so worse either.

tlaurion:
And its a merge!

srgrint:
Just confirming I have flashed heads-x220-maximized-v0.2.0-1705-gedf200e.rom.  Seems to work fine, although not extensively tested yet)

srgrint:
Also have tested heads-t430-maximized-v0.2.0-1705-gedf200e.rom on my t430 which also seems to work fine.

PR #1350, opened 2023-03-24
update GnuPG and friends to 2.4.0

saper:
gnupg 2.2.21 -> 2.4.0
libgcrypt 1.8.6 -> 1.10.1
libassuan 2.5.3 -> 2.5.5
libksba 1.4.0 -> 1.6.3
libgpg-error 1.46

PR #1420, opened 2023-06-23
dasharo flashrom: remove dummy and mtd for x86, put ast1100 as optional 

tlaurion:
all x86 boards now only builds and pack internal programmer.

kgpe-d16 boards add ast1100 in their board config.

Talos II only builds and pack mtd programmer.

----

This is to see size reduction when combined with #1381 and #1350


tlaurion:
TODO: 

- [ ] create flashrom-dasharo module. 
- [ ] have kgpe-d16 boards rely on it

PR #1424, opened 2023-06-27
Makefile: have sizes report output on screen and into sizes.txt

tlaurion:
This adds an easy way to compare sizes of packed stripped built binaries and libs in cpios, compressed under initrd and the kernel bzImage and coreboot rom, both on screen and into sizes.txt.

This is basically a copy paste + `stat -c "%8s:%n"` under Makefile & modules/linux, reusing HASHES code but duplicating for SIZES.

Note that the sizes.txt can be sorted for human eyes through a simple `cat sizes.txt | sort -h`.

I would love to have this merged now so that master can be compared to branches that are rebased upon it.

---

@JonathonHall-Purism : can you review+approve asap please?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,11 +70,13 @@ PACKAGES	:= $(shell mkdir -p "$(packages)")
 
 # record the build date / git hashes and other files here
 HASHES		:= $(board_build)/hashes.txt
+SIZES		:= $(board_build)/sizes.txt
 
 # Create the board output directory if it doesn't already exist
 BOARD_LOG	:= $(shell \
 	mkdir -p "$(board_build)" ; \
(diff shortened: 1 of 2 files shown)


JonathonHall-Purism:
Looks really handy, let's merge if CI passes and the output looks right :100:

JonathonHall-Purism:
I like the right aligned sizes to make the big stuff pop out, let's merge it :100:

tlaurion:
modules/linux needed change to also report on bzImage

tlaurion:
Finally a clean compable point. Will merge so that we have a master comparison point, this is building cache on my own CircleCI instance otherwise.

PR #1425, opened 2023-06-27
CircleCI: have sizes output kept in CircleCI forever just as hashes

tlaurion:
CircleCI keeps steps output forever as opposed to artifacts.
This permits us to keep sizes.txt output as a step for a board forever, just like for hashes.

@JonathonHall-Purism


JonathonHall-Purism:
Love it, then just need to make a dashboard plotting it over time :joy:  Let's merge this

PR #1426, opened 2023-06-27
kexec-parse-boot,kexec-boot: Concatenate multiple initrds

JonathonHall-Purism:
If a GRUB entry has more than one initrd, concatenate them.

Previously, kexec-parse-boot created multiple |initrd...| segments and kexec-boot would pass multiple --initrd arguments to kexec.  Kexec takes the last one and ignores the others.

Instead, concatenate all the initrds when more than one is given.  A single initrd is still used directly without a temporary file.

Fixes NixOS boot with encrypted root filesystem, which adds a second "secrets" initrd after the ordinary initrd. (#1348)

### DIFF
--- a/initrd/bin/kexec-boot
+++ b/initrd/bin/kexec-boot
@@ -66,12 +66,20 @@ if [ "$CONFIG_DEBUG_OUTPUT" = "y" ];then
 	kexeccmd="$kexeccmd -d"
 fi
 
+cut_first_rest() {
+	# The space after $1 in "$1 " avoids a special-case in cut that passes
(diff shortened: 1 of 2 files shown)


tlaurion:
No regression observed for Q4.1 boot, which ads another --module and passes cryptab pointing to key for OS to use when unlocking disk.

LGTM!

tlaurion:
@JonathonHall-Purism non-likely to be ever merged since #1348 was discussed upstream which will lead to another installer media to be created for users to reinstall/cryptsetup-reencrypt. Closing

PR #1428, opened 2023-06-30
initrd/bin/tpmr: replay PCR values from event log 

krystian-hebel:
initrd/bin/tpmr: replay PCR values from event log instead of assuming their values.


tlaurion:
hotp/non-hotp boards:
- Test TOTP seal/unseal + HOTP seal/unseal: ok
- Test TPM disk unlock key sealing/unsealing on default boot 

usb keyboard enabled board:
- PCR5 consistent on disk unlock key seal/unseal

- [ ] x230-maximized
- [ ] x230-hotp-maximized_usb-kb
- [ ] Talos
- [ ] librems

tlaurion:
At this point, no regression on x86 @krystian-hebel

krystian-hebel:
I've successfully got through https://osresearch.net/Configuring-Keys/#tpm-disk-encryption-key-tpm-disk-unlock-key for Debian 12 on encrypted LVM. As a confirmation that TPM is actually involved, I went to recovery shell (to get PCR4 extended), started `/bin/gui-init` again and attempted default boot - got `!!! Failed to unseal the TPM LUKS disk key`, which I assume is expected outcome.

As such, I think this PR is ready for review.

tlaurion:
Retesting!

tlaurion:
Combined both #1419 and #1428 under https://app.circleci.com/pipelines/github/tlaurion/heads for testing.
Will edit here with results.

@JonathonHall-Purism you wight want to review as well since this impacts all platforms.

tlaurion:
> I've successfully got through https://osresearch.net/Configuring-Keys/#tpm-disk-encryption-key-tpm-disk-unlock-key for Debian 12 on encrypted LVM. As a confirmation that TPM is actually involved, I went to recovery shell (to get PCR4 extended), started `/bin/gui-init` again and attempted default boot - got `!!! Failed to unseal the TPM LUKS disk key`, which I assume is expected outcome.

For testing
- Once TPM disk unlock key is sealed 
  - Typing a bad TPM nvram region passphrase will result in a bad passphrase result
  - Typing with a good TPM nvram passphrase with bad measurements will result in unseal errors
    - Tampering with LUKS header consequently will also result in bad measurements and the same unseal error as above.

Will test against 

- [ ] qemu-coreboot-whiptail-tpm1
- [ ] qemu-coreboot-whiptail-tpm2 locally
- [ ] x230-maximized
- [ ] talos

initrd/bin/tpmr
```diff
@@ -86,45 +100,179 @@ tpm1_pcrread() {
...
+			LINES=8
+			;;
+	* )		>&2 echo "Unknown TPM event log format"
+			return
+			;;
+	esac
+	PCR="${PCR}$1"
+	shift
+	extend_sha1 $sha1_zeros \
+	`echo "$LOG" | grep "$PCR" -A$LINES | grep $ALG | cut -d " " -f 3` \
```
JonathonHall-Purism, line 238: Since this is trying to carry state from line to line, awk would be a better fit than multiple-grep.

The number of 'after' lines is fragile because it must be long enough to ensure that the $ALG line for that entry is found, but not so long that it accidentally picks up $ALG for the _next_ entry (which might be a different PCR).  This seems likely to break in subtle ways if the format changes in coreboot, either due to added or removed lines.

With awk you could set a variable when finding a matching PCR line, clear it when finding a non-matching PCR line, and emit the $ALG match only when we are in a section for the correct PCR.

Also, I suggest `$(...)` over backticks: https://www.shellcheck.net/wiki/SC2006

krystian-hebel: Unfortunately, whole format of log printed by coreboot is fragile, and during upstreaming effort we encountered some resistance from the maintainers that initially didn't seem to understand the importance of following standards. Ideally all formats should have constant, well-documented, machine-parsable format, but we can't do anything about "old" coreboot format, and there still are (and probably will be for some time, because nobody cares) boards that use it.

I'll change this to use awk, for some reason I thought that it wasn't available here.

> Also, I suggest $(...) over backticks

Right, I had it that way originally but changed to backticks to debug missing `"` that was in fact missing `;;` and forgot to change it back.

krystian-hebel:
@JonathonHall-Purism thank for the review, I'll try to change all mentioned issues on next round.

tlaurion:
Tested with https://github.com/osresearch/heads/pull/1434 merged on master for qemu-coreboot-whiptail-tpm1 and works as intended.

initrd/bin/tpmr
```diff
@@ -192,86 +158,61 @@ coreboot TPM log:
...
-get_log_format () {
-	echo "$1" | head -n 1 | cut -d " " -f 1
+# awk script to handle all of the above. Note this gets squashed to one line so
+# semicolons are required.
+AWK_PROG='
+BEGIN {
+	getline;
+	hash_regex="([a-fA-F0-9]{40,})";
+	if ($0 == "TPM2 log:") {
+		RS="\n[^[:space:]]";
```
JonathonHall-Purism, line 168: That's a clever way to treat each log entry as a single record, I will have to remember this :wink: :100:

initrd/bin/tpmr
```diff
@@ -192,86 +158,61 @@ coreboot TPM log:
...
+	}
+}
+$0 ~ pcr {
+	match($0, alg);
+	print gensub(alg, "\\1", "g", substr($0, RSTART, RLENGTH));
 }
+'
 
-# usage: tpmr calcfuturepcr <pcr_num> [ <input_file>|<input_hash> ... ]
+# usage: tpmr calcfuturepcr <alg> <pcr_num> [ <input_file>|<input_hash> ... ]
```
JonathonHall-Purism, line 189: Could you please update this to reflect the current name of the function?

initrd/bin/gui-init
```diff
@@ -174,10 +174,22 @@ update_totp()
...
     TOTP="NO TPM"
   else
     TOTP=`unseal-totp`
+    # On platforms using CONFIG_BOOT_EXTRA_TTYS multiple processes may try to
+    # access TPM at the same time, failing with EBUSY. The order of execution
+    # is unpredictable, so the error may appear on main console, secondary one,
+    # or neither of them if the calls are sufficiently staggered. Try up to
+    # three times (including previous one) with small delays in case of error,
+    # instead of immediately scaring users with "you've been pwned" message.
+    while [ $? -ne 0 -a $tries -lt 2 ]; do
```
JonathonHall-Purism, line 188: I'm OK with this as-is, and I know we have this all over in Heads right, but since I saw it tweaked... :wink: `[ ... ] && [ ... ]` is better than `[ ... -a ... ]`.  `-a` introduces ambiguity if any of the variable arguments happen to be `!` or start with `-`.  The 2- and 3-argument forms of `[` are never ambiguous even if the variables contain stuff like that.

With that said, in this case the variables should never contain that stuff, and we have it all over in Heads already, so this is just a general comment for the future :grin:

JonathonHall-Purism:
This looks great to me @krystian-hebel !  I have not tested yet, will try to carve out a bit of time to test but IMO that doesn't need to block if others have tested.  This is not strongly board-specific and I do comprehensive testing for each PureBoot release.

tlaurion:
@krystian-hebel and I arrived to a confortable conclusion so that introspection would be possible with the following change on current PR

```

diff --git a/initrd/bin/tpmr b/initrd/bin/tpmr
index d075e437..76b0fcd4 100755
--- a/initrd/bin/tpmr
+++ b/initrd/bin/tpmr
@@ -211,9 +211,17 @@ replay_pcr() {
        # SHA-256 hashes are 64 chars
        if [ "$alg" = "sha256" ] ; then alg_digits=64; fi
        shift 2
-       extend_pcr_state $alg $(printf "%.${alg_digits}d" 0) \
+       replayed_pcr=$(extend_pcr_state $alg $(printf "%.${alg_digits}d" 0) \
        $(echo "$log" | awk -v alg=$alg -v pcr=$pcr -f <(echo $AWK_PROG)) \
-       $@ | hex2bin
+       $@)
+       echo $replayed_pcr | hex2bin
+       DEBUG "Replayed cbmem -L clean boot state of PCR=$pcr ALG=$alg : $replayed_pcr"
+       #To manually introspect current PCR values under bash shell:
+       #PCR-2: bash -c "tpmr calcfuturepcr 2"
+       #PCR-4: bash -c "tpmr calcfuturepcr 4 <(echo -n recovery)" #In case of landing in recovery shell
+       #PCR-4: bash -c "tpmr calcfuturepcr 4 <(echo -n generic)"  #In case of normal boot passing through kexec-select-boot
+       #PCR-5: bash -c "tpmr calcfuturepcr 5 module0.ko module1.ko module2.ko" #Depending on which modules are loaded for given board
+       #PCR-6 and PCR-7: similar to 5, but with different files passed (6: luks header, 7: user related cbfs files loaded from cbfs-init)
 }
 
 tpm2_extend() {
```

@JonathonHall-Purism comments?
This permits the following output when debug is activated:



While also permitting to do introspection call manually as the code snippet suggests:
```
~ # bash -c "tpmr calcfuturepcr 2"
TRACE: Under /bin/tpmr
TRACE: Under /bin/tpmr:replay_pcr
3�_�h@�`mq&�g����DEBUG: Replayed cbmem clean boot state of PCR=2 ALG=sha1 : 331dd75f06f96840b4606d7126dc67ebfadeecb6
~ # cat /sys/class/tpm/tpm0/pcrs
PCR-00: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
PCR-01: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
PCR-02: 33 1D D7 5F 06 F9 68 40 B4 60 6D 71 26 DC 67 EB FA DE EC B6 
PCR-03: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
PCR-04: 8A 6A 96 FD E1 A8 DD 96 27 14 79 DC 40 74 2B 36 AB A3 C2 B3 
PCR-05: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
PCR-06: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
PCR-07: A3 50 76 B9 40 EB 58 92 F2 C1 D8 40 04 2E 00 02 5B 86 32 6B 
PCR-08: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
PCR-09: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
PCR-10: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
PCR-11: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
PCR-12: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
PCR-13: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
PCR-14: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
PCR-15: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
PCR-16: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 
PCR-17: FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF 
PCR-18: FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF 
PCR-19: FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF 
PCR-20: FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF 
PCR-21: FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF 
PCR-22: FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF FF 
PCR-23: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
```


I think this is good enough, but note that binary output is shown on console when nothing consumes the output of calcfuturepcr. @JonathonHall-Purism ?

initrd/bin/gui-init
krystian-hebel: Changed. Thanks for explanation, I'm learning a lot today :slightly_smiling_face:

krystian-hebel:
@tlaurion I added slightly modified version of your snippet. Mostly wrapping, but also with a warning about binary output.

I don't know about `clean boot state` in the output. I've left it for now, but whether it's just clean boot or something more depends on invocation. It may be possible to print what "something more" is, especially if those are files and not hashes, but I'm not sure if anyone will be needing it.

JonathonHall-Purism:
@tlaurion @krystian-hebel Re: PCR inspection - the binary output _is_ the PCR value, that's the intent of that command.  So you should inspect a PCR with `tpmr calcfuturepcr 2 | xxd -p`.

So I'd suggest in the debugging comments:
```
	# To manually introspect current PCR values from recovery console:
	# PCR-2:
	#     tpmr calcfuturepcr 2 | xxd -p
	# PCR-4, in case of landing in recovery shell:
	#     tpmr calcfuturepcr 4 <(echo -n recovery) | xxd -p
	# PCR-4, in case of normal boot passing through kexec-select-boot:
	#     tpmr calcfuturepcr 4 <(echo -n generic)" | xxd -p
	# PCR-5, depending on which modules are loaded for given board:
	#     tpmr calcfuturepcr 5 module0.ko module1.ko module2.ko | xxd -p
	# PCR-6 and PCR-7: similar to 5, but with different files passed
	#  (6: luks header, 7: user related cbfs files loaded from cbfs-init)
```

You shouldn't need `bash -c` because the script has a shebang.  (And the redirects for recovery, etc. should work in ash too IIRC.)  Include xxd -p because the output is the PCR value like I said, and you'll be able to see the PCR even if debug output is not enabled.

krystian-hebel:
CI fails due to https://review.coreboot.org/c/coreboot/+/76399, and probably will fail for other PRs too until that change is backported everywhere. @tlaurion CC

tlaurion:
@krystian-hebel #1419 fixes all librems since bumping to a coreboot 4.19 based coreboot fork.

Talos 2 will need to point to a coreboot commit having this fix in.

tlaurion:
> @tlaurion @krystian-hebel Re: PCR inspection - the binary output _is_ the PCR value, that's the intent of that command. So you should inspect a PCR with `tpmr calcfuturepcr 2 | xxd -p`.
> 
> So I'd suggest in the debugging comments:
> 
> ```
> 	# To manually introspect current PCR values from recovery console:
> 	# PCR-2:
> 	#     tpmr calcfuturepcr 2 | xxd -p
> 	# PCR-4, in case of landing in recovery shell:
> 	#     tpmr calcfuturepcr 4 <(echo -n recovery) | xxd -p
> 	# PCR-4, in case of normal boot passing through kexec-select-boot:
> 	#     tpmr calcfuturepcr 4 <(echo -n generic)" | xxd -p
> 	# PCR-5, depending on which modules are loaded for given board:
> 	#     tpmr calcfuturepcr 5 module0.ko module1.ko module2.ko | xxd -p
> 	# PCR-6 and PCR-7: similar to 5, but with different files passed
> 	#  (6: luks header, 7: user related cbfs files loaded from cbfs-init)
> ```
> 
> You shouldn't need `bash -c` because the script has a shebang. (And the redirects for recovery, etc. should work in ash too IIRC.) Include xxd -p because the output is the PCR value like I said, and you'll be able to see the PCR even if debug output is not enabled.

I like the absence of binary output, but recommendation of not using `bash -c` won't work since redirection is a bashism.

tlaurion:
@krystian-hebel this PR needs to be rebased on master otherwise build fails.

JonathonHall-Purism:
Current state looks great to me, thanks for improving that, and thanks also for clarifying that ash doesn't actually have process redirection like I thought it did :bow:

tlaurion:
Testing roms from #1419 #1428 board images at https://app.circleci.com/pipelines/github/tlaurion/heads/1819/workflows/683cf547-010b-4dea-bb5f-3420931203ea

- [ ] qemu-coreboot-whiptail-tpm1
- [ ] qemu-coreboot-whiptail-tpm2
- [ ] qemu-coreboot-fbwhiptail-tpm1-hotp
- [ ] qemu-coreboot-fbwhiptail-tpm2-hotp
- [ ] Talos-2
- [ ] x230-maximized

PR #1434, opened 2023-07-06
Non-Intel linux configs: make sure AES is enabled (CONFIG_CRYPTO_AES_NI_INTEL won't work there)

tlaurion:
kgpe-d16 linux configs: disable CONFIG_CRYPTO_AES_NI_INTEL (not avail on AMD), enable generic AES.
qemu linux config: adding generic AES support, not removing CONFIG_CRYPTO_AES_NI_INTEL (in case someone wants to explore with intel coreboot qemu support later on)

This applied to Q35 qemu board which is AMD, not intel. generic AES needs to be enabled on non-intel boards, otherwise cryptsetup doesn't know how to deal with xts-plain

Test point:
`cryptsetup benchmark`

Otherwise cryptsetup fails call fails.


JonathonHall-Purism:
Whoops, looks fine for merge to me :+1:

PR #1436, opened 2023-07-07
Kexec cosmetic fixes

tlaurion:
So that actual boards and future boards show that they might stay in the dark until OS loads the proper DRM/FB drivers if unsupported by kexec

Linux kernel changed since 4.19.
Since then, buffer addresses are generally hidden and cannot be directly passed to another kernel through kexec anymore.

Workarounds have been applied to linux kernel and kexec patch for i915drmfb, while others are currently unsupported out of the box.

Transition out of this is to have all boards rely on linux kernel's simplefb frambebuffer.
Meanwhile, other boards relying on kexec will show an error like the following that is shown for qemu boards as of today:

The `kexec -l` call outputs an error about virtio_gpudrmfb driver, which cannot expose its fb address to kexec and therefore, the user stays in the dark until the next kernel loads that driver and drives the console again.

@JonathonHall-Purism : please review

### DIFF
--- a/initrd/bin/kexec-boot
+++ b/initrd/bin/kexec-boot
@@ -146,6 +146,17 @@ echo "$kexeccmd"
 eval "$kexeccmd" \
 || die "Failed to load the new kernel"
 
+if [ "$CONFIG_DEBUG_OUTPUT" = "y" ];then
+	#Repeat kexec command that will be executed since in debug
+	DEBUG "kexeccmd= $kexeccmd"
+
+	read -n 1 -p "[DEBUG] Continue booting? [Y/n]: " debug_boot_confirm
+	if [ "${debug_boot_confirm^^}" = N ]; then
+		# abort
+		die "Boot aborted"
+	fi
+fi
+
 if [ "$CONFIG_TPM" = "y" ]; then
 	tpmr kexec_finalize
 fi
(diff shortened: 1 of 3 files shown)


initrd/bin/kexec-boot
```diff
@@ -155,4 +155,8 @@ if [ -x /bin/io386 -a "$CONFIG_FINALIZE_PLATFORM_LOCKING_PRESKYLAKE" = "y" ]; th
 fi
 
 echo "Starting the new kernel"
+if [ "$CONFIG_DEBUG_OUTPUT" = "y" ];then
+        #Repeat kexec command that will be executed since in debug
+        DEBUG "kexeccmd= $kexeccmd"
```
JonathonHall-Purism, line 160: @tlaurion I was just adding this too :joy: Would you mind adding:
```
	read -n 1 -p "[DEBUG] Continue booting? [Y/n]: " debug_boot_confirm
	if [ "$debug_boot_confirm^^" = Y ]; then
		# abort
		die "Boot aborted"
	fi
```

So we could abort the boot in debug for inspection.  Should also move it above the `tpmr kexec_finalize` and preskylake lock then IMO

tlaurion: done

tlaurion:
Adjusted per review @JonathonHall-Purism. Saying n N or enter returns to main menu.

JonathonHall-Purism:
Looks great @tlaurion, and I won't have to comment out the kexec -e to iterate on initrd manipulations in the future! :100:

Issue #1421, opened 2023-06-23
Legacy boards considered to be deprecated


tlaurion:
This is x220/t430/x230 boards mot showing "maximized" on Heads main menu.

How many of you that represents?

merge:
I haven't used (or at least changed) a Heads installation in quite a while now. What exactly is the meaning of the issue? - I'm not entierly sure. Skulls users "usually" would unlock IFD, but I consider making this the default behaviour even - as I think ideally we want Heads to enable WP before booting an OS.

tlaurion:
> Skulls users "usually" would unlock IFD, this is the default behaviour even - as I think ideally we want Heads to enable WP before booting an OS.

@merge that is exactly my concern. If ifd is locked by default today, those users can only flash Heads legacy ROMs for t430/x230/x220.

And those legacy boards per this ticket are thought to be dropped from main repository because its extremely hard to pack everything under 7mb of normally defined IFD BIOS region.

So my question to you is to how many users out there from skulls user base are having IFD locked, ME non-neutered/not unlocked which won't be able to internally flash to maximized boards.


As for WP protection, io386 and maximized boards are pushing PR0 locking, aka SMM platform locking through SMI chipset locking really recently.

merge:
no, IFD is *unlocked* by default when installing Skulls. Most users neuter ME, but it's not yet the default. maximize the "bios" region is not yet done by Skulls, but iiuic it could be done without disrupting the user-workfow. what would it roughly be? coreboot config size-change, new layout-file + ifdtool run on the image? And if the *actual* size of the Skulls image remains <4M, external hardware-flashing is still possible to only the one 4M-chip? patches and testing welcome :)

good to hear WP is progressing

bwachter:
I only have boards with neutered ME.

tlaurion:
@merge 
> good to hear WP is progressing

This was merged per https://github.com/osresearch/heads/commit/6ec0c814430c5ad0df54f834c89ee72a904466b2 for maximized boards only as of now.

It is to note that further disabling kernel config network options under linux-x230-legacy.config per https://github.com/osresearch/heads/pull/1381/files#diff-2af02345f256b80007d10fb6a49cfaaf8dd4c2d96f044f9976f8f5f1396b3fb6 delays deprecation for legacy boards (which all use linux-x230-legacy.config 5.10.5 from now on).

[...]

PR #1439, opened 2023-07-11
coreboot 4.11 needs acpica which moved from acpica.org to intel. 

tlaurion:
Download from distfiles.macports.org instead, same hash.

kgpe-d16 and librem-l1um depend on coreboot 4.11 still today in tree, even though building is successful only on debian-10 and those boards are not built on CircleCI anymore (debian-11). 
Fixing so people building 4.11 today are still successful on debian-10.

4.19+ already depends on github.com releases tarballs. 
REF for upstream issue: https://review.coreboot.org/c/coreboot/+/76399

@JonathonHall-Purism please make sure it works, local build taking forever here.

Note: note that none the past coreboot versions in tree outside of 4.19+ will build. from now on.

Ranting begins here: maybe reproducible builds is just a piped dream in current software supply chain ecosystem. Ranting ends here.


JonathonHall-Purism:
@tlaurion I cherry-picked this onto the PureBoot `Release-27` branch so I could build L1UM v1 on Debian 11.  Cleaned entire repo to build with no cache, built successfully.  Let's merge it :100:

tlaurion:
Ok..... Will provide patch for master coreboot and talos-2 coreboot commit....

tlaurion:
#1421 updated but untested kgpe-d16 targets here for survival.

PR #1440, opened 2023-07-11
Use Intel mirror for acpi-unix2 20220331

JonathonHall-Purism:
I built Librem 14 locally to verify the coreboot-4.17 patch.  Copied verbatim to coreboot-talos_2, will check CI build status.


tlaurion:
Note: the decision here was to point to intel.com instead of applying upstream fix, needing to also change naming scheme and hashes, since github releases are not having the same hashes then hosted copies at intel.com which were having the same hashes.

All pending PRs will need to rebased on master after this is merged otherwise CI checks will fail.

#1419 #1439 #1428 most importantly, other PRs under pending review will need to be rebased on master and have boards added under CircleCI config as well.

Issue #1441, opened 2023-07-15
Add `export CONFIG_AUTO_BOOT_TIMEOUT=5` to all board configs


3hhh:
**Is your feature request related to a problem? Please describe.**
heads doesn't start the boot process without user interaction.

**Describe the solution you'd like**
Add an option like `GRUB_TIMEOUT` to execute the default boot after some time, if that's defined and all verifications succeed.

**Additional context**
I'd also like an option to boot after that timeout even if verification fails as that would help with testing.
~~However imho such an option breaks the heads security model as an attacker could simply enable it and then lure the user into auto-booting.~~
EDIT: After giving it some more thought, I believe that such an option should be possible as an attacker can modify the firmware any way s/he wants anyway, The user will still have to manually check the HOTP/TOTP dongle to make sure that doesn't happen.

tlaurion:
@3hhh can you enable restricted boot and test how you you would want to do this?

Recent purism merge add those possibilities, just would need a use case and a matching PR

tlaurion:
Restricted boot stops attacker to get to recovery shell and flash

tlaurion:
Autoboot could be activated but would stop you at TPM disk unlock key?

tlaurion:
Otherwise autoboot would just kexec to OS which prompts the disk decryption password. The whole idea of heads is to type that passphrase in a trusted environment, not the OS.

Waiting for your use case in which default boot should boot securely aftet timeout

3hhh:
> Otherwise autoboot would just kexec to OS which prompts the disk decryption password. The whole idea of heads is to type that passphrase in a trusted environment, not the OS.

Hm I'v never used heads for that. Can I not consider the OS kernel trusted once it was verified?

> Waiting for your use case in which default boot should boot securely aftet timeout

A few use cases:
1. On the machine where I use heads I currently need to manually step to the machine to hit Return on the heads prompt every day as USB is apparently deactivated and my USB keyboard doesn't work. The USB issue is fine for me for security reasons, but it's a bit annoying anyway.
[...]

PR #1442, opened 2023-07-17
Qemu boards: typo correction in comment to manually enable Basic Boot mode

tlaurion:
Was CONFIG_BASIC_BOOT instead of CONFIG_BASIC expected.

@3hhh I would recommend playing with this in the goal of #1441.
@JonathonHall-Purism simple review please.

Also some tweaks seem needed to enable autoboot on basic basic boot: current basic boot enablement thinks autoboot is configured because not negated?


JonathonHall-Purism:
Yes looks right, sorry for overlooking this the first time @tlaurion . Looks good to me

PR #1403, opened 2023-05-11
libgfxinit/nativegfx init: efifb enforced fb (+coreboot ramstage enabled bootsplash)

tlaurion:
EDIT: simplefb support (previous effort) was replaced by efifb for simplicity and possible generalization of usage across FSP GOP/native gfx init/libgfxinit under coreboot/heads usage as a payload under linuxboot (without added compexity of hacking kexec, have coreboot taint kernel on linux command line options ot Heads payload).

- This frees 500kb of firmware space, as detailed in following comments, removing DRM+FB drivers of the kernel.
- It also provides a common base to support GOP enabled graphcis boards, while still untested on my side.
- It also permits to have splashscreen that are smaller then board defined HEIGHT/WIDTH under coreboot config, centering that bootsplash correctly.
- **dGPU fb is now linear, and enforces bootsplash from optionrom fb setuped for 1280x1024** which needs testing

Bonus it works also on qemu target boards (bochs native graphic init + libgfxinit).
Leaving traces below for history.Changes qemu call from -vga virtio to -vga std (to use bochs coreboot native graphic init + fb)

efifb discussions and changes from https://github.com/osresearch/heads/pull/1403#issuecomment-1636219081 and on

Testing needed and **why**:
- [x] x230-legacy/x230-legacy-flash: @tlaurion :  **libgfxinit efifb prepared fb + linux efifb + centered 1024x768 bootsplash (3 second boottime reduction)** fixed from https://github.com/osresearch/heads/pull/1403#issuecomment-1677476841
- [ ] t430-legacy/t430-legacy-flash: @daringer : **libgfxinit efifb prepared fb + linux efifb + centered 1024x768 bootsplash (3 second boottime reduction)** fixed from https://github.com/osresearch/heads/pull/1403#issuecomment-1677476841
- [x] t430-maximized : @daringer : **libgfxinit efifb prepared fb + linux efifb + centered 1024x768 bootsplash (3 second boottime reduction)** fixed from https://github.com/osresearch/heads/pull/1403#issuecomment-1677476841 Thanks to @srgrint for last minute testing which saved t430-maximized and t430-hotp-maximized from going untested prior of merging this PR.

----

- [x] x220 (xx20): @Thrilleratplay @BlackMaria @srgrint : **libgfxinit efifb prepared fb + linux efifb + centered 1024x768 bootsplash (3 second boottime reduction)** Thanks @srgrint for testing!
[...]

PR #1459, opened 2023-08-09
Scale fbwhiptail and console font for high resolution displays

JonathonHall-Purism:
Improve output on high resolution displays.

Mini v1/v2 work significantly better on 3840x2160.  (Previously, the console font was very tiny, and fbwhiptail would mode-switch every time it started and exited.  For example, on my Samsung TV the mode switches take >10 seconds, and it often failed to sync to the new mode entirely.)

Librem 15v4 no longer has to force a 1080p mode on the eDP panel, both console and fbwhiptail are sharper.

---

fbwhiptail detects a scale factor automatically, aiming to approximate logical height 1080 with an 0.5x scale increment.  `FBWHIPTAIL_SCALE` can be set to force a particular scale factor for testing (within limits).  fbwhiptail no longer looks for a 1080p mode when the default is 2160p.  https://source.puri.sm/firmware/fbwhiptail/-/merge_requests/5

Currently, fbwhiptail scaling starts to take effect at height 1350 lines (1.5x, so 900 lines logical height).

The Linux console font size is doubled for height >= 1350 lines.  The only options here are 1x or 2x (I did not add any fonts, just doubled the built-in font at runtime).  This goes to 2x when fbwhiptail starts using 1.5x, which favors "too large" over potentially "too small".

---

@Tonux599 @househead @pcm720 x230 eDP 1440p owners!  How does this change look for you - do you like 1.5x in fbwhiptail or prefer to stay at 1.0x?  Is the Linux console OK at 1.0x?

Scaling based on resolution is just a heuristic, in general it depends on the physical size of the screen, how close you are to it, and a bit personal preference.  We can tweak it if needed.

I enabled console scaling too as I saw some mentions that 1600p panels are possible with mods, but I don't know that anyone here has one.  ~~1440p won't be affected by the console font size change currently.~~  Based on feedback, 1440p does get the 2x font now.   (To try 2x when it doesn't apply by default, do from a recovery console: `setfont -O /tmp/font; setfont -d /tmp/font`.)


pcm720:
Just flashed [774955c](https://github.com/osresearch/heads/pull/1459/commits/774955cc048c7e0b5438c2c21d7f63176c66380a).
fbwhiptail looks pretty nice at 1.5x, maybe a little better than 1.0x. Linux console at 1.0x is a little too small for me, especially when fbwhiptail scaling is set to 1.5. I think that 2x would be much more readable

JonathonHall-Purism:
Thanks for testing @pcm720 !  I reduced the threshold so 1440p displays will get the 2x console font.

tlaurion:
@JonathonHall-Purism needs rebase on master!

Should this be merged before/after https://github.com/osresearch/heads/pull/1403 ?

JonathonHall-Purism:
Thanks, rebased.

Ordering vs. #1403 doesn't matter as long as the modules/fbwhiptail conflict is resolved correctly.  Resolution scaling works for both DRI and linuxfb, and the linuxfb improvements included in this branch won't have any effect if you are still on DRI.

If you think the fbwhiptail changes look good, I will merge the fbwhiptail MR, tag another release, and update this PR

tlaurion:
I like what I see and thanks to @pcm720 we have an input permitting to merge!

PR #1462, opened 2023-08-11
Enable reusing coreboot release toolchains for forks

JonathonHall-Purism:
Forks can indicate what release they are based on to reuse the toolchain build from that release.  Forks no longer attempt to share build directories.

'purism' and 'talos_2' are both based on coreboot 4.20.1 but they have different architectures, so there is little to be shared - but future forks for the same architecture can share.  No boards on regular releases use 4.20.1 yet, but when updated they can share the toolchain with 4.20.1 forks as well.

Also cleaned out coreboot releases/patches that were unused.

---

@krystian-hebel If you incorporate this into the work adding the 'nitrokey' fork, you should be able to add it as another 4.19 fork in CircleCI to share the toolchain build.

@tlaurion I don't think there's anything to change in CircleCI at this point since there is no actual toolchain sharing to occur.  But if you update the Thinkpads to 4.20.1, then we should mark the purism fork as based on 4.20.1 so it will share the toolchain.

---

This works by defining each coreboot version as a separate module instead of defining the 'coreboot' module differently based on the configured version.  Only the needed module(s) (one or two) will become dependencies of the current build - one for the toolchain and one for the ROM, which are the same if not building from a fork.  XGCCPATH is used to direct coreboot toward the release's toolchain.

At the moment no sharing actually occurs - we are building 4.11, 4.19, purism 4.20.1 x86, and talos_2 4.20.1 ppc64, none of which share toolchains.  Toolchain sharing was tested and can be enabled in the relevant branches adding more forks.


JonathonHall-Purism:
`talos-2` is failing, will check it out:

```
make[3]: powerpc64-linux-gnu-gcc: No such file or directory
make[3]: *** [/root/project/build/ppc64/coreboot-talos_2/payloads/external/skiboot/skiboot/Makefile.rules:65: asm/asm-offsets.s] Error 127
make[2]: *** [Makefile:17: /root/project/build/ppc64/coreboot-talos_2/payloads/external/skiboot/build/skiboot.elf] Error 2
make[1]: *** [payloads/external/Makefile.inc:377: payloads/external/skiboot/build/skiboot.elf] Error 2
make[1]: *** Waiting for unfinished jobs....
```

JonathonHall-Purism:
>It should still check the toolchain version, so in the event something was misconfigured the build should break as intended. (Note to self - check this)

Actually it doesn't check the toolchain - I pointed the purism fork to the 4.19 toolchain and it still built.  But the fork commit and toolchain version are right next to each other, so there's not much risk of forgetting to update one.

JonathonHall-Purism:
librem_mini and librem_mini_v2 fail due to the certificate for `https://people.redhat.com/` having expired yesterday (and I invalidated the package cache) - they fetch ioport from here.

x230-maximized-fhd_edp and x230-hotp-maximized-fhd_edp both have `make[1]: *** No rule to make target 'src/mainboard/lenovo/haswell/variants/x230_edp/data.vbt', needed by 'x230-hotp-maximized-fhd_edp/coreboot.pre'.  Stop.`  @tlaurion Know offhand where this is supposed to come from?  I'm guessing there is another failure earlier that didn't break the build right away

JonathonHall-Purism:
The talos-2 build works for me locally, but fails in CI. Might be due to CROSS= which was originally there for historical reasons, but the skiboot payload does check it.

Running CI with CROSS= over lunch, marked WIP until talos-2 is sorted out.

tlaurion:
@JonathonHall-Purism rebase on master! ioport was switched, which is why builds are failing here.

JonathonHall-Purism:
Yup on it thanks, looks like talos-2 works with CROSS= so I will clean that up too, then this is ready for review.

tlaurion:
@JonathonHall-Purism why would talos2 not be able to share toolchain as of x86 boards right now?

JonathonHall-Purism:
> @JonathonHall-Purism why would talos2 not be able to share toolchain as of x86 boards right now if another ppc64 based board was added?

@tlaurion It could, since the `CROSS=` variable fixed building skiboot.  But since there is only one PPC64 build, it will only make the build a little longer/slower to also prepare coreboot 4.20.1 for PPC64, we are better off just building the toolchain from the talos-2 commit.

If we add another ppc64 board based on the talos-2 fork, then we don't need to do anything, they'll share the toolchain since they use the same coreboot fork.

If we add another ppc64 board based on an upstream release or a different fork, then it would be beneficial to prepare a common toolchain.

JonathonHall-Purism:
@tlaurion Ready for review as long as CI passes.  Mini builds should be fixed since it was rebased, x230 eDP was fixed, librem_l1um builds locally (with make 4.2.1), talos-2 worked in CI after adding `CROSS=`.

The Purism fork of coreboot now prepares and uses a 4.20.1 toolchain in anticipation of other boards using it (though none do yet).

JonathonHall-Purism:
On Friday I triggered another pipeline for this branch to test cache reuse: https://app.circleci.com/pipelines/github/JonathonHall-Purism/heads/88/workflows/e4bd226b-4ed2-4587-ae84-477ab86ba925

Looks like it worked!  x230-hotp-maximized reproduced an identical ROM and the toolchains were reused from cache.

tlaurion:
Will do a testing PR basing #1417 on top of this just to make sure everything is kosher. @krystian-hebel you have something against reusing coreboot 4.19 buildstack to build nv41/ns50?

krystian-hebel:
I haven't really worked on nv41/ns50 implementation, but I would say that as long as it builds with 4.19 it should be good. The code seems to be from around that time period so I don't expect any problems.

tlaurion:
> I haven't really worked on nv41/ns50 implementation, but I would say that as long as it builds with 4.19 it should be good. The code seems to be from around that time period so I don't expect any problems.

@krystian-hebel I expect some later on the line; sooner then later, if dasharo is not having an internal util/crossgcc  buildstack that is self-maintained to match upstream. As of now, Heads will benefit reusing 4.19 buildstack, simply because sandy/ivy/haswell is building from upstream 4.19. But what will happen when Heads bump coreboot version to 4.20.1+? Well, Dasharo/coreboot/clevo_release will then be used to build itself, since no other common buildstack will be usable. Not a problem per se, taking for granted that clevo_release includes upstream crossgcc buildstack requirements to build itself. This is expected normally for any coreboot fork and expected to be maintained to build itself. Meaning that a fork's HEAD commit is expected to contain crossgcc changes to have the tip of the branch being able to be built.

I will pinpoint this discussion in PR aimed to merge nv41/ns50 nitropad boards to keep track of this issue.

tlaurion:
So as of now, the "time" cost of passing buildstack on CircleCI is really similar to the cost of building (cpu=time) nv41's own buildstack. There is only a difference of one minute difference and will need to be addressed later on.

One discussed idea would be to have step under CircleCI which would be reponsible to build the different buildstacks we need, and pass it as workspace to all boards which could then be parallelized. It is also thought that it would be possible to do a "combine caches" step prior of save_cache so that the cache contains all build modules caches to be reused if modules don't change.

Otherwise, as of now:
- A clean build takes around 2h without cache
- A build/rebuild with available full cache as with today dependencies with CircleCI caching takes 1h22 if we were to reuse 4.19 buildstack to or 1:40 if we use each forks buildstack. 

Since we want ot be warned early if forks buildstacks  fail, I suggest we use forks buildstacks ofr now and try to implement a better parallelization as described above in parallel in further steps.

I will consequently merge this PR and rebase my branch to be used as a base for #1417.

PR #1463, opened 2023-08-11
ioport: changing url to debian, same hash. First need to go into #1198 direction

tlaurion:
Annoying.

@JonathonHall-Purism will try to reach you to have this approved but otherwise will push without review, works locally and stops builds from succeeding in other PRs [otherwise](https://app.circleci.com/pipelines/github/tlaurion/heads/1901/workflows/ad521f6c-e272-4b13-8b8c-358b9e07f4da/jobs/29234?invite=true#step-103-26502_94):

```
2023-08-10 21:03:53-04:00 WGET https://people.redhat.com/rjones/ioport/files/ioport-1.2.tar.gz
if ! wget -O "/root/project/packages/x86/ioport-1.2.tar.gz.tmp" https://people.redhat.com/rjones/ioport/files/ioport-1.2.tar.gz ; then exit 1 ; fi ; mv "/root/project/packages/x86/ioport-1.2.tar.gz.tmp" "/root/project/packages/x86/ioport-1.2.tar.gz" 
--2023-08-10 21:03:53--  https://people.redhat.com/rjones/ioport/files/ioport-1.2.tar.gz
Resolving people.redhat.com (people.redhat.com)... 209.132.178.26
Connecting to people.redhat.com (people.redhat.com)|209.132.178.26|:443... connected.
ERROR: The certificate of 'people.redhat.com' is not trusted.
ERROR: The certificate of 'people.redhat.com' has expired.
The certificate has expired
```

Issue #1198, opened 2022-08-03
Have all modules point to either github/debian/gentoo/alternatives archives for tarballs


tlaurion:
Heads had several build breakages in the past because repositories go either down, have certificate problems etc.

Pointing to github releases or Debian archives (ref: https://github.com/osresearch/linux-builder/issues/1#issuecomment-1189779961) would mitigate this issue for reproducibility in the future.
Neither of those should have either cert/DNS issues that go unnoticed.

Those may require checksums update, though under modules, as opposed to bouncing packages from mirror to mirrors like we do from time to time. But building an old Heads commit after that change should be possible as long as the recipes from CircleCI are respected to have the same docker image (OS version) and installed packages.

Edit: Need arose at https://github.com/osresearch/heads/pull/1188#issuecomment-1203239597

tlaurion:
Becomes relevant again, see #1439

JonathonHall-Purism:
Maybe we could add locations as secondary mirrors instead of switching everything over to another source?

If we have another place (or more than one) to store archives, we could upload whatever version we are using and have the build system try those sources as secondary mirrors.  Mirrors could be down at any given time (builds only break if there are no reachable sources for a given archive).  Digests don't have to change right away so it has a lower initial impact.

Even if we're not perfect about uploading everything we use, we usually have a copy of the relevant archive when the primary source goes down (and build system checks the digest), we could upload it reactively.

Tricky part there is finding place(s) to put the archives I think.  As a secondary mirror, I might be able to do something on cloud.puri.sm, or maybe source.puri.sm could handle it (does not sound significantly different from anything already done for PureOS).

tlaurion:
Hey @githubisnonfree (Leah), we discussed over libreboot channel a while back the possibility of hosting, and mirroring, tarballs used by Heads. Does the offer still stands?

The easiest way to do this as of today is to simply build Heads locally:
- one server variant (kgpe-d16)
- one hotp variant (x230-hotp-maximized) 

And doing so, cairo/pixman/whiptail/fbwhiptail and all other dependencies would be downloaded locally with their checksums validated.

The result could then be uploaded to what is then mirrored to/from others. And then Heads could find a way either statically/better, dynamically, to point/parse to libreboot mirror list to get new mirror list in case prior sources are failing to download.

Thoughts?

---

Sidenote, does libreboot plan to rebridge to Matrix? Otherwise, how can we foster better collaboration?
[...]

PR #1464, opened 2023-08-11
ioport: changing url to debian https server, same hash. First need to go into #1198 direction

tlaurion:
Sorry for the noise.

Went to fast: https://github.com/osresearch/heads/commit/a276b05a445dc9cec03cd2e88d3e82e30e6a8556 was an http url, not https.

This is why reviewing is important :/

Supersedes #1463, pushing now.


tlaurion:
At least this implements https://github.com/osresearch/heads/issues/1198

PR #1489, opened 2023-09-01
[WIP] Z690 Wifi - Msi ms7d25 ddr5 tpm2

ThePlexus:
This is a WIP. There was some older discussion, with pertinent detail around the way we can may get dTPM supported. Please see #1371 

**status**: 
Serial works, it boots and we see hello world on the monitor, but no TPM as yet (see below)
The board seems to have all regions unlocked for internal programmer.
Not added circleci yet. 

**connections**

- Flash chip is MX25U25635F WSON8, soldered on (though has been seen with a Winbond too, per @miczyg1 )
- External flashing is not amazingly easy, but possible [with a WSON8 clip](https://docs.dasharo.com/variants/msi_z690/recovery)
- External flashing is a bit less fiddly by using the[ SPI bus via the TPM2 header](https://docs.dasharo.com/variants/msi_z690/development/#hardware-connection)
- Dasharo also show how to get [read only ttyS0 console](https://docs.dasharo.com/variants/msi_z690/development/#hardware-connection)
- If you find you have issues with you version of flashrom, please update to a [newer version for W685 chipset compatibility 
](https://github.com/osresearch/heads/pull/1371#discussion_r1165512080)

**ME**
- Cant me_clean and reduce ME firmware on this generation of CPU. just how it is. We can use ME hap bit to disable ME https://github.com/6d6178667269747a/me_cleaner/commit/d6b313ad75ac4ccda3668042d0a8dacd4fef2f0f
- I set HAP the bit in download_BIOS_clean.sh by flipping 0x1DE to 0x11 - validated that setting this bit on factory firmware shows ME version 0.0.0.0 in factory BIOS. 
- As we cant shrink ME, it may be an idea to use VSCC mod by default on this board 
- could we make a patch and have coreboot measure the ME region for changes as part of measured boot?
- the fTPM is not available with HAP set, though thats probably a good thing ;)
-  With HAP set we [should be able to use dTPM module](https://github.com/osresearch/heads/pull/1371#issuecomment-1680218506), with coreboot patches 

**Findings/WIP stuff**

- Switched to using kernet 5.9.17 tied to the x230 maximised config. This is because existing versions do not support the GPU. There are some versions before 5.9 which do support the GPU but require a kernel command switch which taints. 
- [This is interesting](https://mail.coreboot.org/hyperkitty/list/coreboot@coreboot.org/thread/4OS45H7JRVFOC447YYMLCDWUF42X4RHV/) for putting something on the screen during RAM training, which can take a bit of time - uGOP.
- Coreboot VBOOT seems to use the onboard fTPM even with a dTPM attached. The above mentioned error still remains. This is going to take a bit more work. Please see #1371 for possible pathway
-Network at boot is untested and unlikely. dashro extract a binary blob to use in a UEFI payload https://github.com/Dasharo/coreboot/blob/msi_ms7d25/release/build.sh#L16


tlaurion:
@ThePlexus This is sold so should be upstreamed and merged. Newer version of this code somewhere?

ThePlexus:
> @ThePlexus This is sold so should be upstreamed and merged. Newer version of this code somewhere?

Hi @tlaurion - ive not been involved in any development since the code in this PR. New job and new baby have given me no time to work on much of anything not working hours or family time. I was super happy to see the Dasharo tech preview as the folks there and I did speak some time ago about taking this work forward, which is really great to see happening. Im looking forward to trying it out in the near future. I'd recommend reaching out to those folks about any upstreaming

tlaurion:
Abandoned. See #1753

PR #1493, opened 2023-09-05
initrd/bin/flash-gui.sh: Exit instead of errant return

JonathonHall-Purism:
Return is not valid outside of a function - exit instead.


tlaurion:
Nice catch!

PR #1494, opened 2023-09-05
modules/coreboot: Update Purism coreboot to 24e2f7e4

JonathonHall-Purism:
This is 4.21-Purism-1 plus a fix for native graphics init on Mini v1/v2: HDMI1 is enabled so passive DisplayPort to DVI/HDMI adapters will work.

### DIFF
--- a/boards/librem_mini/initrd/bin/set_ec_poweron.sh
+++ b/boards/librem_mini/initrd/bin/set_ec_poweron.sh
@@ -7,10 +7,16 @@
 BRAMADDR=0x360
 BRAMDATA=0x361
 
-outb "$BRAMADDR" 0x29 # Select byte at offset 29h
 if [ "$1" = "y" ]; then
-	outb "$BRAMDATA" 0x00 # 0 -> automatic power-on
+	BRAM_VALUE="0x00" # 0 -> automatic power-on
 else
-	outb "$BRAMDATA" 0x01 # 1 -> stay off
+	BRAM_VALUE="0x01" # 1 -> stay off
 fi
 
+outb "$BRAMADDR" 0x29 # Select byte at offset 29h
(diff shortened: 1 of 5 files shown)


tlaurion:
Can you drop some notes here as example for others to follow so that anything special for boards can be encapsulated in board specific initrd overrides?


Maybe just point to entry point here in PR or link to rpi or commit code explaining how to reuse those concepts later on!


If it builds, looks good to me! Can merge!

JonathonHall-Purism:
In this commit I updated the board-specific initialization for Librem Mini v1/v2 to account for changes in the CMOS layout.

Boards can have an `initrd` directory underneath the board directory.  If they do, any files from that directory are overlaid onto the Heads initrd.  Boards can add scripts (`bin/`), config (`etc/`), or in general any files needed.  (Makefile packs these in `board.cpio` which becomes part of the combined initrd - https://github.com/osresearch/heads/blob/master/Makefile#L618).

If the board provides `bin/board-init.sh`, it's invoked by `init` after initialization is performed and before the regular bootscript starts.  (https://github.com/osresearch/heads/blob/master/initrd/init#L165)  Mini v1/v2 use this to update the CMOS power-on setting to reflect the current Heads setting.

I don't recommend trying to override files from Heads in the board-specific area - only add files, and add hooks in Heads files if needed, like the hook in `init` checking for `board-init.sh`.

Some good future work might be to add a hook to `config-gui.sh` to present board-specific configuration options.  For the Mini automatic power-on feature, the config option is present in `config-gui.sh`, and the board enables it with `CONFIG_SUPPORT_AUTOMATIC_POWERON`.  That wouldn't scale very well if we gain a lot of board-specific options though.  Although on the other hand, other boards might be able to support the automatic power-on setting too, like servers, so it might not be board-specific forever.

PR #1495, opened 2023-09-05
initrd/bin/flash.sh: Improve speed and reliability

JonathonHall-Purism:
Improve speed by pre-filtering only for lines containing any tokens of interest to flashrom_progress_tokenize().

Improve reliability by avoiding dropping tokens that cross a stream buffer boundary.  Occasionally, a token could be missed if it crosses a stream buffer boundary, due to read timing out too quickly before the next buffer is flushed.  If this was a state-changing token, flashrom_progress() would hang forever.

---

Purism's coreboot utility contains a copy of this same code (https://source.puri.sm/firmware/utility/-/blob/master/coreboot_util.sh#L775).  During the PureBoot 28 release cycle, I encountered a situation where coreboot_util.sh would _consistently_ hang due to the "contents..." token always crossing a stream buffer boundary - the script read "c" and then "ontents..." separately and got stuck in the initializing state.

I believe this also happens in Heads/PureBoot.  I encountered a similar hang once when flashing in PureBoot 27.1, although with no way to break out of the script I could not troubleshoot it (that device did not have serial).  I believe it was the same issue.

---

Unfortunately, I could not carry over the best part of this fix (a `grep --line-buffered ''` in the pipeline) since busybox grep lacks `--line-buffered` and I could not find anything similar in busybox.  I think the pre-filtering still makes it less likely to fail, but it might still be possible.

### DIFF
--- a/initrd/bin/flash.sh
+++ b/initrd/bin/flash.sh
@@ -22,12 +22,13 @@ case "$CONFIG_FLASHROM_OPTIONS" in
 esac
 
 flashrom_progress() {
-    # The ichspi programmer now spews register status lines constantly that are brutally slow
-    # to feed through the parser in flashrom_progress_tokenize.  Exclude them.
-    # flashrom_progress_tokenize operates on individual tokens (not lines), so it splits by
-    # spaces in 'read'.  But we also need to separate the last word on a line from the next
-    # line, so replace newlines.
-    grep -v -e '^HSFS:' -e '^HSFC:' | tr '\n' ' ' | flashrom_progress_tokenize "$1"
+    # The ichspi programmer now spews register status lines constantly that are
+    # brutally slow to feed through the parser in flashrom_progress_tokenize.
+    # Filter the input with grep for only lines containing at least one token
+    # that we care about.
+    grep -E -e 'contents\.\.\.' -e 'done\.' -e '0x[0-9a-f]+-(0x[0-9a-f]+):' \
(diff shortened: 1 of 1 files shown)


JonathonHall-Purism:
@tlaurion suggested to me in DM to try using `sed` to get line buffering, so I wrote this test based on the critical parts of `flashrom_progress`: https://gist.github.com/JonathonHall-Purism/4e5ac97088a9b72cb53326901b44a084

The good news is that it does not seem to have any problem with fragmented tokens in Heads.  I suspect busybox grep and sed are line buffering even when stdout is not a TTY, although I did not dig into the source to confirm.  It doesn't seem to be a musl behavior - this post says musl disables line-buffering when stdout is not a TTY just like glibc (https://www.openwall.com/lists/musl/2014/10/21/5).

So the `grep --line-buffered` isn't necessary on busybox.  (It is needed on GNU coreutils though.)

The bad news is that the one hang I observed when flashing probably wasn't caused by this issue then.  It has not occurred again since, I'll just have to monitor for it to occur again.

tlaurion:
I observe a hang when trying to flash from legacy to maximized boards and regions were not rw.

Not sure how to deal with that in the future. But that is good news in the sense that flashrom bails, but nothing outputs to the user. Scary for the user. But no flashing is attempted when writing whole image not just bios region.

No regression here. Just a note!
Merging!

PR #1496, opened 2023-09-05
initrd/bin/unseal-hotp: Prevent script errors if unseal fails

JonathonHall-Purism:
If the secret can't be unsealed, die immediately rather than continuing on to generate errors.


tlaurion:
Nice catch

PR #1430, opened 2023-07-04
Support Thinkpad W541

gaspar-ilom:
allows building heads for the w541 based on the t440p which has a sufficiently similar architecture.

I tested this with my the blobs from the test board and did not encounter any issues so far.

PR #1500, opened 2023-09-11
W541 Support

resende-gustavo:
Adds support for building heads for the W541 model, based on the T440p.

I have tested it with my board blobs, and so far, there have been no issues.


tlaurion:
@resende-gustavo doenload-clean-me script misses chmod +x permission as reported by Circleci when the script is attempted to be ran

resende-gustavo:
Sorry about that, the blobs were also missing. I have already set the GBE MAC address to 00:de:ad:c0:ff:ee

tlaurion:
@resende-gustavo if based on t440p Linux config, which is basically x230-maximized, is there any reason to not take t440p Linux config?


I saw as comments into board config a note about creating common ThinkPad config, which as today is not really possible because revisions of boards will have certain optimizations from what is provided by CPU extensions, nvme presence etc.


If next steps is to specialize those, which could happen under chipset specialization for example, then for this board I would recommend for now to check what is needed to not reuse the t440p Linux config directly, and/or create an Haswell variation. But I think as if now, since xx20/xx30 (Ivy/Sandy) are currently sharing x230-maximized, I think Haswell could share t440p-maximized.

The reasoning of this is that legacy and flash boards are sonner then later going to be deprecated and eventually we could have sandy/ivy/Haswell maximized Linux configs that are shared across chipsets being the same and offering same cpu extensions.

The longer version of why it's not really feasible is that for example even across ivy/Sandy there is people expecting to be able to use I3 CPUs which for example won't have NI-AES nor virtualization which was an issue leading to add support for non accelerated AES kernel crypto backend. Not an issue if we get rid of legacy boards, not an issue if kernel configs are shared across chipsets, but those chipset related configs need to take for granted that enough space is available in SPI and ease comparison of activated kernel options so that only what is needed across all possible configs is activated for a specific chipset supported CPU if that makes sense.

Tldr: if w541 kernel config is the same as t440p, to ease maintaintership, w541 board config should point to t440p for now or be different only because it is. Makes sense?

tlaurion:
> Sorry about that, the blobs were also missing. I have already set the GBE MAC address to 00:de:ad:c0:ff:ee

@resende-gustavo awesome. After previous comment addressed, please confirm that by flashing Circleci image everything works for you.

Ideally commits are a bit cleaner and could be squashed all together to add support of the W451. Quick review here seems to take work and reviewed work of t440p, I'll try to review more deeply in the next weeks prior of merging.

tlaurion:
Duplicate of #1430?

tlaurion:
#1430 has been rebased on master.

PR #1505, opened 2023-09-29
Add support for Librem 11

JonathonHall-Purism:
Add support for the Librem 11 convertible tablet.

Librem 11 has a detachable keyboard cover, which connects with USB.  If the keyboard is not attached, the volume and power keys can navigate menus.

loadkeys (from kbd) is used to load a key map configuring the volume up/down as up/down arrows, and power as Enter.  This enables navigation in fbwhiptail as well as executing history commands from the recovery shell (so it is possible to at least `poweroff` or `reboot` if you are dropped to the recovery shell without a keyboard nearby).

fbwhiptail is updated to 1.3, which has improved navigation, all menus can be navigated with either up/down or left/right ("yes/no" prompts would not have been navigable otherwise, since they needed left/right arrows).  There is a lot of internal cleanup to fbwhiptail's layout and navigation also.

coreboot-purism is updated to 4.21-Purism-2, adding Librem 11.  Blob jail binaries are now separated by board, and Librem 11 device firmware is added.

Librem 11 uses coreboot graphics init with Linux efifb; coreboot initializes graphics with FSP GOP since libgfxinit currently lacks Jasper Lake support.


.gitignore
```diff
@@ -16,12 +16,10 @@ config/*.old
...
 crossgcc
 clean
-*.map
 *.sec
 *.dep
 *.ffs
 *.vol
 *.lz
 *.fv
-*.bin
```
tlaurion, line 26: Not sure about this. Adding it manually for devels is one thing, but not getting "dirty" builds when modified is another? This is more for debating then refusal of that change.

For example on older boards where ifd.bin is modified but ifd is in git, the user will get a dirty build where git status would inform that ifd was changed under blobs dir. The dev/user can then add the ifd and sign to get rid of "dirty" builds or accept that the build he flashed is dirty.

blobs/librem_jail/README
```diff
@@ -1 +1,17 @@
...
+
+This directory contains firmware/microcode needed for peripherals in some Librem boards, for OSes that do not provide device firmware.
+
+When the blob jail feature is enabled, PureBoot provides device firmware to the OS by adding it to the initrd, then copying it to /run/firmware (see initrd/bin/inject_firmware.sh).
+
+## Librem Mini v2
+
+Librem Mini v2 ships with an Atheros Wi-Fi/Bluetooth card, which does not require firmware for Wi-Fi (but it does for Bluetooth).  However, some whitelabel variants ship with Intel AX200, which requires firmware for both Wi-Fi and Bluetooth.
+
+Device firmware is provided in a preconfigured variant buidl for librem_mini_v2, the basic_usb_autoboot_blob_jail build.
```
tlaurion, line 11: typo in build

.gitignore
JonathonHall-Purism: @tlaurion I'm _removing_ *.bin.  It's ignored in master now.  It doesn't affect files added to the index already AFAIK, but does affect whether dropping a new .bin file shows up as a change and whether `git add` will actually add it or silently ignore it.

initrd/.ash_history
```diff
@@ -15,3 +15,6 @@ seal-totp
 #Verify Intel ME state:
 cbmem --console | grep '^ME'
 cbmem --console | less
+# Reboot/power off (important for devices with no keyboard to escape recovery shell)
+reboot
```
tlaurion, line 19: I would suggest adding comment after comment so that user can see those+ comments when going up and down where comment would be neglicted at call.

tlaurion:
@JonathonHall-Purism stepped into https://review.coreboot.org/c/coreboot/+/78239

Affected?

JonathonHall-Purism:
Looks like the original change was in 4.21, so yeah anything based on 4.21 would be affected, the Purism branch is already affected.  Looks like the only consequence is that diffoscope doesn't work, which isn't perfect, but I think I will just let this land in 4.22, not worth cherry-picking and re-testing.

blobs/librem_jail/README
JonathonHall-Purism: Fixed, thanks :bow:

initrd/.ash_history
JonathonHall-Purism: Added that, thanks, good suggestion :bow:

PR #1508, opened 2023-10-11
WiP: Improve checksum verification upon totp unseal errors

tlaurion:
Raw notes:
- Writing multi-line output to /dev/kmsg didn't work. We now output one line of message in DEBUG calls
- When TOTP unseal errors happened, the use needed to also sign /boot without knowing if things were right.
  - Now integrity report is given
   - If gpg detached signature or hashes are invalid, we give the user a detailed report of hashes being different is detaches signature is good
- There were multiple instances of verify_global_hashes
  - The one from gui-init has been moved to etc/functions, and calls requiring to update the hashes now pass optional parameter
- Are some /tmp/hash_output relevance missed somewhere?

What should be improved further?

### DIFF
--- a/initrd/etc/functions
+++ b/initrd/etc/functions
@@ -35,9 +35,9 @@ DO_WITH_DEBUG() {
 
 pcrs() {
 	if [ "$CONFIG_TPM2_TOOLS" = "y" ]; then
-		tpm2 pcrread sha256
+		tpm2 pcrread sha256 || die "tpm2 pcrread failed"
 	elif [ "$CONFIG_TPM" = "y" ]; then
-		head -8 /sys/class/tpm/tpm0/pcrs
+		cat /sys/class/tpm/tpm0/pcrs
 	fi
 }
 
@@ -322,7 +322,8 @@ check_config() {
 
 	if [ "$2" != "force" ]; then
 		if ! sha256sum $(find $1/kexec*.txt) | gpgv $1/kexec.sig -; then
-			die 'Invalid signature on kexec boot params'
+			warn "Invalid signature on $1/boot/kexec*.txt files"
+			die "Please verify hashes/signatures or use call check_config $1 force to check hashes without signature"
 		fi
 	fi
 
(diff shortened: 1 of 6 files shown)


tlaurion:
Played around with qemu-fbwhiptail-tpm1-hotp, qemu-fbwhiptail-tpm2-hotp and love the improvements.

tlaurion:
Currently fails on default boot and select boot option. Putting as draft

JonathonHall-Purism:
From reviewing verify_global_hashes, it looks like there were significant differences between these two functions.  We can't just replace one with the other.

In particular:
* The one in gui-init returned 0 or 1 for success/failure.  The one in kexec-select-boot always returned 0, instead it set valid_hash=y and valid_global_hash=y for success and left them untouched for failure.
* The one in kexec-select-boot also checks root file hashes.  The one in gui-init did not.
* The one in kexec-select-boot used "$gui_menu" to control whether differences are shown or just checked silently, the one in gui-init did not.

I do think this needs to be refactored, but we can't just replace one with the other.

initrd/etc/functions
```diff
@@ -440,6 +441,177 @@ update_checksums() {
...
+			UPDATE_INITRAMFS_PACKAGE=$(grep '^UPDATE_INITRAMFS_PACKAGE' $TMP_PACKAGE_TRIGGER_POST | cut -f 2 -d '=' | tr -d '"')
+
+			if [ "$UPDATE_INITRAMFS_PACKAGE" != "" ]; then
+				TEXT="The following files failed the verification process AFTER package updates ran:\n${CHANGED_FILES}\n\nThis is likely due to package triggers in$UPDATE_INITRAMFS_PACKAGE.\n\nYou will need to update your checksums for all files in /boot."
+				TEXT+="$UPDATE_TEXT"
+			else
+				TEXT="The following files failed the verification process AFTER package updates ran:\n${CHANGED_FILES}\n\nThis might be due to the following package updates:\n$LAST_PACKAGE_LIST.\n\nYou will need to update your checksums for all files in /boot."
+				TEXT+="$UPDATE_TEXT"
+			fi
+		else
```
JonathonHall-Purism, line 505: This logic isn't new (moved from gui-init), but while we're here....

This looks like logic to allow an OS package manager to place files on /boot indicating:
* what files failed validation _before_ updating packages
* what files failed validation _after_ updating packages
* what packages were updated

This then controls the prompts shown by Heads if files do not validate.  But `kexec_package_trigger_{pre/post}.txt` aren't themselves authenticated in any way (which would have to somehow "authenticate" that your package manager really made these changes).  So as far as I can tell, somebody tampering /boot could just place these files here with whatever content they want and thereby control what Heads will say when it observes the tampering.

Does this package manager support exist in any OS?

I noted some additional details from my read-through here, but really the key question to me is - why should we trust these kexec_package_trigger_* files?

initrd/etc/functions
```diff
@@ -440,6 +441,177 @@ update_checksums() {
...
+			HASH="OK"
+		else
+			HASH="ALTERED"
+		fi
+
+		#Show results
+
+		#If the system is in a bad state, show the user a warning
+		whiptail $MAIN_MENU_BG_COLOR --title "Measured Integrity Report" --msgbox "$date\n\nTOTP: $TOTP\nHOTP: $HOTP\nBOOT INTEGRITY SIGNATURE: $SIGNED\nBOOT INTEGRITY HASHES: $HASH\n\nPress OK to continue with detailed information in case of errors" 0 80
+
```
JonathonHall-Purism, line 606: IMO:

* All the CAPS introduced in the error text, as well as  the much longer labels for integrity, make this difficult to read.  There's no longer any separation between the "labels" and "content" for errors (screenshot below)
* I think the separation between "signature" and "hashes" should be an implementation detail of Heads - that's how we authenticate /boot.  The normal user flow should just say whether /boot integrity was authenticated or not, so it's easier to understand if you are not an expert in Heads.  (I'd be OK with providing detailed information as an optional thing or on the console, etc. to cater to advanced users too, but putting it front-and-center IMO makes it a requirement to be an knowledgeable in Heads internals to use Heads.)

Re: the caps, this is what I was greeted with when firing this up in qemu:

Screenshot_20231012_171341

From a glance it is really hard to parse this and know what to look at.  The text-only prompt is a bit limiting but I'd suggest:
* Don't use all caps for the errors, this doesn't add anything and just makes it harder to read
* Go back to just "INTEGRITY: <status>" instead of the last two lines, which would make the labels all similar-length again and reduce a lot of verbosity
* Take out the colon from the HOTP error so the colons form a bit of separation between the labels and contents

JonathonHall-Purism: Oh and also, it did not show me any detailed information when I pressed Enter :-/

JonathonHall-Purism:
Thinking back to the actual intent of the change - why do you want to report /boot integrity when TOTP/HOTP unseal fails?  Seems like this is adding another prompt in an already nontrivial flow, and I think it'd be hard for users who aren't experts in Heads to understand what this information has to do with the issue at hand.

>When TOTP unseal errors happened, the use needed to also sign /boot without knowing if things were right. 

TOTP/HOTP re-seal doesn't affect the integrity of /boot.  Why would you need to sign /boot without knowing the prior status in this case?

tlaurion:
The intent of this is once again for tpm disk unlock key specifically but more globally for users to not sign /boot content when totp unsealing fails and resealing totp. 

When one updates totp, tpm Disk Unlock Key needs to be updated, but doing so, LUKS header changes and its local hash kept under /boot changes which is signed under kexec.sig.

On tpm2, primary handle needs to be updated as well.

As for pre trigger and post trigger files under /boot, I will have to check history to see which os populates those and if they are relevant. I do not use any OS doing those but thought probably wrongly that pureos was the intent of support here. Will check later but as of now I think the unification of those functions will need more thinking.

PR #1509, opened 2023-10-11
newt(whiptail): fix code that was doing toupper of input

### LINKED ISSUE #1455: Make sure whiptail prompt choices are unique (eg. no 'M' and 'm' in same menu)
Otherwise, this makes automated testing difficult since whiptail doesn't differentiate lowercase and uppercase.

tlaurion:
This fixes #1455 without needing to change all previous menu choices.

### DIFF
--- a/patches/newt-0.52.20.patch
+++ b/patches/newt-0.52.20.patch
@@ -1,0 +1,20 @@
+--- ./listbox.c.orig	2023-10-11 15:26:20.365000000 -0400
++++ ./listbox.c	2023-10-11 15:27:26.762000000 -0400
+@@ -652,7 +652,7 @@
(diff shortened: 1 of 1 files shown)


tlaurion:
@krystian-hebel is that good enough? Tests could now type F and f instead of changing all prompts.
At the end of the day, that's your call for automated tests. As of today, https://github.com/Dasharo/open-source-firmware-validation/blob/main/dasharo-compatibility/heads-bootloader-support.robot would not need change.

JonathonHall-Purism:
If we have to have differently-cased menu items in the same prompt, this looks OK.  I still think having differently-cased menu items adds pointless complication, but we ship fbwhiptail on everything so this doesn't impact me much.

krystian-hebel (approved):

PR #1485, opened 2023-08-31
add Nitropad NV41/NS50 TPM2 boards (2nd)

daringer:
This is the continuation PR, originating from: #1417
Mainly adds the CI related fixes from here: #1469
The Nitrokey 3 support is extracted into this PR: #1483 

CircleCI builds are also activated

PR #1518, opened 2023-10-28
POC: Add tooling for built-in ROM digest verification

JonathonHall-Purism:
Proof of concept for built-in ROM digest verification.  This method is backward compatible with older Heads and also supports external/manual flashing with the same artifact (since the artifact is still just the ROM).

@tlaurion @daringer Following up on discussion in #1514, I propose this method instead of the NPF format.  (Please check out my last few comments there for some findings on the NPF support.)  Please let me know what you think :grin: 

I have not yet integrated this into the existing flashing scripts, it's just standalone scripts to calculate the digest or check it.  Check out the usage notes in those scripts to try them out.

I added a "--replace" operation to `cbfs` to have a clear guarantee that it would replace the digest data in-place without altering any other part of the ROM, which is necessary for digest verification.  A delete+add generally does work at this point since the file placement/header are exactly the same, but adding "--replace" is pretty easy and much more robust as tooling might change in the future.

We should be able to pretty easily calculate the digest during the build so no further manual steps are needed too.

edit: Whoops tagged the wrong person, had intended to tag @daringer re: #1485 which introduced NPF, my mistake @krystian-hebel

### DIFF
--- a/initrd/bin/calc_digest.sh
+++ b/initrd/bin/calc_digest.sh
@@ -1,0 +1,46 @@
+#! /bin/bash
+
+set -eo pipefail
+
+ARG_ZERO=
+
+if [ "$1" = "-z" ]; then
+	ARG_ZERO=y
+	shift
+fi
+
+if [ "$#" -lt 2 ]; then
+	echo "usage: $0 [-z] <original-rom> <updated-rom>" >&2
+	echo
+	echo "Updates (or adds) rom_digest in the ROM."
+	echo
+	echo "By default, rom_digest is set to the ROM's SHA-256 digest (with"
+	echo "rom_digest set to all-0)."
+	echo
+	echo "With -z, just zero rom_digest, so integrity can be checked."
+	exit 1
+fi
+
+ORIGINAL_ROM="$1"
+UPDATED_ROM="$2"
+
+cp "$ORIGINAL_ROM" "$UPDATED_ROM"
+dd if=/dev/zero bs=32 count=1 of=/tmp/digest.bin status=none
+
+# Ensure there is a zeroed rom_digest file, but don't delete any existing file
+if ! cbfs -l -o "$UPDATED_ROM" | grep -q '^rom_digest$'; then
+	# Add the file
+	cbfs -a rom_digest -o "$UPDATED_ROM" -f /tmp/digest.bin
+else
+	# Replace the file content
+	cbfs -p rom_digest -o "$UPDATED_ROM" -f /tmp/digest.bin
+fi
+
+# If we are just zeroing the digest, we're done, otherwise calculate a digest
+# and set it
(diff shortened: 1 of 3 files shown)


tlaurion:
Ok so if I get this right, the idea would be to inject the calculated rom hash at build time, invalidating builds hashes.txt produced at build time? It would also require flash to dump rom, remove cbfs added hash prior of validating hash of rim without it?

JonathonHall-Purism:
> Ok so if I get this right, the idea would be to inject the calculated rom hash at build time

Yes, the idea is to insert this at build time.

>invalidating builds hashes.txt produced at build time?

No, if the integrity digest is inserted before hashes.txt is calculated, hashes.txt will be fine.

daringer:
Generally we have no strict preference for one or another approach. We chose the `.npf` idea because this delivers two features: 
* it is a plain `zip` so we can tell customers to unzip it in order to get the original `.rom`
* one can verify the rom by using default tools (`sha256sum` + `unzip`) 

Nevertheless, including the checksum info the cbfs is clearly the more backwards compatible way. Please correct me, but the cbfs-based approach does something like this to verify the image:
* extract the checksum
* copy the rom and nullify the checksum 
* calculate checksum for nullified-checksum-rom
* compare with extracted checksum

This is quite some added complexity for the sake of backwards compatibility, which is not necessarily a bad thing. Without heads tooling it is not really possible to verify the integrity of the `.rom` - this feels like a drawback.

On the other side we could also improve the `npf` approach. 

Under the line I personally cannot see that one approach is clearly "better" - for me this looks like a design-decision.

tlaurion:
> No, if the integrity digest is inserted before hashes.txt is calculated, hashes.txt will be fine.

Yes, but the rom won't match that verification checksum by itself and will require additional tooling/scripts to validate. Not against, but will require transaparency and Makefile modifications at least stating on CI and on builder's machines that the rom is modified with added cbfs file and that that cbfs file needs to be removed prior of checking rom for hash match.

Will continue to think on this, but I would prefer a simpler approach
- distribute zip files on CI with hash files in
- generalize npf codepath and verywhere npf searched for/prmpted for/messages for user, just *.npf/*.zip

Go on with this #1518 concept in parallel. But as stated under #1514, master state with this npf prompt without Heads providing such npf files is making me unconfortable. I would simply continue to have CI provide rom+zip files and modify that PR so that npf/zip files are promoted as of now. And I think continuing promoting rom files as of today with manual user verification works. People flashing #1514 merged result, when we agree, would permit people to flash zip/npf later on, where we should adapt our things to provide such files downloadable/buildable.

JonathonHall-Purism:
Closing this in favor of #1526 - our support team agrees with plans to slowly migrate to the ZIP format, so the new PR improves that instead.

PR #1526, opened 2023-11-13
flash-gui.sh: Extend NPF archive format to ZIP, improve workflow

JonathonHall-Purism:
The NPF file extension can be configured on a per-board basis, default is ZIP.  Fixed some issues in the implementation (don't require /tmp/verified_rom/ in sha256sum.txt, show integrity error instead of dying if ZIP itself is corrupt as that is more likely, etc.), details in commit messages.

Update package is built in default target now, removed create_npf.sh.

@daringer This should still be compatible with any npf archives already published (/tmp/verified_rom/ prefix is removed if present).  Do you want `export CONFIG_BRAND_UPDATE_PKG_EXT=npf` added to your boards?

@tlaurion I refactored prompts a little so .tgz only appears for talos-2 (and .rom does not appear on talos-2).  I can't really do any further refactoring to better integrate talos-2 with the zip method.  If you want to improve anything there I'm happy to review/discuss, we could also merge this and improve later as this is better than the current master for all boards.

So help me, I added a confirmation prompt :scream:  zip/npf somehow ended up without one, went straight to flash after the file selector.  It's just too easy IMO to accidentally select the wrong file, and flashing is sensitive/irreversible.  I brought back the prompt message from before the NPF introduction.

The prompt after rebooting now shows the file that was selected, not the file that was flashed (which is some temp file when selecting an update package).

The .rom prompt is unchanged, I expect to carry a patch downstream in PureBoot for a while softening the message there until we are confident that .zip support has rolled out widely enough to switch to it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,6 +52,11 @@ CONFIG_TARGET_ARCH := x86
 
 include $(CONFIG)
 
+# Default update package extension is 'zip' unless a brand wants a branded
+# extension
+CONFIG_BRAND_UPDATE_PKG_EXT ?= zip
+CB_UPDATE_PKG_FILE := $(CB_OUTPUT_BASENAME).$(CONFIG_BRAND_UPDATE_PKG_EXT)
+
 # Unless otherwise specified, we are building for heads
 CONFIG_HEADS	?= y
 
@@ -159,7 +164,18 @@ payload: $(build)/$(BOARD)/bzImage $(build)/$(initrd_dir)/initrd.cpio.xz
 
 ifeq ($(CONFIG_COREBOOT), y)
 
-all: $(board_build)/$(CB_OUTPUT_FILE)
+# Coreboot targets create an update package that can be applied with integrity
+# verification before flashing (see flash-gui.sh).  The ZIP package format
+# allows other metadata that might be needed to added in the future without
+# breaking backward compatibility.
+$(board_build)/$(CB_UPDATE_PKG_FILE): $(board_build)/$(CB_OUTPUT_FILE)
+	rm -rf "$(board_build)/update_pkg"
+	mkdir -p "$(board_build)/update_pkg"
+	cp "$<" "$(board_build)/update_pkg/"
+	cd "$(board_build)/update_pkg" && sha256sum "$(CB_OUTPUT_FILE)" >sha256sum.txt
+	cd "$(board_build)/update_pkg" && zip -9 "$@" "$(CB_OUTPUT_FILE)" sha256sum.txt
+
+all: $(board_build)/$(CB_OUTPUT_FILE) $(board_build)/$(CB_UPDATE_PKG_FILE)
 ifneq ($(CONFIG_COREBOOT_BOOTBLOCK),)
 all: $(board_build)/$(CB_BOOTBLOCK_FILE)
 endif
--- a/Makefile
+++ b/Makefile
@@ -183,9 +183,10 @@ $(board_build)/$(CB_UPDATE_PKG_FILE): $(board_build)/$(CB_OUTPUT_FILE)
 	cp "$<" "$(board_build)/update_pkg/"
 	cd "$(board_build)/update_pkg" && sha256sum "$(CB_OUTPUT_FILE)" >sha256sum.txt
 	cd "$(board_build)/update_pkg" && zip -9 "$@" "$(CB_OUTPUT_FILE)" sha256sum.txt
-endif
 
 all: $(board_build)/$(CB_OUTPUT_FILE) $(board_build)/$(CB_UPDATE_PKG_FILE)
+endif
+
 ifneq ($(CONFIG_COREBOOT_BOOTBLOCK),)
 all: $(board_build)/$(CB_BOOTBLOCK_FILE)
 endif
--- a/Makefile
+++ b/Makefile
@@ -50,6 +50,12 @@ endif
 # By default, we are building for x86, up to a board to change this variable
 CONFIG_TARGET_ARCH := x86
 
+# Legacy flash boards have to be handled specifically for some functionality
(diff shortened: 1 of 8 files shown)


tlaurion:
Building and testing here locally for flash/reflash

Makefile
```diff
@@ -174,6 +183,7 @@ $(board_build)/$(CB_UPDATE_PKG_FILE): $(board_build)/$(CB_OUTPUT_FILE)
 	cp "$<" "$(board_build)/update_pkg/"
 	cd "$(board_build)/update_pkg" && sha256sum "$(CB_OUTPUT_FILE)" >sha256sum.txt
 	cd "$(board_build)/update_pkg" && zip -9 "$@" "$(CB_OUTPUT_FILE)" sha256sum.txt
+endif
 
 all: $(board_build)/$(CB_OUTPUT_FILE) $(board_build)/$(CB_UPDATE_PKG_FILE)
```
tlaurion, line 188: This needs to be before endif
```


user@heads-tests-deb12:~/heads$ git diff
diff --git a/Makefile b/Makefile
index 591d3b31fe6..3455d5ecf37 100644
--- a/Makefile
+++ b/Makefile
@@ -183,9 +183,9 @@ $(board_build)/$(CB_UPDATE_PKG_FILE): $(board_build)/$(CB_OUTPUT_FILE)
        cp "$<" "$(board_build)/update_pkg/"
        cd "$(board_build)/update_pkg" && sha256sum "$(CB_OUTPUT_FILE)" >sha256sum.txt
        cd "$(board_build)/update_pkg" && zip -9 "$@" "$(CB_OUTPUT_FILE)" sha256sum.txt
+all: $(board_build)/$(CB_OUTPUT_FILE) $(board_build)/$(CB_UPDATE_PKG_FILE)
 endif
 
-all: $(board_build)/$(CB_OUTPUT_FILE) $(board_build)/$(CB_UPDATE_PKG_FILE)
 ifneq ($(CONFIG_COREBOOT_BOOTBLOCK),)
 all: $(board_build)/$(CB_BOOTBLOCK_FILE)
 endif

```

JonathonHall-Purism: Fixed in 51e2d789, thanks

tlaurion:
> > @daringer This should still be compatible with any npf archives already published (/tmp/verified_rom/ prefix is removed if present). Do you want `export CONFIG_BRAND_UPDATE_PKG_EXT=npf` added to your boards?
> 
> no, don't think this is useful for upstream - this will only confuse people, we'll do this in our releases thanks a lot for adding it and all the improvements

@daringer you sure? I would have thought that nitrokey would want npf files poposed to them instead of zip files for their boards, built through Heads or released from Nitrokey. Otherwise as this PR, zip files will be proposed to them from Heads, and npf from Nitrokey releases?

tlaurion:
@JonathonHall-Purism 
Unfortunately, the flow is broken for Talos-2 here since tgz is considered a "plain" rom from flash-gui.sh, where logic for validation happens under flash.sh

signal-2023-11-14-151718
signal-2023-11-14-151745

JonathonHall-Purism:
> @JonathonHall-Purism Unfortunately, the flow is broken for Talos-2 here since tgz is considered a "plain" rom from flash-gui.sh, where logic for validation happens under flash.sh

We discussed this, this issue is in master (from #1485).  In retrospect, when we discovered the issues in these changes we probably should have backed it out so it could be fixed properly rather than rushing in more changes on top (#1514, which when through several iterations, and now this PR).

I don't think a revert is necessarily needed at this point if this PR is mergeable.  talos-2 has a regression originating from #1485 that still isn't fixed, but I'm not sure the impact is great enough to justify a revert going that far back.

JonathonHall-Purism:
The last talos-2 CI failure just failed to download a toolchain source archive (I touched Makefile so it invalidated the whole cache).  I'll retry failed when the pipeline completes

JonathonHall-Purism:
Had to push something to trigger a new CircleCI pipeline, so re-signed the last commit and force pushed.

I guess CircleCI changed they way there Github integration worked, I had to create a new account, which means I couldn't access the old pipeline to retry it.  Had to set it up again and trigger a new pipeline.

daringer:
The question is maybe to be asked from another direction: Should HEADS maybe accept both npf & zip as provided file, this would reduce migration issues. 

Generally, I see no issue in migrating to `.zip` also mainly to be consistent with upstream - accepting both might just reduce pain here for the users.

tlaurion:
Ok so with [b8428f3](https://github.com/tlaurion/heads/commit/b8428f3728480ff359dc4e7575cb3387b6221042) talos flashed through bmc and then attemtping upgrades from usb tgz:

signal-2023-11-15-105153
signal-2023-11-15-105232
signal-2023-11-15-105343

JonathonHall-Purism:
Thanks all :+1: 

@tlaurion Picked the fixes for talos-2.  Per chat we'll hold off on the caching change for now.

@daringer Thanks for the consideration and keeping us informed, I've removed the brand config for the package extension, we'll use zip everywhere.

Tested this quickly on Librem 15v3 to confirm no regressions flashing .zip or .rom.  If I can get CI to cooperate, this should be good to go.

I think CI will probably cooperate now - ftpmirror.gnu.org is back up, and I reluctantly gave CircleCI access to private repos too since there seems to be no other option.  (I don't actually have any at the moment, but annoyed to give access that isn't needed...)  I tried switching to the new "GitHub App" method but that had problems cloning dependencies that I did not want to troubleshoot, hopefully the OAuth method sticks around for a while.  We'll see :crossed_fingers:

tlaurion (approved):
Tested extensively. We might want to readd confirmation for talos-2 later on, but at least this is coherent as is and put things in better state then master.

Let's merge and thank you for this!

PR #1452, opened 2023-08-04
modules/coreboot: update rev to Dasharo v.0.7.0 release for RC Talos II

macpijan:
The diff between these two commits is literally a change in defconfig, so no actual change in terms of codebse from heads perspective, since heads is maintaining defconfig on it's now.

PR #1530, opened 2023-11-17
Talos-2: bring changes to a working state outside of usage of GPG key material backup as of now

tlaurion:
- Closes https://github.com/linuxboot/heads/pull/1452
- coreboot: Take Talos II 0.7 release coreboot config file that was inside of cbfs and use it as a base upstream.
- linux: Readd sysctl and proc requirements for cbmem to work.

TODO: fix gpg2 module so that the following doesn't happen (a ppc64 thing. Can't figure out why):

```
Adding generated key to current firmware and re-flashing...

Board talos-2 detected, continuing...
37281653053696daf2e40a8efe9451b557d9d6ab586830dc85f814bf2e03a05f  /tmp/talos-2.rom
Initializing Flash Programmer
Reading old flash contents. Please wait...
Flashing: [##################################################\] (100%)
Verifying flash contents. Please wait...
The flash contents were verified and the image was flashed correctly.

Signing boot files and generating checksums...

180726119: 000E452213510000005A
gpg: error running '//bin/dirmngr': probably not installed
gpg: failed to start dirmngr '//bin/dirmngr': Configuration error
gpg: can't connect to the dirmngr: Configuration error
gpg: no default secret key: No dirmngr
gpg: signing failed: No dirmngr
```
dirmngr is deactivated per configure statement --disable-dirmngr, and works as expected on x86


tlaurion:
Rebased on master that just merged #1526

Issue #1438, opened 2023-07-11
ISO signing GUI


JonathonHall-Purism:
We could have a "signing GUI" to sign an ISO with your key.  Potentially this could also allow configuring which OS keys (if any) are trusted for ISO signatures, which integrates nicely with Restricted Boot.

Ideally, this should verify the ISO against sha256sums on the device while signing, or otherwise present the sha256 to the user for verification.

Original discussion: https://github.com/osresearch/heads/pull/1419#discussion_r1258631584

tlaurion:
Some considerations.
OSes are now providing hashes in separated file be be checked upon. Historically, fedora was provided detached gpg signatures and stopped doing that, providing detached signatures of hashes instead.

To me, providing gpg detached signature does both steps in one which was better, but OSes slowly stopped doing that.

I will have to gather notes on that and replicate but i remember some kind if issue with busybox trying to support the new format.

Just to note that providing hashes before signing requires two times processing ISO which are nomally big files. If going in that direction we might as well try to support validation against OSes provided digests and hack around busybox limitations, while detached signatures still provide authenticity and integrity validation.

tlaurion:
gpg --print-md SHA256 --detach-sign test.txt

tlaurion:
@JonathonHall-Purism i don't seem to find a hack to have gnupg output contain the 256sum calculated and be able to just prompt the user to confirm hash is as expected prior of accept created detached signature (otherwise deleting created detached signature at this point).

That would be ideal but haven't found a way. Ideas?

JonathonHall-Purism:
@tlaurion It seems like what we really want here would be:
1. hash the ISO
2. ask user if the hash is correct
3. If so, detach sign with the hash already computed

But there's no way to do step 3, as far as I can tell - there's no way to have GPG create a signature for an already-known digest, it wants to digest the file itself.

In theory maybe that could be added to GPG?  Not sure though.  Otherwise, only options I can think of:
1. Hash the ISO, then sign the hash file (a la SHA256SUMS) instead of signing the ISO directly - I don't love the extra level of indirection, but maybe this is why ISO signatures are often published this way?
2. Sign the ISO, then somehow dump the hash from the signature (it's in there).  GPG doesn't seem to support this as-is either - `--list-packets` on a signature unhelpfully abbreviates the digest as `begin of digest cd a3` (and doc says --list-packets is not considered a stable interface).  I don't love creating a signature that we might throw away either.
[...]

PR #1540, opened 2023-11-27
initrd/etc/functions: Refine list_usb_storage logic to enable booting from a USB device with an isohybrid-created image

P5vc:
This pull request modifies the `list_usb_storage` function within `/initrd/etc/functions` with the primary goal of correcting a bug that prevents a user from booting from certain USB media.

For example, [here's a post](https://forum.qubes-os.org/t/qubesos-4-2-rc3-clean-install-issue-on-librem-14/21051) from the Qubes OS forum of users encountering this bug, and [here's an article](https://github.com/P5vc/general-wiki/wiki/USB-Boot-Fails-%28Due-to-Hybrid-ISO-Format%29) outlining, in depth, the root cause of this issue.

In addition to correcting the aforementioned bug, this pull request takes advantage of having to rewrite a chunk of the core logic anyways, to:
- Generalize the `fat32` workaround to apply to theoretically any filesystem causing the same type of error
- Simplify some of the logic to better-conform to the POSIX standard, and remove dependence on more-complex utilities (such as `sed` and `awk`), when simpler utilities will do the trick
- Standardize some of the terminology; different parts of the function used different terms when referring to the USB block storage devices (such as "devices", "disks", "drives", etc.)
- Write additional comments/descriptions to better-document the purpose of each utility. This should be especially helpful given the long chain of piped output found in this function

As a final note, you may notice that main addition to the pre-existing logic was to add a mount test for all detected USB block storage devices (but not for their partitions). Although perhaps not the prettiest way to address this issue, I was unfortunately unable to find a better solution within the confines of the busybox environment. For example, I initially attempted to simply add logic looking for the ISO 9660 filesystem on the block devices, however utilities that can easily do this on most systems (such as `blkid`), do not support this functionality in their busybox versions. In fact, the only busybox utility I could find with this functionality was `mount`. Therefore, this pull request could theoretically be modified to add the `-t iso9660` option to the `mount` command, and as a result only end up mounting (and including) devices with the ISO 9660 filesystem (fixing the specific error from the links above). However, adding this option would not significantly impact efficiency (at least it didn't during my limited testing), and may unnecessarily limit the fix by making it overly-specific to the above issues. That being said, if you think that adding this option is preferred, or if you know of a better approach that doesn't require using `mount`, please let me know!

### DIFF
--- a/initrd/etc/functions
+++ b/initrd/etc/functions
@@ -132,15 +132,19 @@ enable_usb_storage() {
 	fi
 }
 
+# Output, one per line, paths to all detected USB block device partitions. If a
+# block device does not contain any valid partitions and/or has had a filesystem
+# directly written to it, the device itself will be included in the list.
+# Alternatively, the "devices" argument may be passed to list only the USB block
+# devices themselves, ignoring any partitions they may contain.
 list_usb_storage() {
 	TRACE "Under /etc/functions:list_usb_storage"
-	# List all USB storage devices, including partitions unless we received argument stating we want drives only
-	# The output is a list of device names, one per line.
 
-	if [ "$1" = "disks" ]; then
-		DEBUG "Listing USB storage devices (disks only) since list_usb_storage was called with 'disks' argument"
-	else
-		DEBUG "Listing USB storage devices (including partitions)"
+	if [ "${1-}" = 'devices' ]
+	then # "devices" argument passed; list block devices, ignoring partitions
+		DEBUG 'Listing USB block devices (excluding partitions, due to "devices" argument)'
+	else # "devices" argument missing; list only block device partitions, if found
+		DEBUG "Listing valid USB block device partitions (and devices with a filesystem or no paritions)"
 	fi
 
 	stat -c %N /sys/block/sd* 2>/dev/null | grep usb |
@@ -175,9 +179,9 @@ list_usb_storage() {
 				# No partition table, include this device
 				DEBUG "USB storage device without partition table: $b"
 				echo "$b"
-			#Bypass the check for partitions if we want only disks
-			elif [ "$1" = "disks" ]; then
-				# disks only were requested, so we don't list partitions
+			# Bypass the check for partitions if we want only devices
+			elif [ "${1-}" = 'devices' ]
+			then # Devices only were requested, so we don't list partitions
 				DEBUG "USB storage device with partition table: $b"
 				DEBUG "We asked for disks only, so we don't want to list partitions"
 				echo "$b"
--- a/initrd/etc/functions
+++ b/initrd/etc/functions
@@ -132,11 +132,11 @@ enable_usb_storage() {
 	fi
 }
 
-# Output, one per line, paths to all detected USB block device partitions. If a
-# block device does not contain any valid partitions and/or has had a filesystem
-# directly written to it, the device itself will be included in the list.
-# Alternatively, the "devices" argument may be passed to list only the USB block
-# devices themselves, ignoring any partitions they may contain.
+# Output, one per line, paths to all detected USB block storage device
+# partitions. If a block device does not contain any valid partitions and/or has
+# had a filesystem directly written to it, the device itself will be included in
+# the list. Alternatively, the "devices" argument may be passed to list only the
+# USB block devices themselves, ignoring any partitions they may contain.
 list_usb_storage() {
 	TRACE "Under /etc/functions:list_usb_storage"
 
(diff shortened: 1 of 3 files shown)


tlaurion:
New requirement from linuxboot org: commits need to be --sign-off

Sorry about that @P5vc . Reading from phone will check code tomorrow

tlaurion:
@P5vc testing https://app.circleci.com/pipelines/github/linuxboot/heads/701/workflows/90e1bf0d-f35a-43fe-ae46-f1c9975a7b73/jobs/13647 's https://output.circle-artifacts.com/output/job/521ebb13-af2b-469e-8eed-31954831d3dc/artifacts/0/build/x86/x230-hotp-maximized/heads-x230-hotp-maximized-v0.2.0-1919-g73c790e.zip

tlaurion (dismissed):
@P5vc I love what I see. @JonathonHall-Purism : veto?

P5vc:
> New requirement from linuxboot org: commits need to be --sign-off
> 
> Sorry about that @P5vc . Reading from phone will check code tomorrow

@tlaurion No worries! Would you like me to force push signed versions of those commits? If so, I don't mind, just let me know!

---

> testing... against Q4.2 rc5 dd'ed to usb thumb drive.

Just a heads up, it appears that the Qubes OS devs [reverted the switch](https://github.com/QubesOS/qubes-builderv2/pull/79) to using isohybrid just before Qubes OS 4.2 Release Candidate 5 was built, in response to [this issue](https://github.com/QubesOS/qubes-issues/issues/8717). Therefore, using one of the two previous RC builds ([Qubes-R4.2.0-rc3-x86_64.iso](https://ftp.qubes-os.org/iso/Qubes-R4.2.0-rc3-x86_64.iso) or [Qubes-R4.2.0-rc4-x86_64.iso](https://ftp.qubes-os.org/iso/Qubes-R4.2.0-rc4-x86_64.iso)) may be a better test of the bug fix. :+1: 

---

> I prefer to [detach sign isos with my key after having verified their integrity once](https://osresearch.net/InstallingOS/#generic-os-installation), and then attest of their state by detach-signing them. This permits to use and reuse a single USB thumb drive with plenty of ISOs that I just need to replace when a new version is released instead of keeping individual thumb drives per installation method.

This is objectively a much better method that I will certainly need to start using more often!! Thanks for the tip!

tlaurion:
Well I'm happy regression tests worked. Can do one last test with older release on test laptop.

- [ ] test q4.2 rc4


----

@P5vc yes please sign-off. Info is availle under failed test

tlaurion (approved):

P5vc:
The commits have been signed off on, as requested!

JonathonHall-Purism:
@P5vc Thanks a lot for the cleanup and consistency, and for identifying the problem here.  Thanks also for the context from Qubes.

TL;DR - Unfortunately, using the device iso9660 filesystem creates problems for a number of install media I tried today.  I do not think it is correct to boot isohybrid from flash media using the device iso9660 filesystem, we should use the partition table - that's the entire reason isohybrid exists, otherwise it would not exist and we'd just use iso9660 everywhere.

The real issue here was that Qubes 4.2 RC media had a broken partition table, as mentioned in the linked change reverting it.  Using the device iso9660 filesystem was a workaround to ignore the broken partition table.  They've fixed it, so we should not do this.

The cleanup is great though, I would like to merge that.  Below I listed the problems I ran into, so if we can revert the behavior of allowing the device to mount when partitions exist, this would be a great improvement.

I liked the idea of doing a test mount of the device to see if it contains a filesystem, the test before was a bit fragile.  However, since we need to favor the partitions if both exist, I don't think we can go that direction.

## Images offering both disk and partitions

Several images that used to offer two partition choices now offer both the disk itself and the partitions: memtest86+ GRUB, Debian 12 netinstall, PureOS byzantium Plasma, Debian 11 GNOME live, Librem EC update ISO.  That makes the partition selection more complex.  In a few cases it used to be able to select the partition automatically.

## Old remnant ISO9660 found

I have one flash drive in particular that behaves strangely with this change.  It has a GPT partition table, one ext4 partition, and that partition contains a bunch of ISOs I use to test Heads/PureBoot.  Current master works fine, finds the partition automatically and asks to select an ISO.

Strangely, with the MR it now finds remnants of an old iso9660 filesystem and mounts that automatically - it doesn't even offer me a choice between that or the legitimate ext4 partition.  Apparently I had a Fedora 36 image on this at one point, because those are the boot options I get.  I don't think we can blame the partitioning here, we can't require zeroing the entire disk to repartition it.

Restricting the device mount to iso9660 won't solve this (tried it, that's what it's finding).

initrd/etc/functions
```diff
@@ -132,66 +132,85 @@ enable_usb_storage() {
...
+
+			# Check if the USB block device has had a filesystem written directly
+			# to it, and include the device in the outputted list if so. Although
+			# this will not apply to the majority of block devices, it covers
+			# important edge cases, such as when bootable media has had an ISO
+			# created using the "isohybrid" utility, written to it.
+			test_mntpt='/tmp/direct-mount-test'
+			mkdir -p "$test_mntpt"
+			blkdev_included='false'
+			if mount -r "$blkdev" "$test_mntpt" >/dev/null 2>/dev/null
```
JonathonHall-Purism, line 182: Per my other comment, we need to go back to the strategy of checking if the device has a partition table first, and using only the partitions if it does.

I liked the idea of doing a mount test - maybe we could apply that to the partitions or to unpartitioned media in the future - but since it can apparently find remnant filesystems after repartitioning I don't think we can go this route.

JonathonHall-Purism (changes requested):

P5vc:
@JonathonHall-Purism Thank you so much for the feedback! I spent the latter half of my day tweaking these commits per your comments, and then conducting extensive testing. I also managed to go down an embarrassing number of rabbit holes in the process, but hey, it made for a fascinating and educational day! :) I should hopefully be ready to push these new changes in the next couple of days, schedule permitting, and will likely have a couple of follow up questions for you as well. Thanks again!

tlaurion:
Main question on my side is: what other OS is doing hybrid ISOs that would not boot from heads if user guided into detach signing them.

That would be https://github.com/linuxboot/heads/issues/1438#issue-1798894993 instead.

The reasoning being: should ISOs dd'ed or prepared by external tools be deprecated and replaced by booting ISOs that don't show the bug in question? Is supporting raw iso over USB thumb drive simply a need of users not aware that booting from ISOs is better and would be improved by Heads trying to find iso.asc iso.sig for same iso and if not found, prompt user to verify sha256sum once and then detach sign that iso?

JonathonHall-Purism:
> should ISOs dd'ed or prepared by external tools be deprecated and replaced by booting ISOs that don't show the bug in question?

No!  Practically all instructions for installing any Linux distribution tell users to prepare a bootable USB flash drive this way.  It's great to support additional means, but if that most common method does not work in Heads, it will bring a lot of pain to users.

IMO, we should work toward closing the gap between "things that boot with BIOS/UEFI" and "things that boot with Heads", not widening it.

tlaurion (commented):
@JonathonHall-Purism rejected?

P5vc:
@tlaurion I'll go ahead and close this pull request. The original premise was mistaken, and although I ended up making some good changes, I got wrapped up in a bunch of other projects and haven't had a chance to revisit that work and update this pull request accordingly. In any event, those changes would probably be more appropriate to propose in separate pull requests. Thank you both again for your feedback and suggestions!

Issue #1539, opened 2023-11-25
Choose stronger encryption by default and/or re-use encryption parameters of LUKS container


UndeadDevel:
**Is your feature request related to a problem? Please describe.**
1. Currently re-encrypting the LUKS container from Heads seems to choose default parameters, which are no longer recommended (argon2i with 1 GB memory cost). E.g. the [IRTF recommendation](https://www.rfc-editor.org/rfc/rfc9106.html) states that argon2id with a 6 GB memory cost and parallelity of 4 is recommended for disk encryption; 6 GB might be problematic due to dom0 having only 4 GB of RAM by default, so it would be difficult to change the passphrase from dom0 later in that case (but maybe the user shouldn't do that anyway), but 2 GB would definitely work, or even 3 or perhaps 4, as there is typically 4 GB of swap in dom0 also. 2 GB is recommended in that RFC for general use cases. Other OSs Heads runs on shouldn't have a problem even with 6GB, but see below under "Additional context".
2. Another problem with the current, weaker defaults is the theoretical possibility of compromising the container using an old header or copy of the current volume encryption key encrypted with weaker parameters, even after the user has changed to stronger parameters later in dom0, if Heads is encrypting containers on COW media, such as SSDs, which are widely used these days and even recommended by QubesOS. See e.g. the [cryptseptup FAQ](https://gitlab.com/cryptsetup/cryptsetup/-/wikis/FrequentlyAskedQuestions#5-security-aspects) point 5.19 and others.
3. There's also an [open issue](https://gitlab.com/cryptsetup/cryptsetup/-/issues/785) in the cryptsetup gitlab that is related: does Heads *first* create the new key (for the keyslots, not the volume key) and *only then* re-encrypt (which always creates a new volume key first)? This is the order that is recommended, because otherwise the chance of the old keyslots lying around is greater and the new volume key would first be encrypted with weaker settings; when it is later re-encrypted in the header upon changing the params and keyslots, on COW media, such as SSDs, the old, more weakly encrypted copy of the new volume key may linger.

**Describe the solution you'd like**
[...]

PR #1542, opened 2023-12-01
Luks functions text patch & change order of reencrypt and passphrase change

UndeadDevel:
Fixes point 3. from #1539 and other minor points in discussion there. Summary of changes:

1. If user selects both reencrypt and change passphrase during OEM factory reset, then the passphrase will be changed first so as not to leave a copy of the new volume key encrypted with potentially compromised defaults (e.g. "PleaseChangeMe" passphrase from OEM) lying around on disk if disk is COW media such as SSD and does not garbage collect properly. This fixes point 3. from #1539.
2. Added check for exported new passphrase to reencrypt function, so that it uses the new one if it was set.
3. Removed all instances of incorrect term "Recovery Disk Key" and replaced with "Disk Recovery Key".
4. Changed the warning message of reencrypt function to better reflect the process and warn user not to reboot or shut off, which is especially important if we will use `--resilience=none` in the future.
5. Replaced occurrences of "TPM Disk unlock key" with "TPM Disk Unlock Key" to be more consistent with naming conventions.
6. Fixed a lot of grammatical errors.

### DIFF
--- a/initrd/bin/oem-factory-reset
+++ b/initrd/bin/oem-factory-reset
@@ -1148,8 +1148,8 @@ fi
 
 if [ -n "$luks_new_Disk_Recovery_Key_desired" -a -n "$luks_new_Disk_Recovery_Key_passphrase_desired" ]; then
     #Reencryption of disk, disk recovery key and Disk Recovery Key passphrase change is requested
-    luks_reencrypt
     luks_change_passphrase
+    luks_reencrypt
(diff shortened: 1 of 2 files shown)


tlaurion:
If terminology is to be changed, then Disk Recovery Key could be changed to Disk Installation Encryption key or something more generic. Reviewing this PR adds a new term being Disk Key which I think will confuse even more?

JonathonHall-Purism (approved):
Thanks, look good to you too @tlaurion ?

tlaurion:
Better to change passphrase first yes. Merging

PR #1543, opened 2023-12-01
Use a wider window to show the secrets

### LINKED ISSUE #1537: Summary window of all secrets after OEM factory reset doesn't display long passphrases in full
### A. Provide Hardware Details

Nitropad NV41 with iGPU and Nitrokey 3A Mini using HOTP.

### B. Identify how the board was flashed

Flashed at OEM with Heads 2.1, not updated so far.

### C. Identify the rom related to this bug report

Nitrokey Heads 2.1 maximized.

## Please describe the problem

**Describe the bug**
When doing the OEM factory reset on Heads 2.1 there will be a window that summarizes and displays all the secrets (PINs and passphrases) that were chosen. However, when a very long passphrase (about 60 characters or more) is used, it does not seem to be able to change in size or break the passphrase across multiple lines, but instead the passphrase is shown incompletely (cut off at the window's right side).

**To Reproduce**
Do OEM factory reset and choose a very long disk recovery passphrase.

**Expected behavior**
The window should either "grow" to fit the passphrase or break the passphrase across multiple lines.

UndeadDevel:
This partially fixes #1537, but while the increased width wouldn't be a problem on the NV41 AFAICT, I don't know about other machines.

I don't know what @tlaurion means with "busybox's folding", which may be a better solution.

### DIFF
--- a/initrd/bin/oem-factory-reset
+++ b/initrd/bin/oem-factory-reset
@@ -1326,8 +1326,8 @@ fi
 
 ## Show to user current configured secrets prior of rebooting
 whiptail --msgbox "
-    $passphrases" \
-    $HEIGHT $(($WIDTH*2)) --title "Configured secrets"
+    $(echo "$passphrases" | fold -w $((WIDTH-20)))" \
+    $HEIGHT $WIDTH --title "Configured secrets"
 
 ## all done -- reboot
 whiptail --msgbox "
--- a/initrd/bin/oem-factory-reset
+++ b/initrd/bin/oem-factory-reset
@@ -1326,7 +1326,7 @@ fi
 
 ## Show to user current configured secrets prior of rebooting
 whiptail --msgbox "
-    $(echo "$passphrases" | fold -w $((WIDTH-20)))" \
+    $(echo -e "$passphrases" | fold -w $((WIDTH-5)))" \
     $HEIGHT $WIDTH --title "Configured secrets"
 
 ## all done -- reboot
--- a/initrd/bin/oem-factory-reset
+++ b/initrd/bin/oem-factory-reset
@@ -1327,7 +1327,7 @@ fi
 ## Show to user current configured secrets prior of rebooting
 whiptail --msgbox "
     $passphrases" \
-    $HEIGHT $WIDTH --title "Configured secrets"
+    $HEIGHT $(($WIDTH*2)) --title "Configured secrets"
 
 ## all done -- reboot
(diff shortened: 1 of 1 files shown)


tlaurion:
fold is a normal command which permits to put output on multiple lines.

You cn try it on bash command line on common OSes.

String="This is a really too long string of too many characters that won't display on a single line to be shown on a single line and needs to be folded to be showed correctly without hacking around"

echo "$String" | fold

Fold accepts parameters, which are different from busybox. Drop on recovery shell and try fold --help

UndeadDevel:
Thanks for the clarification! Changed to fold-based solution now; tested in recovery shell of NK Heads 2.1.

JonathonHall-Purism:
Thanks for checking this out!

---

I think this needs `echo -e` just before fold, i.e. `echo -e "$passphrases" | fold <...>`.  Otherwise fold could break the `\n` and it will break later lines incorrectly not realizing there was a line break escape (it would think it is one giant line).  The '\n' is handled by whiptail currently.

Adding echo -e might compound problems if passphrases themselves contain `\` (now they'll be incorrectly substituted _twice_) but that is already broken unfortunately, so I think `echo -e` is OK to make this mergeable.

I didn't test any of it, that's what it looks like from review, please correct me if I missed something :wink: 

---

IMO, I don't think we need to fold to WIDTH-20 (=60) columns, I think we can use all 80 columns.  With 60 columns, some reasonable-length passphrases will word wrap unnecessarily.

We're spending 35 chars on `LUKS Disk Recovery Key passphrase: `, at 60 that leaves just 25 for the passphrase before it wraps.  Most 4-word passphrases with no separators are longer than that.  At width 80 we'd have 45 chars.

UndeadDevel:
@JonathonHall-Purism Thank you for the feedback! You are right about `grep` needing `-e`; I missed that in testing at first; it's not just about backslash, but also `;` chars. Regarding the `WIDTH`, however, I found that using `80` still hides some characters, because whiptail includes a margin around the edges, within which characters are invisible, but after some more experimentation in recovery shell I found that subtracting 5 from `WIDTH` will fix this issue, i.e. use all available (visible) space but not miss any characters.

I also rebased and signed off now.

JonathonHall-Purism (approved):
Looks great to me!  @tlaurion any objections?

tlaurion:
Calculations seem right, cannot test but should be improvement nonetheless. Merging

PR #1541, opened 2023-11-28
Cryptsetup toolstack version bump (2.3.3-> 2.6.1) + reencryption cleanup (LUKS v1/v2 proper support + reencryption on Q4.2 + BTRFS dual LUKS containers install)

tlaurion:
- Bump kernel from 5.10.5 -> 5.10.214
  - Cloudfare patches to speed up LUKS encryption were upstreamed into linux kernel and backported to 5.10.9+. 
    - See https://github.com/cloudflare/linux/issues/1#issuecomment-763085915
  - cryptsetup2 toolstack version bump and script fixes to support multi-LUKS containers (BTRFS QubesOS 4.2)
    - Argon2 used externally and internally: requires a lot of RAM and CPU to derivate passphrase to key validated in key slots.
      - This is used to rate limit efficiently bruteforcing of LUKS key slots, requiring each offline brute force attempt to consume ~15-30 seconds per attempt
      - Of course, strong passphrases are still recommended, but bruteforcing LUKSv2 containers with Argon2 would require immense time, ram and CPU even to bruteforce low entropy passphrase/PINs.
     - passphrase change doesn't reuse LUKS key slot  anymore: cryptsetup passhprase change enforces key slot rotattion (new one consusumed per hot operation: the old one wiped internally after change. EG: LUKS key slot 1 created, then 0 deleted)
     - reencryption doesn't permit old call arguments. No more direct-io; inadmissibly slow through AIO (async) calls, need workarounds for good enough perfs (arguments + newer kernel with cloudfare fixes in tree)
  - cryptsetup required whole toolstack version bump:
    - cryptsetup requires lvm2 2.03.23+
    - cryptsetup requires libaio, which is also included in this PR (could be hacked out but deep dependency at first sight: left in, **deprecates all legacy boards**)
       - which requires util-linux 2.39
     - patches for reproducible builds are included for above 4 packages.

- luks-functions was updated to support the new cryptsetup 2.6.1 version calls/changes as opposed to old 2.3 in master
  - reencryption happen in direct-io, offline mode and without locking, requiring linux 5.10.9+ to bypass linux queues
    - from tests, this is best for performance and reliability in single-user mode, as before, op cannot be interrupted.
  - LUKS container ops now validate Disk Recovery Key (DRK) passphrase and DRK key slot prior of going forward if needed, failing close early.
    - Heads don't expect DRK to be in static key slot anymore, and finds and keeps track of the used DRK key slot dynamically.
[...]

PR #1547, opened 2023-12-07
Improve reencryption UX

### LINKED ISSUE #1538: Improve LUKS container re-encryption and OEM factory reset process
**Is your feature request related to a problem? Please describe.**
1. When doing the LUKS container re-encryption, either by itself or as part of OEM factory reset, the new password(s) are displayed as plain text on screen and will stay there until the process is complete, which can take some time with a large drive that needs to be fully re-encrypted; this is fine when there is no chance of anyone seeing the screen in this time, but many people live together with others in the same space and in those cases it's inconvenient to have the secrets displayed for so long and have to cover the screen with a towel or similar.
2. After doing the re-encryption the user must sign all checksums again, but I didn't see this being communicated (maybe I missed it?). This should be part of the information given to the user or the user should be prompted to do it after the process completes (this is regarding LUKS re-encryption by itself, as I think OEM factory reset already does that).

**Describe the solution you'd like**
1. I would suggest clearing the screen at once (perhaps a warning about this and ask user confirmation; there is still the summary window anyway, though currently there's a [bug](https://github.com/linuxboot/heads/issues/1537) about it).
2. Prompt the user to sign all files and checksums after doing a LUKS-container re-encryption.

**Describe alternatives you've considered**
1. Perhaps having the re-encryption process "push" the text up (i.e. insert newlines) periodically, e.g. once every 10 seconds so that the secrets will all disappear from screen in 2 or 3 minutes.
2. Put some info on screen telling user to sign checksums after LUKS container re-encryption.

UndeadDevel:
Fixes #1538 

With the first commit, which addresses point 1 in #1538:
- Refactored the `luks_reencrypt` function (superfluous else clause with no unique code)
- Added `clear` to the `luks_reencrypt` function to hide the new secrets as re-encryption starts to avoid having the secrets stuck on the screen in plaintext. Since #1543 fixed the final secrets whiptail window, it's appropriate to remove the secrets from screen for the potentially long re-encryption process so that people aren't stuck trying to cover their screen for an hour or more if they are not physically isolated.

This PR also addresses point 2 in #1538, using the alternative solution suggested there, namely it informs the user that after a passphrase change or re-encryption the checksums need to be updated and the TPM resealed, if applicable, thus giving pointers to an inexperienced user, who may be wondering why the default boot fails after using the separate "change passphrase" or "re-encrypt LUKS container" functions.


initrd/bin/gui-init
```diff
@@ -465,10 +465,12 @@ show_options_menu()
     C )
       luks_reencrypt
       luks_secrets_cleanup
+      prompt_update_checksums
```
tlaurion, line 468: Is that enough? Afk still.

tlaurion (changes requested):
@UndeadDevel reviewed!

initrd/bin/gui-init
UndeadDevel: Oh right, I didn't think of that case. I guess there should be a window asking the user whether a TPM DUK should be set up again after passphrase change? Does there need to be a new TPM DUK if only the volume key was changed (I think no, but I'm not sure)?

initrd/etc/luks-functions
```diff
@@ -349,14 +349,13 @@ luks_reencrypt() {
...
-			cryptsetup-reencrypt -B 64 --use-directio "$LUKS" --key-slot 0 --key-file /tmp/luks_current_Disk_Recovery_Key_passphrase
 		fi
+		echo -n "$luks_current_Disk_Recovery_Key_passphrase" >/tmp/luks_current_Disk_Recovery_Key_passphrase
+		#make secrets disappear from screen as reencryption can take a long time (we show these to the user again later in whiptail anyway)
+		printf "\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n"
```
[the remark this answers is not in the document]
tlaurion: Damnit.
2.1 nitrokey doesn't include it
https://github.com/Nitrokey/heads/blob/f7019e80af3549fb38a5f661534b851a7bf9404f/config/busybox.config#L364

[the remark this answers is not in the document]
tlaurion: It was added under heads under https://github.com/linuxboot/heads/commit/d3ec6ac11fa2b68a41fc1d572a998337054b3c66

So 2.1 nitrokey release is July and adding clear is October.

initrd/bin/gui-init
tlaurion: Header will change on either pass change or reenncryption.

gui-init calls this function when tpm is reset or totp is resealed, doing needed magic

https://github.com/linuxboot/heads/blob/6b936e76aa941841148d34b1626c5fd2dd259617/initrd/etc/functions#L83

UndeadDevel: > Header will change on either pass change or reenncryption.

Agreed. But that wasn't my question; I was wondering if a new TPM DUK needs to be generated if the user only does a re-encryption of the LUKS container, without changing his passphrase (I'm thinking no, but wanted confirmation). For the case (or cases) where it does need to be set, the reseal function seems to be the correct one, as it also calls `kexec-seal-key`, which will generate a new TPM DUK, though it wouldn't be the best UX, as that function will ask the user to enter the Disk Recovery Key passphrase, which he had just entered earlier, at the start of the process.

Anyway, the TPM adds quite a bit of complexity to this PR...maybe I should remove that commit and make the PR only about the "clear" command? I can't test any of the code that involves TPM DUK stuff.

tlaurion: >>Header will change on either pass change or reenncryption.
>Agreed. But that wasn't my question; I was wondering if a new TPM DUK needs to be generated if the user only does a re-encryption of the LUKS container, without changing his passphrase (I'm thinking no, but wanted confirmation). 

@UndeadDevel : the sealing function takes into consideration hash of luks header, so yes, in both cases TPM DUK needs to be resealed as per PCR6 measurement

https://github.com/linuxboot/heads/blob/6b936e76aa941841148d34b1626c5fd2dd259617/initrd/bin/kexec-seal-key#L134

Created by
https://github.com/linuxboot/heads/blob/6b936e76aa941841148d34b1626c5fd2dd259617/initrd/bin/qubes-measure-luks#L18

and then part of TPM DUK unsealing at 
https://github.com/linuxboot/heads/blob/6b936e76aa941841148d34b1626c5fd2dd259617/initrd/bin/kexec-unseal-key#L36

Called by
https://github.com/linuxboot/heads/blob/6b936e76aa941841148d34b1626c5fd2dd259617/initrd/bin/kexec-insert-key#L46

UndeadDevel:
Okay so I fixed `clear` vs. `printf` as discussed.

The other issue I'm not touching beyond adding a message to the user as per my alternative suggestion in #1538. The reason is that too much would have to be refactored in order not to make UX worse (e.g. by having user enter the same new passphrase multiple times) and I can't test anything that involves TPM DUK functionality.

tlaurion:
@UndeadDevel please update op olcomment reflecting stare. Will test in next following days

UndeadDevel:
Done.

tlaurion (approved):

tlaurion:
Need to test but LGTM

tlaurion:
Was included as part of https://github.com/linuxboot/heads/pull/1541. Closing

PR #1554, opened 2023-12-16
WiP: Ease cohesion of Heads expected features and rebranding

tlaurion:
It is now clear that users use interchangeably upstream Heads and downstream Pureboot/Nitrokey forks produced roms.
As of today, features of expected upstream Heads are not generally available across boards depending of brand related policy decisions. The question arises on why users would switch between firmware sources if its not to test features of the different brandings. Therefore, upstream (Heads) need to be consistent in features that are deployed by default, and branding needs to enable/disable such features in their forks board configurations so that experienced users can at least see what is different in the different repositories.

-----

Boards config sources: 
- Generic second hand buyable machines without specific branding (full featureset)
- Librems (disable TPM Disk Unlock Key usage and on screen QR code presentation on sealing since USB security dongle usage expected upstream)
- Nitropads disable TPM Disk Unlock Key usage upstream)

Therefore, the following config options should be reverted upstream (Heads) and changed in forks:
- export CONFIG_TPM_NO_LUKS_DISK_UNLOCK=y
- export CONFIG_TOTP_SKIP_QRCODE=y

This PR inverts those choices that do not fit upstream featureset enabled by default to ease downstream forks to be reapplied with a simple sed and a single patch to be maintained on top of Heads. 

This might not be the best way to do this, but this is a WiP to expose the problem and propose a solution to fix it.
@JonathonHall-Purism @daringer @jans23: let's debate in code/counter propositions. )


JonathonHall-Purism:
@tlaurion I agree that upstream builds should be consistent across different boards.  The changes in config here are minor too, I don't think any upstream Heads users would mind the change (and they can be added to config.user to restore the customized behavior).

I don't think it's a good strategy though to have to maintain those distribution customizations spread out among lots of boards.  I don't think BRAND_NAME= actually works in board configs currently (ROM names are determined using BRAND_NAME before board config is included).

Instead I think we could maintain distribution-specific configuration in a specific file that's maintained downstream and not board-specific, like coreboot does with `site-local`.  Put together a PR here (but still checking CI failures, so draft for the moment, works on qemu and Librem 13v4): #1556

If we go with that strategy, then I think you should delete those `export CONFIG_...=y` lines from the affected boards, no need to duplicate the default value.

daringer:
> Instead I think we could maintain distribution-specific configuration in a specific file that's maintained downstream and not board-specific

+1 this might also help to maintain downstream changes while touching less upstream files

tlaurion:
@daringer @JonathonHall-Purism I love #1556 approach. Will adapt this PR to remove unneeded customizations, so those two approaches become complementary

Next commit will remove the current changes:
- Upstream boards will not deactivate TPM DUK
- Upstream will not force BRAND_NAME which currently defaults to Heads
- Upstream will not deactivate Qr code on screen output on HOTP sealing

Those will need to be overridden downstream.

JonathonHall-Purism (approved):
This looks good to me, we'll be able to maintain the config customizations downstream using site-local/config in #1556 :+1:

PR #1556, opened 2023-12-18
Makefile: Allow downstreams to put local customizations in site-local/config

JonathonHall-Purism:
This relates to #1554 .

Instead of maintaining downstream customizations spread among many board configs, allow downstreams to create `site-local/config` (which can be versioned normally).  Include `site-local/config` at a strategic point in `Makefile` to allow it to override board options and BRAND_NAME.  Other things might get hooks in site-local later as well (maybe `site-local/initrd`, etc.)  The same strategy is used in coreboot: https://doc.coreboot.org/tutorial/managing_local_additions.html

For an example of brand customizations using this mechanism, see this commit: https://github.com/JonathonHall-Purism/heads/commit/e550e37f0e77100b186bc6db19ac873df2dbdb20

The sample `AcmeBoot` brand uses text-based whiptail on all boards and has its own bootsplash.  No changes to board configs need to be maintained downstream.  A brand-specific configuration variable is exported to /etc/config as well to demonstrate the mechanism.

ROM output names were moved later to allow site-local/config to override BRAND_NAME (and still override variables from board config).

Several boards used the ROM output names in the board config, almost all of these were in duplicated logic that was moved to targets/:
* qemu boards (targets to prepare and run in qemu)
* 12MB boards (split top/bottom ROMs)
* legacy-flash boards (top ROMs)
* talos-2 (TGZ update package)

This ended up requiring more changes than I initially expected, but it was a good opportunity to deduplicate a lot of code.  Qemu documentation was moved to targets/qemu.md as well.


JonathonHall-Purism:
Looking at CI failures, draft until fixed

JonathonHall-Purism:
There were more boards using ROM output names in the board config than I realized.  That was an issue because there is no possible point where site-local could override BRAND_NAME and board config; BRAND_NAME had to be defined before the board config and could not be changed at that point.

Almost all of this was logic duplicated among many boards, so moving these things to targets/ and including it a little later in Makefile deduplicated a lot of code.

tlaurion:
@daringer @JonathonHall-Purism As long as downstream are clear about this.
Should we collaborate into creating a documentation page under heads-wiki to make this clear for downstream first and then when agreed, merge this after that documentation page created?

tlaurion (commented):
I love this. Way simpler approach.
Only thing missing is documentation page under heads-wiki to make it clear for any downstream fork to deal properly with this.

@JonathonHall-Purism Can you create and we can collaborate clearly until @daringer knows clearly what to do as direct stakeholder/user of this feature?

tlaurion:
@JonathonHall-Purism might merge master into this first so this PR is a representation of what needs to be adapted?

JonathonHall-Purism:
@tlaurion Merged master (no conflicts); heads-wiki: https://github.com/linuxboot/heads-wiki/pull/137

tlaurion:
> @tlaurion Merged master (no conflicts); heads-wiki: [linuxboot/heads-wiki#137](https://github.com/linuxboot/heads-wiki/pull/137)

Perfect to me. Merged.

@daringer : Can you look at https://osresearch.net/Distributions/ and tell use if that answers all your questions for next NK release? Will ping you on matrix. If that's ok, merging.

daringer:
> @daringer : Can you look at https://osresearch.net/Distributions/ and tell us if that answers all your questions for next NK release? Will ping you on matrix. If that's ok, merging.

This looks really good for me, good stuff, thx @JonathonHall-Purism

tlaurion (approved):
LGTM

PR #1576, opened 2024-01-05
Makefile: Support mirrors for dependency source packages

### LINKED ISSUE #1198: Have all modules point to either github/debian/gentoo/alternatives archives for tarballs

JonathonHall-Purism:
Try to download dependency source packages from mirrors if the primary source fails or the archive has changed.

Move the download and verify logic to bin/fetch_source_archive.sh.  The mirror list is here, currently only
https://storage.puri.sm/heads-packages/, but others can be added.  The mirror list is randomized to load each mirror equally.

The verify logic is moved to this script too so it can fail over to a mirror (or another mirror) if a mismatched archive is served, not just for a failure.  Makefile no longer needs to verify separately and there are no separate .*-_verify files any more, the archive is only moved to its final place once verified.

Add `packages` target to just fetch all needed packages for a board, facilitates seeding a mirror.

Closes #1198

### DIFF
--- a/bin/fetch_coreboot_crossgcc_archive.sh
+++ b/bin/fetch_coreboot_crossgcc_archive.sh
@@ -25,7 +25,7 @@ fi
 
 COREBOOT_DIR="$1"
 PKG_NAME="$2"
-PKGS_DIR="$3"
+PKGS_DIR="$(realpath "$3")" # Make sure it's an absolute path
 
 # Get the result of a glob that should match a single thing, or die if it
 # doesn't match exactly one thing.
@@ -126,5 +126,7 @@ BIN_DIR="$(dirname "${BASH_SOURCE[0]}")"
 # Copy to the tarballs directory so coreboot's toolchain build will use this
 # archive
 mkdir -p "$COREBOOT_DIR/util/crossgcc/tarballs"
-cp "$PKGS_DIR/coreboot-crossgcc-$PKG_BASENAME" \
-	"$COREBOOT_DIR/util/crossgcc/tarballs/$PKG_BASENAME"
+(
+	cd "$COREBOOT_DIR/util/crossgcc/tarballs"
+	ln -s "$PKGS_DIR/coreboot-crossgcc-$PKG_BASENAME" "$PKG_BASENAME"
+)
--- a/bin/fetch_coreboot_crossgcc_archive.sh
+++ b/bin/fetch_coreboot_crossgcc_archive.sh
@@ -1,0 +1,130 @@
+#! /usr/bin/env bash
+set -eo pipefail
+
+usage()
+{
+	cat <<USAGE_END
+usage:
+	$0 <coreboot-dir> <pkg-name> <pkgs-dir>
+	$0 --help
+
+	Downloads the source archive for <pkg-name> needed by coreboot's
+	crossgcc toolchain.  The package version and digest are found in
+	coreboot-dir.  The package is downloaded to <pkgs-dir>, then placed in
+	the coreboot directory.
+
+	Uses fetch_source_archive.sh, so mirrors are used and WGET can override
+	the path to wget.
+USAGE_END
+}
+
+if [ "$#" -lt 3 ]; then
+	usage
+	exit 1
+fi
+
+COREBOOT_DIR="$1"
+PKG_NAME="$2"
+PKGS_DIR="$3"
+
+# Get the result of a glob that should match a single thing, or die if it
+# doesn't match exactly one thing.
+single() {
+	if [ "$#" -eq 1 ]; then
+		if [ -f "$1" ]; then
+			echo "$1"
+		else
+			echo "$1: no matches" >&2
+			exit 1
+		fi
+	else
+		echo "multiple unexpected matches for glob:" "$@" >&2
+		exit 1
+	fi
+}
+
+# Delete prefix and suffix from a value
+delete_prefix_suffix() {
+	local value prefix suffix
+	value="$1"
+	prefix="$2"
+	suffix="$3"
+	value="${value/#$prefix/}"
+	value="${value/%$suffix/}"
+	echo "$value"
+}
+
+# Find the checksum file for this package
+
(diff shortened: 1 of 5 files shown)


JonathonHall-Purism:
I tested this initially by creating a dummy directory locally full of the packages, ran a web server with `python -m http.server`, and used that as a mirror.  Also created a directory full of bad files with the same names to test the mismatch failover.  Tested the download script directly quite a few times and ran a few times from the Makefile.

I synced packages to https://storage.puri.sm/heads-packages and tested fetching everything for x230-maximized from the mirror (commented out the primary download in the script).

I seeded the package directory with:
```
make packages BOARD=qemu-coreboot-fbwhiptail-tpm1-hotp
make packages BOARD=talos-2 # newt, PPC
make packages BOARD=librem_l1um_v2 # TPM2
make packages BOARD=librem_l1um # coreboot 4.11
make packages BOARD=x230-maximized # io386
cp packages/ppc64/* packages/x86/* <mirror-directory>
```

After that, `make packages BOARD=librem_14` didn't have to download anything so I think I have covered most of it.  Let me know if there is anything else I might have missed.  Even if a primary goes down and we realize it was missed, we can always add it to the mirror later without altering the build.

tlaurion:
@JonathonHall-Purism only thing I think about is dependencies of crossgcc dependencies of coreboot. I checked real quick https://github.com/coreboot/coreboot/commit/e59868c6b6b69262adad27686f8fbf410edba6d0 and https://github.com/coreboot/coreboot/commit/3b32af950d704b1d14ddca917dd480f5ebbf09a5 a d I'm afk for the moment.

Not sure how do deal with that but to hack our way into adding patches to coreboot under Heads.

tlaurion:
@JonathonHall-Purism only thing I think about is dependencies of crossgcc dependencies of coreboot. I checked real quick https://github.com/coreboot/coreboot/commit/e59868c6b6b69262adad27686f8fbf410edba6d0 and https://github.com/coreboot/coreboot/commit/3b32af950d704b1d14ddca917dd480f5ebbf09a5 a d I'm afk for the moment.

Not sure how to deal with that but to hack our way into adding patches to coreboot under Heads so that those tarballs (which are historically the ones which fail the most) are also cached.

JonathonHall-Purism:
> crossgcc dependencies of coreboot

:thinking: Yeah, that's not addressed right now.  Nor are git repo sources (but those don't seem to disappear as much).  I sent this up since it is an improvement already, IMO no sense holding back benefits for other things that are not made worse.

That said, I had a little inspiration here, maybe the coreboot toolchain's task could preload those archives using fetch_source_archive.sh before calling into coreboot's build for the toolchain.  I think if the archives are already there, coreboot will use them.  I should be able to make a little time today to try it.

JonathonHall-Purism:
Toolchain builds failed in CI for talos-2 and x230-hotp-maximized, checking it out

JonathonHall-Purism:
I missed one occurrence of $(COREBOOT_TOOLCHAIN) that is now $(COREBOOT_TOOLCHAIN_DIR)/.heads-toolchain, it worked locally because I was building the toolchain target explicitly.  Fixed

tlaurion:
@JonathonHall-Purism Awesome work. Your Make fu is inspiring, really. Will review tomorrow but there doesn't seem to have anything weird to my eyes.

bin/fetch_coreboot_crossgcc_archive.sh
```diff
@@ -0,0 +1,130 @@
...
+BIN_DIR="$(dirname "${BASH_SOURCE[0]}")"
+
+# Download to packages/<arch>
+"$BIN_DIR/fetch_source_archive.sh" "$PKG_BASEURL$PKG_BASENAME" \
+	"$PKGS_DIR/coreboot-crossgcc-$PKG_BASENAME" "$PKG_DIGEST"
+
+# Copy to the tarballs directory so coreboot's toolchain build will use this
+# archive
+mkdir -p "$COREBOOT_DIR/util/crossgcc/tarballs"
+cp "$PKGS_DIR/coreboot-crossgcc-$PKG_BASENAME" \
```
tlaurion, line 129: @JonathonHall-Purism `ln -s` ? `cp -a` ?

tlaurion (changes requested):

bin/fetch_coreboot_crossgcc_archive.sh
JonathonHall-Purism: Yeah a symlink should be fine here, might save me a few GB of disk space given all the PB/Heads workspaces I have :grin:   I'll change it

tlaurion: @JonathonHall-Purism cp -l might be better (hardlink)

JonathonHall-Purism: (Discussed out of band, agreed on symlink)

JonathonHall-Purism: Changed to `ln -s` in https://github.com/linuxboot/heads/pull/1576/commits/93ccf25d24bb73c89ff51e17d847bccc78bc61e2

tlaurion:
Building from clean cache at https://app.circleci.com/pipelines/github/tlaurion/heads/2208/workflows/101c01d7-f8e3-407b-a5d8-05eeae53fb4e

tlaurion:
Awesome work here.

PR #1580, opened 2024-01-09
gpg2: make sure dirmngr is not spawn to refresh keys under initrd/.gnupg/gpg.conf

tlaurion:
Otherwise under debian-12 qemu/talos II (not sure why not on other platforms?!):

```
~ # gpg --detach-sign /tmp/debug.log 
gpg: error running '//bin/dirmngr': probably not installed
gpg: failed to start dirmngr '//bin/dirmngr': Configuration error
gpg: can't connect to the dirmngr: Configuration error
gpg: no default secret key: No dirmngr
gpg: signing failed: No dirmngr

```

Alternative: pack dirmngr, but as of now, doesn't make sense since we implicitely state we do not want dirmngr but that isn't respected. When/if we need key download and have on-demand networking for more advanced/dynamic Heads config: revisit.

### DIFF
--- a/initrd/.gnupg/gpg.conf
+++ b/initrd/.gnupg/gpg.conf
@@ -1,1 +1,2 @@
 use-agent
+keyserver-options no-auto-key-retrieve
--- a/modules/gpg2
+++ b/modules/gpg2
@@ -5,6 +5,7 @@ gpg2_dir := gnupg-$(gpg2_version)
 gpg2_tar := gnupg-$(gpg2_version).tar.bz2
 gpg2_url := https://www.gnupg.org/ftp/gcrypt/gnupg/$(gpg2_tar)
 gpg2_hash := 1d79158dd01d992431dd2e3facb89fdac97127f89784ea2cb610c600fb0c1483
+gpg2_depends := libgpg-error libgcrypt libksba libassuan npth libusb-compat $(musl_dep)
 
 # For reproducibility reasons we have to override the exec_prefix
 # and datarootdir on the configure line so that the Makefiles will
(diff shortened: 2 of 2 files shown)


JonathonHall-Purism:
@tlaurion So this only happens on talos-2, and only when firmware was built on Debian 12?  Probably gpg2 leaking something detected from the host into the build.  Change seems reasonable, let me give it a quick try.

tlaurion:
> @tlaurion So this only happens on talos-2, and only when firmware was built on Debian 12?  Probably gpg2 leaking something detected from the host into the build.  Change seems reasonable, let me give it a quick try.

Also on qemu

JonathonHall-Purism:
@tlaurion Works, will approve.  Looked into this a bit, have some thoughts but nothing concrete, this mitigation seems appropriate for hardening even if we figure out and fix the root issue.

I can reproduce this in qemu.  I wonder if the bit leaking into the build is actually coming from bin/inject_gpg_key.sh - it could be leaking something strange into the keyring or doing something incorrectly.

If you attempt to detach-sign in qemu just after OEM reset (without rebooting), you get a different error - `gpg: signing failed: No pinentry`.  It doesn't seem to try to invoke dirmngr.  I'm not sure what the difference is exactly.  I didn't see anything obviously wrong in the script compared to OEM reset.

I'm pretty sure it's not the gpg build itself, I built against Librem 14 too and verified that GPG and the libg* libraries are identical on both.  Still don't see this on Librem 14, which IMO points to the keyring.  Probably could try copying actual keyrings around to confirm, but at any rate this mitigation is good and makes sense even if we fix the root issue.

I can't explain talos-2 though (but also can't actually test it myself :wink: )

JonathonHall-Purism (approved):
Mitigation works in qemu for me, and makes sense, I think we would want this for hardening even if we later identify the root issue.

PR #1581, opened 2024-01-09
Show error messages for various media errors in flash-gui.sh

JonathonHall-Purism:
Show messages for media errors in flash-gui.sh, either when enumerating ROMs or reading a plain ROM.  (Reading a ZIP package is already OK, the unzip fails and it falls through to the integrity check error.)

A user had I/O errors when enumerating ROMs, which just returns to the main menu with no explanation.  (find errors will be on the recovery console.)  Show an error in that case suggesting to check the disk, reformat, or try a different disk.

I believe the user's case was due to filesystem corruption, but I can reproduce the same find errors with the following:
* With an ancient/slow flash drive, produce a lot of dummy files so the 'find' will take long enough to yank the disk
    * `for i in $(seq 1 50000); do dd if=/dev/urandom bs=1K count=1 of=<disk>/$(printf rom%05i.rom $i) status=none; done` (this ended after ~31K files for me, out of inodes on a 128 MB disk)
* Plug in this flash drive and at least one other (to be sure the disk is yanked after mounting)
* Go through the flash GUI menus, select the dummy disk
* While the 'find' is enumerating, yank the disk
    * master will just kick back to the main menu, errors can be seen on the console
    * PR will show a message

Additionally, show a similar message if the ROM itself is unreadable.  Copy the ROM to /tmp before using it, handle I/O errors there, then use the copy in /tmp from then on to avoid any problems with the medium.  This is reproducible just by yanking the disk during the ROM menu, then selecting any ROM.  This didn't occur in the field, but it's a similar failure mode.

### DIFF
--- a/initrd/bin/flash-gui.sh
+++ b/initrd/bin/flash-gui.sh
@@ -60,10 +60,19 @@ while true; do
       --yesno "You will need to insert a USB drive containing your BIOS image (*.zip or\n*.$UPDATE_PLAIN_EXT).\n\nAfter you select this file, this program will reflash your BIOS.\n\nDo you want to proceed?" 0 80); then
       mount_usb
       if grep -q /media /proc/mounts; then
+        # 'find' parameters to match desired ROM extensions
+        FIND_ROM_EXTS=(\( -name "*.$UPDATE_PLAIN_EXT" -o -type f -name "*.zip" \))
         if [ "${CONFIG_BOARD%_*}" = talos-2 ]; then
-          find /media ! -path '*/\.*' -type f -name "*.$UPDATE_PLAIN_EXT" | sort >/tmp/filelist.txt
-        else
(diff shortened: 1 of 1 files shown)


tlaurion:
@JonathonHall-Purism please sign all commits :)

tlaurion (changes requested):
LGTM if all commits were verified! (Will plug Talos II and test and merge when done)

JonathonHall-Purism:
Thanks @tlaurion , signed now.  It's odd, I cherry-picked these commits from my release branch, definitely had the -S (checked history), but I guess it only signed the last one.  I'll cherry pick one at a time next time :shrug:

tlaurion:
@JonathonHall-Purism Just because it would be easier to test everything and you just rebased/resigned: can you rebase on master so roms produced can be directly tested on Talos on next test run please?

JonathonHall-Purism:
Sure, rebased and pushed.  Yeah, all the ones from me were 'ready' at the same time, I was working on all these fixes together to reduce testing cycles for the next PureBoot release :grin:

tlaurion (approved):

PR #1582, opened 2024-01-10
bin/fetch_coreboot_crossgcc_archive.sh: ln -s again if link exists

JonathonHall-Purism:
Changes in things like modules/coreboot will check the coreboot toolchain archives again.  We reuse the cached archive already, but the final ln -s may fail if the link already exists.  Remove it first and link again.


tlaurion:
@JonathonHall-Purism Just hit that one rebuilding for talos-2. Thanks!

tlaurion (approved):

Issue #1588, opened 2024-01-12
Improve TPM DUK passphrase UX


UndeadDevel:
**Is your feature request related to a problem? Please describe.**
There are two issues IMHO:
1. Capslock doesn't work, but there is no warning, so if the user enters the passphrase using capslock for some characters during setup, thinking that those characters are now capitalized and then later during passphrase entry on boot enters those characters with Shift instead of capslock, the passphrase won't be accepted, because it's not the same.
2. When the wrong passphrase (for TPM DUK) is entered the op fails immediately, falling back to recovery key. Have mercy! Give the user at least another couple of tries in case of mistyping (is known to happen occasionally)...otherwise the presumably much more complex Disk Recovery Passphrase needs to be entered or a reboot triggered, which is bad UX.

**Describe the solution you'd like**
for 1.: either allow capslock under Heads or notify the user that capslock has no effect
for 2.: if TPM DUK unlocking fails due to bad passphrase, allow the user to retry a couple of times (maybe twice, so three tries in total).

**Describe alternatives you've considered**
Alternative for 1 is described above.
For 2: I guess returning to the main menu would be a better alternative than dropping to the recovery shell, though that may not be desirable as it makes it too easy to retry passphrases indefinitely.

UndeadDevel:
Had a situation again when I mistyped the TPM DUK passphrase but this time it didn't fall back to recovery shell for some reason (I still had to reboot from the main menu, but it was better UX-wise)...I'm not sure exactly what I did differently this time, but I suppose the more important issue is CAPSLOCK, which is at least partially covered by #1486.

tlaurion:
@UndeadDevel 
> I suppose the more important issue is CAPSLOCK

@UndeadDevel I really thought it would be easy to fix capslock, but it seems that we will need additional tooling to do this. 
One can expose for example the capslock leds, and/or see state of those lighs for keyboards having those leds... But in our case, we can expose the led but since EC is not toggling those leds because they actully don't exist, it seems we would have to interact with the tty to get the state of modifiers applied to see if capslock is triggered for next input. That would be through kbd tool, and or would need a new minimalist c tool being made to address this, inteacting with ioctl and reporting state to produce a warning and then add that test prior of all user input in code.

> When the wrong passphrase (for TPM DUK) is entered the op fails immediately, falling back to recovery shell. Have mercy! 

[...]

PR #1592, opened 2024-01-18
WiP: Fix TPM DUK retries/vocabilary (again) and CapsLock warning

tlaurion:
Addresses partly #1588 until proper CapsLock detection code is created.

TODO:

- [ ] Cleanup commit logs. This is raw at creation of PR: just works in Q4.1(lvm) and Q4.2(brtfs).
- [ ] Would love this to work out of the box to pass additional LUKS discovered LUKS devices, but as of now, Qubes asks for Disk Recovery Key if the whole setup chain for that additional device is not completed on OS side.
- [ ] So in current state, all LUKS discovered devices are part of "suggested" disks for user to select. This means that if those devices are setuped with different DRK, it will fail as well. TLDR: all DUK to be part of TPM DUK need to have the same DRK to be enrolled as TPM DUK and OS needs to have proper setup (/etc/crypptab+ dracut regen + grub regen).

Putting as draft upon creation.
@UndeadDevel : comments can be added either in issue or here if you have code comments(here) or other requirements(in issue).


tlaurion:
This contains WiP where #1595 will be more consensual for now. Leaving this one as draft and will rework when I have a fix for Capslock

PR #1595, opened 2024-01-19
Fix TPM DUK retry loop (bogus), uniformize related vocabulary

tlaurion:
#1588 related: fixes retry. Left behind: capslock key detection and fix.

Tested:
- lukskillslot of slot 1 when two available.
- lukschangepassword without specifying slot so that slot 0 LUKS Disk Recovery Key is put under slot1 where no slot 0 exist.
  -  on TPM DUK addition, warning showed to user to repair manual config which is unsupported
  - when manually fixed as suggested, both slot 1 and slot 0 exist and Heads detects and warns correctly on both LUSKv1 and LUKSv2
- TPM DUK loops for 1/3 to 3/3 and warns that capslock might be the cause of failing.
- After 3 bad attempts, normal codepath continue proposing user to type LUKS Disk Recovery Key password in OS.

Basically, the loop code was bogues because of pipefail, TPM DUK passphrase failing to unseal caused pipefail and exiting without retrying input as codepath expected.

TODO: find a nice way to get capslock state from tty. Only way found as of now would be through kbd, and extracting ioctl call to tty to get state of key modifiers. That is estimated to be 7kb of compiled c code that would need to be an independent app. I have not found any other easy way to get capslock state. Reminder, there is no capslock led on old lenovos, and exposing those leds through kernel modules successfully exposes capclock but since EC firmware doesn't interact in any way with abent leds, the state cannot be read from LEDs so not an alternative solution. If anybody has an idea that could go through exposed /sys /proc or even agetty/stty, I tried a lot but didn't succeed. @JonathonHall-Purism  ideas there?

### DIFF
--- a/initrd/bin/kexec-seal-key
+++ b/initrd/bin/kexec-seal-key
@@ -74,12 +74,31 @@ dd \
 for dev in $(cat "$KEY_DEVICES" | cut -d\  -f1); do
 	DEBUG "Checking number of slots used on $dev LUKS header"
 	#check if the device is a LUKS device with luks[1,2]
-	slots_used=$(cryptsetup luksDump $dev | grep -c 'luks[0-9]*' || die "Unable to get number of slots used on $dev")
+	# Get the number of key slots used on the LUKS header.
+	# LUKS1 Format is :
+	# Slot 0: ENABLED
+	# Slot 1: ENABLED
+	# Slot 2: DISABLED
+	# Slot 3: DISABLED
+	#...
+	# Slot 7: DISABLED
+	# Luks2 only reports on enabled slots.
+	# luks2 Format is :
+	# 0: luks2
+	# 1: luks2
+	#  Meaning that the number of slots used is the number of lines returned by a grep on the LUKS2 above format.
+	# We need to count the number of ENABLED slots for both LUKS1 and LUKS2
+	# create regex pattern for both LUKS1 and LUKS2
+	regex="Slot [0-9]*: ENABLED"
+	regex+="\|"
+	regex+="[0-9]*: luks2"
+	slots_used=$(cryptsetup luksDump "$dev" | grep -c "$regex" || die "Unable to get number of slots used on $dev")
+
 	DEBUG "Number of slots used on $dev LUKS header: $slots_used"
 	# If slot1 is the only one used, warn and die with proper messages
-	if [ $slots_used -eq 1 ]; then
+	if [ "$slots_used" -eq 1 ]; then
 		# Check if slot 1 is the only one existing
-		if cryptsetup luksDump $dev | grep -q "Slot 1: ENABLED"; then
+		if [ "$(cryptsetup luksDump "$dev" | grep -c "Slot 1: ENABLED")" -eq 1 ] || [ "$(cryptsetup luksDump "$dev" | grep -c "1: luks2")" -eq 1 ]; then
 			warn "Slot 1 is the only one existing on $dev LUKS header. Heads cannot use it to store TPM sealed LUKS Disk Unlock Key"
 			warn "Slot 1 should not be the only slot existing on $dev LUKS header. Slot 0 should be used to store Disk Recovery Key/passphrase"
 			die "You can safely fix this before continuing through Heads recovery shell: cryptsetup luksAddKey $dev"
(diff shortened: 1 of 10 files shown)


initrd/bin/kexec-unseal-key
```diff
@@ -25,22 +25,23 @@ DEBUG "CONFIG_TPM2_TOOLS: $CONFIG_TPM2_TOOLS"
 DEBUG "Show PCRs"
 DEBUG "$(pcrs)"
 
+failed=0
 for tries in 1 2 3; do
-	read -s -p "Enter LUKS Disk Unlock Key passphrase (blank to abort): " tpm_password
+	read -s -p "Enter LUKS LUKS TPM Disk Unlock Key passphrase (blank to abort): " tpm_password
```
UndeadDevel, line 30: LUKS word repeating.

initrd/bin/kexec-unseal-key
```diff
@@ -25,22 +25,23 @@ DEBUG "CONFIG_TPM2_TOOLS: $CONFIG_TPM2_TOOLS"
...
 
 	DO_WITH_DEBUG --mask-position 6 \
 		tpmr unseal "$TPM_INDEX" "0,1,2,3,4,5,6,7" "$TPM_SIZE" \
-		"$key_file" "$tpm_password"
+		"$key_file" "$tpm_password" || failed=1
 
-	if [ "$?" -eq 0 ]; then
+	if [ "$failed" -eq 0 ]; then
 		exit 0
 	fi
```
JonathonHall-Purism, line 42: It looks to me like if you enter an incorrect passphrase once, it won't exit correctly if you get it right on try 2, because `failed` is never reset to 0.  I didn't try it though, did you try it?

A better structure would be to test the command success directly:

```
if DO_WITH_DEBUG --mask-position 6 tpmr unseal <yada yada>; then
        exit 0
fi
```

initrd/bin/kexec-seal-key
```diff
@@ -74,36 +74,55 @@ dd \
...
+	# luks2 Format is :
+	# 0: luks2
+	# 1: luks2
+	#  Meaning that the number of slots used is the number of lines returned by a grep on the LUKS2 above format.
+	# We need to count the number of ENABLED slots for both LUKS1 and LUKS2
+	# create glob pattern for both LUKS1 and LUKS2
+	glob="Slot [0-9]*: ENABLED"
+	glob+="\|"
+	glob+="[0-9]*: luks2"
+	slots_used=$(cryptsetup luksDump "$dev" | grep -c "$glob" || die "Unable to get number of slots used on $dev")
```
JonathonHall-Purism, line 95:
```suggestion
	# We need to count the number of ENABLED slots for both LUKS1 and LUKS2
	# create regex for both LUKS1 and LUKS2
	regex="Slot [0-9]*: ENABLED"
	regex+="\|"
	regex+="[0-9]*: luks2"
	slots_used=$(cryptsetup luksDump "$dev" | grep -c "$regex" || die "Unable to get number of slots used on $dev")
```

I like breaking up the pattern this way but it is technically a regex, not a glob :grin:

initrd/etc/luks-functions
```diff
@@ -339,30 +339,30 @@ luks_reencrypt() {
...
-			echo -e "\nEnter the current Disk Recovery Key passphrase:"
+			--msgbox "This will replace the encrypted container content and its LUKS Disk Recovery Key.\n\nThe passphrase associated with this key will be asked from the user under\nthe following conditions:\n 1-Every boot if no Disk Unlock Key was added to the TPM\n 2-If the TPM fails (hardware failure)\n 3-If the firmware has been tampered with/upgraded/modified by the user\n\nThis process requires you to type the current LUKS Disk Recovery Key passphrase\nand will delete the LUKS TPM Disk Unlock Key slot, if set up, by setting a default\n boot LUKS key slot (1) if present.\n\nAt the next prompt, you may be asked to select which file corresponds to\nthe LUKS device container.\n\nHit Enter to continue." 0 80
```
JonathonHall-Purism, line 349: Did you check this prompt with fbwhiptail to make sure the lines don't run off the edge?

JonathonHall-Purism (changes requested):
Thanks @tlaurion, good find, improvements to wording consistency are great too.  Posted a few notes but generally looks good.  Not sure if the fix is 100% correct from a read over but made a suggestion, details on kexec-unseal-key.

initrd/etc/luks-functions
tlaurion: I thought the 0 made us able to dismiss those checks. Checking

tlaurion:
@JonathonHall-Purism all comments should be addressed!

initrd/etc/luks-functions
JonathonHall-Purism: @tlaurion So this prompt is OK in fbwhiptail?

initrd/bin/kexec-unseal-key
JonathonHall-Purism: @tlaurion Please fix this too even though I overlooked it.  (Good catch @UndeadDevel)

initrd/etc/luks-functions
tlaurion: @JonathonHall-Purism I confirm that setting 0 0 instead of 0 80 doesnt't have any impact with fbwhiptail. Playing with fold to do PoC that can be reused later on. This simplifies a lot prompt preparation, actually.

tlaurion: @JonathonHall-Purism Look at commit log of https://github.com/linuxboot/heads/pull/1595/commits/3f956f57847941064104bda80c80eefb40848f66 for unifirmization notes for later on of long fbwiptail messages

initrd/bin/kexec-unseal-key
tlaurion: @UndeadDevel thanks! @JonathonHall-Purism fixed under [4f2b1b6](https://github.com/linuxboot/heads/pull/1595/commits/4f2b1b68b0081c9e37526a8f8359d1a762a7c2ba)

tlaurion:
@JonathonHall-Purism Ready for merge?

PR #1602, opened 2024-02-02
init: Silence exFAT errors when mounting iso9660; reorder exfat last

JonathonHall-Purism:
Since exFAT support was enabled, spurious exFAT errors appear in the console in some cases:
* mounting iso9660 hybrid USB
* searching for /boot when a bios-grub, LVM, or LUKS partition precedes boot

In both cases, avoid trying to mount things as exFAT that aren't exFAT.  exfat is moved below iso9660 in automatic mount detection by creating /etc/filesystems (using /proc/filesystems).  Detecting /boot skips some things it knows are not mountable as /boot (bios-grub, LUKS, LVM PV), and also avoids trying the same device twice.  Details in commits.

Tested with mount-usb:
* mount iso9660 hybrid image on USB, verify no more exFAT errors
* mount exFAT partition
* format exFAT over iso9660 hybrid image with mkfs.exfat in Heads (without partition table), mount, verify new exFAT is found correctly (not remnant iso9660)
* format exFAT over iso9660 as above, but with mkfs.exfat from Debian 12
* mount ext4 partition

Tested with detect_boot_device:
* GPT disk with bios-grub, ext4 /boot, and LUKS root - tested setting CONFIG_BOOT_DEV to each to verify bios-grub and LUKS do not emit exFAT errors
* MBR disk with ext4 /boot and LVM root (Qubes default) - tested setting CONFIG_BOOT_DEV to each to verify the LUKS PV does not emit exFAT errors

### DIFF
--- a/initrd/bin/root-hashes-gui.sh
+++ b/initrd/bin/root-hashes-gui.sh
@@ -159,28 +159,6 @@ check_root_checksums() {
   fi
 }
 
-# Check if a device is an LVM2 PV, and if so print the VG name
-find_lvm_vg_name() {
-  TRACE_FUNC
-  local DEVICE VG
-  DEVICE="$1"
-
-  mkdir -p /tmp/root-hashes-gui
-  if ! lvm pvs "$DEVICE" >/tmp/root-hashes-gui/lvm_vg 2>/dev/null; then
-    # It's not an LVM PV
-    return 1
-  fi
-
-  VG="$(tail -n +2 /tmp/root-hashes-gui/lvm_vg | awk '{print $2}')"
-  if [ -z "$VG" ]; then
-    DEBUG "Could not find LVM2 VG from lvm pvs output:"
-    DEBUG "$(cat /tmp/root-hashes-gui/lvm_vg)"
-    return 1
-  fi
-
-  echo "$VG"
-}
-
 # Open an LVM volume group, then continue looking for more layers in the 'root'
 # logical volume.
 open_block_device_lvm() {
--- a/initrd/etc/functions
+++ b/initrd/etc/functions
@@ -154,6 +154,28 @@ enable_usb_storage() {
 	fi
 }
 
+device_has_partitions() {
+	local DEVICE="$1"
+	# fdisk normally says "doesn't contain a valid partition table" for
+	# devices that lack a partition table - except for FAT32.
+	#
+	# FAT32 devices have a volume boot record that looks enough like an MBR
+	# to satisfy fdisk.  In that case, fdisk prints a partition table header
+	# but no partitions.
+	#
+	# This check covers that: [ $(fdisk -l "$b" | wc -l) -eq 5 ]
+	# In both cases the output is 5 lines: 3 about device info, 1 empty line
+	# and the 5th will be the table header or the invalid message.
+	local DISK_DATA=$(fdisk -l "$DEVICE")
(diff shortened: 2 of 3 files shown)


JonathonHall-Purism:
Testing on more devices I found another situation where spurious exFAT errors appear, and this PR makes them worse :facepalm:  Setting to draft while I solve that.

On Librem 11, the default OS image has a GPT partition table with /dev/nvme0n1p1 as the bios-grub partition (holds stage 2), the actual /boot is /dev/nvme0n1p2.  But CONFIG_BOOT_DEV is defaulted to /dev/nvme0n1p1 and it relies on autodetect to find out that it's actually nvme0n1p2.

detect_boot_device tries to mount /dev/nvme0n1p1 twice (once since it's CONFIG_BOOT_DEV, then again when trying all partitions in order).  This used to generate two sets of exFAT errors since we can't mount that partition at all.  Now it generates four, since each mount tries exFAT twice (once from /etc/filesystems, once from /proc/filesystems).  There's no point trying both lists here, but busybox always uses both (GNU coreutils' mount only uses both lists if /etc/filesystems ends with a '*' line).

tlaurion:
I tested real quick under debian-12 on qemu and boot messages were still showing this for vda1 (normal since non-supported fs by Heads. Ugly but this is how it is).

Tried to rewrite to /proc/filesystem but changes didnt stick. Now comparing linux configs for cleanup and make differences between configs only as needed per features of the boards.

JonathonHall-Purism:
Yeah @tlaurion that's the same thing I saw on Librem 11.  I think I am going to exclude partitions based on type before trying to automatically mount them as /boot, e.g. there is no point trying a 'grub' partition as /boot.  Or LUKS or LVM, currently, etc.

What is your vda1 in that case exactly? (Check the last column of `fdisk -l /dev/vda`)

JonathonHall-Purism:
@tlaurion Addressed detecting the /boot device and updated OP.  If your /dev/vda1 is LUKS or an LVM PV, it shouldn't emit these errors any more.

It was a fair amount of code to work around some warnings that I wish the kernel just didn't emit.  But it seems preferable to the alternatives:
* patch kernel...would have to maintain this for every kernel version
* don't emit kernel errors on console...maybe someday, but not sure I want to make such a change for this specific issue (could discuss though if you lean this way)
* leave it...I really don't want spurious errors showing up all the time that I have to tell users to ignore, dilutes any real errors that might appear

The refactoring / improved logic seem beneficial anyway to improve detection of /boot and add some tracing.

tlaurion:
I will have to deep check this. I I stalled debian with nitrokey oem image to fast create installation with oem LUKS defsukt LUKS key a while back but never checked their deployment. Will do.

JonathonHall-Purism:
> Another path would be to check level of Linux reported errors from heads kernel l and check if that error level reported can be dismissed by decreasing verbosity of heads to show only valuable errors outside of debug boards config?

I believe this trace occurs at error level, so we would essentially silence all kernel output from the console.  Maybe it is time to consider that.  We do have configurable debug logging now.

> My intuition points to not defaulting boot device anymore and let the logic of autodetecting boot device at factory default do the job properly without boards hardcoding it but I will double check this.

I agree, but that's not what's causing the errors to appear.  The errors appear when you try to mount something that isn't an understood filesystem (with the PR promoting iso9660 above exfat, anyway).  Automatically detecting /boot does that to find the partition with grub files on it.

Issue #1608, opened 2024-02-13
cbmem -L broken under coreboot 4.21+


tlaurion:
> @JonathonHall-Purism 

>> Well, TBH I'm not sure a lot of other use so much measured boot outside of Heads @JonathonHall-Purism ?
This shows its not tested a lot, or as you said, since cbmem is built by musl-cross-make related doubts, which is compiler buillding everything tools.cpio related under Heads, including cbmem here), that might as well be an undesired artifact of those? (modified quote)

> I took the time to use rom produced here (4.22.01 tag release codebase built with measured boot and coreboot custom event log format) and test cbmem built on top of ubuntu 23.10/debian trixie OS (to remove musl-cross-make doubts).

> Interestingly enough:
> - `sudo ~/Downloada/coreboot-4.22.01/util/cbmem/cbmem -L` :  `Unknown TPM log specification:` : FAIL
> - `sudo ~/Downloada/coreboot-4.19/util/cbmem/cbmem -L` : `coreboot TPM log: [...]` : OK

> So now, let's test cbmem behavior in between those release versions:
> - `sudo ~/Downloada/coreboot-4.21/util/cbmem/cbmem -L` :  `Unknown TPM log specification:` : FAIL
> - `sudo ~/Downloada/coreboot-4.20.1/util/cbmem/cbmem -L` : `coreboot TPM log: [...]` : OK

> **So `cbmem -L` broke somewhere between 4.20.1 and 4.21.**

> ---
> Raised awareness under #coreboot channel at https://matrix.to/#/!EhaGFZyYcbyhdSgStq:matrix.org/$DJW_mF5d1PfhJIvoM-_HvX8s3BpzNvtNu77auhvKBo0?via=matrix.org&via=sibnsk.net&via=datanauten.de. @miczyg1 said he would take a look at https://matrix.to/#/!EhaGFZyYcbyhdSgStq:matrix.org/$34SHe96Ef5Qrv2pIXtmi2lqpSwnh1GsKXlZCR9I7PpY?via=matrix.org&via=sibnsk.net&via=datanauten.de

_Originally posted by @tlaurion in https://github.com/linuxboot/heads/issues/1604#issuecomment-1942087004_

miczyg1:
Issued a PR that should address the issue: https://github.com/linuxboot/heads/pull/1609

Why `cbmem -L` does not work with the standard TPM log per specifications is a separate matter (I suspect the first event in the event log is not SpecID event, but why? not sure yet).

[...]

PR #1604, opened 2024-02-05
4.22.01 fhd patch merged upstream (4.24) + bump all 4.19 boards to 4.22.01

tlaurion:
Revisiting 4.22 coreboot bump with fhd patch (superseeds and replaces #1524), Might fix #1603)

    - patches/coreboot-4.22.01/0001-x230-fhd-variant.patch created per
      - git fetch https://review.coreboot.org/coreboot refs/changes/50/28950/27 && git format-patch -1 --stdout FETCH_HEAD > ~/heads/patches/coreboot-4.22.01/0001-x230-fhd-variant.patch
    - all boards configs bumped with:
      - sed 's/4.19/4.22.01/g' boards/*/*.config -i
      - grep -Rn 4.22 boards/ | awk -F "/" {'print $2'}| while read line; do make BOARD=$line coreboot.save_in_oldconfig_format_in_place ; done
      - sed 's/4.19/4.22.01/g' .circleci/config.yml -i
    - CircleCI caches updated
    - modules/coreboot: remove 4.19 references
    - All 4.19 boards switched to TCG event log format which segfaults `cbmem -L` https://github.com/linuxboot/heads/issues/1608#issuecomment-1953036042
      - Note that Purism boards already switched to TCG format where https://github.com/linuxboot/heads/pull/1609 doesn't fix anything (coreboot custom event log format is broken since 4.21
        - **More details under https://github.com/linuxboot/heads/issues/1608 which isn't blocking this PR** 

Testing report at https://github.com/linuxboot/heads/pull/1604#issuecomment-1953128002

----

4.22 didn't include the patchset in time as https://github.com/linuxboot/heads/issues/1523 related.
~Hopefully this is the last time we pick a patchset. Please board owners, report your edp/fhd kit (PRECISELY) and passing/failing results under https://review.coreboot.org/c/coreboot/+/28950~ EDIT: https://review.coreboot.org/coreboot refs/changes/50/28950/27 was merged.

PR #1609, opened 2024-02-14
config/coreboot-librem: Use coreboot's custom TPM event log

miczyg1:


miczyg1:
Should address issue: https://github.com/linuxboot/heads/issues/1608

JonathonHall-Purism:
The issue in #1608 is actually occurring the boards changed in #1604 (@tlaurion is working on that).  On Librem boards, cbmem does segfault, but it emits the complete and correct TPM log before faulting, so Heads functionality works fine.  I had not noticed it was faulting until it was mentioned in #1604.

That said I'm happy to test this and see if it addresses the segfault (may be a bit before I can get to it).

miczyg1:
I had no problem dumping TPM log in TCG format with cbmem on 4.21+ too, but segfault is something new to me.

`cbmem -L` will also fail to dup the log if the first event is not SpecID event for TCG format. Not sure about that, but there might be a case where the first entry in the log is not SpecID event, i.e. vboot. If vboot is used (as separate stage IIRC) it may happen that measurement of verstage is first, then CRTM initialization and SpecID event is created later... I would have to double check.

But other than that, the SpecID should be always first if vboot is not used (which should be true for all heads builds). Maybe if preram TCG LOG is full, then something is being overwritten?

miczyg1:
> So not sure why your behavior would be any different unless your buildsystem is not the one of coreboot (which I think is the case).

Always using the coreboot toolchain inside docker. In case of heads we also have docker to get all cross toolchains to be built properly, but that still uses heads-built musl-cross and coreboot toolchain still...

miczyg1:
```
TCPA log entry 10:
	PCR: -967035966
	Event type: Unknown (0xace82028 >= 19)
	Digest: d97f7b94ead60f73575cdf71010ef9fd99ed6b28
	Event data: 
```

Looks like coreboot didn't terminate the event log... cbmem checks for a zero_block of one TCPA entry length, if it isn;t found, then it goes further... We could add more safety checks to cbmem to avoid that, but also make sure coreboto properly terminates the eventlog (or at least clear the cbmem entry with the log while creating it...)

@SergiiDmytruk cc

SergiiDmytruk:
> Of course, new format not recognized under 4.19 nor 4.20.1. And segfaults there for TPM1.

Each log is passed in a separate CBMEM entry, old versions of `cbmem` should just not find the log and be done.

> coreboot is clearing the cbmem areas with TCG logs in `src/security/tpm/tspi/log-tpm1.c` and `src/security/tpm/tspi/log-tpm2.c`: `memset(tclt, 0, tpm_log_len);`, so a memory corruption?

Memory issue or some other bug is always possible.  Endianness should be fine (inverting bytes doesn't seem to result in anything sensible).  But given that the entry follows `Event data: CBFS: fallback/payload`, maybe something other than `coreboot` added it?

> If vboot is used (as separate stage IIRC) it may happen that measurement of verstage is first, then CRTM initialization and SpecID event is created later... I would have to double check.

It isn't appending of bytes after bytes, those entries are copied individually into the area allocated for entries:
```c
void tpm1_log_copy_entries(const void *from, void *to)
{
	const struct tpm_1_log_table *from_log = from;
	struct tpm_1_log_table *to_log = to;
	int i;

	for (i = 0; i < le16toh(from_log->vendor.num_entries); i++) {
		if (le16toh(to_log->vendor.num_entries) >= le16toh(to_log->vendor.max_entries)) {
			printk(BIOS_WARNING, "TPM LOG: log table is full\n");
			return;
		}

		struct tpm_1_log_entry *tce =
			&to_log->entries[le16toh(to_log->vendor.num_entries)];
		memcpy(tce, &from_log->entries[i], sizeof(*tce));

		to_log->vendor.num_entries = htole16(le16toh(to_log->vendor.num_entries) + 1);
	}
}
```
That area is after the header which contains SpecID:
```c
struct tpm_1_log_table {
	/* The first entry of the log is inlined and describes the log itself */
	uint32_t pcr;
	uint32_t event_type;
	uint8_t digest[TPM_1_LOG_DIGEST_MAX_LENGTH];
	uint32_t spec_id_size;
	struct spec_id_event_data spec_id;
	struct tpm_1_vendor vendor;

	struct tpm_1_log_entry entries[]; /* Variable number of entries */
} __packed;
```

tlaurion:
@SergiiDmytruk seems like patches are needed  on coreboot master and this behavior affects all measured boot configured platforms. Where can that be tracked?

SergiiDmytruk:
@tlaurion Not necessarily, we still don't know why this happens.  If my guess about something else updating event log is correct, it might be happening due to standard logs, unlike coreboot-specific one, being published in ACPI tables where they can be found by e.g. Linux.  If patching `src/acpi/acpi.c` to do nothing in `acpi_create_tcpa()` and `acpi_create_tpm2()` changes anything, then it could be it.  However, not publishing the logs seems like a workaround rather than a fix if coreboot doesn't do anything wrong.

tlaurion:
@SergiiDmytruk @miczyg1 @krystian-hebel @JonathonHall-Purism  I started a discussion at https://matrix.to/#/!pAlHOfxQNPXOgFGTmo:matrix.org/$CQDLxAckANKmXAu8ZDtcDIHUn1dgvgfzyifZidHr7SA?via=matrix.org&via=nitro.chat&via=fairydust.space to ease the process of understanding the current issue at stake tagging peple that should be involved.

tlaurion:
Issue opened with logs under #1608. This PR doesn't fix anything at all.

PR #1625, opened 2024-03-27
LUKS header change validation upon sealing and unsealing ops

### LINKED ISSUE #1092: LUKS header is not verified directly when a TPM Disk Unlock Key setuped/verified on default boot and fails unsealing

tlaurion:
Fixes #1092 
Supersedes #1093 

- Rework of https://github.com/linuxboot/heads/commit/ed1c23aaa3f8d5264c5b6e5f66ad359406bed166 (credit to @root-hardenedvault) thank you!)
- Addresses and correct self-review under https://github.com/linuxboot/heads/pull/1093 (@root-hardenedvault : you don't answer often here!)
  - ~kexec-unseal-key: Warn a user who attempts to default boot while his Disk Unlock Key passphrase fails to unseal because LUKS headers changed (https://github.com/linuxboot/heads/pull/1093 (comment))~ EDIT: everything move to kexec-insert-key to check header checksum part of detached signed hash digest, before attempting TPM DUK unseal op. 
  - kexec-seal-key: Identical as in https://github.com/linuxboot/heads/commit/ed1c23aaa3f8d5264c5b6e5f66ad359406bed166
  - ~kexec-add-key: Tell the user that the Headers did not change when changing TPM released Disk Unlock Key
    (Through changing default boot at Options->Boot Options -> Show OS boot options: select a new boot option
    and set a Disk Unlock Key in TPM, accept to modify disk and sign /boot options)~
    - ~Here, we cancel the diff output shown on screen https://github.com/linuxboot/heads/pull/1093 (comment)~
      - EDIT: irrelevant in current workflow, since sealing/unsealing/tpm reset triggers TPM DUK resealing. We expect the last boot was not tampered with here prior of firmware upgrade. And then after firmware upgrade, the user should expect firmwware to have changed and engage in resealing TPM DUK (https://github.com/linuxboot/heads/pull/1482)
    - ~And we change the warning given to the user to past tense "Headers of LUKS containers to be unlocked via TPM Disk Unlock Key passphrase did not change." The user knows that unsealing error is not due to LUKS header having changed.~
      - Edit: verbiage expose to end user less technical when unneeded: we talk about Encrypted disk keys, where sealing ops are related still to TPM Disk Unlock (DUK) verbiage. 
    
  Workflow in pictures at https://github.com/linuxboot/heads/pull/1625#issuecomment-2050368105

### DIFF
--- a/initrd/bin/kexec-insert-key
+++ b/initrd/bin/kexec-insert-key
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Unseal a disk key from TPM and add to a new initramfs
+# Unseal a LUKS Disk Unlock Key from TPM and add to a new initramfs
 set -e -o pipefail
 . /etc/functions
 
@@ -28,7 +28,7 @@ if [ -r "$TMP_KEY_LVM" ]; then
 		die "$VOLUME_GROUP: unable to activate volume group"
 fi
 
-# Measure the LUKS headers before we unseal the disk key
+# Measure the LUKS headers before we unseal the LUKS Disk Unlock Key from TPM
 cat "$TMP_KEY_DEVICES" | cut -d\  -f1 | xargs /bin/qubes-measure-luks ||
 	die "LUKS measure failed"
 
@@ -40,13 +40,13 @@ SECRET_CPIO=/tmp/secret/initrd.cpio
 bootdir=$(dirname "$INITRD")
 mkdir -p "$INITRD_DIR/etc"
 
-# Attempt to unseal the disk key from the TPM
+# Attempt to unseal the Disk Unlok Key from the TPM
 # should we give this some number of tries?
 unseal_failed="n"
 if ! kexec-unseal-key "$INITRD_DIR/secret.key"; then
 	unseal_failed="y"
 	echo
-	echo "!!! Failed to unseal the TPM LUKS disk key"
+	echo "!!! Failed to unseal the TPM LUKS Disk Unlock Key"
 fi
 
 # Override PCR 4 so that user can't read the key
@@ -57,8 +57,8 @@ tpmr extend -ix 4 -ic generic ||
 # Check to continue
 if [ "$unseal_failed" = "y" ]; then
 	confirm_boot="n"
-	if diff "$(dirname $INITRD)/kexec_lukshdr_hash.txt" /tmp/luksDump.txt > /dev/null 2>&1; then
-		echo "Headers of LUKS containers to be unlocked via TPM Disk Unlock Key passphrase did not change."
+	if cmp -s "$bootdir/kexec_lukshdr_hash.txt" /tmp/luksDump.txt > /dev/null 2>&1; then
+		echo "Encrypted disk keys(s) have not been changed since sealed in TPM Disk Unlock Key"
 	fi
 	read \
 		-n 1 \
@@ -69,7 +69,7 @@ if [ "$unseal_failed" = "y" ]; then
 		-a "$confirm_boot" != 'Y' \
 		-a -n "$confirm_boot" ] \
 		; then
-		die "!!! Aborting boot due to failure to unseal TPM disk key"
+		die "!!! Aborting boot due to failure to unseal TPM Disk Unlock Key"
 	fi
 fi
 
--- a/initrd/bin/kexec-insert-key
+++ b/initrd/bin/kexec-insert-key
@@ -57,6 +57,9 @@ tpmr extend -ix 4 -ic generic ||
 # Check to continue
 if [ "$unseal_failed" = "y" ]; then
 	confirm_boot="n"
+	if diff "$(dirname $INITRD)/kexec_lukshdr_hash.txt" /tmp/luksDump.txt > /dev/null 2>&1; then
+		echo "Headers of LUKS containers to be unlocked via TPM Disk Unlock Key passphrase did not change."
+	fi
 	read \
 		-n 1 \
 		-p "Do you wish to boot and use the LUKS Disk Recovery Key? [Y/n] " \
--- a/initrd/bin/kexec-insert-key
+++ b/initrd/bin/kexec-insert-key
@@ -40,7 +40,22 @@ SECRET_CPIO=/tmp/secret/initrd.cpio
 bootdir=$(dirname "$INITRD")
 mkdir -p "$INITRD_DIR/etc"
 
-# Attempt to unseal the Disk Unlok Key from the TPM
+if [ -e /boot/kexec_lukshdr_hash.txt ] && [ -e /tmp/luksDump.txt ]; then
+	if ! cmp -s /boot/kexec_lukshdr_hash.txt /tmp/luksDump.txt >/dev/null 2>&1; then
+		#LUKS header hash part of detached signed hash digest under boot doesn't match qubes-measure-luks tmp file
+		warn "Encrypted disk keys have changed since the TPM Disk Unlock Key was sealed. If you did not make this change, the disk may be compromised"
+		exit 1
+	else
+		#LUKS header hash part of detached signed hash digest matches
+		echo "+++ Encrypted disk keys have not been changed since sealed in TPM Disk Unlock Key"
+		#TODO: remove "+++" with boot info helper when added, same with "!!!" currently for info.
+	fi
+else
+	warn "Could not check for tampering of Encrypted disk keys"
+	warn "Re-seal the TPM Disk Unlock Key by re-selecting your default boot option to enable this check (Options -> Boot Options -> Show OS boot menu)."
+fi
+
+# Attempt to unseal the Disk Unlock Key from the TPM
 # should we give this some number of tries?
 unseal_failed="n"
 if ! kexec-unseal-key "$INITRD_DIR/secret.key"; then
@@ -57,9 +72,6 @@ tpmr extend -ix 4 -ic generic ||
 # Check to continue
 if [ "$unseal_failed" = "y" ]; then
 	confirm_boot="n"
-	if cmp -s "$bootdir/kexec_lukshdr_hash.txt" /tmp/luksDump.txt > /dev/null 2>&1; then
-		echo "Encrypted disk keys(s) have not been changed since sealed in TPM Disk Unlock Key"
-	fi
 	read \
 		-n 1 \
 		-p "Do you wish to boot and use the LUKS Disk Recovery Key? [Y/n] " \
--- a/initrd/bin/kexec-save-default
+++ b/initrd/bin/kexec-save-default
@@ -195,7 +195,7 @@ if [ "$CONFIG_TPM" = "y" ] && [ "$CONFIG_TPM_NO_LUKS_DISK_UNLOCK" != "y" ] && [ 
 		DEBUG "LUKS TPM Disk Unlock Key was previously set up from $KEY_DEVICES"
 		read \
 			-n 1 \
-			-p "Do you want to reseal a disk key to the TPM [y/N]: " \
+			-p "Do you want to reseal a Disk Unlock Key in the TPM [y/N]: " \
 			change_key_confirm
 		echo
 
--- a/initrd/bin/kexec-select-boot
+++ b/initrd/bin/kexec-select-boot
@@ -70,6 +70,8 @@ if [ "$CONFIG_TPM2_TOOLS" = "y" ]; then
 	else
 		warn "Hash of TPM2 primary key handle does not exist"
 		warn "Please rebuild the boot hash tree"
+		warn "Select Options-> Update checksums and sign all files in /boot"
+		#TODO: Simplify/Automatize TPM2 firmware upgrade process. Today: upgrade, reboot, reseal(type TPM owner pass), resign, boot
 		default_failed="y"
 		DEBUG "Hash of TPM2 primary key handle does not exist under $PRIMHASH_FILE"
 	fi
@@ -340,7 +342,7 @@ do_boot() {
 		fi
 
 		kexec-insert-key $INITRD ||
-			die "!!! Failed to insert disk key into a new initrd"
+			die "!!! Failed to prepare TPM Disk Unlock Key for boot"
 
 		kexec-boot -b "$bootdir" -e "$option" \
 			-a "$add" -r "$remove" -o "/tmp/secret/initrd.cpio" ||
--- a/initrd/bin/kexec-unseal-key
+++ b/initrd/bin/kexec-unseal-key
@@ -38,15 +38,14 @@ for tries in 1 2 3; do
 		exit 0
 	fi
 
-	DEBUG $(pcrs)
-	warn "Unable to unseal disk encryption key"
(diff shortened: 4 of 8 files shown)


tlaurion:
Simulating a LUKS attack/undesired key slot addition on Heads deployment with DUK setup (qemu with ubuntu deployed as OS):
`cryptsetup luksKeyChange --key-slot 0 /dev/vda4`

Result:

Debug trace showing LUKS header measured under PCR6, mismatch resulting in DUK unsealing failure and verification failing between detached signed LUKS header vs live `cryptsetup luksHeaderBackup`:




@JonathonHall-Purism please review (wording suggestions welcome as usual)

JonathonHall-Purism:
Regarding the three key messages, how about the following wording:

* OK: "Encrypted disk keys have not been changed."
* Altered: "Encrypted disk keys have been changed.  If you did not make this change, the disk may be compromised."
* Can't check: "Could not check for tampering of encrypted disk keys."
    * "Re-seal the TPM Disk Unlock Key by re-selecting your default boot option to enable this check (Options -> Boot Options -> Show OS boot menu)."

Original for comparison:
* OK: "Headers of LUKS containers to be unlocked via TPM Disk Unlock Key passphrase did not change since sealed in TPM."
* Altered: "Encrypted LUKS(es) container(s) headers changed since they were measured and sealed in TPM for Disk Unlock key. You might want to investigate."
* Can't check: "No encrypted LUKS container(s) headers were found/comparable under /boot/kexec_lukshdr_hash.txt"
    * "You might need to setup a new boot default and TPM Disk Unlock Key from Options->Boot Options->Show OS boot menu."

I get that the LUKS header contains other things besides the volume key and key slots, but virtually everything relates in some way to the keys (e.g. changing the cipher changes the way the key is applied to the data).  This differentiates it from the encrypted disk _data_, which we do not check and could still have been tampered (though not in a predictable manner if the attacker does not know a key).

As usual I don't think we should require the user to know what a LUKS container is, what its header is, or what individual files in /boot mean to use Heads, which led me to the wording above.

initrd/bin/kexec-insert-key
```diff
@@ -57,6 +57,9 @@ tpmr extend -ix 4 -ic generic ||
 # Check to continue
 if [ "$unseal_failed" = "y" ]; then
 	confirm_boot="n"
+	if diff "$(dirname $INITRD)/kexec_lukshdr_hash.txt" /tmp/luksDump.txt > /dev/null 2>&1; then
+		echo "Headers of LUKS containers to be unlocked via TPM Disk Unlock Key passphrase did not change since sealed in TPM."
```
tlaurion, line 61: @JonathonHall-Purism 

Recap.
When sealing DUK, PCR6 is extended with LUKS header backup prior of the resulting measurements sealed into seperate TPM nvram+DUK passphrase.

So here, by attempting to unseal TPM nvram DUK passphrase protected region, two things could happen:
- DUK won't unseal because DUK passphrase is wrong, while measurements are good
- DUK won't unseal because measurements are wrong, while DUK passphrase is good (but cannot know because measurements+passphrase needs to be as in sealed state

So here, if undeal fails, we improve upon @root-hardenedvault PR by reporting not only diff of current header vs detached signed header backup that was given only after the 3 bad useal attempts. As of today, DUK would'nt unseal, but the user would not know that its because of the LUKS header having changed.

JonathonHall-Purism (commented):
Also a few detailed review comments - but some of the details may change if the logic is moved around as I suggested a couple of comments up, let me know what you think

tlaurion:
@JonathonHall-Purism I think #1092 is badly named and would go forward changing its name since tits content and goals in description are all fixed by this PR.

As for having LUKS header verified outside of DUK, it would require refactoring with that goal in mind.
As for this PR, it fullfills its goals of telling user that DUK unsealing failed because of the LUKS header having failed.

I think validating LUKS header could be done independently without it being unsealed, but as of now, this is never the case.

tlaurion:
@JonathonHall-Purism addressed review, opened #1629 and corrected name of #1092 which this PR should still close.

initrd/bin/kexec-insert-key
```diff
@@ -57,6 +57,9 @@ tpmr extend -ix 4 -ic generic ||
 # Check to continue
 if [ "$unseal_failed" = "y" ]; then
 	confirm_boot="n"
+	if diff "$(dirname $INITRD)/kexec_lukshdr_hash.txt" /tmp/luksDump.txt > /dev/null 2>&1; then
```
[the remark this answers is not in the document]
tlaurion: [97328b2](https://github.com/linuxboot/heads/pull/1625/commits/97328b2de5ecc1c5cb2298fcb97cf32febaf6367)

initrd/bin/kexec-save-default
```diff
@@ -195,7 +195,7 @@ if [ "$CONFIG_TPM" = "y" ] && [ "$CONFIG_TPM_NO_LUKS_DISK_UNLOCK" != "y" ] && [
 		DEBUG "LUKS TPM Disk Unlock Key was previously set up from $KEY_DEVICES"
 		read \
 			-n 1 \
-			-p "Do you want to reseal a disk key to the TPM [y/N]: " \
+			-p "Do you want to reseal a Disk Unlock Key in TPM [y/N]: " \
```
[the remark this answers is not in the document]
tlaurion: [97328b2](https://github.com/linuxboot/heads/pull/1625/commits/97328b2de5ecc1c5cb2298fcb97cf32febaf6367)

initrd/bin/kexec-select-boot
```diff
@@ -340,7 +342,7 @@ do_boot() {
 		fi
 
 		kexec-insert-key $INITRD ||
-			die "!!! Failed to insert disk key into a new initrd"
+			die "!!! Failed to insert TPM Disk Unlock Key into new initrd to be passed though kexec call to final OS"
```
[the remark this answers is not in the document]
tlaurion: [97328b2](https://github.com/linuxboot/heads/pull/1625/commits/97328b2de5ecc1c5cb2298fcb97cf32febaf6367)

initrd/bin/kexec-unseal-key
```diff
@@ -38,8 +38,15 @@ for tries in 1 2 3; do
 		exit 0
 	fi
 
-	DEBUG $(pcrs)
-	warn "Unable to unseal disk encryption key"
+	warn "Unable to unseal LUKS Disk Unlock Key from TPM"
+	if [ -e /boot/kexec_lukshdr_hash.txt -a -e /tmp/luksDump.txt ]; then
+		if ! diff /boot/kexec_lukshdr_hash.txt /tmp/luksDump.txt > /dev/null 2>&1; then
```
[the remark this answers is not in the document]
tlaurion: [97328b2](https://github.com/linuxboot/heads/pull/1625/commits/97328b2de5ecc1c5cb2298fcb97cf32febaf6367)

JonathonHall-Purism:
@tlaurion Did you see this part of the comment above?  (Not trying to be snarky, just might have been overlooked)

>Regarding implementation - why are the LUKS headers checked in two separate places: kexec-unseal-key / kexec-insert-key? It seems like kexec-insert-key could check it before even prompting for the passphrase, and it could tell you right away.

>If the headers don't match, should we even bother trying to unlock DUK? Maybe we should assume DUK can't be unlocked then and ask whether to boot with the recovery passphrase?

>It also seems like 'luksDump.txt' really should exist at this point, so we shouldn't ignore kexec_lukshdr_hash.txt if it doesn't - that would mean something has gone really wrong. (But if kexec_lukshdr_hash.txt doesn't exist due to having been created by a much older Heads, that seems reasonable, and the signatures on /boot cover it to prevent tampering.)

Would like to know your thoughts on these, did not see anything in the comments above :thinking:

initrd/bin/kexec-insert-key
```diff
@@ -57,6 +57,9 @@ tpmr extend -ix 4 -ic generic ||
 # Check to continue
 if [ "$unseal_failed" = "y" ]; then
 	confirm_boot="n"
+	if cmp -s "$bootdir/kexec_lukshdr_hash.txt" /tmp/luksDump.txt > /dev/null 2>&1; then
+		echo "Encrypted disk keys(s) have not been changed since sealed in TPM Disk Unlock Key"
```
JonathonHall-Purism, line 61: IMO: `keys` instead of `keys(s)` (3x).  I get that the idea is to permit the possibility that there's only one key, but I think `key(s)` is unnecessarily awkward.  (keys(s) has two Ss too, reads like "Hobbitses" to me :sweat_smile: )

I think I'd only really use `(s)` when we're asking the user something and it needs to be abundantly clear that this is the option for one or more "thing(s)" - and there's no better way to rewrite the prompt, etc.  Otherwise it could be unclear what to do for exactly one "thing".

I think the meaning is clear and we're watering down the real point by making it more awkward.  It doesn't agree with `have` as written either, etc.

initrd/bin/kexec-unseal-key
```diff
@@ -38,8 +38,15 @@ for tries in 1 2 3; do
 		exit 0
 	fi
 
-	DEBUG $(pcrs)
-	warn "Unable to unseal disk encryption key"
+	warn "Unable to unseal LUKS Disk Unlock Key from TPM"
+	if [ -e /boot/kexec_lukshdr_hash.txt ] && [ -e /tmp/luksDump.txt ]; then
+		if ! cmp -s /boot/kexec_lukshdr_hash.txt /tmp/luksDump.txt > /dev/null 2>&1; then
+			warn "Encrypted disk keys(s) have changed since sealed in TPM Disk Unlock Key. You might want to investigate."
```
JonathonHall-Purism, line 44: I like the precision of specifying that they have not been changed relative to a specific point in time (sealing the TPM Disk Unlock Key).  But IMO "You might want to investigate." is unnecessarily vague, we can say _why_ they may want to investigate.

So I'd suggest:
`Encrypted disk keys have changed since the TPM Disk Unlock Key was sealed. If you did not make this change, the disk may be compromised.`

initrd/bin/kexec-insert-key
```diff
@@ -40,13 +40,13 @@ SECRET_CPIO=/tmp/secret/initrd.cpio
 bootdir=$(dirname "$INITRD")
 mkdir -p "$INITRD_DIR/etc"
 
-# Attempt to unseal the disk key from the TPM
+# Attempt to unseal the Disk Unlok Key from the TPM
```
tlaurion, line 43: Typo: Unlok Key

tlaurion:
> @tlaurion Did you see this part of the comment above? (Not trying to be snarky, just might have been overlooked)
> 
> > Regarding implementation - why are the LUKS headers checked in two separate places: kexec-unseal-key / kexec-insert-key? It seems like kexec-insert-key could check it before even prompting for the passphrase, and it could tell you right away.
> 
> > If the headers don't match, should we even bother trying to unlock DUK? Maybe we should assume DUK can't be unlocked then and ask whether to boot with the recovery passphrase?
> 
> > It also seems like 'luksDump.txt' really should exist at this point, so we shouldn't ignore kexec_lukshdr_hash.txt if it doesn't - that would mean something has gone really wrong. (But if kexec_lukshdr_hash.txt doesn't exist due to having been created by a much older Heads, that seems reasonable, and the signatures on /boot cover it to prevent tampering.)
> 
> Would like to know your thoughts on these, did not see anything in the comments above 🤔

Answer was at https://github.com/linuxboot/heads/pull/1625#discussion_r1543449421.
Unseal is done by insert, therefore the comparisons and check for file existence is done in unseal, where insert just confirms it?
@JonathonHall-Purism If you insist on this, then checks of unseal could be moved to insert code. Chain is:
- key file exists because it was sealed and that file is used to construct override of crypttab
- key file checked by detached signature, so we cannot go that far if detached signed integrity fails (header from sealing moment)
- arriving at unseal, if key file exist, we expect header backup. Where qubes measure script creates the tmp file so both should exists and can be compared against. No check done there, deletaged to unseal op.
- unseal will fail from tpm unseal op, where cmp is made to confirm header difference (verbiage now a bit less precise on this)
- unseal fails, so we go back to insert which gave more context, now also a bit diluted but I still find ok.

@JonathonHall-Purism still want the refactoring? Security is chain of trust here where higher level is TPM and then files are used to give context. I prefer to trust TPM here and give context on unseal failure more then use tpm file creating without TPM.

initrd/bin/kexec-insert-key
```diff
@@ -40,13 +40,28 @@ SECRET_CPIO=/tmp/secret/initrd.cpio
...
+		#LUKS header hash part of detached signed hash digest under boot doesn't match qubes-measure-luks tmp file
+		warn "Encrypted disk keys have changed since the TPM Disk Unlock Key was sealed. If you did not make this change, the disk may be compromised"
+		exit 1
+	else
+		#LUKS header hash part of detached signed hash digest matches
+		echo "+++ Encrypted disk keys have not been changed since sealed in TPM Disk Unlock Key"
+		#TODO: remove "+++" with boot info helper when added, same with "!!!" currently for info.
+	fi
+else
+	warn "Could not check for tampering of Encrypted disk keys(s)"
```
JonathonHall-Purism, line 54:
```suggestion
	warn "Could not check for tampering of Encrypted disk keys"
```

Please change `keys(s)` to `keys` here

tlaurion:
> still want the refactoring? Security is chain of trust here where higher level is TPM and then files are used to give context. I prefer to trust TPM here and give context on unseal failure more then use tpm file creating without TPM.

Has been refactored. Now luks headers hashes compared before unsealing attempt.

JonathonHall-Purism:
@tlaurion I don't see any changes to any initrd scripts now, were they lost in a force-push?

tlaurion:
Normal boot (notice `+++ Encrypted disk keys have not been changed since sealed in TPM Disk Unlock Key`)


Tampering with the header:


Booting default again:

tlaurion:
@JonathonHall-Purism Sorry about the past mess in commits (and thanks for the pointer). Should be clean now...

JonathonHall-Purism (approved):
LGTM, thanks @tlaurion

Issue #1641, opened 2024-04-17
Ubuntu 24.04  LTS/24.10 misses efifb in initramfs (upstream bug: WON'T FIX?!?!?!)


nestire:
After a successful install via USB Drive Ubuntu is not booting and hangs  at the point of "Starting new Kernel".
It reacts to CRTL+ALT+ENTF but will not continue to boot in any case . 
Tried also the "Unsafe Boot" Option.  No error message is shown, also adding "-d"   [here](https://github.com/linuxboot/heads/blob/82179e4e9894c5cb503f85522ff57088bdd444dc/initrd/bin/kexec-boot#L179C13-L179C14) did not bring up any message. 

Test it with the Nitrokey Heads Release 2.4.1 and with the test build from #1640 on nitropad ns70/nv41/t430

tlaurion:
@nestire thanks for reporting this.

#1640 removed `CONFIG_FB_VESA=y` https://github.com/linuxboot/heads/pull/1640/files#diff-782b88c1e0e03988fb8336bd99c65310869be9f3c1e3a88a1be57bcd5ab7c4e8R1759 since `CONFIG_FB_EFI=y` is supposed to be enough to boot from coreboot initialized ligfxinit GOP FB compatible with efifb driver.

I will diff 2.4.1 release linux config with master and #1640 further more. 

Meanwhile, can you post a screenshot of the kexec command line where the kexec leads to a black screen please?
Is master permitting to boot into 24.04 BETA LTS in current state?

tlaurion:
@nestire https://github.com/linuxboot/heads/pull/1640/commits/fa70cba7ec9df887b24992550c961dbbc708d1c1 was added under #1640 

After flashing, go under configuration settings, activate DEBUG+Tracing option, save in flash (will flash config back to flash) and then try to boot up to last prompt asking you if you really want to boot, on which please say no. You will land in recovery shell.

put a supported FS based usb thumdrive, call 
```
mount-usb --mode rw 
cp /tpm/debug.log /media
umount /media
```

And drop the log here please.

tlaurion:
> Tried also the "Unsafe Boot" Option. No error message is shown, also adding "-d" [here](https://github.com/linuxboot/heads/blob/82179e4e9894c5cb503f85522ff57088bdd444dc/initrd/bin/kexec-boot#L179C13-L179C14) did not bring up any message.

https://github.com/linuxboot/heads/pull/1642#issuecomment-2065315700 ?

[...]

PR #1640, opened 2024-04-16
Nitrokey board cleaning+ unification cleanup (enable htop validated autoboot + tethering)

tlaurion:
Edit: UX changes related to this PR [state](https://github.com/linuxboot/heads/pull/1640#issuecomment-2093569453) in pictures.

---

NS50/NV41:
- Fixed nitropad-x shared linux config to not have technological debt vs other linux configs under tree (librem config/x230 config received a lot of time for those optimizations, which nitropads were not benefiting)
  - no more time wasted building modules that are never included in modules.cpio
  - cpu optimization enablement for randomization and kernel crypto backend
  - Tethering modules inclusion (25kb cost, next steps is use it to sync time from GUI)
  - enable better/disable inefficient crypto algos since kernel is used as crypto backend for cryptsetup 
  - ... (so much more from past merged pr)
- fixed nv41/ns50 coreboot configs to reflect currently used oldconfig (see commit for replication notes with helpers)

qemu//nitropads:
- board config uniformization between nitropad/x230/qemu boards, which should be considered reference boards, making it clear what does what until someone gets funds to switch to Kconfig or something.

---

@daringer : testing notes under commit. Running this on my nv41: all good. 
Please approve and use master's commit for your tests/next version.

Todo: firmware upg from zip+ TPMTOTP/HOTP reseal + TPM DUK reseal + TPM DUK based boot + network-init-recovery(usb network tethering+ time sync [auto])
- [x] nv41
- [x] x230-hotp-maximized
- [x] qemu-coreboot-tpm2-hotp
- [x] qemu-coreboot-tpm1-hotp
- [ ] ns50

PR #1642, opened 2024-04-18
Improve DEBUG and DO_WITH_DEBUG output handling to also keep output of DO_WITH_DEBUG commands

tlaurion:
Goal here was to have `kexec -l` output to kmesg so also to console when BOARD is in DEBUG+TRACE mode (configuration settings menu + flash)

@JonathonHall-Purism  please review

A variation of this commit is under #1640 to troubleshoot #1641 more easily and permit users to provide /tmp/debug.log

### DIFF
--- a/boards/qemu-coreboot-fbwhiptail-tpm1-hotp/qemu-coreboot-fbwhiptail-tpm1-hotp.config
+++ b/boards/qemu-coreboot-fbwhiptail-tpm1-hotp/qemu-coreboot-fbwhiptail-tpm1-hotp.config
@@ -8,56 +8,83 @@ export CONFIG_COREBOOT=y
 export CONFIG_COREBOOT_VERSION=4.22.01
 export CONFIG_LINUX_VERSION=5.10.5
 
+CONFIG_COREBOOT_CONFIG=config/coreboot-qemu-tpm1.config
+CONFIG_LINUX_CONFIG=config/linux-qemu.config
+
 #Enable only one RESTRICTED/BASIC boot modes below to test them manually (we cannot inject config under QEMU (no internal flashing)
 #export CONFIG_RESTRICTED_BOOT=y
 #export CONFIG_BASIC=y
 
+#Enable HAVE_GPG_KEY_BACKUP to test GPG key backup drive (we cannot inject config under QEMU (no internal flashing))
+#export CONFIG_HAVE_GPG_KEY_BACKUP=y
+
 #Enable DEBUG output
 export CONFIG_DEBUG_OUTPUT=y
 export CONFIG_ENABLE_FUNCTION_TRACING_OUTPUT=y
+#Enable TPM2 pcap output under /tmp
+#export CONFIG_TPM2_CAPTURE_PCAP=y
 
-CONFIG_COREBOOT_CONFIG=config/coreboot-qemu-tpm1.config
-CONFIG_LINUX_CONFIG=config/linux-qemu.config
+#On-demand hardware support (modules.cpio)
+CONFIG_LINUX_USB=y
+CONFIG_LINUX_E1000=y
+#CONFIG_MOBILE_TETHERING=y
 
+
+
+#Modules packed into tools.cpio
 ifeq "$(CONFIG_UROOT)" "y"
 CONFIG_BUSYBOX=n
 else
-CONFIG_KEXEC=y
-CONFIG_QRENCODE=y
-CONFIG_TPMTOTP=y
-CONFIG_POPT=y
-CONFIG_FLASHTOOLS=y
-CONFIG_FLASHROM=y
-CONFIG_PCIUTILS=y
-CONFIG_UTIL_LINUX=y
 CONFIG_CRYPTSETUP2=y
+CONFIG_FLASHROM=y
+CONFIG_FLASHTOOLS=y
 CONFIG_GPG2=y
+CONFIG_KEXEC=y
+CONFIG_UTIL_LINUX=y
 CONFIG_LVM2=y
 CONFIG_MBEDTLS=y
-CONFIG_DROPBEAR=y
+CONFIG_PCIUTILS=y
+#Runtime tools to write to EC/MSR
+CONFIG_IOTOOLS=y
 CONFIG_MSRTOOLS=y
+#Remote attestation support
+# TPM2 requirements
+#CONFIG_TPM2_TSS=y
+#CONFIG_OPENSSL=y
+#Remote Attestation common tools
+CONFIG_POPT=y
+CONFIG_QRENCODE=y
+CONFIG_TPMTOTP=y
+#HOTP based remote attestation for supported USB Security dongle
+#With/Without TPM support
 CONFIG_HOTPKEY=y
-export CONFIG_AUTO_BOOT_TIMEOUT=5
-
-#Uncomment only one of the following block
-#Required for graphical gui-init (FBWhiptail)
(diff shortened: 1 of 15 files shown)


tlaurion:
Without debug on qemu-coreboot-whiptail-tpm2-hotp

```
~ # kexec-select-boot -b /boot -f
 *** WARNING: Hash of TPM2 primary key handle does not exist ***
 *** WARNING: Please rebuild the boot hash tree ***
+++ Found verified kexec boot params
+++ Scanning for unsigned boot options
+++ Select your boot option:
1. Ubuntu [kernel /vmlinuz-6.5.0-14-generic]
2. Ubuntu, with Linux 6.5.0-14-generic [kernel /vmlinuz-6.5.0-14-generic]
3. Ubuntu, with Linux 6.5.0-14-generic (recovery mode) [kernel /vmlinuz-6.5.0-14-generic]
4. Ubuntu, with Linux 5.15.0-48-generic [kernel /vmlinuz-5.15.0-48-generic]
5. Ubuntu, with Linux 5.15.0-48-generic (recovery mode) [kernel /vmlinuz-5.15.0-48-generic]
6. Memory test (memtest86+.bin, serial console) [kernel /memtest86+.bin]
Choose the boot option [1-6, a to abort]: 1
Enter LUKS TPM Disk Unlock Key passphrase (blank to abort): 
!!! ERROR: Aborting unseal disk encryption key !!!

!!! Failed to unseal the TPM LUKS disk key
Do you wish to boot and use the LUKS Disk Recovery Key? [Y/n] y

+++ Building initrd
232902+1 records in
232903+0 records out
119246336 bytes (113.7MB) copied, 3.421565 seconds, 33.2MB/s
Loading the new kernel:
kexec -l /boot/vmlinuz-6.5.0-14-generic --initrd=/tmp/secret/initrd.cpio --append="root=/dev/mapper/vgubuntu-root ro console=ttyS0 console=tty systemd.zram=0 console=ttyS0 console=tty systemd.zram=0  vt.default_red=0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff,0xff"
Locking TPM2 platform hierarchy...
Starting the new kernel
qemu: terminating on signal 2

```

No observed regression and really big step forward to have useful user generated bug reports. :)

JonathonHall-Purism:
Thanks for splitting this up, taking a look :eyes:

JonathonHall-Purism:
@tlaurion Suggestion here: https://github.com/JonathonHall-Purism/heads/commit/015af7e6c7befe9835dfa542e73f05bb5ca538db

TL;DR - DO_WITH_DEBUG can be inserted anywhere with no impact to the script (exit status / stdout / stderr all stay the same from the script's perspective) and with minimal spam (trace exit status only when nonzero, trace PATH only for exit status 127, trace output only when non-empty, etc.)

This way you can add/remove a DO_WITH_DEBUG just about anywhere without risking subtle breakage due to stdout/stderr being consumed, splitting or not splitting a single argument, etc.

I improved the documentation and examples in the comment too, so have a look over there :slightly_smiling_face: 

Specifically:
* It looks to me like kexec-unseal-key would have had failed:
    * --mask-position was breaking the password (it executed the masked command)
    * I think set -e would have caused it to terminate
* Do not split a single argument, I think that's a bad idea as I said in https://github.com/linuxboot/heads/pull/1640/files#r1571385205 (and it violates the "just add/remove DO_WITH_DEBUG with no impact" design)
* Exit status is only trace if unsuccessful to reduce spam
* Command output is only traced if unsuccessful to reduce spam
* Out/err are clearly marked in the debug log ($1 out / $1 err)
* PATH is only traced if the command was (probably) not found to reduce spam (exit status 127)
* Out/err still go to the caller as they would without DO_WITH_DEBUG so you can use it anywhere

----

I didn't test Heads logic extensively yet (help appreciated!) but I did test DO_WITH_DEBUG itself.

Silent success - traces command, nothing else:
```
~ # bash -c '. /etc/functions; DO_WITH_DEBUG true; echo $?'
[   32.377401] DEBUG: true
0
```

Silent failure - traces command and exit status, caller still sees status:
```
~ # bash -c '. /etc/functions; DO_WITH_DEBUG false; echo $?'
[   28.408719] DEBUG: false
[   28.415424] DEBUG: false: exited with status 1
1
```

stdout - caller still observes output, traced with label:
```
~ # bash -c '. /etc/functions; DO_WITH_DEBUG ls /; echo $?'
[   22.737953] DEBUG: ls /
bin
boot
dev
etc
init
lib
media
mount-boot
proc
run
sbin
sys
tmp
0
~ # [   22.749170] DEBUG: ls out: bin
[   22.752434] DEBUG: ls out: boot
[   22.758151] DEBUG: ls out: dev
[   22.761524] DEBUG: ls out: etc
[   22.764724] DEBUG: ls out: init
[   22.767905] DEBUG: ls out: lib
[   22.772845] DEBUG: ls out: media
[   22.776752] DEBUG: ls out: mount-boot
[   22.779941] DEBUG: ls out: proc
[   22.783067] DEBUG: ls out: run
[   22.786146] DEBUG: ls out: sbin
[   22.790383] DEBUG: ls out: sys
[   22.793391] DEBUG: ls out: tmp

```

stderr - still printed normally, caller observes exit status
```
~ # bash -c '. /etc/functions; DO_WITH_DEBUG ls /dummy; echo $?'
[   40.044565] DEBUG: ls /dummy
ls: /dummy: No such file or directory
[   40.051851] DEBUG: ls err: ls: /dummy: No such file or directory
[   40.055193] DEBUG: ls: exited with status 1
1
```

command does not exist - stderr still printed, PATH shown, caller observes status
```
~ # bash -c '. /etc/functions; DO_WITH_DEBUG dummy; echo $?'
[   54.542449] DEBUG: dummy
/etc/functions: line 97: dummy: command not found
[   54.546724] DEBUG: dummy err: /etc/functions: line 97: dummy: command not found
[   54.548941] DEBUG: dummy: exited with status 127
[   54.550579] DEBUG: dummy: PATH=/sbin:/bin:/usr/sbin:/usr/bin:/usr/local/bin
127
```

Also ran all of the above with `set -eo pipefail`, output is essentially the same (other than set -e preventing the `echo $?` for failures)

JonathonHall-Purism:
Oh and one more I just remembered - masking a parameter (showing it still executes with the unmasked value) and output that doesn't end in a line break:

```
~ # bash -c '. /etc/functions; DO_WITH_DEBUG --mask-position 2 echo -n foo " "; 
echo $?'
[  234.772789] DEBUG: echo -n <hidden>
foo  0
~ # [  234.781664] DEBUG: echo out: foo
```

JonathonHall-Purism:
From out of band discussion:
* We tweaked the labeling of stdout/stderr in the logs to be clearer ("$1 err:" looked like an error, "$1 stderr:" is better)
* stdout/stderr no longer go to kmsg since we never really want these on the console, log file only (added LOG for verbose tracing like this), silenced kexec stderr from the console too with 2>/dev/null
* restored eval in front of $kexeccmd, DO_WITH_DEBUG should be the same as regular command invocation so just add DO_WITH_DEBUG to the existing eval
* board and version are debug-logged when entering recovery shell

JonathonHall-Purism (approved):

tlaurion:
Testing first prior of merging.

JonathonHall-Purism (approved):

tlaurion:
Ci builds, if failing, would be CI related. Merging.

PR #1646, opened 2024-04-21
Address inconsistency between docs and OEM factory reset User GPG PIN minimum length requirement

### LINKED ISSUE #1645: UX issues during OEM factory reset PIN / Password questionnaire
### A. Provide Hardware Details

**1. What board are you using (see list of boards [here](https://github.com/eganonoa/heads/tree/master/boards))?**
NV41
**2. Does your computer have a dGPU or is it iGPU-only?**
- [ ] dGPU
- [x] iGPU-only

**3. Who installed Heads on this computer?**
- [ ] Insurgo
- [x] Nitrokey (original; internally updated to NK Heads 2.4.1 myself later)
- [ ] Purism
- [ ] Other provider
- [ ] Self-installed

**4. What PGP key is being used?**
- [ ] Librem Key
- [ ] Nitrokey Pro 2
- [ ] Nitrokey Storage
- [ ] Yubikey
- [x] NK3

**5. Are you using the PGP key to provide HOTP verification?**
- [x] Yes
- [ ] No
- [ ] I don't know

### B. Identify how the board was flashed

**1. Is this problem related to updating heads or flashing it for the first time?**
- [ ] First-time flash
- [ ] Updating heads 

**2. If the problem is related to an update, how did you attempt to apply the update?**
- [ ] Using the Heads GUI
- [ ] Flashrom via the Recovery Shell
- [ ] External flashing

**3. How was Heads initially flashed**
- [x] External flashing
- [ ] Internal-only / 1vyrain
- [ ] Don't know

**4. Was the board flashed with a maximized or non-maximized/legacy rom?**
- [x] Maximized
- [ ] Non-maximized / legacy
- [ ] I don't know

**5. If Heads was externally flashed, was IFD unlocked?**
- [x] Yes
- [ ] No
- [ ] Don't know

### C. Identify the rom related to this bug report

**1. Did you download or build the rom at issue in this bug report?**
- [x] I downloaded it
- [ ] I built it

**2. If you downloaded your rom, where did you get it from?**
- [ ] Heads CircleCi
- [ ] Purism
- [x] Nitrokey
- [ ] Somewhere else (please identify)

*Please provide the release number or otherwise identify the rom downloaded*
NK Heads 2.4.1
**3. If you built your rom, which repository:branch did you use?**
- [ ] Heads:Master
- [ ] Other (please identify)

**4. What version of coreboot did you use in building?**
- [ ] 4.8.1 (current default in heads:master)
- [ ] 4.13
- [ ] 4.14
- [ ] 4.15
- [ ] Other (please specify)
- [ ] I don't know

**5. In building the rom where did you get the blobs?**
- [ ] No blobs required
- [ ] Provided by the company that installed Heads on the device
- [ ] Extracted from a backup rom taken from this device
- [ ] Extracted from another backup rom taken from another device (please identify the board model)
- [ ] Extracted from the online bios using the automated tools provided in Heads
- [ ] I don't know

## Please describe the problem

**Describe the bug**
There are two UX issues with the PIN / password entry part of OEM-factory-reset:
1. The question about setting distinct PINs / passwords, which comes after the question about setting a single custom password if the user answers no to that, has the default "No", even though that will set the default passwords ("123456" etc.) for those components, which is obviously not recommended.
2. The minimum user PIN length is set to 8, even though the NitroKey docs explain that it is 6; the NitroKey app also requires only 6, not 8 and even Heads will set "123456" as default user PIN (while "12345678" is the default admin PIN). This can be an issue for people who have already memorized a 6 character / digit PIN as their user PIN (e.g. they've been using their dongle already previously) and suddenly Heads requires that they add two more characters / digits. I think this is probably an oversight (code likely just copied from the admin PIN section).

**To Reproduce**
1. Run an OEM-factory-reset
2. answer "n" to default
3. answer "n" to custom password
4. answer "n" or `Enter` to question about distinct passwords for point 1 above / answer "y" to see incongruent user PIN requirement

**Expected behavior**
For 1.: The default should be "Yes".
For 2.: Heads behaves consistently with itself and the NitroKey app / docs.

UndeadDevel:
Addresses part of #1645:

Lowers the User GPG PIN minimum length to 6 to be consistent with NitroKey app / docs and default PIN length (Admin GPG PIN minimum length is still 8).

### DIFF
--- a/initrd/bin/oem-factory-reset
+++ b/initrd/bin/oem-factory-reset
@@ -994,7 +994,7 @@ if [ "$use_defaults" == "n" -o "$use_defaults" == "N" ]; then
         echo
         if [ "$prompt_output" != "n" \
             -a "$prompt_output" != "N" ]; then
-            echo -e "\nThey must be each at least 8 characters in length.\n"
+            echo -e "\nThe TPM Owner Password and Admin PIN must be at least 8, the User PIN at least 6 characters in length.\n"
             echo
             if [ "$CONFIG_TPM" = "y" ]; then
                 while [[ ${#TPM_PASS} -lt 8 ]]; do
--- a/initrd/bin/oem-factory-reset
+++ b/initrd/bin/oem-factory-reset
@@ -994,7 +994,7 @@ if [ "$use_defaults" == "n" -o "$use_defaults" == "N" ]; then
         echo
         if [ "$prompt_output" != "n" \
             -a "$prompt_output" != "N" ]; then
-            echo -e "\nThe TPM Owner Password and Admin PIN must be at least 8, the User PIN at least 6 characters in length.\n"
+            echo -e "\nThey must be each at least 8 characters in length.\n"
             echo
             if [ "$CONFIG_TPM" = "y" ]; then
                 while [[ ${#TPM_PASS} -lt 8 ]]; do
--- a/initrd/bin/oem-factory-reset
+++ b/initrd/bin/oem-factory-reset
@@ -1002,8 +1002,8 @@ if [ "$use_defaults" == "n" -o "$use_defaults" == "N" ]; then
                     read TPM_PASS
                 done
             fi
-            while [[ ${#ADMIN_PIN} -lt 6 ]] || [[ ${#ADMIN_PIN} -gt $MAX_HOTP_GPG_PIN_LENGTH ]]; do
-                echo -e -n "\nThis PIN should be between 6 to $MAX_HOTP_GPG_PIN_LENGTH characters in length.\n"
+            while [[ ${#ADMIN_PIN} -lt 8 ]] || [[ ${#ADMIN_PIN} -gt $MAX_HOTP_GPG_PIN_LENGTH ]]; do
+                echo -e -n "\nThis PIN should be between 8 to $MAX_HOTP_GPG_PIN_LENGTH characters in length.\n"
                 echo -e -n "Enter desired GPG Admin PIN: "
                 read ADMIN_PIN
             done
@@ -1011,8 +1011,8 @@ if [ "$use_defaults" == "n" -o "$use_defaults" == "N" ]; then
             # That is, if keys were NOT generated in memory (on smartcard only) or
             #  if keys were generated in memory but are to be moved from local keyring to smartcard
             if [ "$GPG_GEN_KEY_IN_MEMORY" = "n" -o "$GPG_GEN_KEY_IN_MEMORY_COPY_TO_SMARTCARD" = "y" ]; then
-                while [[ ${#USER_PIN} -lt 8 ]] || [[ ${#USER_PIN} -gt $MAX_HOTP_GPG_PIN_LENGTH ]]; do
-                    echo -e -n "\nThis PIN should be between 8 to $MAX_HOTP_GPG_PIN_LENGTH characters in length.\n"
+                while [[ ${#USER_PIN} -lt 6 ]] || [[ ${#USER_PIN} -gt $MAX_HOTP_GPG_PIN_LENGTH ]]; do
+                    echo -e -n "\nThis PIN should be between 6 to $MAX_HOTP_GPG_PIN_LENGTH characters in length.\n"
                     echo -e -n "Enter desired GPG User PIN: "
                     read USER_PIN
                 done
--- a/initrd/bin/oem-factory-reset
+++ b/initrd/bin/oem-factory-reset
@@ -989,11 +989,11 @@ if [ "$use_defaults" == "n" -o "$use_defaults" == "N" ]; then
             luks_new_Disk_Recovery_Key_passphrase=${CUSTOM_SINGLE_PASS}
         fi
     else
-        echo -e -n "Would you like to set distinct PINs/passwords to configure previously stated security components? [y/N]: "
+        echo -e -n "Would you like to set distinct PINs/passwords to configure previously stated security components? [Y/n]: "
         read -n 1 prompt_output
         echo
-        if [ "$prompt_output" == "y" \
-            -o "$prompt_output" == "Y" ]; then
+        if [ "$prompt_output" != "n" \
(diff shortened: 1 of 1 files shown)


tlaurion:
@UndeadDevel no problem with changing the pins lengthts. I have no problem changing the default to setting custom PIN as well, but that is for those enforcing OEM factory reset prior of shipping, which is oem-factory-reset main use case, otherwise should be considered under https://github.com/linuxboot/heads/discussions/1521 bigger discussions. 

@daringer @JonathonHall-Purism

UndeadDevel:
As requested by @tlaurion [here](https://github.com/linuxboot/heads/issues/1645#issuecomment-2070909995) this PR now only addresses the inconsistency between the docs / NK app and Heads OEM factory reset minimum User PIN length requirement.

tlaurion:
@UndeadDevel, I’ve realized that the user experience (UX) for the OEM factory reset limits PIN/password/passphrase lengths to a minimum of 8 characters and remembered why. This isn’t due to misreading of smartcard/GPG docs but because a comprehensive re-ownership process shouldn’t rely on a PIN, especially not one shorter than 8 characters. And because in default settings, if choosing a single passphrase for TPM/GPG/DRK, it shouldn't be of 6 characters even if software permits it.

When users opt for default settings, this 8-character minimum isn’t enforced. The GPG factory reset defaults to a user PIN of ‘123456’ (6 characters) and an admin PIN of ‘12345678’ (8 characters).

While you might prefer PIN sizes to align with GPG/OpenPGP smartcard documentation, we’re focusing on users who reject these defaults on first use. We encourage users to choose passwords or, ideally, passphrases, which are more secure according to the documentation: [Configuring Keys.](https://osresearch.net/Configuring-Keys/#oem-factory-resetre-ownership)

A single PIN/password could not meet the security needs of a LUKS Disk Recovery Key passphrase with earlier version of LUKsv1. I mean today you could use a PIN of 6 digits for LUKSv2, but I still won't vet to go that way and would actually get rid of the word PIN if it was possible, but that is what gpg uses, so the wording tries to push users into choosing passphrase, not password and even less PIN. PIN refers to a numeric code. But under GPG/OpenPGP smartcard its a bit different since 3 attempts is permitted prior of locking that user/admin. Still no reason to use a "PIN" to me, where a 2 words EFF short list would be way better http://www.highprogrammer.com/alan/diceware/ even for a single shared passphrase for all secrets if never disclosed.

Conclusion:
We won’t be merging the PIN size changes and I hope the reasoning is clear. You can appeal of my decision and I could, with good arguments, accept to change the minimal size to 6 alone as said in the issue, but other stakeholders would need to agree with that as well.

If passphrases were aotomatically generated, they should be random and meet the minimum word requirements as suggested in the documentation. They could be shared with the user via a QR code or, ideally, not shared at all except for the Disk Recovery Key (DRK) passphrase necessary for re-ownership. Other secrets are "non need to know" and would force the user to do a re-ownership upon reception of hardware and remove accountability on the OEM at that point in time for any leak that could happen, also including reencryption of disk. My personal recommendation is to only share the DRK passphrase with the customer, if that is not of course PleaseChangeMe.  Whether the OEM decides to generate and store additional secrets is up to them. However, shipping devices with a shared passphrase for all secrets should motivate users to take ownership ASAP. If Heads prompts users to change default PINs at each boot with devices shipped seperately, that's also good to me: we expect users to follow this guidance.

TLDR:
The decision to set these design parameters was to simplify the code. The terminology around PINs, passwords, and passphrases can be confusing, which is why we use ‘password’ in the UX. The system tolerates a minimum of 8 characters for a PIN, even though the default is 6. This doesn’t affect the DRK passphrase.




Next Steps:
I concur with your previous comment. We’ll consider modifying the OEM factory reset once upstream bugs are resolved and we can assess the expected behavior vs desired. Personally, I’d prefer to eliminate the term ‘PIN’ altogether, but we can't because used by gpg toolstack. If we adopt passphrases using diceware, even with the EFF small wordlist, the minimum passphrase length would average more than 8 characters anyway, making it more secure and easier to remember than a numeric PIN.

UndeadDevel:
@tlaurion I just noticed that I changed the wrong limit in the PR; fixed with the last commit.

I do disagree with your reasoning, because much of it, it seems to me, does not apply here: this PR does not change the limit for the unified password (what you referenced as "single passphrase for TPM/GPG/DRK"), but only for the User GPG PIN. The USB Security Dongle enforces a 3-try limit before it locks the ability to enter more User PINs; in this case the user or attacker need to enter the Admin PIN (still at 8 characters) to unlock the User PIN again...though here, as well, the dongle enforces a 3-try limit.

So I think there is a misunderstanding here. I simply want Heads to be in line with the recommendation and your entire reasoning, while perfectly valid for the "single passphrase" or TPM DUK or LUKS DRK or TPM Owner passphrase, is invalid here because of how the USB Security Dongles work; one could argue that it's tangentially valid for the Admin GPG PIN, because the new(ish) mechanism with USB sticks as GPG vaults for Heads sets the LUKS vault volume key passphrase to the Admin GPG PIN, but even that argument is irrelevant for the User GPG PIN, which is the only thing this PR affects.

So, I request you review your decision, please.

tlaurion:
> As requested by @tlaurion [here](https://github.com/linuxboot/heads/issues/1645#issuecomment-2070909995) this PR now only addresses the inconsistency between the docs / NK app and Heads OEM factory reset minimum User PIN length requirement.

Sorry I missed this comment. Agreed. This simple change makes total sense.

Tested and then reviewed





The commit log trace is not so clean, but I encourage external contributions and therefore the final changeset on one single file applied shows the exact reasoning leading to the final change, which I love.

Thanks for spotting the discrepency and fixing it in the proper way. Based on those bases, I merge since its a minimal change and I can hear @JonathonHall-Purism in my mind saying this is good enough. If otherwise, sorry, I need things to go forward.

tlaurion (approved):
LGTM! Merging!

PR #1650, opened 2024-04-22
Fix HOTP verification logic (and counter increment) in gui-init and oem-factory-reset scripts

### LINKED ISSUE #1648: HOTP mismatch when booting computer without HOTP USB Security dongle 5 times in a row
An old customer recontacted me today again for this issue which is kinda cryptic and doesn't make sense.

The hotp counter is not related in any way to the TPM counter, is a file under /boot which is not part of the kexec.sig file because kexec* not being kexec*.txt (normal since it changes every boot) and its purpose is unclear.

I would have swore there was an issue opened for that before. But couldn't find it.

Was even countacted a while back for a security report since that counter could be changed (+1/-1) and decreased to be falsified without notice.

What is the purpose of that counter today? Should we remove it?
I remember vaguely a discussion with @JonathonHall-Purism on that, probably off-channel.

But if there is a purpose to that counter, then it should be accompanied with a really big warning to the user reminding him that if the computer is booted more then 5 times in a row, the computer will be confused, since local increments not matching counter on HOTP dongle anymore.

Users are then confused on what to do.

I also vaguely recall that in most OEM setups/rebrandings, the TPMTOTP secret is not shown as QR code to be scanned by end users. So if thsi is the case and then HOTP counter falls out of sync, what is the end user supposed to do? Reset TPM since he cannot validate against his phone?

The only suggestion to those users today (not having TPMTOTP Qr code scanned) is to flash a clean firmware image and  regenerated HOTP from option menus.

But that is a mitgation to an implementation flaw. I think I also remember a discussions with @JonathonHall-Purism about only incrementing the counter when HOTP is actually verified, and not on every boot? Otherwise the UX is missing the big fat warning resulting is way too many questions about that.

@jan23 @daringer @JonathonHall-Purism @wessel-novacustom: what SHOULD be the good behavior here? My take is increment HOTP only when HOTP dongle is connected and continue to do what we do today, limiting user support requests, or discuss on what wwe throught this counter did, what we want it to do in 2024 and make it do what it should do.

tlaurion:
fixes #1648

- Modify unseal-hotp callers (gui-init, oem-factory-reset) to unseal only when HOTP dongle detected
  - Only increment HOTP counter if dongle is actually detected
- Make sure oem-factory-reset die on HOTP boards not having HOTP dongle connected for provisioning.

### DIFF
--- a/initrd/bin/gui-init
+++ b/initrd/bin/gui-init
@@ -250,8 +250,8 @@ update_totp()
 update_hotp()
 {
   TRACE_FUNC
+  HOTP="Unverified"
   if [ -x /bin/hotp_verification ]; then
-    HOTP=`unseal-hotp`
     if ! hotp_verification info ; then
       if [ "$skip_to_menu" = "true" ]; then
         return 1  # Already asked to skip to menu from a prior error
@@ -265,6 +265,7 @@ update_hotp()
         return
       fi
     fi
+    HOTP=`unseal-hotp`
     # Don't output HOTP codes to screen, so as to make replay attacks harder
     hotp_verification check "$HOTP"
     case "$?" in
--- a/initrd/bin/oem-factory-reset
+++ b/initrd/bin/oem-factory-reset
@@ -783,12 +783,22 @@ report_integrity_measurements() {
 
         # Check and report on HOTP status
         if [ -x /bin/hotp_verification ]; then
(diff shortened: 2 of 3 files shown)


JonathonHall-Purism (approved):
Looks like this should do it, thanks :+1:

PR #1688, opened 2024-05-24
Updated README.md for kernel.yama.ptrace_scope issues

123ahaha:
README.md updated in build instructions and in particular an error encountered because of kernel.yama.ptrace_scope during docker image build

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,12 +48,26 @@ Build docker from nix develop layer locally
     * `mkdir -p ~/.config/nix`  
     * `echo 'experimental-features = nix-command flakes' >>~/.config/nix/nix.conf`  
 
+
 #### Build image
 
 * Build nix developer local environment with flakes locked to specified versions  
     * `nix --print-build-logs --verbose develop --ignore-environment --command true`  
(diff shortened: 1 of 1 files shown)


tlaurion:
@JonathonHall-Purism I'm ok with that change since it fixed user's issue in his usage context I cannot replicate. Are you?

@123ahaha thanks!

JonathonHall-Purism (approved):
Thanks for writing this up @123ahaha !

I'm curious to know why ptracing would occur during the docker image build, did you happen to notice where it happens?  (It's fine if not, certainly appreciate the PR for this either way!)

123ahaha:
> Thanks for writing this up @123ahaha !
> 
> I'm curious to know why ptracing would occur during the docker image build, did you happen to notice where it happens? (It's fine if not, certainly appreciate the PR for this either way!)

With the knowledge I have it's either :
- for docker profiling: I don't know whether strace, perf or execve are used at some point
- for sandbox mechanisms on which docker relies
- or for system calls, but I don't see why you'd need syscall to build the image or to go into it...

PR #1693, opened 2024-05-29
All Intel CPU based boards: add cpufreq=xen:hwp in an attempt to fix xen related powersave issues that is not enabled by default under QubesOS

tlaurion:
WiP

### DIFF
--- a/boards/librem_l1um/librem_l1um.config
+++ b/boards/librem_l1um/librem_l1um.config
@@ -32,7 +32,7 @@ export CONFIG_TPM=y
 export CONFIG_BOOTSCRIPT=/bin/gui-init
 export CONFIG_BOOT_REQ_HASH=n
 export CONFIG_BOOT_REQ_ROLLBACK=n
-export CONFIG_BOOT_KERNEL_ADD="intel_iommu=on"
+export CONFIG_BOOT_KERNEL_ADD="intel_iommu=oncpufreq=xen:hwp"
 export CONFIG_BOOT_KERNEL_REMOVE="plymouth.ignore-serial-consoles"
 export CONFIG_BOOT_DEV="/dev/nvme0n1p1"
 export CONFIG_BOARD_NAME="Librem Server L1UM"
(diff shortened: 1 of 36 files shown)


boards/librem_l1um/librem_l1um.config
```diff
@@ -32,7 +32,7 @@ export CONFIG_TPM=y
 export CONFIG_BOOTSCRIPT=/bin/gui-init
 export CONFIG_BOOT_REQ_HASH=n
 export CONFIG_BOOT_REQ_ROLLBACK=n
-export CONFIG_BOOT_KERNEL_ADD="intel_iommu=on"
+export CONFIG_BOOT_KERNEL_ADD="intel_iommu=oncpufreq=xen:hwp"
```
JonathonHall-Purism, line 35: This needs a space assuming we go ahead with the change (thoughts in review summary)

boards/qemu-coreboot-fbwhiptail-tpm1-hotp/qemu-coreboot-fbwhiptail-tpm1-hotp.config
```diff
@@ -89,7 +89,7 @@ export CONFIG_BOOTSCRIPT=/bin/gui-init
 export CONFIG_BOOT_REQ_HASH=n
 export CONFIG_BOOT_REQ_ROLLBACK=n
 export CONFIG_BOOT_RECOVERY_SERIAL="/dev/ttyS0"
-export CONFIG_BOOT_KERNEL_ADD="console=ttyS0 console=tty systemd.zram=0"
+export CONFIG_BOOT_KERNEL_ADD="console=ttyS0 console=tty systemd.zram=0cpufreq=xen:hwp"
```
JonathonHall-Purism, line 92: This doesn't apply to qemu boards does it?  (If so it would need a space though.)

JonathonHall-Purism (commented):
@tlaurion Thanks for finding this.

Does it negatively impact non-Xen distributions?

For Qubes, it's not specifically related to Heads is it?  I'm on the fence here because users/vendors could also enable it by adding it to grub.cfg.  If there's any downside, we're preventing the user from changing it if we enable it in Heads, it is not easy to override CONFIG_BOOT_KERNEL_ADD.  (Somebody on the thread said typing input was noticeably slower, haven't tested it myself yet.)

tlaurion:
Actually, `cpufreq=xen:hwp` is now default and if anything, might need to be deactivated for older boards through KERNEL_REMOVE instead, confirmed by @marmarek on chat (eg instability for some, like the x230 maybe?)

@JonathonHall-Purism let's remember that anything added under KERNEL_ADD will passthrough init if not kernel related and then initramfs of OSes need to take it into consideration to apply it.

Therefore, this would impact Xen based distributions which for today is QubesOS. Other OSes woulnd not be impacted, just as we saw in past PR when some parameters reached init and then normally init will tell that some parameters were not applied.

I

PR #1699, opened 2024-06-07
Update preserve_rom function to check if /tmp/cbfs-init.rom should be used

mdrobnak:
On the MSI boards, the "internal" CBFS region, which is the default when calling `cbfs` without arguments, returns no files, or when called with `-v`, errors:

```
Header delta          : -29261268
Header magic          : ffffffff
Header version        : ffffffff
Header ROM size       : ffffffff
Header boot block size: ffffffff
Header align          : ffffffff
Header offset         : ffffffff
Header arch           : ffffffff
Failed to find valid header
```

This patch is an extension of https://github.com/linuxboot/heads/commit/4d533b36976d1a00d1964449b199e6dad1858a85 changes. That change uses flashrom to extract the image of the CBFS if enabled. Here it uses that extracted image to add files to a new image which is to be flashed.

This has been tested on my MSI Z790-P DDR4 board.

### DIFF
--- a/initrd/etc/ash_functions
+++ b/initrd/etc/ash_functions
@@ -50,14 +50,19 @@ fw_version() {
 
 preserve_rom() {
 	TRACE "Under /etc/ash_functions:preserve_rom"
+	if [ "$CONFIG_CBFS_VIA_FLASHROM" = "y" ]; then
+		CBFS_ARG=" -o /tmp/cbfs-init.rom"
+	else
+		CBFS_ARG=""
+	fi
 	new_rom="$1"
-	old_files=`cbfs -t 50 -l 2>/dev/null | grep "^heads/"`
+	old_files=`cbfs -t 50 -l $CBFS_ARG 2>/dev/null | grep "^heads/"`
 
 	for old_file in `echo $old_files`; do
 		new_file=`cbfs.sh -o $1 -l | grep -x $old_file`
 		if [ -z "$new_file" ]; then
 			echo "+++ Adding $old_file to $1"
-			cbfs -t 50 -r $old_file >/tmp/rom.$$ \
+			cbfs -t 50 $CBFS_ARG -r $old_file >/tmp/rom.$$ \
 			|| die "Failed to read cbfs file from ROM"
 			cbfs.sh -o $1 -a $old_file -f /tmp/rom.$$ \
 			|| die "Failed to write cbfs file to new ROM file"


JonathonHall-Purism:
@mdrobnak Thanks for looking into this!

That commit - https://github.com/linuxboot/heads/commit/4d533b36976d1a00d1964449b199e6dad1858a85 - isn't in master, so nothing reads cbfs-init.rom, this can't be merged as-is.

I have some 32 MB flash boards (L1UM v2 and the WIP Librem 16) but the COREBOOT CBFS region is set to 16 MB on both of those so it's still in the memory-mapped region.  Does this board actually need a larger COREBOOT region?  It may be simpler just to change the region size.

In theory it'd be amazing if we could extend `cbfs` to know how to read the SPI flash using the PCH, but I'm sure that's not trivial.  If we do need this, the workaround to read the whole region ahead of time with flashrom is better than not working at all.

tlaurion:
@JonathonHall-Purism @mdrobnak I wasn't aware of  @mkopec's https://github.com/linuxboot/heads/commit/4d533b36976d1a00d1964449b199e6dad1858a85 and other associated work downstream but this seems to show need of extending flashtool's https://github.com/osresearch/flashtools/blob/master/cbfs.c to be able to read things directly instead of abusing of flashrom which is slow at applying workaround here.

@JonathonHall-Purism going flashrom path would make this work out of the box today, but this approach is less than ideal.

mdrobnak:
I took a quick look at the file and that's out of my expertise at the moment, so I'm unable to help there unfortunately.

mkopec:
I agree that the ideal fix would be to extend flashtools with support for alderlake's extended BIOS window. The original patch to use flashrom was more of a workaround than a real fix.

tlaurion:
@mkopec @mdrobnak so best course of action is to revive PR on dasharo side and have the board and config upstreamed under linuxboot/heads as previously discussed. That board and config doesn't exist under Heads upstream which is an issue for proper collaboration and upstrwamig and bleeding edge testing of fixes.

mdrobnak:
> @mkopec @mdrobnak so best course of action is to revive PR on dasharo side and have the board and config upstreamed under linuxboot/heads as previously discussed. That board and config doesn't exist under Heads upstream which is an issue for proper collaboration and upstrwamig and bleeding edge testing of fixes.

Ok will revive the Dasharo PR for now then.

tlaurion:
Opened https://github.com/Dasharo/dasharo-issues/issues/897 to track issue downstream. Closing this PR, needs to come from Dasharo team.

PR #1702, opened 2024-06-19
circleci: Add Librem L1UM to CI, in front of unmaintained 4.11 boards

JonathonHall-Purism:
Librem L1UM (v1) works in CI, re-enable it in CircleCI.

I put it in front of the KGPE-D16 boards just because I don't think a maintained board should depend on unmaintained boards, but I can tweak it if there's a reason to have them in a different order.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -250,8 +250,8 @@ workflows:
 
       # coreboot 4.11
       - build_and_persist:
-          name: UNMAINTAINED_kgpe-d16_workstation
-          target: UNMAINTAINED_kgpe-d16_workstation
+          name: librem_l1um
+          target: librem_l1um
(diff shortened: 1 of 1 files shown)


tlaurion (approved):
Perfectly fine with me. Merging

tlaurion:
@JonathonHall-Purism unrelated to this but from really limited tests, it seems that builds might not be reproducible there, haven't investigated though.

tlaurion:
That was https://matrix.to/#/!pAlHOfxQNPXOgFGTmo:matrix.org/$7_qW3cWuYY-i50x8EHyiX6PErMU6Bo66EcPZskOzhnI?via=matrix.org&via=nitro.chat&via=envs.net

PR #1703, opened 2024-06-19
modules/coreboot: 24.02.01-Purism-1, remove CFLAGS overrides, needs nss

JonathonHall-Purism:
Update Purism coreboot to 24.02.01-Purism-1.

Remove CFLAGS overrides when building coreboot.  These overrides break 24.02.01, which added (and needs) --param=min-pagesize=1024.  This has happened repeatedly in the past since Heads has to duplicate coreboot's CFLAGS if it overrides them.

Specifically, the build fails with this error:
```
src/commonlib/include/commonlib/endian.h:27:26: error: array subscript 1 is outside array bounds of 'void[0]' [-Werror=array-bounds=]
   27 |         *(uint8_t *)dest = val;
      |         ~~~~~~~~~~~~~~~~~^~~~~
In function 'setup_default_ebda':
cc1: note: source object is likely at address zero
```

That's because coreboot is attempting to write to EBDA at physical address 0x40e, just above 1024.  That is a valid address for x86, but it's too close to 0 by default for GCC, --param-min-pagesize=1024 allows writes to physical addresses above 1024.

coreboot shouldn't need any of the usual Heads CFLAGS overrides for reproducibility; it is already reproducible.

Fix indentation in modules/coreboot.  Make accepted it before because the indented lines followed a variable assignment, so they couldn't be part of a recipe.  That assignment is now gone, so they're now interprted as part of a recipe for the `.configured` target just above, they should not be indented.

Add nss to flake.nix, needed as of 24.02.01.

Update Librem coreboot configs for 24.02.01-Purism-1.  Notably, the board Kconfig changed for Mini v2 in coreboot, so this is needed for correct builds.


JonathonHall-Purism:
@tlaurion The build environment needs `nss` now.  I've added it to flake.nix in this branch, could you publish a new Docker image and let me know the version so I can update CircleCI config?

tlaurion:
@JonathonHall-Purism sorry about seeing this late. Doing

tlaurion:
@JonathonHall-Purism : CircleCI building https://github.com/tlaurion/heads/commit/bb4cbc4450964abf2c8227fcb7f8bbfe8502a710

tlaurion (approved):
LGTM!

tlaurion:
Pushed docker image to latest

PR #1715, opened 2024-07-10
coreboot version bump from 4.22.01 to 24.02.01 (Q35/ivy/sandy/haswell/librems) 

tlaurion:
Coreboot version bump to 24.02.01 for all platforms that depended on coreboot 4.20.01. 

**Testable roms available under green mark, select board, then go under artifacts and select zip file to internally upgrade. Do backup first from recovery shell and be ready to externally flash backup if bricked.** 

EDIT: [8d3c5cf](https://github.com/linuxboot/heads/pull/1715/commits/8d3c5cf355d8d2777dd4ab10c7cf95f2d1f9c7f1) inclues @JonathonHall-Purism patches (that can be seen under `patches/coreboot-24.02.01` of this PR):

Tested:

- [ ] qemu tpm1/tpm2: ok!
- [ ] x230

-----
Old (prior of [8d3c5cf](https://github.com/linuxboot/heads/pull/1715/commits/8d3c5cf355d8d2777dd4ab10c7cf95f2d1f9c7f1))

Tested
- [ ] t430 : thanks @aluciani https://github.com/linuxboot/heads/pull/1715#issuecomment-2222381749 **NO BOOTSPLASH** 
- [ ] x230 (@tlaurion) : https://output.circle-artifacts.com/output/job/a3a29799-197a-43a6-ad76-ec3b0e1b238d/artifacts/0/build/x86/x230-maximized/heads-x230-maximized-v0.2.0-2215-g21ab781.zip **BOOTSLASH IS GONE**

----
All boards untested will be prepended UNTESTED, then all boards UNTESTED will go UNMAINTAINED.

PR #1716, opened 2024-07-11
xxx0 : coreboot version bump to 24.05 for qemu q35/sandy/ivy/haswell : regression still

tlaurion:
Same regression observed under #1715


tlaurion:
Will need to be redoneon top of #1715. Closing for now.

PR #1745, opened 2024-08-06
targets/qemu.mk: Fix skip-if-exist in swtpm_setup

JonathonHall-Purism:
Commit 46cad549 ("WiP flake.nix: make docker image usable for...") added 'root' to the swtpm_setup call, but broke skip-if-exist because the flags are supposed to be comma-separated.  swtpm_setup was ignoring skip-if-exist and would fail if the config files exist.

Put a comma there so it works again.


tlaurion (approved):
Good catch. Merging

tlaurion:
Good catch. Merging

PR #1746, opened 2024-08-06
keys: Remove PureOS signing key

JonathonHall-Purism:
PureOS ISOs still are not detach-signed and it is misleading that there is a key here.  Remove it, we'll add it when the ISOs are signed.


tlaurion (approved):
Commit to  be reverted hopefully sooner than later @JonathonHall-Purism

PR #1737, opened 2024-07-29
Alexgithublab: change time (superseeds #1730)

tlaurion:
This is PR superseding #1730 

This add helper to change system time clock engaging user into coonsulting external source of time in UTC and setting system time accordingly quite easily, thanks to @alex-nitrokey contribution under #1730.
Screenshots of changed UX at https://github.com/linuxboot/heads/pull/1737#issuecomment-2256118049

---

@alex-nitrokey please comment and approve PR. 
@JonathonHall-Purism please acknowledge/suggest changes so that you are ok with wording prior of merge.

PR #1748, opened 2024-08-06
Alexgithublab: change time, 3.0 (supersedes #1737)

JonathonHall-Purism:
Nice work on this under #1737 @alexgithublab and @tlaurion .  I made some improvements:

# UX

* Improve prompt wording when entering fields
* get_date: Complete input with enter, allow Backspace, don't require leading zeros
    * Not allowing backspace is frustrating if you make a typo, and worse that it actually inputs a character :sweat_smile: 
    * Moving on without requiring Enter is surprising, and if you press Enter quickly expecting it to be required, you skipped the next prompt inadvertently
    * Users with imperfect typing might double a digit unintentionally, it's really frustrating to have to complete the steps and set a wrong time since you couldn't edit.  (TBH this happens to me, occasionally my hands twitch, this is important for accessibility)
* get_date: Don't offer minimum field value by default
    * It's not useful very often, you're only on minute 0 for 1/60 of the time :stuck_out_tongue:
    * Pressing '0[enter]' is intuitive and no longer requires '00[enter]'
    * Eliminates a lot of wordiness that would have to be read (or realistically, ignored)
* Ask whether to retry rather than forcing to repeat until it works
* Improve wording of TOTP/HOTP mismatch prompt (show current time, suggest what to do, actually include a question since it's yes/no :) )

# Scripting

* Use named variables in get_date (clearer implementation and helps document usage)
* Test exit status of date directly rather than trying to parse stdout/stderr
* Infer digits from length of maximum value
* Loop instead of recurse
* Add .sh

Approved changes of UX, new prompt looks like


alexgithublab (approved):

tlaurion:
@JonathonHall-Purism unfortunately inconsistent !

tlaurion:
@JonathonHall-Purism Also, current date should be given in expected format so user knows approximately which offset he needs to work agaist to resolve current time drift causing TOTP mismatch.

JonathonHall-Purism:
> @JonathonHall-Purism unfortunately inconsistent !

Wow, nice catch on the octal shenanigans, will fix that :100:

JonathonHall-Purism:
> @JonathonHall-Purism Also, current date should be given in expected format so user knows approximately which offset he needs to work agaist to resolve current time drift causing TOTP mismatch as said under [#1737 (comment)](https://github.com/linuxboot/heads/pull/1737#issuecomment-2258883792) for @alexgithublab to address.

What would you suggest?  The current prompt displays the numbers in the same order that we prompt for them.

E.g. it says: `The current UTC time is: 2024-08-06 16:33:16 UTC"

You would enter:
* 2024
* 08
* 06
* 16
* 33
* 16

I suppose there's a redundant "UTC" there that's not needed.  Is there something else you'd suggest to change?

tlaurion:
> > @JonathonHall-Purism Also, current date should be given in expected format so user knows approximately which offset he needs to work agaist to resolve current time drift causing TOTP mismatch as said under [#1737 (comment)](https://github.com/linuxboot/heads/pull/1737#issuecomment-2258883792) for @alexgithublab to address.
> 
> What would you suggest? The current prompt displays the numbers in the same order that we prompt for them.
> 
> E.g. it says: `The current UTC time is: 2024-08-06 16:33:16 UTC"
> 
> You would enter:
> 
> * 2024
> * 08
> * 06
> * 16
> * 33
> * 16
> 
> I suppose there's a redundant "UTC" there that's not needed. Is there something else you'd suggest to change?

That would be perfect, nothing more needed, just a reminder of the time just prior of entering the time in UTC format (normally its minutes and seconds that are problematic with time drift if CMOS battery not garbage), which most end users will need to fix manually. So simple timestamp is fine. Testing last changes and outputting UX changes and report if ok.

If ok, will merge.

tlaurion:
@JonathonHall-Purism just a timestamp reminder where %Z is already giving UTC would be fine, but missing

JonathonHall-Purism:
@tlaurion You want that in `The system date has been successfully set to <time> *UTC*`, right?

initrd/bin/gui-init
```diff
@@ -547,7 +547,7 @@ prompt_totp_mismatch()
 {
   TRACE_FUNC
   if (whiptail_warning --title "TOTP/HOTP code mismatched" \
-    --yesno "TOTP/HOTP code mismatches could indicate TPM tampering or clock drift.\n\nThe current UTC time is: $(date "+%Y-%m-%d %H:%M:%S %Z")\nIf this is incorrect, set the correct time and check TOTP/HOTP again.\n\nDo you want to change the time?" 0 80) then
+    --yesno "TOTP/HOTP code mismatches could indicate TPM tampering or clock drift.\n\nThe current UTC time is: $(date "+%Y-%m-%d %H:%M:%S")\nIf this is incorrect, set the correct time and check TOTP/HOTP again.\n\nDo you want to change the time?" 0 80) then
```
tlaurion, line 550: This is from whiptail menu, not console.
This might be redundant and most probably more relevant in change-time directly @JonathonHall-Purism, since this is where time change will occur anyway? Or duplicate, I don't really mind. Its just that change-time is called from whiptail and from another menu which doesn't show time right now, and call clear in the script. So would be better under change-time in my opinion.

tlaurion (changes requested):

initrd/bin/change-time.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
+#change time using hwclock and date -s
+
+clear
+
+echo -e -n "Please enter the current date and time in UTC\n"
+echo -e -n "To find the current date and time in UTC, please check https://time.is/UTC\n\n"
```
tlaurion, line 7: @JonathonHall-Purism here would be ideal IMO

initrd/bin/gui-init
tlaurion: https://github.com/linuxboot/heads/pull/1748/files#r1705908889

initrd/bin/gui-init
```diff
@@ -430,6 +430,7 @@ show_options_menu()
     --menu "" 0 80 10 \
     'b' ' Boot Options -->' \
     't' ' TPM/TOTP/HOTP Options -->' \
+    'h' ' Change system time' \
```
tlaurion, line 433: Otherwise the two entry points show different behavior: one with datestamp one without. change-time should be where date reminder occur IMO

initrd/bin/change-time.sh
JonathonHall-Purism: Screenshot_20240806_140058

Added that, let me know what you think.

I kept the current time reminder in the "TOTP/HOTP mismatch" prompt also because it is relevant to answering the question "do you want to change the time?"

tlaurion: Good! You're the picky one on UX, so if you're ok, i'm ok :)

tlaurion (approved):

tlaurion:
Approved changes of UX, new prompt looks like

tlaurion:
@JonathonHall-Purism I took liberty to add UX screenshot in OP. Merging

tlaurion:
@JonathonHall-Purism @alexgithublab thanks! This improves TOTP mismatch resolution a lot! Merged!

alexgithublab:
@JonathonHall-Purism @tlaurion No problem, I'm happy to help and thank you for improving my code.

PR #1750, opened 2024-08-06
move w541 back to tested

gaspar-ilom:
W541 was tested and can be moved back to the tested boards.


tlaurion:
@gaspar-ilom : thanks for doing exactly the ideal contribution: move untested boards to tested boards through a pull request, not having me to do anything at all but pressing the merge button! Much appreciated.

This can be used as an example to move back untested/unmaintained boards back to tested boards.

PR #1753, opened 2024-08-09
Bring dasharo+heads MSI boards from downstream Dasharo/heads to upstream

tlaurion:
- files: boards + coreboot + linux, borrowed directly from https://github.com/Dasharo/heads/commit/cb430394e289fcfdaace8123aed4f76148041759 tip
- cbfs-init modified as per downstream fork dasharo+heads used modifications (flashrom)
- ash_functions modified as per downstream fork dasharo+heads used modifications (CBFS)
- network-init-recovery modified as per downstream fork dasharo+heads used modifications (igc)
- modules/linux modified as per downstream fork dasharo+heads used modifications (igc)
- modules/coreboot modified as per downstream fork dasharo+heads used modifications (also impact nv41/ns50: coreboot version bump)
- modules/flashrom modified to user flashrom 1.4 + https://review.coreboot.org/c/flashrom/+/83854
- Circleci: added boards being dependent of nv41

NOTES:
- [This requires Nk3 firmware to be 1.7.1+](https://www.nitrokey.com/blog/2024/heads-v25-and-nitrokey-3-firmware-v171-security-updatehttps://www.nitrokey.com/blog/2024/heads-v25-and-nitrokey-3-firmware-v171-security-update)  
- Requires external programmer in case of bricking, untested and not owning the platforms.
- dGPU support known problematic
- kgpe-d16 patch that was still under master was unapplied on master (permitting to flash openbmc from Heads)

---

DISCLAIMER: UNTESTED
Sorry, not gonna cherry-pick commits from dahsharo/heads here, way too messy.

----

after merge, WILL superseed https://github.com/linuxboot/heads/pull/1489 abandonned effort.

----

TODO: 

- [x] kgpe-d16 misses a working patch for internal flashing? In this PR, unapplied patch deleted. issue #1753 created

-----

OLD
~DO NOT FLASH YET IF NO EXTERNAL PROGRAMMER. THIS VERSION BUMP FLASHROM SO FLASHING MASTER FROM THIS IS NEEDED~ Edit: fixed with #1755 commit cherry-picked.

PR #1755, opened 2024-08-23
flash.sh: drop inhouse progress output on console when flashing with flashrom

tlaurion:
Flashrom dropped --progress output under 1.2.
flashrom 1.4 still has no --progress output.

---

Heads implemented in-house progress output as a workaround until fixed upstream.
That never happened.
This is still opened issue https://ticket.coreboot.org/issues/390
---

As per https://github.com/linuxboot/heads/pull/1753#discussion_r1725694439 @JonathonHall-Purism comment:
> What would you think about removing the progress bar? The dream would be to rework it (maybe doing it in flashrom itself, or at least providing a reliably-parseable progress output), but the current implementation is a huge pain, I don't feel it adds enough value to outweigh the pain. For comparison the read step takes a comparable amount of time and does not have a progress bar.

---

So here, we drop progress output and leave user in the dark while they erase/write the firmware internally when upgrading Heads.

### DIFF
--- a/initrd/bin/flash.sh
+++ b/initrd/bin/flash.sh
@@ -21,106 +21,6 @@ case "$CONFIG_FLASHROM_OPTIONS" in
   ;;
 esac
 
-flashrom_progress() {
-    # The ichspi programmer now spews register status lines constantly that are
-    # brutally slow to feed through the parser in flashrom_progress_tokenize.
-    # Filter the input with grep for only lines containing at least one token
-    # that we care about.
-    grep -E -e 'contents\.\.\.' -e 'done\.' -e '0x[0-9a-f]+-(0x[0-9a-f]+):' \
-        -e 'identical' -e 'VERIFIED\.' -e 'FAILED' | \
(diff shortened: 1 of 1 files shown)


tlaurion:
@JonathonHall-Purism works as intended. Merge?

JonathonHall-Purism:
:100: Let's merge, thank you @tlaurion 

I'm thrilled to see this go.  While it's not awesome to leave users with an indefinite wait, it is far better than sometimes freezing during a critical potentially-bricking operation, and this code was a nightmare to maintain.

tlaurion:
> :100: Let's merge, thank you @tlaurion 
> 
> I'm thrilled to see this go.  While it's not awesome to leave users with an indefinite wait, it is far better than sometimes freezing during a critical potentially-bricking operation, and this code was a nightmare to maintain.

@JonathonHall-Purism please approve!

JonathonHall-Purism (approved):

tlaurion:
Master flashrom output for 1.3 with e typing on keyboard in scary dark moment, in between flashrom stating it will do write erase operations prior of doing them.

Issue #1709, opened 2024-06-26
kgpe-d16: Make sure those patches are under Heads


tlaurion:
https://github.com/mrothfuss/coreboot-D16

tlaurion:
Commit were not signed, created PR at https://github.com/mrothfuss/heads/pull/1

@mrothfuss CC

tlaurion:
Superseeded by https://github.com/linuxboot/heads/issues/1709

PR #1760, opened 2024-08-26
coreboot-4.11: add fixes to KGPE-D16 raminit

tlaurion:
The added patches fix bugs in fam15h ram DQS timing and configure the motherboard to restart gracefully if raminit fails instead of booting into an unstable state and/or crashing.

Superseeds #1709 with signed commits to pass DCO from CI.

### DIFF
--- a/patches/coreboot-4.11/0075-northbridge-amd-Fixed-errors-in-fam15h-DQS-timing.patch
+++ b/patches/coreboot-4.11/0075-northbridge-amd-Fixed-errors-in-fam15h-DQS-timing.patch
@@ -1,0 +1,62 @@
+From f6c818898b3f978bd22ed2829a881322e0eadaf9 Mon Sep 17 00:00:00 2001
+From: Mike Rothfuss <6182328+mrothfuss@users.noreply.github.com>
+Date: Fri, 23 Aug 2024 19:54:54 -0600
+Subject: [PATCH 1/2] northbridge/amd: Fixed errors in fam15h DQS timing
+
+Fixed two errors in determining whether valid values were
+found for read DQS delays in raminit.
+---
+ src/northbridge/amd/amdmct/mct_ddr3/mctdqs_d.c | 17 ++++++-----------
+ 1 file changed, 6 insertions(+), 11 deletions(-)
+
+diff --git a/src/northbridge/amd/amdmct/mct_ddr3/mctdqs_d.c b/src/northbridge/amd/amdmct/mct_ddr3/mctdqs_d.c
+index d34b2dc2ba..6cf67afa4f 100644
+--- a/src/northbridge/amd/amdmct/mct_ddr3/mctdqs_d.c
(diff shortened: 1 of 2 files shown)


tlaurion:
So no observed regression on at least one supported HCL. Merging.

patches/coreboot-4.11/0075-northbridge-amd-Fixed-errors-in-fam15h-DQS-timing.patch
```diff
@@ -0,0 +1,62 @@
+From f6c818898b3f978bd22ed2829a881322e0eadaf9 Mon Sep 17 00:00:00 2001
+From: Mike Rothfuss <6182328+mrothfuss@users.noreply.github.com>
+Date: Fri, 23 Aug 2024 19:54:54 -0600
+Subject: [PATCH 1/2] northbridge/amd: Fixed errors in fam15h DQS timing
+
+Fixed two errors in determining whether valid values were
+found for read DQS delays in raminit.
```
paulmenzel, line 7: Nice. As raminit is quite complex issue and a long known problem, as long as you have memory about your debugging and analysis, could you please elaborate. What was the problem? What type of RAM was affected? How can it be verified? What read DQS delay values were considered incorrect and are now correct?

tlaurion: Cc @mrothfuss

mrothfuss: > Nice. As raminit is quite complex issue and a long known problem, as long as you have memory about your debugging and analysis, could you please elaborate. What was the problem? What type of RAM was affected? How can it be verified? What read DQS delay values were considered incorrect and are now correct?

Three fixes are included in this patch that improve the reliability of raminit for all RAM.

1.  A logic error required only a single memory lane to pass DQS timing for read delays, this patch fixes it so all lanes must pass for the timing values to be added to the list of potential timing configurations.
2. When raminit is searching for timing values, it now discards negative DQS values instead of adding the faulty parameters to the list of candidate configurations.
3. Previously when raminit fails, coreboot continues to boot into an extremely unstable state (often resetting or freezing before coreboot can even finish booting). The patch just has the board reset and try again when raminit fails. This is a very lazy work around. After adding fixes 1&2, this patch only got triggered ~1 in 50 boots IIRC.

These patches improve the boot consistency. At times I'd have as many as 5 out of 10 boots fail under stock coreboot-4.11 (8xCT16G3ERSLD4160B, 2x6328), with these patches I had 1000+ boots without any issue. I do not know if these patches enable any new RAM configurations. It is possible that the "bad" RAM configs were more prone to incorrectly passing DQS timing with bad read delays (Fix 1). If that is the case, this could enable previously unusable RAM configs.

Issue #1765, opened 2024-08-27
Porting heads to Galago Pro 5, built ROM fails to boot


loftlifter31:
## Context of the Build

**1. What board are you trying to build?**
Trying to port heads to a new board, the system76 galp5 which is already supported by coreboot.

**2. What repository:branch are you using to build from?**
- [ X] Heads:Master
- [ ] Other (please specify)

**3. What version of coreboot are you trying to build**
4.22.01

**4. In building the rom where did you get the blobs?**
- [ ] No blobs required
- [ ] Provided by the company that installed Heads on the device
- [ ] Extracted from a backup rom taken from this device
- [ ] Extracted from another backup rom taken from another device (please identify the board model)
- [ ] Extracted from the online bios using the automated tools provided in Heads
- [ ] I don't know
- [X] Provided by system board manufacturer

**5. If using the automated tools to get the blobs did you run the relevant scripts in the blobs directory**
- [ ] Yes
- [ ] No

**6. What operating system are you using**

Debian 11

## Please describe the problem

No errors during build process, but boots into black screen. Unsure what steps to take to troubleshoot at this point.

**To Reproduce**
1. Build rom
2. flash using hardware programmer
3. reboot
**Expected behavior**
Boot to Heads first time boot screen

**Screenshots**

**Additional context**
[heads_build.log](https://github.com/user-attachments/files/16768085/heads_build.log)

loftlifter31:
Coreboot config files I am using come from manufacturer 
[coreboot-galp5.config.txt](https://github.com/user-attachments/files/16768098/coreboot-galp5.config.txt)
[galp5.config.txt](https://github.com/user-attachments/files/16768100/galp5.config.txt)
[linux-galp5.config.txt](https://github.com/user-attachments/files/16768102/linux-galp5.config.txt)

tlaurion:
@loftlifter31 where can I see a branch you work on?

loftlifter31:
Work on? I'm building from master if that's your question.

tlaurion:
> Work on? I'm building from master if that's your question.

Where can I see your code as a branch?
Ideally do a WiP pull request on master, so that I can comment in code, just like all other community ports before se we can iterate on top. Board config, Linux config coreboot config that I will be able to review comparing other boards of same or similar family, nothing what is HCL, iGPU, dGPU and other things that could prevent booting or have console output

loftlifter31:
I added pull request  #1766 with all the config files.

[...]

PR #1766, opened 2024-08-29
Port heads to system76 galago pro 5

loftlifter31:
Config files which came from other sources on github have been annotated with source. Config files without annotations are customizations of existing heads config files modified by me to the requirements of the board.


tlaurion:
@loftlifter31  from #1765 
>Board config, Linux config coreboot config that I will be able to review comparing other boards of same or similar family, nothing what is HCL, iGPU, dGPU and other things that could prevent booting or have console output

So what is "non-booting" ? what is iGPU etc? what is HCL?
Please rpovide details not expecting me to work on this prior of understanding at least in theory what is "not working".

loftlifter31:
@tlaurion GPU is Intel Iris Xe. I'm not sure what HCL is. Hardware compatibility list? Maybe the info you want is here? https://tech-docs.system76.com/models/galp5/README.html

Not booting in this case means black screen on power on. Video backlight comes on and power led changes from amber to green but nothing is displayed on screen. I suspect there may be an issue with the VBT based on a conversation with one of the System76 developers. Apparently the "parser" needs to match the VBT. I tried replacing the vbt.rom I had been using from the System76 github repository with the VBT extracted from the working firmware which shipped with the machine. Also tried using the data.vbt coreboot includes under src/mainboard/system76/tgl-u/variants/galp5/. Also tried commenting out the CONFIG_INTEL_GMA_VBT_FILE line to let coreboot use whatever it thinks is best. None of those solutions result in any video output.

tlaurion:
@loftlifter31 are you under Heads matrix channel?

This is one of the subject where dynamic discussions would be way more effective then trying to fix your itch through PR.
Community page: https://osresearch.net/community/

Otherwise the reason why I pointed you to check HCL (yes hardware compatibility list) is that you could see that other platforms are really similar. For example you could basically borrow a librem similar to that hardware, and then start comparing configs and make it work.

As you saw with your own upstream discussions, you seem to come near the conclusion that using an old fork of coreboot won't work. So you could check history of modules/coreboot file, and then incororate system76 coreboot fork.

Then borrow librem coreboot config, linux config and then arrive to a booting system using coreboot libgfxinit to drive iGPU up to linux efifb driver and have something working probably really quick.

Reach on matrix channel and open a thread there, I can point you to things, and link those discussions here. 


Direct pointers:
- You will need libgfxinit, efifb and vbt yes. so next step is use their fork under modules/coreboot

tlaurion:
Put as draft. Please #1774 as a base to force push commits here from git from command line.

tlaurion:
Deprecated in favor of #1774. To propose superseeding PR, you will need to create a branch, not use master of your fork.
Closing.

PR #1772, opened 2024-09-02
init: make sure config.user is overriden only with new values so that going DEBUG/undoing can work reproducibly

tlaurion:
As of today, somone deciding to Config -> Enable DEBUG + flash cannot simply go back to Config-> Disable DEBUG + flash and have its system permit unsealing previous TPMTOTP + TPM DUK and go on.

This fixes config.user being injected over and over the same TPM related settings by making sure they are not preent prior of reinjectig them.

### DIFF
--- a/initrd/init
+++ b/initrd/init
@@ -140,8 +140,14 @@ fi
 #
 # Values in user config have higher priority during combining thus effectively
 # changing the value for the rest of the scripts which source /tmp/config.
-echo "export CONFIG_TPM=\"$CONFIG_TPM\"" >> /etc/config.user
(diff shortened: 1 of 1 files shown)


tlaurion:
Merging as tested as well under #1773 thoroughly. This is actually a bugfix. Merging without needing review.

There i absolutely no good reason for having those entries multiple times, fused in and injected under CBFS as on master.

Merging bypassing review process.

JonathonHall-Purism (commented):
Thanks for this change, agree 100%.  Sorry for the delay looking over this.

PR #1776, opened 2024-09-03
key-init: force user to change time if <2024, give warnings on key import errors

### LINKED ISSUE #1775: Another time drift issue impacting key import at key-init
Reported at https://matrix.to/#/!pAlHOfxQNPXOgFGTmo:matrix.org/$DljnyH05rmRWvb1Ym33e-GADdH_5r4UMdijh4e856Zo?via=matrix.org&via=nitro.chat&via=envs.net

tlaurion:
Fixes #1775 : without this, a machine that had CMOS battery disconnected for initial flashing will show weird and unexpected behaviors, early under key-init since gpg keys imported there have a creation date in the future. 

This is a bug, and since we now have https://github.com/linuxboot/heads/pull/1737 merged in, we can force users to change system clock if in an impossible past, in all case (<2024. Could also be 1970, but why not <2024 just as the GUI forces, today).

----
This PR:
- Check if system clock year < 2024. If so force user to set time with new GUI merged at https://github.com/linuxboot/heads/pull/1737 
- Do not send gpg errors to /dev/null and pretend everything is ok... warn instead
- only import individual user keys if directory present (legacy: TODO deprecate totally as well as under oem-factory-reset and flash scripts)
  - oem-factory-reset and flashing scripts with config preservation takes trustdb+pub keyring+user config overrides and inject them back in firmware when internally flashing. The old individual key import are still there, but not used since years. We could delte at some point. @JonathonHall-Purism : we do this separately? Left a TODO in code under key-init


tlaurion:
```
[    4.643655] DEBUG: Debug output enabled from board CONFIG_DEBUG_OUTPUT=y option (/etc/config)
[    4.669303] TRACE: Under init
[    4.711578] DEBUG: Applying panic_on_oom setting to sysctl
[    4.858228] TRACE: /bin/tpmr(32): main
[    4.910196] TRACE: /bin/tpmr(336): tpm2_startsession
[    5.542699] TRACE: /bin/cbfs-init(5): main
[    5.659485] DEBUG: Extending TPM PCR 7 with /.gnupg/pubring.kbx
[    5.730724] TRACE: /bin/tpmr(32): main
[    5.765843] TRACE: /bin/tpmr(232): tpm2_extend
[    5.867261] DEBUG: tpm2 pcrread sha256:7
[    6.078541] DEBUG: Extending TPM PCR 7 with /.gnupg/trustdb.gpg
[    6.145530] TRACE: /bin/tpmr(32): main
[    6.177971] TRACE: /bin/tpmr(232): tpm2_extend
[    6.286832] DEBUG: tpm2 pcrread sha256:7
[    6.455275] TRACE: /bin/key-init(5): main
[    6.539725]  *** WARNING: Importing user's keys failed ***
[    8.503260] TRACE: Under /etc/ash_functions:combine_configs
[    8.570345] TRACE: Under /etc/ash_functions:pause_recovery
!!! Hit enter to proceed to recovery shell !!!

```

tlaurion:
@JonathonHall-Purism please approve

tlaurion:
@macpijan Docs should be clearer on effect of disconnecting CMOS battery resulting in being back in 1970 at https://docs.dasharo.com/unified/novacustom/initial-deployment/#installing-dasharo

tlaurion:
@JonathonHall-Purism forgot to tag you here for review

tlaurion:
Goes alongside documentation fix at https://github.com/Dasharo/dasharo-issues/issues/1040

@JonathonHall-Purism open internal issue if Purism documentation misses that step too, while less important because Pureboot should pick up on that change on less release automatically once merged?

JonathonHall-Purism:
@tlaurion Suggestion: https://github.com/linuxboot/heads/commit/92396ca6773dad6e160f5475694f570f6e0cb7d9

- The 'warn' message isn't very effective, change-time.sh clears the screen right after.  Prompt with whiptail instead like we do for TOTP reset and everything else.  This is also less jarring IMO, the user knows what is happening before we drop them into a bunch of prompts.
- Allow skipping if the user really wants to.  While they should usually set the time, IMO it's pretty annoying if I know I don't want to do this right now but Heads forces me to anyway.

This is great, thanks for putting this together :muscle: 

Surprisingly, on Librem 14 (and others IIRC), the clock resets to 2070, not 1970. Maybe if I enable the RTC century bit in CMOS it'll reset to 1970 instead.

I think I would rather try that before something like asking to reset if the clock is set to 2070.  46 years seems like a long time, but who knows :wink: 

Anyway this is great, so don't hold it up while I figure out another improvement for Librems.  If you agree with the changes I suggested, let's merge, otherwise feel free to discuss :+1:

tlaurion:
> @tlaurion Suggestion: [92396ca](https://github.com/linuxboot/heads/commit/92396ca6773dad6e160f5475694f570f6e0cb7d9)
> 
> * The 'warn' message isn't very effective, change-time.sh clears the screen right after.  Prompt with whiptail instead like we do for TOTP reset and everything else.  This is also less jarring IMO, the user knows what is happening before we drop them into a bunch of prompts.
> * Allow skipping if the user really wants to.  While they should usually set the time, IMO it's pretty annoying if I know I don't want to do this right now but Heads forces me to anyway.
> 
> This is great, thanks for putting this together 💪
> 
> Surprisingly, on Librem 14 (and others IIRC), the clock resets to 2070, not 1970. Maybe if I enable the RTC century bit in CMOS it'll reset to 1970 instead.

That would help, should most probably set it everywhere if it works for you.
> 
> I think I would rather try that before something like asking to reset if the clock is set to 2070. 46 years seems like a long time, but who knows 😉

Again should not trigger bug, will create weird logs in rootfs, but won't frezze nor create weird behavior under heads, which are because keys imported > 1970 < 2070. I still belive we should sync clock from network but this is not possible nor easy as part of a traditionnal OEM factory reset. IDeally, we would have all computers be in 1970 if possible after CMOS battery disconnection otherwise we do not have a stable behavior to correct upon.
> 
> Anyway this is great, so don't hold it up while I figure out another improvement for Librems. If you agree with the changes I suggested, let's merge, otherwise feel free to discuss 👍

@JonathonHall-Purism merged master+cherry-picked [cbd1f28](https://github.com/linuxboot/heads/pull/1776/commits/cbd1f285c6d1cb2aa2481e4a85bce4dd1d887433)

tlaurion:
@JonathonHall-Purism let's merge whenever you are ready to push merge button

JonathonHall-Purism (approved):

JonathonHall-Purism:
Button pushed!  Thanks

PR #1777, opened 2024-09-04
WiP : Add tracing and debugging to better locate d16 segfault on oem-factory-reset without USB Security dongle connected

tlaurion:
@arhabd test internal flashing of d16 workstation with usb please and provide logs

PR #1698, opened 2024-06-06
Add support for modifying background color when using Newt. Adds ncurses to provide terminfo file.

mdrobnak:
- Tested on text mode MSI dGPU board.
- Tested on Chromebook GOP framebuffer mode.

I think I got all of the occurrences in which colors are modified. Open to modification of variable names etc.

Issue #1395, opened 2023-05-03
KGPE-d16 board status


tlaurion:
Work done under #1378 has not been tested under that board.
Meaning that it is not expected that a flat framebuffer is passed through kexec.

Interesting enough, this board is 5.10.5 linux based for a while.
Seems like users of that board are ok not having no vga until kexec'ed OS is initializing kernel.

Unfortunately, my kgpe-d16 doesn't boot anymore, looping into coreboot raminit.

Someone else still using that board under Heads? (Tagging per #692 board owners)
@0rb677 @Tonux599 @zifxify @blobless ?

Tonux599:
Hi @tlaurion, long time no speak and I hope your well.

This message comes at a convenient time actually as I have recently starting using Heads again after about an 8 month hiatus due to employment requirements.

I installed heads on my KGPE-D16 at commit 77b5933 a couple of weeks ago. Using `kgpe-d16_workstation` and injected some GPU firmware files into the initrd for `amdgpu` module.

One thing that was worth noting was the internal flash was not working, so could not add my GPG/PGP keys the Heads way. I didn't investigate too much but the "flashing" progress bar just did not progress over many minutes. I resolved it by adding the pubring and trustdb with cbfs tool outside of heads and flashing again.

Re the PR above, my install is older than when it was merged. My system has the onboard VGA disabled via the header and Linux inits the GPU and the GUI renders okay.

I would also be keen to see Dasharo used for KGPE-D16 in Heads too. I see you have draft PR #1303 open which I'll try and have a play with at some point.

My ability to test is however restrained by time, lack of DIP8 chips (I'll buy some spares at somepoint) and that it's my main workhorse. So generally, testing for me would be flash a chip, turn off my main computer, swap, test real quick, swap back and report. I'm hesitant to test too deeply / commit myself on my primary system.

I'll ping you a message / comment here when I've restocked and if there is anything I can do let me know :slightly_smiling_face:

tlaurion:
> I didn't investigate too much but the "flashing" progress bar just did not progress over many minutes.

@Tonux599 flash.sh, called by flash-gui.sh, is drawing progress from flashrom -V output in a file.

@Tonux599 It would be nice to just have output from flashrom -p internal -w /path/to/ROM to see what is happening there.

build-cool91:
@tlaurion Hello, I have a KGPE-D16 board with a few extra chips for testing and although I'm not an expert but would be happy to help out. What can I do? 

[...]

PR #1778, opened 2024-09-04
oem-factory-reset gui-init: fix whiptail_error segfaulting because selfcalling itself; fix typo in gui-init

tlaurion:
fix bugs introduced by #1698:
- whiptail_error was self calling itself in case no usb security dongle was connected instead of asking the user to plug it in and try again
- whiptail_error and whiptail_warning are overrides of sourced /etc/gui_functions
  - therefore, whiptail_error calling whiptail_error was self calling passing wrong paramters and whiptail prompt was never showed.
  - fixed by renaming oem-factory-reset's whiptail_error into local_whiptail_error function and changing whiptail_error_die and the only place where this fonction was called directly

--- 

Superseeds #1777 

Addresses https://github.com/linuxboot/heads/issues/1395#issuecomment-2308014938 : thanks for reporting @arhabd

---

No need for review after testing, "emergency" bugfix

### DIFF
--- a/initrd/bin/gui-init
+++ b/initrd/bin/gui-init
@@ -363,7 +363,7 @@ check_gpg_key()
     option=$(cat /tmp/whiptail)
     case "$option" in 
       g )
-        gpg-gui.sh && BG_COLOR_MAIN_MENU="normnal"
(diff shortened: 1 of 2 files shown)


tlaurion:
Tested on qemu and x230. Merging.

PR #1779, opened 2024-09-04
gui_functions: add newline after message telling user that automatic boot will happen unless interrupted

tlaurion:
Aesthetic fix. Otherwise interrupting default boot and next action write on same line.

@JonathonHall-Purism quick review. Approve & Merge when done!


JonathonHall-Purism (approved):
Such a tiny, yet wonderful change :bow: Thank you!

PR #1735, opened 2024-07-27
nitropad-nv41/ns50: readd Intel WIFI otherwise fsp error to fix previous PR regression (preventing master to build)

tlaurion:
Enabling CONFIG_DRIVERS_INTEL_WIFI=y in turn enables CONFIG_DRIVERS_WIFI_GENERIC=y which is needed.

Bugfix, pushed on master without review: this is regression preventing master to report succcess for last master commit.

nv41:
CC         generated/ramstage.o
/home/user/heads/build/x86/coreboot-dasharo/util/crossgcc/xgcc/bin/i386-elf-ld.bfd: nitropad-nv41/romstage/soc/intel/alderlake/romstage/fsp_params.o: in function `fill_fspm_misc_params':
/home/user/heads/build/x86/coreboot-dasharo/src/soc/intel/alderlake/romstage/fsp_params.c:235: undefined reference to `wifi_generic_cnvi_ddr_rfim_enabled'
make[1]: *** [src/arch/x86/Makefile.inc:196: nitropad-nv41/cbfs/fallback/romstage.debug] Error 1

ns50:
CC         generated/ramstage.o
/home/user/heads/build/x86/coreboot-dasharo/util/crossgcc/xgcc/bin/i386-elf-ld.bfd: nitropad-ns50/romstage/soc/intel/alderlake/romstage/fsp_params.o: in function `fill_fspm_misc_params':
/home/user/heads/build/x86/coreboot-dasharo/src/soc/intel/alderlake/romstage/fsp_params.c:235: undefined reference to `wifi_generic_cnvi_ddr_rfim_enabled'

PR #1781, opened 2024-09-05
Librem 11: Fix build, add to CI

JonathonHall-Purism:
Disabling Wi-Fi drivers broke the Librem 11 build, its device tree describes its built-in Wi-Fi card.  We didn't notice because it wasn't in CI; add it.


tlaurion:
Consider readding CONFIG_DRIVERS_INTEL_WIFI=y to all boards, since it is now needed per #1781 and was needed as emergency fix for https://github.com/linuxboot/heads/pull/1735

@JonathonHall-Purism merging here, leaving this note to turn into issue.

tlaurion (approved):
LGTM

Issue #1783, opened 2024-09-06
Heads fails to find /boot partition with Qubes OS 4.3


marmarek:
### A. Provide Hardware Details

1. What board are you using? (Choose from the list of boards [here](https://github.com/eganonoa/heads/tree/master/boards))

This was found on Nitropad T430 (maximized), but I believe it applies to others too.

2. Does your computer have a dGPU or is it iGPU-only?
    - [ ] dGPU (Distinct GPU other then internal GPU)
    - [x] iGPU-only (Internal GPU, normally Intel GPU)

3. Who installed Heads on this computer?
    - [ ] Insurgo (Issues to be reported at https://github.com/linuxboot/heads/issues)
    - [ ] Nitrokey (Issues to be reported at https://github.com/Nitrokey/heads/issues)
    - [ ] Purism (Issues to be reported at https://source.puri.sm/firmware/pureboot/-/issues)
    - [ ] Novacustom (Issues to be reported at https://github.com/Dasharo/dasharo-issues)
    - [ ] HardnenedVault (Issues to be reported at https://github.com/hardenedvault/vaultboot/issues)
    - [ ] Other provider
    - [x] Self-installed

4. What PGP key is being used?
    - [ ] Librem Key (Nitrokey Pro 2 rebranded)
    - [x] Nitrokey Pro
    - [ ] Nitrokey Pro 2
    - [ ] Nitrokey 3 NFC
    - [ ] Nitrokey 3 NFC Mini
    - [ ] Nitrokey Storage
    - [ ] Nitrokey Storage 2
    - [ ] Yubikey
    - [ ] Other

5. Are you using the PGP key to provide HOTP verification?
    - [x] Yes
    - [ ] No
    - [ ] I don't know

### B. Identify how the board was flashed

1. Is this problem related to updating heads or flashing it for the first time?
    - [ ] First-time flash
    - [x] Updating heads 

2. If the problem is related to an update, how did you attempt to apply the update?
    - [ ] Using the Heads menus
    - [ ] Flashrom via the Recovery Shell
    - [ ] External flashing

3. How was Heads initially flashed?
    - [x] External flashing
    - [ ] Internal-only / 1vyprep+1vyrain / skulls
    - [ ] Don't know

4. Was the board flashed with a maximized or non-maximized/legacy rom?
    - [x] Maximized
    - [ ] Non-maximized / legacy
    - [ ] I don't know

5. If Heads was externally flashed, was IFD unlocked?
    - [ ] Yes
[...]

PR #1784, opened 2024-09-08
board configs: remove CONFIG_BOOT_DEV to always detect_boot_device. Skip too small partition mount attempts 

tlaurion:
Fixes https://github.com/linuxboot/heads/issues/1783

- Remove static  CONFIG_BOOT_DEV in board configs
- Skip attempting mount for smaller than 2mb (4096 sectors) partitions under /etc/functions:mount_possible_boot_device

Tested against https://openqa.qubes-os.org/tests/111529/asset/iso/Qubes-4.3.202409051715-x86_64.iso install


tlaurion:
Revisiting https://github.com/linuxboot/heads/pull/1602 since /boot (and mostly everything else under Heads) should try exfat last.

initrd/etc/functions
```diff
@@ -697,8 +697,17 @@ mount_possible_boot_device() {
 		return 1
 	fi
 
-	TRACE "Try mounting $BOOT_DEV as /boot"
-	if mount -o ro "$BOOT_DEV" /boot >/dev/null 2>&1; then
+	# Get the size of BOOT_DEV in 512-byte sectors
+	sectors=$(blockdev --getsz "$BOOT_DEV")
+
+	# Check if the partition is small (less than 2MB, which is 4096 sectors)
```
tlaurion, line 703: @JonathonHall-Purism @marmarek : kinda a hack, but works preventing exfat error spurring into dmesg fixing https://github.com/linuxboot/heads/issues/1783

tlaurion:
@JonathonHall-Purism alternatively we could fix check for fixated size, unsure what else out there could be skipped one way or the other skipping by size/sectors

initrd/etc/functions
marmarek: alternative solution could be looking for fs signature first (blkid?) but I'm not sure if busybox(?) is capable of that...

marmarek: or explicitly skipping "bios boot" partition (would need to get the partition type - again, not sure if easy in busybox

marmarek: if alternatives are not easy to do, I'm okay with the size check too - 2MB is small enough that no legitimate /boot would be affected.

initrd/etc/functions
```diff
@@ -697,8 +697,17 @@ mount_possible_boot_device() {
...
+	if [ "$sectors" -lt 4096 ]; then
+		TRACE_FUNC
+		DEBUG "Partition $BOOT_DEV is very small, likely BIOS boot. Skipping mount."
+		return 1
+	TRACE_FUNC
+	DEBUG "Try mounting $BOOT_DEV as /boot"
+	elif mount -o ro "$BOOT_DEV" /boot >/dev/null 2>&1; then
```
JonathonHall-Purism, line 710:
```suggestion
		TRACE_FUNC
		DEBUG "Try mounting $BOOT_DEV as /boot"
	elif mount -o ro "$BOOT_DEV" /boot >/dev/null 2>&1; then
```

@tlaurion Could you please fix the indentation here?

initrd/etc/functions
JonathonHall-Purism: I'm OK with this as a band-aid improvement :+1:   It does suppress some spurious warnings and I don't see any downside.

I've already made multiple changes trying to suppress these annoying exFAT warnings - IMO, I think in the long run we either need to patch out that warning from the kernel or stop dumping kernel warnings to the console.

This is a great improvement but overall this is looking like a whack-a-mole game trying to address each situation individually.

Again though, let's merge this and I'll open a separate issue for the exFAT warnings.

JonathonHall-Purism (changes requested):
This is a great change :+1:   The /boot detection is reliable now, and hard-coding a default isn't very reliable any more since it depends on the type of storage installed and the partition layout.

I made one suggestion regarding fixing indentation, otherwise it looks great.

tlaurion:
@JonathonHall-Purism @marmarek : I just added blkid type with low cost, and now investigating if we could just simply use blkid output for lines that at least include a UUID prior of doing mount calls.

In the case of interest for @marmarek, sda1 doesn't even show in blkid. with blkid type added under busybox config, partitions not showing a type should definitely not be attempted to be mounted.

This would be a real fix instead of a bandaid fix, where partition size of less than 2mb might hit us in the butt later.

Thoughts before I go too far in that direction?

initrd/etc/functions
JonathonHall-Purism: Actually, @marmarek, it does look for the partition type above (`is_gpt_bios_grub`) - why didn't that exclude your bios-grub partition?

I still don't have a problem with this sector count check though, it's not going to cause any harm.

JonathonHall-Purism: Opened #1788 for exFAT warnings in general

JonathonHall-Purism:
@tlaurion If it helps, that's great, but I was already checking the partition type with fdisk -l (see `is_gpt_bios_grub`).  I'm not sure why this did not exclude @marmarek 's bios-grub partition.  Is the check with blkid any different?

initrd/etc/functions
tlaurion: @JonathonHall-Purism the identation os fine, since the TRACE_FUNC and DEBUG are placed i such indent because they are meant to show where and what will be done under that elif calling an action. The indent is aimed at showing to the reader that the next call is action, not linked to prior code block. Makes sense?

JonathonHall-Purism: No, it doesn't make sense :thinking: 

As written, it will **not** print "Try mounting $BOOT_DEV as /boot" when it is about to try mounting $BOOT_DEV as /boot.  It will **only** print "Try mounting $BOOT_DEV as /boot" when it is **not** about to try mounting $BOOT_DEV as /boot.

That is the opposite of what you want.

Clearly I didn't read the message well enough to connect it to the statement below, that's my fault.  Fixing the indentation reveals that the code is wrong, but actually the code needs to be reworked to be correct.

marmarek:
> @tlaurion If it helps, that's great, but I was already checking the partition type with fdisk -l (see `is_gpt_bios_grub`). I'm not sure why this did not exclude @marmarek 's bios-grub partition. Is the check with blkid any different?

That's a very good question, maybe due to the hardcoded `CONFIG_BOOT_DEV`? If skipping bios boot is already there, perhaps just removing static `CONFIG_BOOT_DEV` is enough?

tlaurion:
> @tlaurion If it helps, that's great, but I was already checking the partition type with fdisk -l (see `is_gpt_bios_grub`).  I'm not sure why this did not exclude @marmarek 's bios-grub partition.  Is the check with blkid any different?

JonathonHall-Purism:
> That's a very good question, maybe due to the hardcoded `CONFIG_BOOT_DEV`? If skipping bios boot is already there, perhaps just removing static `CONFIG_BOOT_DEV` is enough?

Ah, yes, that could be it.  The partition type is probably only checked by the automatic detection, not by the attempt to mount CONFIG_BOOT_DEV.

So @tlaurion maybe we don't need the small partition check, and blkid doesn't seem to be needed either.  (I'd still be OK with the small partition check if you want, it doesn't do any harm and is testable.  You'd have to create a <2MB partition _without_ the bios-grub flag to test it.)

Removing CONFIG_BOOT_DEV defaults everywhere might be enough to cover this case.

tlaurion:
Unfortunately, without mount skip and only removing CONFIG_BOOT_DEV from all boards, on clean boot (wiped all settings from config menu+flash wiped settings) exfat warnings still there

initrd/etc/functions
tlaurion: @JonathonHall-Purism fixed under [1724584](https://github.com/linuxboot/heads/pull/1784/commits/1724584fecaaf512ef707b2a7e071b328b22c13c)

JonathonHall-Purism (approved):

tlaurion:
@JonathonHall-Purism there was still an indentation error. fixed under [faa77d4](https://github.com/linuxboot/heads/pull/1784/commits/faa77d40643ec51f583c9edf5b1cd402f2443a3e)

JonathonHall-Purism:
> @JonathonHall-Purism there was still an indentation error. fixed under [faa77d4](https://github.com/linuxboot/heads/pull/1784/commits/faa77d40643ec51f583c9edf5b1cd402f2443a3e)

This is the diff I see for the latest force-push:

Screenshot_20240909_104942

The indentation was right before, this introduced an error.  But this change showed up just as I was merging, so it already got merged :shrug:

PR #1800, opened 2024-09-26
bin/fetch_source_archive.sh: Add storage.puri.st mirror

JonathonHall-Purism:
storage.puri.st is an alternate host name for storage.puri.sm, in case there is another issue with the .sm name registration.


tlaurion (approved):
Awesome! Thanks @JonathonHall-Purism merging.

PR #1785, opened 2024-09-08
Dell Optiplex 7019/9010 SFF (and DT form factor) + TXT inclusion based on coreboot master 24.02.01 for now

tlaurion:
This pull request includes various updates and additions to support the Dell Optiplex 7010/9010 systems, including new workflows, configuration files, and scripts for handling firmware blobs. The most important changes include updating the Docker image version, adding new jobs and workflows in the CircleCI configuration, and introducing new configuration files and scripts for the Optiplex 7010/9010 systems.

### CircleCI Configuration Updates:
* Updated Docker image version to `tlaurion/heads-dev-env:v0.2.3` in multiple jobs (`prep_env`, `build_and_persist`, `build`, `save_cache`). [[1]](diffhunk://#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47L48-R48) [[2]](diffhunk://#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47L114-R118) [[3]](diffhunk://#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47L142-R146) [[4]](diffhunk://#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47L163-R167)
* Added a new job to download Optiplex 7010/9010 blobs.
* Introduced new workflows for building various configurations of the Optiplex 7010/9010.

### New Configuration Files:
* Added configuration files for different Optiplex 7010/9010 builds (`optiplex-7010_9010-hotp-maximized.config`, `optiplex-7010_9010-maximized.config`, `optiplex-7010_9010_TXT-hotp-maximized.config`, `optiplex-7010_9010_TXT-maximized.config`). [[1]](diffhunk://#diff-bdda7ef01145d01ecf1999aafab35d5a6a286f7cf30cda1ee545086bd744aa8fR1-R93) [[2]](diffhunk://#diff-62051791f14a4d3a4327d3f022f9fd21bb3aedcb398cd17d1e51dda809a47e04R1-R93) [[3]](diffhunk://#diff-67a2e2cb6adb2861636bd135ccdbc59f53e51d0cd9b862d325c454375198bd7fR1-R93) [[4]](diffhunk://#diff-2406de558247565a482b2c9efe9fc44ea924d13381f78a9ed62a2810b9236216R1-R93)

### New Scripts and Documentation:
[...]

PR #1806, opened 2024-10-11
coreboot configs: remove coreboot console logging

tlaurion:
As flagged under #1785 : there is no good reason to keep coreboot console logging. Outside of giving debug info and slowing boot, there is no use for production ready boards.

Removing it for all boards.

@JonathonHall-Purism thoughts?


JonathonHall-Purism:
@tlaurion I think the slowdown noted in #1785 was due to printing the logs to the serial console, not due to generating the logs entirely.

We do sometimes need to ask customers to collect the coreboot log with cbmem after boot, e.g. if there was a problem initializing graphics during boot, but the boot does complete so we can collect the log.

Could we disable the serial console output on boards with serial ports instead of eliminating the logs entirely?

tlaurion:
Superseeded by #1807

PR #1816, opened 2024-10-20
first draft of a kernel configuration transformation test

aluciani:
I was looking at heads' kernel config (for nitropad-nv41 in my case), and I thought there might be some options that would be worth changing (“y”->“is not set”, “is not set”->“y” and some literal values).
This idea comes to me from a13xp0p0v's project [kernel-hardening-checker](https://github.com/a13xp0p0v/kernel-hardening-checker), which aims to verify the security of a linux kernel. It's true that this project is more for server or desktop linux, but I think some options could be useful in the case of head.
I'm talking about the following options:
From "y" to "is not set" :
```
CONFIG_SLAB_MERGE_DEFAULT
CONFIG_MODULES
CONFIG_DEVMEM
CONFIG_LDISC_AUTOLOAD
CONFIG_FB
CONFIG_VT
CONFIG_DEVPORT
CONFIG_IO_URING
CONFIG_KCMP
~CONFIG_X86_IOPL_IOPERM //"needed by flashrom"
CONFIG_ACPI_TABLE_UPGRADE
CONFIG_EFI_CUSTOM_SSDT_OVERLAYS
~CONFIG_MAGIC_SYSRQ // "sysrq for safe belts" 
~CONFIG_MAGIC_SYSRQ_SERIAL // "sysrq for safe belts"
```

From to "is not set" to "y" :
```
CONFIG_INIT_STACK_ALL_ZERO
CONFIG_DEBUG_WX
CONFIG_X86_KERNEL_IBT
CONFIG_BUG_ON_DATA_CORRUPTION
CONFIG_SLAB_FREELIST_HARDENED
CONFIG_SLAB_FREELIST_RANDOM
CONFIG_SHUFFLE_PAGE_ALLOCATOR
CONFIG_FORTIFY_SOURCE
CONFIG_DEBUG_SG
CONFIG_INIT_ON_ALLOC_DEFAULT_ON
CONFIG_SCHED_CORE
CONFIG_KFENCE
CONFIG_KFENCE_SAMPLE_INTERVAL
CONFIG_INIT_ON_FREE_DEFAULT_ON
CONFIG_EFI_DISABLE_PCI_DMA
CONFIG_GCC_PLUGIN_STACKLEAK
CONFIG_STACKLEAK_METRICS
CONFIG_STACKLEAK_RUNTIME_DISABLE
CONFIG_RANDOMIZE_KSTACK_OFFSET_DEFAULT
CONFIG_PAGE_TABLE_CHECK
[...]

PR #1817, opened 2024-10-20
Aluciani:new_config_file (teest under qemu over 6.1.8 kernel bump branch)

tlaurion:
Superseeds #1816 with documentation in commit logs.

@aluciani: please review commit logs and challenge.
Putting as draft.


aluciani:
We might also want to reduce the CONFIG_BLK_DEV_LOOP_MIN_COUNT from 8 to 2.
Each block is around 256 (up to 512) bytes (please correct me if I'm wrong), so 
8 * 256 = 2048 bytes
2 * 256 = 512 bytes
Worth the space.

I wanted to reduce the value of CONFIG_NR_CPUS for the same reason, each supported CPU adds 8KB, so :
For 32 CPUs: 32 CPUs×8192 bytes/CPU =262,144 bytes (≈256KB)
For 24 CPUs: 24 CPUs×8192  bytes/CPU =196,608 bytes (≈192KB)
Memory Savings:
262,144bytes−196,608bytes=65,536bytes(≈64KB)

But I understand that it will be a pain to update every linux configuration for every card every time.

aluciani:
The [kernel-hardening-checker](https://github.com/a13xp0p0v/kernel-hardening-checker)  also talks about CFI, I think it would also be a plus to have it, but I think it would slow head down too much.
CONFIG_CFI_AUTO_DEFAULT

aluciani:
> @aluciani this is nice exercise. Wondering if https://github.com/a13xp0p0v/kernel-hardening-checker should be added under nix docker image and some self-test should be added in CI in long term to make those checks automatic and warn of security regressions, somehow.

https://github.com/linuxboot/heads/pull/1816#issuecomment-2425067431

The tool is good and complete. But the problem is that it has a lot of checks, most of which don't apply here.The ones suggested here are in the minority.
I don't know if integrating it into nix is a good idea. Or it could be for a limited number of configurations.

aluciani:
Seccomp, Stack Leak Protection, and Strict Memory Access could also be enable with thoses parameters :
- `CONFIG_SECCOMP` and `CONFIG_SECCOMP_FILTER` to enhance security by enabling syscall filtering.
- Switch to `CONFIG_EFI_DISABLE_PCI_DMA` for better PCI DMA security in EFI environments.
- Switch IOMMU default behavior to strict mode with `CONFIG_IOMMU_DEFAULT_DMA_STRICT` to enforce stricter DMA rules.
-  `CONFIG_SECURITY_DMESG_RESTRICT` to limit access to kernel messages and reduce attack surface.
-  `CONFIG_STRICT_DEVMEM` and `CONFIG_IO_STRICT_DEVMEM` to restrict access to `/dev/mem` and prevent unintended access to hardware memory.

tlaurion:
A quick reminder that Heads is a controlled flow environment, GUI based where user input are related to secret provisioning. 

Other areas criticized for Heads are filesystem parsing/mounting where not much more can be done then print_tree sanitization in code right now, where luks/ext3/ext4/exfat is a bit out of reach of further sanitization as of now. Also noted that Heads parses syslinux/grub configs, where some additional sanitization could happen, with opened issues. 

I'm open to further kernel hardening, but Heads use case of kernel needs to be understood. Across platforms, different uses of sysfs and devmem happens, where "ultimate lockdown" is currently optional : that is Authenticated wall guarded recovery shell access /usb boot. 

We could tune kernel further more, but corner case impact will be discovered only with extensive testing, which is to be dodged for ease of maintainability and generalization of kernel config across all boards.

@aluciani what additional prevention is thought to be added by hardening further more the kernel with your past comment?

tlaurion:
> Seccomp, Stack Leak Protection, and Strict Memory Access could also be enable with thoses parameters :
> 
> * `CONFIG_SECCOMP` and `CONFIG_SECCOMP_FILTER` to enhance security by enabling syscall filtering.

I don't see why not.

> * Switch to `CONFIG_EFI_DISABLE_PCI_DMA` for better PCI DMA security in EFI environments.

I see possible problems, would need extensive testing (coreboot fb -> efifb basic 2d fb prior of final OS doing drm+3d modules loading and own graphic card on its own with kexec passing coreboot fb to final OS)

> * Switch IOMMU default behavior to strict mode with `CONFIG_IOMMU_DEFAULT_DMA_STRICT` to enforce stricter DMA rules.

Same as above for same reasoning

> * `CONFIG_SECURITY_DMESG_RESTRICT` to limit access to kernel messages and reduce attack surface.

I don't see interest, Heads is single root user where dmesg is used to even dump debug info to be outputted under /tmp/debug.log if put in TRACE+DEBUG mode through config settings to understand what Heads does. No go.

> * `CONFIG_STRICT_DEVMEM` and `CONFIG_IO_STRICT_DEVMEM` to restrict access to `/dev/mem` and prevent unintended access to hardware memory.

I see problems here as well.

PR #1823, opened 2024-10-29
config-gui.sh: fix bug happening when clearing all user config settings/calling config-gui.sh from recovery shell

tlaurion:
Early Recovery Shell access (`r` pressed around seeing asciiart): /boot is not mounted, so need to call helper first before attempting to do operations on /boot.

Also, code was using pipe (`|`) instead of or pipe (`||`) to make sure pipefail would not happen on previous error commands.

Flagged because ./gnupg contains directories which failed to be deleted throwing error on recovery console, breaking script.

### DIFF
--- a/initrd/bin/config-gui.sh
+++ b/initrd/bin/config-gui.sh
@@ -182,19 +182,21 @@ while true; do
                   \n\nDo you want to proceed?" 0 80) then
         read_rom /tmp/config-gui.rom
         # clear local keyring
-        rm /.gnupg/* | true
+        rm -rf /.gnupg/* || true
+
         # clear /boot signatures/checksums
+        detect_boot_device
         mount -o remount,rw /boot
-        rm /boot/kexec* | true
+        rm -f /boot/kexec* || true
         mount -o remount,ro /boot
+        
         # clear GPG keys and user settings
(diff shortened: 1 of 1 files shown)


JonathonHall-Purism (approved):
Thanks for this @tlaurion .  Let's merge it for the bug fix :+1: 

I have a couple of tips for future iterations, but the change here is strictly better than the existing code so we do not have to hold the fixes up.

`        rm -rf /.gnupg/* || true`
I don't really think you need `|| true` for this - `-f` already tolerates nonexistent files, and I would think if it was actually unable to remove a file we'd want to bail.  (Existing code already ignored this though, so no harm in this change.)

`        rm /boot/kexec* || true`
We never really want `rm` to be interactive so you'd pretty much always want `-f`, which also ignores nonexistent files.  So `|| true` wouldn't be needed then.  (But again, existing code already had this so the change is still better.)

tlaurion:
> ` rm /boot/kexec* || true`
> We never really want `rm` to be interactive so you'd pretty much always want `-f`, which also ignores nonexistent files. So `|| true` wouldn't be needed then. (But again, existing code already had this so the change is still better.)

Added ` rm -f /boot/kexec* || true` under [dca3a09](https://github.com/linuxboot/heads/pull/1823/commits/dca3a096f57892501bc942909527929f169b0b56)


Merging

PR #1838, opened 2024-11-05
Enable optional USB keyboard for all current and future boarfs (Requires on-demand enablement through Config Options + flash to persist across reboot/firmware upgrades)

### LINKED ISSUE #1780: Offer USB keyboard support by default (and require optional enablement of USB Keyboard in configuration settings to enable and flash config change to persist across reboot and upgrades)
> `CONFIG_SUPPORT_USB_KEYBOARD=y` allows the user to enable USB keyboard support in config.  It's off by default.  I provide this for Librems due to popular demand.

> You're thinking of `CONFIG_USB_KEYBOARD_REQUIRED=y` which means the device _only_ has USB keyboards.

>I'm not familiar with this hardware, but typically the internal keyboard is a PS/2 keyboard (well, really provided by the EC as an 8042 interface, no actual PS/2 port or signaling, but I digress).  So I think you have this right @loftlifter31 

> _Originally posted by @JonathonHall-Purism in https://github.com/linuxboot/heads/pull/1774#discussion_r1745617700_

tlaurion:
Fixes #1780

- Enables usb hid kernel module to be packed under modules.cpio unconditionally (build errors for board linux config not packing usb hid from CircleCI if not compiled as module or path changed between kernel bumps)
- change config settings script so that usb support is taken for granted (under modules.cpio per build time guarantee)
  - menu toggle to turn on/off USB Keyboard support (warning is explicit), sets user desired usb keyboard as variable exported, no module usb keyboard module loaded until reboot:
    - Requires flashing resulting config.user to cbfs, which makes the setting persisten across reboot/upgrades.
    - It's possible to disable, requiring the same flashing of config.user file to CBFS to not be applied on reboot (=n) 
- Removes usb support in boards that were asking now deprecated support from board config (librems)


tlaurion:
Tested on w530. Ready to merge whenever you are @JonathonHall-Purism . You can press the merge button after approving.

We should probably bring the Configuration Settings menu more up in the list of Settings.

marmarek:
I haven't tested this version, but looks okay and should solve my needs.

JonathonHall-Purism (approved):
Wonderful, thanks @tlaurion .  Looks perfect, merging now

PR #1840, opened 2024-11-06
d16 boards: bump CONFIG_CRYPTSETUP=y to CONFIG_CRYPTSETUP2=y 

tlaurion:
Were still using 1.7.3 which failed to build with newer muslc version bump (Other pending PR incoming)


tlaurion:
Builds as expected, merging as bugfix without review (reuses what works and tested for other boards needing cryptsetup2.6+ anyway.

Issue #1822, opened 2024-10-24
Implement a "Quiet mode", hiding console output into LOG mode to be seen under /tmp/debug.log from recovery console


tlaurion:
The idea here is that Heads is getting more and more mainstreamed and aims to get easier/less scary to newcomers.

History:
- Heads was once used only by technical users and collaborated upon by security oriented people.
- Nowadays, this is shifting a bit where advanced users still wants to see TPM extend operations (that were cleaned and more spot on relative information is now provided on console prior of TPM PCR extend ops) but **Heads newcomers do not want to see verbose information on console (TMI!!!)**: This is the current technical information provided to all on each boot:
  - On early boot: Heads asciiart
  - Followed by user material being extracted from cbfs and measured prior of being used
  - Followed by early, board specific kernel modules measurements + load (usb controlloers kernel modules, usb keyboard)
  - Followed by late runtime state PCR extension/recovery shell access PCR extension to invalidate secret unsealing from TPM nvram
  - Followed by late optional TPM Disk Unlock Key (DUK) additional measurements (LUKS header, some platform TPM event log replay) measurements
  - Followed by /boot signed hash digest validation
  - Followed by HOTP enablement + uatomatic default boot "press any key to exit default boot" prompt
  - Followed by optional DUK passphrase prompt if everything above kosher
  - Followed by board specific chipset locking of SPI chip (PR0)  
  - Followed by TPM cleanup (TPM2)
  - Followed by kexec call
  - OS takeover output

---

With HOTP and automatic boot now being default in all HOTP boards, which are the ones aimed at less technical users as opposed to non-HOTP (TOTP only remote attestation over phone), only required output on default boot could be, for quiet mode:
- HOTP default boot "press any key to exit default boot" prompt to enter GUI
- OS takeover output

With such implementation, OEM could, as part of their rebranding commit on top of chosen master commit, overload board config with something like `export CONFIG_QUIET_MODE=y` and have everything advanced hidden from end users. OF course this would require paid development time, but would reduce noise that is currently scary for non technical users. Log traces will always stay available through recovery shell under /tmp/debug.log, where everything above would be trapped behind LOG helper (see /initrd/etc/ash_functions's LOG) to output in debug.log even if no TRACE/DEBUG is enabled under board's config, leaving otherwise default output to console under file instead. 

---

Thoughts? Comments?
This would be effort to address concerns for downstream OEM users and to help mainstream Head usage and reduce noise for less advanced users.

[...]

PR #1845, opened 2024-11-12
boards/qemu*: add "prod" variants, not built by CircleCI as of now, to test Heads prod console output

tlaurion:
Those basically turn off TRACING, DEBUGGING and PCAP output as opposed to their standard non `-prod` counterpart:
```
#Enable DEBUG output
#export CONFIG_DEBUG_OUTPUT=y
#export CONFIG_ENABLE_FUNCTION_TRACING_OUTPUT=y
#Enable TPM2 pcap output under /tmp
#export CONFIG_TPM2_CAPTURE_PCAP=y
```

This is to facilitate non-platform specific features in virtualized environment.
Note that :
- coreboot config still enables high verbosity kernel output, but once Heads is running, the output is similar to production for tpm1/tpm2/hotp/non-hotp variants.
- OEM only ship HOTP variants
- HOTP variants still not play well with qemu tcg mode (non-kvm/non-nested virt environements like QubesOS: can't fix).

---
Will be used to develop #1822

### DIFF
--- a/boards/qemu-coreboot-fbwhiptail-tpm1-hotp-prod/qemu-coreboot-fbwhiptail-tpm1-hotp-prod.config
+++ b/boards/qemu-coreboot-fbwhiptail-tpm1-hotp-prod/qemu-coreboot-fbwhiptail-tpm1-hotp-prod.config
@@ -1,0 +1,97 @@
+# Configuration for building a coreboot ROM that works in
+# the qemu emulator in console mode thanks to Whiptail
+#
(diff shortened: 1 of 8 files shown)


tlaurion:
Would be helpful to test all sorta things, including but not limited to #1203 when that happens both into flashrom/flashprog and qemu

tlaurion:
Does not impact anything else, need it to provide output for #1849 for screenshots, merging

PR #1875, opened 2024-12-09
TESTING NEEDED: STAGING PR  (quiet mode + diceware + nk3 fixes)

tlaurion:
Latest demos for Request For Comment (RFC) at https://github.com/linuxboot/heads/pull/1875#issuecomment-2552362845
**This changes massively the User eXperience an is an important change for feedback and testing**

know board owners, testing would be better, but commenting on the demo videos would be as much appreciated (tested in qemu and nv41 on my side for both tpm1 and tpm2, with total hosnesty of it all being more tested under tpm2 boards though)
@natterangell @alexmaloteaux @akfhasodh @doob85 @srgrint @Thrilleratplay @nestire @lsafd @bwachter @shamen123 @eganonoa @nitrosimon @jans23 @icequbes1 @weyounsix @zifxify @jnscmns @computer-user123 @tlaurion @osresearch @merge @MrChromebox @n4ru @Tonux599 @househead @pcm720 @fhvyhjriur @3hhh @ThePlexus @akunterkontrolle @rbreslow @ResendeGHF @gaspar-ilom @JonathonHall-Purism @daringer @arhabd @d-wid

---

Summary:
Staging PR including diceware automatic passphrase generation for early 'o' OEM factory reset mode, quiet mode and hotp-verification changes so that secrets app PIN is finally setup per oem-factory-reset prior of use, info output consistent so nk3 is no regression compared to <nk3 security dongles under Heads use case for remote attestation. 
This pull request introduces a new configuration option to enable quiet mode across various board configuration files. The quiet mode ensures that technical information is logged under `/tmp/debug.log` instead of being displayed.

Details:
- early 'o' launches oem-factory-reset in oem mode (menu option could be splitted into User Re-Ownership doing individual secret provisioning. Here oem mode randomizes one single diceware passphrase for all secrets: TPM Owner/GPG Admin/GPG User/Secrets app PIN(nk3).
- Quiet mode enalbed on all boards per https://github.com/linuxboot/heads/pull/1875/commits/65d6fc48ee4f9e8b61bd59f102b60cd01f7a3a39 (Q: do we want this to be the default? Toherwise will be reverted prior of merge and optional, turned on by Options->Configuration Settings menu, where quiet disable DEBUG and vice versa))
- nk3 hotp-verification changes bettering info output (number of retries before locking on GPG Admin/User PIN/Secrets app PIN (nk3 only)
[...]

PR #1888, opened 2025-01-16
doc: Start documenting Heads logging and configuration variables

JonathonHall-Purism:
Start documenting some design aspects of Heads to aid future PR discussions.

Logging has changed quite a few times, and it tends to be difficult to review because we don't have established expectations for how logging behaves.  Logging guidelines are helpful when adding features to keep logging consistent.

We have a lot more config variables now than in the past.  Only some of them are intended to be configured by the user in config.user, and it's not always clear which those are.

The logging documentation is fairly complete if we all agree on it :slightly_smiling_face:   Comments welcome, let's establish a common expectation.

The config documentation is not complete, there are many more variables not documented.  Help appreciated :slightly_smiling_face:  If we agree on the structure, we could merge and do this incrementally.  Some doc is better than none, it doesn't have to be blocked waiting for "perfect" docs IMO.

I've documented the updated logging behavior in #1875 since we agree on the changes in that PR.

- [ ] Agree on logging conventions
- [ ] Agree on config document structure
- [ ] Document more config variables, or agree to do this incrementally as we find time :wink:

### DIFF
--- a/initrd/etc/functions
+++ b/initrd/etc/functions
@@ -12,6 +12,9 @@ die() {
 	exit 1
 }
 
+# Use warn only for output that indicates a _likely_ problem, is _actionable_ to
+# correct, and when we are able to continue with degraded functionalty.
+# Do not overuse this!  See doc/logging.md.
 warn() {
 	if [ "$CONFIG_DEBUG_OUTPUT" = "y" ]; then
 		echo -e " *** WARNING: $* ***" | tee -a /tmp/debug.log /dev/kmsg >/dev/null
@@ -21,6 +24,10 @@ warn() {
 	sleep 1
 }
 
+# Use DEBUG to track decisions made in script/function logic and the context
+# relating to those decisions.  Generally, focus on decision points, because
+# straight-line execution can usually be followed without further tracing.  See
+# doc/logging.md.
 DEBUG() {
 	if [ "$CONFIG_DEBUG_OUTPUT" = "y" ]; then
 		# fold -s -w 960 will wrap lines at 960 characters on the last space before the limit
@@ -30,13 +37,30 @@ DEBUG() {
 	fi
 }
 
+# Use TRACE to trace control flow through Heads.  This is usually called by
+# TRACE_FUNC, but you can use it to additionally trace parameter values, etc.
+# Usually, use this to display unprocessed parameters that your script or
+# function received.  For information about the logic occurring in your script
+# or function, use DEBUG.  See doc/logging.md.
 TRACE() {
 	if [ "$CONFIG_ENABLE_FUNCTION_TRACING_OUTPUT" = "y" ]; then
 		echo "TRACE: $*" | tee -a /tmp/debug.log /dev/kmsg >/dev/null
 	fi
 }
 
-# Function to manage information output level to the console/debug.log
+# Use NOTE to explain behaviors that are _likely_ to be unexpected or confusing.
+# Unlike INFO, this cannot be hidden, as the explained behavior would be too
+# confusing without this output.
+# Don't overuse this - too much NOTE output will cause users to ignore it.  See
+# doc/logging.md.
+NOTE() {
+	echo "NOTE:" "$@" | tee -a /tmp/debug.log
+}
+
+# Use INFO for contextual information that might make sense to non-developers,
+# but that isn't generally needed to use Heads.  Non-developers might use this
+# level to troubleshoot basic problems, so it must make sense without deep
+# knowledge of how Heads works.  See doc/logging.md.
 INFO() {
 	#TODO: add colors to output, here green for INFO?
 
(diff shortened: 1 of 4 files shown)


JonathonHall-Purism:
@tlaurion Thinking on this again, I think this is mergeable if you agree with the logging conventions I described.  Or if not, please comment and let's reach a consensus :slightly_smiling_face: 

I didn't document every variable yet, but docs for some variables are better than docs for none.  We can add them over time.

Thought about this again because I came across some "warnings" that should not be warnings:

```
./etc/functions:        warn "Your GPG User PIN, followed by Enter key will be required for input at: 'Please unlock the card' next prompt"
```

That's not a warning, we are doing exactly what is expected.  This is informational and should be INFO.  (`warn is for output that indicates a problem. We think the user should act on it, but we are able to continue, possibly with degraded functionality.`)

```
./etc/functions:                warn "Rebooting in 3 seconds to enable booting default boot option"
```

Same, in this context "boot default option" is exactly what the user expects us to do.  This message is informational since rebooting to achieve this might be surprising, it should be INFO.

tlaurion:
> @tlaurion Thinking on this again, I think this is mergeable if you agree with the logging conventions I described. Or if not, please comment and let's reach a consensus 🙂
> 
> I didn't document every variable yet, but docs for some variables are better than docs for none. We can add them over time.
> 
> Thought about this again because I came across some "warnings" that should not be warnings:
> 
> ```
> ./etc/functions:        warn "Your GPG User PIN, followed by Enter key will be required for input at: 'Please unlock the card' next prompt"
> ```
> 
> That's not a warning, we are doing exactly what is expected. This is informational and should be INFO. (`warn is for output that indicates a problem. We think the user should act on it, but we are able to continue, possibly with degraded functionality.`)
> 
> ```
> ./etc/functions:                warn "Rebooting in 3 seconds to enable booting default boot option"
> ```
> 
> Same, in this context "boot default option" is exactly what the user expects us to do. This message is informational since rebooting to achieve this might be surprising, it should be INFO.

@JonathonHall-Purism the two examples you bring here is where the crux of the issue is. Turning those warn into INFO would make them silent to the unexperienced user: 
- He would not be warned that the GPG User PIN is what he needs to type on the next gpg prompt. This is why this is currently a warning.
- He would not be warned that the system will reboot in 3 seconds.

This leads to redefining what a warning is. Otherwise I love what was written here and agree with this pr content.

A warning would:
- informs the user of something that is not preventing the system from functionning (not an error), to bring his attention on something that is important. Everything else is either INFO, or DEBUG.

Thoughts?

JonathonHall-Purism:
> @JonathonHall-Purism the two examples you bring here is where the crux of the issue is. Turning those warn into INFO would make them silent to the unexperienced user:
> 
>     * He would not be warned that the GPG User PIN is what he needs to type on the next gpg prompt. This is why this is currently a warning.
> 
>     * He would not be warned that the system will reboot in 3 seconds.

OK, I see your point here.  These two messages are important because the behavior is not clear without them.  The GPG prompts are confusing, so we need to try to clear that up with this message.  The reboot is normal and accomplishes what the user asks, but could be surprising.

So `INFO` is not appropriate - INFO messages are less important and can be hidden.  These messages should always appear, they shouldn't hide.

However, I don't think `warn` is appropriate either.  If we start displaying `*** WARNING: $* ***` for things that are _normal_ and _not actionable_ (even if they are potentially surprising), users will become accustomed to ignoring `*** WARNING: $* ***`.  We won't be able to get their attention when it is important and the user SHOULD take action.

I think this is a justifiable example for a level in between `INFO` and `warn`, say `NOTE`.  NOTE would be for potentially-surprising but normal behaviors.  It cannot be hidden like INFO, but it is not actionable like `warn`.

Let me try putting together a change for this.  BTW, I think it is a huge documentation win if documenting our behavior exposes something that's confusing, and we fix it :muscle:

initrd/etc/functions
```diff
@@ -30,13 +37,30 @@ DEBUG() {
...
 
-# Function to manage information output level to the console/debug.log
+# Use NOTE to explain behaviors that are _likely_ to be unexpected or confusing.
+# Unlike INFO, this cannot be hidden, as the explained behavior would be too
+# confusing without this output.
+# Don't overuse this - too much NOTE output will cause users to ignore it.  See
+# doc/logging.md.
+NOTE() {
+	echo "NOTE:" "$@" | tee -a /tmp/debug.log
+}
```
tlaurion, line 58: @JonathonHall-Purism I love eveyrhing I see here, but the change to warn to note removing the 2 seconds that was added by warn. Not a big blocker, but sometimes, the note happening before a clear, a reboot, or a binary input might just pass as silence.

Your call, but its also part of why warn was used (the sleep) permitting user to acknoeldge in absence of output and action on the console.

tlaurion (changes requested):
warn -> note : no 2 seconds sleep on note?

tlaurion:
Ping @JonathonHall-Purism

tlaurion:
> warn -> note : no 2 seconds sleep on note?

This PR is LGTM and the proposed additional change can be discussed in another PR.

PR #1908, opened 2025-02-12
Coreboot version  bump 24.02.01 ->  24.12 for most board + Thinkpad T480 addition (coreboot upstream in WiP state)

tlaurion:
- **Based and superseeds Heads t480 community port effort under https://github.com/linuxboot/heads/pull/1906**

---

- t480 WiP upstream https://review.coreboot.org/c/coreboot/+/83274 disabled thermal control and broke  build of all other thinkpads
  - **fixed by @gaspar-ilom under https://github.com/linuxboot/heads/pull/1908/commits/f76558601576409c2bb0fc49f0a119a97d8506fe which affects all skylake+ and still need proper fixes upstream by coreboot devs.** 

Leah blocked me on mastodon because I urged to work upstream and lbmk stop building things in bubble, since this harms ecosystem, including coreboot upstream which won't participate much in review process if patchset proposed cause regressions. Oh well. 

---

Since https://github.com/linuxboot/heads/pull/1908/commits/f76558601576409c2bb0fc49f0a119a97d8506fe now permits to version bump all prior 24.02.01 boards to 24.12, including t480, we need to retest all boards prior of merging:

- [x] Optiplex 7010/9010 SFF/DT: @tlaurion (owns DT variant)
  - [x] Thanks @tlaurion for testing report under  https://github.com/linuxboot/heads/pull/1908#issuecomment-2711862625
- [ ] x230t : @fhvyhjriur
  - [x] from coreboot perspective, its a x230. No testing expected needed here
- [ ] t530 (xx30): @fhvyhjriur @3hhh (Opportunity to mainstream and close https://github.com/linuxboot/heads/issues/1682)
  - [x] moved to untested per https://github.com/linuxboot/heads/pull/1908#issuecomment-2711603241 
- [ ] x230-fhd/edp variant: @n4ru @computer-user123 (nitro caster board) @Tonux599 @househead @pcm720 (eDP 4.0 board and 1440p display) @doob85 https://matrix.to/#/@rsabdpy:matrix.org (agan mod board)
  - [x] tested in prior version of this pr, changes following not expected to impact edp, should be equal as tested x230 
- [ ] z220-cmt (HP Z220 CMT): @d-wid
  - [x] moved to untested per https://github.com/linuxboot/heads/pull/1908#issuecomment-2711603241 
- [x] t420 (xx20): @notgivenby @alexmaloteaux @akfhasodh @doob85
[...]

PR #1920, opened 2025-03-03
coreboot 24.12 bump with libreboot patchset (no thermal.asl commented per patch) to see all bumped boards fail compilation in CircleCI

tlaurion:
So that we can show problem to upstream and blocker to https://github.com/linuxboot/heads/pull/1908 per https://review.coreboot.org/c/coreboot/+/83274


patches/coreboot-24.12/0003-mb-lenovo-Add-ThinkPad-T480-and-ThinkPad-T480s.patch
gaspar-ilom, line 1: @tlaurion Why do we keep the patches and this one in particular at all if we are already dropping the T480 from this PR?

gaspar-ilom: Nvm, I should probably left this in #1908

tlaurion:
Fixed by @gaspar-ilom under #1908 commit https://github.com/linuxboot/heads/pull/1908/commits/f76558601576409c2bb0fc49f0a119a97d8506fe

PR #1928, opened 2025-03-12
bin/seed_package_mirror.sh: Update talos-2 -> UNTESTED_talos-2

JonathonHall-Purism:
Board name was changed due to being untested.  It still builds, so the packages still sync to mirrors for now.


JonathonHall-Purism:
I updated the Purism package mirror using the fixed script.

tlaurion (approved):
Thanks @JonathonHall-Purism for reseeding the mirror with coreboot 24.12 crossgcc deps as can be seen under 

https://github.com/linuxboot/heads/blob/b48fca280b3d906fcce56db777213c7d6ab84652/bin/fetch_source_archive.sh#L5-L7

Let's merge!

PR #1943, opened 2025-04-02
Add kconfig to use blobs on 24.12 boards coreboot configs and proper build output to console

tlaurion:
- Fixes #1940 by adding `CONFIG_USE_BLOBS=y` in all boards that switched from coreboot 24.02.01 tarball release to coreboot fit repo with 24.12 (race condition introduced under https://github.com/linuxboot/heads/pull/1908)
- Makefile: have coreboot git repo actions be more verbose on console output
- Other coreboot configs: saved in oldconfig format so that config is as used by coreboot  (see https://github.com/linuxboot/heads/pull/1943/commits/fe4b5e638fa362689648ffa0d6b793508d261bab commit log for repro)

PR #1953, opened 2025-04-10
Revert previous applied patches in reverse order if applying patches fails and reapply patches + canary wipe dev helper improvements

### LINKED ISSUE #1952: Makefile: Have patches reverted if they fail to apply cleanly
Heads currently applies patches that sometimes create files into build dir.

Heads buildsystem Makefile logic implies in build dir of modules/* : 
- .canary : if doesn't exist, unpack tarball/clone repo and touch file
- .patched : if doesn't exist: apply patches/MODULE-VERSION.patch or patches/MODULE-VERSION/NUMBER-DESCRIPTION.patch application and touch file
- .configured: if doesn't exist, configure and touch
- .build: if doesn't exist, build and touch file

A proper rebuild helper would be to wipe .canary files. Doing so would force archive/git repo recreation, but issue arises when previously created files from patches are present in build dir. 

Logic to cleanly attempt to apply patch, and if patch apply fails, revert patches in reverse order and then reapply patches would permit fast rebuilds of just changed source files per modules makefiles.

Unfortunately, this is way more work then expected under #1943 and was removed from there to this separate issue.

----

Original discussion:
>How about `git apply -R` ? If heads build system could loop through patches and revert them, then maybe it would be fine? But you would have to revert them in reversed order, as opposed to the order they were applied...

_Originally posted by @miczyg1 in https://github.com/linuxboot/heads/pull/1943#discussion_r2027170818_

tlaurion:
Fixes #1952 

WiP (currently includes #1943 + work done here)


tlaurion:
@miczyg1 this addresses discussions under https://github.com/linuxboot/heads/issues/1952

Run:
- `./docker_repro.sh make BOARD=x230-hotp-maximized real.remove_canary_files-extract_patch_rebuild_what_changed`
- `./docker_repro.sh make BOARD=x230-hotp-maximized`

modules/coreboot
```diff
@@ -150,6 +150,7 @@ endif
 
 $(coreboot_module)_configure := \
 	mkdir -p "$(build)/$(coreboot_dir)"; \
+	rm -rf $(build)/$(coreboot_dir)/*; \
```
tlaurion, line 153: This poses the question if we should maybe move the coreboot board built artifacts out of coreboot build directory though

ie: this wipes builds/x86/coreboot-24.12/x230-maximized/* here.

Where shoul be coreboot board built artifacts prior of stitching a board rom if not under coreboot build directory?

This has considerations for CircleCI caching as well. We could move this out of build. Maybe tmp/$BOARD ?

That would require to modify `coreboot_dir` so where that would be
@JonathonHall-Purism insights?

tlaurion: Was moved to dev cycle helper to clean coreboot_dir (build/ARCH/coreboot-VERSION/BOARD) and board_dir (build/ARCH/BOARD) to have reproducible builds

JonathonHall-Purism (approved):

Issue #1956, opened 2025-04-16
cbmem -L output with TPM1.2 under coreboot 24.12 segfaults (inconsistently: needs debug)


marmarek:
### A. Provide Hardware Details

1. What board are you using? (Choose from the list of boards [here](https://github.com/eganonoa/heads/tree/master/boards))

x230

2. Does your computer have a dGPU or is it iGPU-only?
    - [ ] dGPU (Distinct GPU other then internal GPU)
    - [x] iGPU-only (Internal GPU, normally Intel GPU)

3. Who installed Heads on this computer?
    - [x] Insurgo (Issues to be reported at https://github.com/linuxboot/heads/issues)
    - [ ] Nitrokey (Issues to be reported at https://github.com/Nitrokey/heads/issues)
    - [ ] Purism (Issues to be reported at https://source.puri.sm/firmware/pureboot/-/issues)
    - [ ] Novacustom (Issues to be reported at https://github.com/Dasharo/dasharo-issues)
    - [ ] HardnenedVault (Issues to be reported at https://github.com/hardenedvault/vaultboot/issues)
    - [ ] Other provider
    - [ ] Self-installed

4. What PGP key is being used?
    - [ ] Librem Key (Nitrokey Pro 2 rebranded)
    - [ ] Nitrokey Pro
    - [x] Nitrokey Pro 2
    - [ ] Nitrokey 3 NFC
    - [ ] Nitrokey 3 NFC Mini
    - [ ] Nitrokey Storage
    - [ ] Nitrokey Storage 2
    - [ ] Yubikey
    - [ ] Other

5. Are you using the PGP key to provide HOTP verification?
    - [x] Yes
    - [ ] No
    - [ ] I don't know

### B. Identify how the board was flashed

1. Is this problem related to updating heads or flashing it for the first time?
    - [ ] First-time flash
    - [x] Updating heads 

2. If the problem is related to an update, how did you attempt to apply the update?
    - [x] Using the Heads menus
    - [ ] Flashrom via the Recovery Shell
    - [ ] External flashing

3. How was Heads initially flashed?
    - [ ] External flashing
    - [ ] Internal-only / 1vyprep+1vyrain / skulls
    - [ ] Don't know

4. Was the board flashed with a maximized or non-maximized/legacy rom?
    - [x] Maximized
    - [ ] Non-maximized / legacy
    - [ ] I don't know

5. If Heads was externally flashed, was IFD unlocked?
    - [ ] Yes
    - [ ] No
[...]

PR #1824, opened 2024-10-30
patches/coreboot-*: fix tpm1 cbmem crash on 'cbmem -L'

tlaurion:
fix #1608 for Dasharo, 24.02.01 and Purism coreboot forks

@JonathonHall-Purism please review

PR #1959, opened 2025-04-17
patches/coreboot-24.12/0010-cbmem_tpm-clear_whole_log_on_creation.patch: readd missing uptream cbmem patch for TPM 1.2

tlaurion:
Fixes https://github.com/linuxboot/heads/issues/1956

This is is still unmerged coreboot upstream since 24.02.01 (meaning more than a year now: February 2024) : https://review.coreboot.org/c/coreboot/+/84926

Cleaner solution would be to clear whole area on creation: https://review.coreboot.org/c/coreboot/+/84927, but this patch doesn't apply cleanly and is not even under review yet. 

@marmarek let me know if that fixes it for you. I cannot reproduce on my side aymore.


tlaurion:
Was part of patches applied by heads on 24.02.01 from https://github.com/linuxboot/heads/pull/1824
Still not merged upstream

marmarek:
Yes, this fixes the issue for me, thanks!

PR #1987, opened 2025-07-16
initrd/bin/reboot bugfix: switch from sysrq instant reset to busybox' s reboot kernel syscall (so tpm2.0 does not increment TPM DA counter by one)

### LINKED ISSUE #1986: /bin/reboot calls cause TPM2 DA increment
Reported under https://github.com/Dasharo/dasharo-issues/issues/1396 and diagnosed under https://github.com/tlaurion/heads/tree/tpm2_da-troubleshoot, issue discovered by @marmarek through OpenQA tests.

Current bug:
https://github.com/linuxboot/heads/blob/08f52af033411286aefd77eca6ec09d993462b7c/initrd/bin/reboot#L26-L27

Current dictionary attack policy, applied per TPM reset (otherwise varies across TPM2 vendors)
https://github.com/linuxboot/heads/blob/08f52af033411286aefd77eca6ec09d993462b7c/initrd/bin/tpmr#L685-L706
- 10 attempts prior of TPM2 DA lockout
- Each attempt has a counter of 3600s (1h) attached to it; expiring in that configured time window
- When the 10 counters are used; TPM2 DA lockout happens and requires TPM reset to unlock (unlock passphrase is currently randomized, and there is no lockout-recovery-time (configured=0)

Which requires:
- /bin/reboot calling busybox reboot to issue proper syscalls (ctrl+alt+delete does the right thing, so does poweroff)
- reviewing TPM reset policy deployed
- Automated testing going away of "pulling the plug" to reset platforms which goes against TPM2 spec part 1, section 19.8.6, aka "Non-orderly Shutdown"

Some notes:
- v540tu TPM2 when reset, prior of applying policy has
- 7200s (2h)
- 32 attempts prior of triggering a TPM2 DA lockout

Waiting for input on what we want to fix here from referred issue and what is proper PR to fix cucrent issue and downstream opened one.

CC @JonathonHall-Purism @marmarek

tlaurion:
This PR removes Heads bug of causing TPM DA lockout counter to increment each time /bin/reboot was called which forced a reset through  sysrq 'b' usage, even though tpmr shutdown was called (which prepares for shutdown, but is not enough for a non "orderly shutdown" to not trigger TPM DA counter to be triggered.

Fixes #1986 
Improves https://github.com/Dasharo/dasharo-issues/issues/1396 (see TODO below to close that issue)

---

Changes:
- config/busybox.config:	add reboot command under busybox
- initrd/bin/reboot:	use 'busybox reboot -f' instead of 'echo b > /proc/sysrq-trigger'
- modules/busybox:	remove /sbin/reboot symlink to busybox so that our /bin/reboot is called, which calls busybox for reboot

---

Todo:
- https://github.com/Dasharo/dasharo-issues/issues/1396 : review tpmr's tpm reset policy for bruteforce/bad auth: 10 TPM DA counters of each 1h each (meaning that a combination of 10 hard reset and/or 10 bad TPM Disk Unlock Key passphrase and/or TPMTOTP reseal into TPM will result in TPM DA if the 10 TPM DA counters can't age out of their 1h window, causing TPM2 DA lockout. a TPM DA lockout currently requires a TPM reset, which in turn requires TPM Owner password, HOTP dongle PIN and LUKS Disk Recovery Key to reconfigure boot default's configured TPM unsealed Disk Unlock Key (DUK) to unlock LUKS container in final OS.


tlaurion:
@marmarek please test and thumbs up this PR OP description to merge.

tlaurion:
Note that this PR takes a subset of changes from https://github.com/tlaurion/heads/tree/tpm2_da-troubleshoot that was used to add DEBUG calls and understand the problem, which led to /bin/reboot consuming 1 TPM DA counter, being same as when power button was held. Also note that ctrl-alt-delete issues the same syscall of busybox's reboot here.

So with that merged, this leaves 10 TPM DA counters of 1h to deal with forced hard/cold reboot/poweroff without issuing proper kernel syscalls, and actual TPM DA lockout caused by bad TPM DUK passphrases, where 10 counters of 1h still seem fine outside of bug fixed by this PR for hard reset/poweroff.

tlaurion:
@JonathonHall-Purism will merge at end of day; this is bugfix.

JonathonHall-Purism:
LGTM.  This is definitely a better approach, thanks all.

It seems possible that the reason it did not work before could have been the kernel interacting with the TPM again for one of the remaining sysrqs that followed TPM2 shutdown.  Not sure though, and offhand I'm not sure why exactly it would.

At any rate, this is a much better approach and I trust the testing you all described.

JonathonHall-Purism (approved):

PR #1993, opened 2025-08-26
initrd/etc/gui_functions pause_automatic_boot: show TOTP code while waiting for automatic boot is set and hotp is valid

tlaurion:
This pull request enhances the user experience during the automatic boot countdown in the `initrd/etc/gui_functions` script. The main improvement is the addition of a dynamic countdown display that updates every second and optionally shows the current TOTP code if TPM is enabled and QR code display is not skipped.

User experience improvements:

* The automatic boot countdown now updates every second, providing a more interactive and visible countdown for the user.
* If TPM is enabled and QR code display is not skipped, the current TOTP code is retrieved and displayed alongside the countdown, giving users immediate access to authentication information.
* The script checks for a keypress each second, allowing the user to interrupt the boot process at any time during the countdown.

---

In action under qemu-coreboot-whiptail-tpm2-hotp-prod (not silent, not debug):


tlaurion:
signal-2025-09-22-144315

Self reviewed. @JonathonHall-Purism comments?

Tonux599:
> @Tonux599 comments?

This is only for HOTP configs? I can't test that, however code looks fine to me. Would be nice if this could be extended to non-hotp configs also.

Tonux599:
> @Tonux599 then if not in pause waiting for automatic boot delay to come to 0, this could be put in place while waiting user to press Esc keyboard key before typing TPM DUK passphrase?
> 
> Otherwise as of now, TOTP is only shown from Heads main menu.

I don't think in TPM DUK, as not everyone will use that (I trust my TPM to measure the boot, but not to store my disk encryption key).

https://github.com/linuxboot/heads/blob/46f3091c6cc6a70e2c4f90ee1fe9bb98ddbd8717/initrd/bin/gui-init#L660-L662

If above was changed to something like (pseudo-ish code):

```
if [ -n "$CONFIG_AUTO_BOOT_TIMEOUT" ]; then 
  if [ "$CONFIG_HOTPKEY" = "y" -a "$HOTP" != "Success" ]; then
    # Return an error here and stop autoboot.
  fi

  prompt_auto_default_boot 
fi 
```

I think that ought to catch all?

PR #2013, opened 2025-10-28
Use docker "immutable identifier" instead of tag

Tonux599:
CircleCI and `docker_repro.sh` should use Docker's immutable identifier (sha256 digest of image) instead of tags.

Currently, using tags, the administrators of Docker Hub could be coerced into modifying `tlaurion/heads-dev-env` to produce malicious ROM's.

@tlaurion the safest way to ensure that CircleCI and local builds with `docker_repro.sh` are not tainted by a malicious images would be to use immutable identifiers instead of tags. Going forward, I would recommend you build your container locally, taking note of the sha256 digest, then pushing to docker hub before creating a signed commit replacing the checksums in `.circleci/config.yml`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -48,7 +48,7 @@ commands:
 jobs:
   prep_env:
     docker:
-      - image: tlaurion/heads-dev-env:v0.2.5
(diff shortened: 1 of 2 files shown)


tlaurion:
To-do : document under README.md with copy paste related commands
- [ ] "Going forward, I would recommend you build your container locally, taking note of the sha256 digest, then pushing to docker hub before creating a signed commit replacing the checksums in .circleci/config.yml."

So that next maintainer can reuse this knowledge.

Issue #2014, opened 2025-10-28
Upgrading Intel based platforms with Intel Ethernet (GBE region) sets MAC adress to 00:DE:AD:C0:FF:EE


nestire:
### A. Provide Hardware Details

1. What board are you using? (Choose from the list of boards [here](https://github.com/eganonoa/heads/tree/master/boards))

T480 mazimized htop

3. Who installed Heads on this computer?
    - [ X] Self-installed

4. What PGP key is being used?
    - [X ] Nitrokey 3 NFC

5. Are you using the PGP key to provide HOTP verification?
    - [X ] Yes

### B. Identify how the board was flashed

1. Is this problem related to updating heads or flashing it for the first time
    - [ X] Updating heads 

2. If the problem is related to an update, how did you attempt to apply the update?
    - [X ] Using the Heads menus

3. How was Heads initially flashed?
    - [X] External flashing

4. Was the board flashed with a maximized or non-maximized/legacy rom?
    - [X ] Maximized

5. If Heads was externally flashed, was IFD unlocked?
    - [X] Yes

### C. Identify the rom related to this bug report

1. Did you download or build the rom at issue in this bug report?
    - [X ] I downloaded it

2. If you downloaded your rom, where did you get it from?
    - [X] Heads CircleCi

    *Please provide the release number or otherwise identify the rom downloaded*

3. If you built your rom, which repository:branch did you use?
    - [X ] Heads:Master

## Please describe the problem

**Describe the bug**
After an Firmeware upgrade the mac adress of the ethernet interface has the dummy mac adresse 00:de:ad:be:ef thats creates problems in networks

**To Reproduce**
upgrade the Firmware on a t480
check mac adress in the installed OS
**Expected behavior**
the mac adress should persist 

**Additional context**

tlaurion:
This is true for all Intel based laptop boards having gbe configuration blobs to produce a "maximized" rim which otherwise would be invalid when flashed externally as a whole. 

You want to change this for all platforms and document globally that the firmware image contains a forged Mac unless initial flash doesn't overwrite gbe region and that internal upgrading doesn't modify MAC afterward unless gbe added into the flasher command?

tlaurion:
> Where should i document this?

@nestire

[...]

PR #2015, opened 2025-10-28
Perserve mac adress in firmware upgrades for t480

nestire:
see issue #2014 

tested with a t480


nestire:
not needed anymore with #2171

PR #2017, opened 2025-10-31
Require sudo for docker build command

Danrancan:
Updated README to require sudo for docker build command. This is necessary and tested for building on Ubuntu 25.04. If sudo is not used you will recieve build errors, and build won't finish.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ Heads now builds with Nix built docker images since https://github.com/linuxboot
 
 The short path to build Heads is to do what CircleCI would do (./docker_repro.sh under heads git cloned directory):
 - Install _docker-ce_ for your OS of choice (refer to their documentation)
-- run `./docker_repro.sh make BOARD=XYZ`
+- run `sudo ./docker_repro.sh make BOARD=XYZ`
 
 Using Nix local dev environement / building docker images with Nix
 ==


README.md
```diff
@@ -37,7 +37,7 @@ Heads now builds with Nix built docker images since https://github.com/linuxboot
 
 The short path to build Heads is to do what CircleCI would do (./docker_repro.sh under heads git cloned directory):
 - Install _docker-ce_ for your OS of choice (refer to their documentation)
-- run `./docker_repro.sh make BOARD=XYZ`
+- run `sudo ./docker_repro.sh make BOARD=XYZ`
```
Tonux599, line 40: This isn't the only place in the README the docker scripts are used. I think it would be better to add a notice that the scripts require root privileges.

tlaurion:
@Danrancan will you update the docs once your local setup works as expected? That would be useful so others that did the same as you arrive to a working state too.

Thanks.

This PR can be closed. See @Tonux599 comments which I second.

Danrancan:
> @Danrancan will you update the docs once your local setup works as expected? 

Sure I can, but what exactly do you mean by "Update the docs"? Do you want me to just find where documantation says "./docker_repro.sh ..." without sudo, and add sudo to the documentation? Or, what exactly are the "docs", and what do you mean "update"?

UPDATE: As we speak, I am currently building heads again from source on Ubuntu 25.10, and I am attempting to do so with, and then without sudo. I will report back the exact errors we get if we fail to use sudo, and any notable lines we get when we DO use sudo. 

I am also writing a tutorial about this process for my website, but I would be more than happy to contribute a tutorial directly to the heads documentation. Well see where I go with that after I increase my understanding of heads a bit more.

Tonux599:
> Do you want me to just find where documantation says "./docker_repro.sh ..." without sudo, and add sudo to the documentation?

No. Not every user on every system will need to use `sudo` for `./docker_repro.sh` or any of the other docker scripts. As previously linked: https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user

> Or, what exactly are the "docs", and what do you mean "update"?

The "docs" are the `.md` files in this repository and https://github.com/linuxboot/heads-wiki/
"Update" is what you are doing by creating this PR.

As I said previously, it would be better to add a note to the docs clarifying that the scripts require root privileges via `sudo` or the user being part of the `docker` group.

tlaurion:
tagged as documentation with help wanted. This PR won't be merged as is.

PR #2019, opened 2025-10-31
ci: enable automatic reruns

Tonux599:
CircleCI seems quite flaky. I've had to re-run from failed a few times to get a successful build.

This PR will automatically attempt to re-run the failed workflow 3 times before failing.

Eventually, issues causing the fails should be investigated and minimised.


tlaurion (approved):
LGTM! Thanks for having digged that up! Merging!

PR #2022, opened 2025-11-05
*WIP* ci: dynamic configuration

Tonux599:
CircleCI now allows for dynamic generation of configuration files. I'm going to be working on generating the CI workflow file using a script, so that we don't need to manually add boards to the CI config file.

This will probably take some time, at the moment the .yml files have be gutted of anything useful. However, right now it does pick 1 of each coreboot boards and branches out from that, as well as determining the `arch` variable.

For the super messy testing branch (which would just spam the PR with the amount of push's I'll have to make), see: https://github.com/Tonux599/heads/tree/circleci-dynamic-configuration-staging


Tonux599:
@tlaurion eventually (no rush, this will take time) you will need to enable dynamic configuration for `linuxboot/heads`:

https://circleci.com/docs/guides/orchestrate/dynamic-config/#enable-dynamic-config

Issue #692, opened 2020-03-09
Platform blobs, collaborators/maintainers/testers for faster problems resolution


tlaurion:
#### Intel
xxx0: [gm45 bridge, Montevina: no FSP, no ME: X200, T400, T500, R500, X300](https://doc.coreboot.org/mainboard/lenovo/montevina_series.html) : **no QubesOS support there** (no proper vt-d2)
[xx20](https://doc.coreboot.org/mainboard/lenovo/x2xx_series.html): [Sandy bridge, no FSP. ME<10: BUP module required only: X220/T420/T520](https://doc.coreboot.org/mainboard/lenovo/Sandy_Bridge_series.html)
xx30: [Ivy bridge, no FSP. ME<10: ROMP and BUP required: X230/T430/W530](https://doc.coreboot.org/mainboard/lenovo/Ivy_Bridge_series.html) Z220 CMT and others

##### Additional required Intel blobs: 
[FSP is present in all Broadwell+ platforms](https://doc.coreboot.org/soc/intel/fsp/index.html)
[MRC: Memory Reference Code blob](https://doc.coreboot.org/northbridge/intel/haswell/mrc.bin.html) required in Broadwell+ (T440p/w541) : [follow ongoing coreboot Native Ram Initialization (NRI effort)](https://github.com/linuxboot/heads/issues/1711)

##### ME status on different boards models
[Removed in ME <=6 (xxx0)](https://libreboot.org/faq.html#intelme)
[Deactivated+Neutered ME in ME  6 <= 10 (xx20 BUP/xx30 BUP+ROMP)](https://github.com/corna/me_cleaner/wiki/How-does-it-work%3F#me-versions-from-60-nehalem-to-10x-broadwell-1)
[Deactivate+Partially Neutered (BUP, RBE, Kernel and syslibs modules **REQUIRED** in ME > 11)](https://github.com/corna/me_cleaner/wiki/How-does-it-work%3F#me-versions-from-11x-skylake-1)
[Soft disable/HAP disable bit possible on ME 12+ (**PoC BE CAUTIOUS**)](https://github.com/corna/me_cleaner/pull/384)

[xx30, xx20](https://github.com/corna/me_cleaner/wiki/How-does-it-work%3F#me-versions-from-60-nehalem-to-10x-broadwell): ME 6 <= 10
[...]

PR #2030, opened 2025-11-20
T480: fix the headphone jack

gaspar-ilom:
Adds an unmerged patch from coreboot to fix the headphone jack for the 480.
See the details in the commit message in the patch upstream: https://review.coreboot.org/c/coreboot/+/90023

Without this patch the headphone jack is not automatically detect and cannot even be switched to when using pipewire but only with pulseaudio.

## Test reports
- [x] @gaspar-ilom  @doritos4mlady @MattClifton76 @notgivenby @akunterkontrolle
  - Thanks for the report @gaspar-ilom https://github.com/linuxboot/heads/pull/2030#issuecomment-3562377305

### DIFF
--- a/blobs/xx80/README.md
+++ b/blobs/xx80/README.md
@@ -54,3 +54,7 @@ A guide on how to flash this board (both the Heads rom and the Thunderbolt `tb.b
 https://osresearch.net/T480-maximized-flashing/
 
 The upstream documentation is available here. It includes a list of known issues: https://doc.coreboot.org/mainboard/lenovo/t480.html
+Please note that some of the listed issues have been fixed under heads by using patches that are not yet merged upstream:
+* headphone jack works as expected and is automatically detected when plugged in
+* thunderbolt works
+* lower USB-C port works


gaspar-ilom:
> @gaspar-ilom what about documenting left known issues under the board configs? Agreed that wiki should tell users to look at board configs for current commit status, otherwise I see disjunction between docs and code.
> 
> https://github.com/linuxboot/heads/blob/master/boards%2FEOL_t480-maximized%2FEOL_t480-maximized.config

@tlaurion my  proposal would be to require all (maintained) boards under heads to have a markdown file per board in some dedicated `doc/` directory that lists the current state of the board in an understandable way for heads end users. This should replace the `README.md` files under `blobs/` and the documentation in the comment in the config. We can still keep documenting things in the config if its technical information meant for developers/ board testers but not for end users. Under `blobs/` we should also only have technical information for devs.

From heads wiki Github CI we can pull that directory and include it when building the website. This would provide us with a way to maintain the board specific documentation here alongside the code and it would be easy to make sure everything is in sync during review, while it would allow us to list all available boards in the wiki and give some specific information on them. Should also include the state (EOL, (un-)tested, unmaintained) We could start by automatically creating the initial version from the board config and `README.md` files or just create a generic placeholder where no info is available for now.

What do you thinks about this idea? Is it worth the effort? If so we should continue the discussion under a dedicated issue.

> Are there still left known issue for t480? As of now, I see the t480s having headset jack issue still. Maybe we should document known issues instead?

Those are minor as listed here https://doc.coreboot.org/mainboard/lenovo/t480.html#known-issues
* dgpu has some issues (but my board does not have one)
* some fn+f keys don't work properly. not sure which ones exactly. the ones I use frequently do work, though.

> Also, @gaspar-ilom if you could link this report to upstream and tell that it worked for you, that would be great

Done: https://review.coreboot.org/c/coreboot/+/90023/comments/daf72534_69d477a4

> @gaspar-ilom https://github.com/linuxboot/heads/blob/master/BOARDS_AND_TESTERS.md:
> `Please see boards/BOARD_NAME/BOARD_NAME.config for HCL details.`
>
> Maybe that is a bit too hidden, as https://github.com/linuxboot/heads/issues/692 OP which should be part of documentation on boards, their blob requirements evolution in time etc?

I think that both are too hidden. As a non-technical user I would only look at the documentation in the wiki and I would hate it if there were multiple files at different locations in source code to gather all the important information from. So I guess we have room for improvement here.

Do you think this issue cannot be merged yet due to the documentation? I do not think that it makes things worse than they are currently in master?!

tlaurion (approved):
LGTM! Thanks @gaspar-ilom

PR #2048, opened 2026-01-26
T480 linux config: add retpoline and USB OHCI HCD requirements for optional USB keyboard support (config menu enabled from ps/2 keyboard required)

notgivenby:
This PR adds an enhancement against spectre. Moreover, it adds linux configs that are necessary to use HID usb keyboard. If the CONFIG_USB_KEYBOARD_REQUIRED=y without the lines the linux modules are not properly build which results in kernel panic. The issue was discovered during the port of m900_tiny board. https://github.com/notgivenby/heads/commit/adafc97f09fc2fb98bbd54d77c580bbe3093f3f9


tlaurion (approved):
LGTM merging

PR #2049, opened 2026-01-27
Merge v0.2.1 release related commit into master (versioning of Heads changes from 0.2.0 to 0.2.1+number of commits since 0.2.1 in versioning info, filename etc)

tlaurion:
@notgivenby found that for unknown reasons of me, https://github.com/linuxboot/heads/commit/7f2eb15b59c07f450f5e22f5699fb59a469b4425 which is commit of release 0.2.1 (see https://github.com/linuxboot/heads/compare/v0.2.0...v0.2.1) was never part of master.

For reference, this old commit goal was for recovery shell to extend pcrs, which is since then made in etc/functions:recovery.
Nothing new in that commit vs master.

This empty commit merges the missing commit (in git terms to track 0.2.1 into master) which is an empty commit (no file changes)

In my local branch I now  see:
```
user@heads-notgivenby:~/heads$ git describe --abbrev=7 --tags --dirty
v0.2.1-2886-g9a402a0
```

Where otherwise, origin/master shows:
```
user@heads-notgivenby:~/heads$ git describe --abbrev=7 --tags --dirty
v0.2.0-2885-ga44a0ab
```

Am I doing something stupid here?
Unfortunately, DCO fails because https://github.com/linuxboot/heads/pull/2049/commits/7f2eb15b59c07f450f5e22f5699fb59a469b4425 is not signoff, while signed (GPG signed) without signoff which DCO is complaining about here.

I can relax the check here but that is that.


macpijan:
I do not think it matters from our perspective. When we made a release, we used to make a tag on our own, in a downstream repo. So we do not rely on this old heads tag.

PR #2080, opened 2026-04-08
docker scripts: Use portable shebang

JohnAZoidberg:
Not all distros have /bin/bash, use env to get bash from the PATH

### DIFF
--- a/docker_latest.sh
+++ b/docker_latest.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # Source shared docker helper functions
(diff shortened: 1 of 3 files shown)


tlaurion:
> Not all distros have /bin/bash, use env to get bash from the PATH

Thanks!

tlaurion (approved):

JohnAZoidberg:
Ci failure looks unrelated

tlaurion:
Yeah, don't know what's up with CircleCI randomly throwing return 1 when there is no build error. 
Let's consider that CircleCI hiccup, builds were all successful yesterday.

PR #2081, opened 2026-04-08
docker: Add HEADS_FORCE_DOCKER_REBUILD and improve reproducibility checks

tlaurion:
## What

- Add `HEADS_FORCE_DOCKER_REBUILD=1` to force rebuild from `flake.nix`/`flake.lock` regardless of git status; also deletes the cached nix store result/link before rebuilding
- Add `--print-build-logs` to `nix build` for build visibility
- Use `docker load -i result` instead of `docker load < result`
- Use script-relative `--out-link` path so `nix build` and `docker load` are deterministic regardless of `$PWD`
- Improve reproducibility check: compare config digests (authoritative, bit-for-bit) rather than manifest digests; explain why they differ
- Show which fetch method was used (`registry+jq`, `registry+sed`, `skopeo+jq`, or `pulled`)
- Add `get_remote_manifest_digest()` to show the Docker Hub URL for manual cross-check
- Fall back gracefully to `docker pull` when registry API is unavailable; tip to install `jq`/`curl` to avoid the pull
- Update `doc/docker.md` explaining config vs manifest digests, with accurate example output
- Normalize indentation to tabs across all docker scripts

## Correctness fixes (addressed during Copilot review)

- `result_target` declared `local` in the `HEADS_FORCE_DOCKER_REBUILD` path
- Dangling `result` symlink always removed (not only when target exists)
- `get_local_manifest_digest()` removed (unused, would have hashed with wrong newline semantics)
- `sha256` regex restricted to exactly 64 hex chars (`{64}`)
- `(via ${remote_method})` reflects actual fetch path instead of hardcoded "Verified via registry API"
- Docker Hub URL uses `sha256-<hex>` path format (not `sha256:<hex>`)
- `curl` availability checked before registry API path; non-Docker-Hub registries short-circuit early
- `curl -D - -o /dev/null` captures only headers, not the full manifest body
- `_parse_docker_image()` normalises `docker.io`/`index.docker.io` to `registry-1.docker.io` and adds `library/` prefix for official images
- `=== End Reproducibility Check ===` only printed at actual return points, not before the pull fallback
- `last_segment` declared `local` to avoid leaking into calling scope
- `pin-and-run.sh`: strip `:tag` before appending `@digest` to produce valid Docker reference (`repo@sha256:...`)

## Example: force rebuild

```
HEADS_FORCE_DOCKER_REBUILD=1 ./docker_local_dev.sh
Developer helper: ./docker_local_dev.sh (local image: linuxboot/heads:dev-env)
Rebuilds local image when flake.nix/flake.lock have uncommitted changes. Opt-out: HEADS_SKIP_DOCKER_REBUILD=1
Force rebuild: HEADS_FORCE_DOCKER_REBUILD=1
For published images use: ./docker_latest.sh; for reproducible builds use: ./docker_repro.sh

HEADS_FORCE_DOCKER_REBUILD=1: forcing Docker rebuild from flake
Verifying Nix environment...
Building Docker image from flake.nix...
[...]
Loading Docker image...
Loaded image: linuxboot/heads:dev-env
Using local dev image: linuxboot/heads:dev-env
```

## Example: reproducibility check

```
HEADS_CHECK_REPRODUCIBILITY=1 ./docker_local_dev.sh

Git repository is clean. Using existing Docker image.
Using local dev image: linuxboot/heads:dev-env

=== Reproducibility Check ===
Local image (linuxboot/heads:dev-env):   sha256:8ae7744cc8b4ff0e959aa6dfeeb40dbd40d20ac6fa1f7071dd21ec0c2d0f9f41
Remote image (tlaurion/heads-dev-env:latest): sha256:8ae7744cc8b4ff0e959aa6dfeeb40dbd40d20ac6fa1f7071dd21ec0c2d0f9f41
(via registry+jq)
✓ MATCH: Config digests identical (bit-for-bit reproducible)
Config digest: sha256:8ae7744cc8b4ff0e959aa6dfeeb40dbd40d20ac6fa1f7071dd21ec0c2d0f9f41
Note: manifest digest differs from config (normal - manifest includes metadata)
Docker Hub: https://hub.docker.com/layers/tlaurion/heads-dev-env/latest/images/sha256-5f890f3d1b6b57f9e567191695df003a2ee880f084f5dfe7a5633e3e8f937479
=== End Reproducibility Check ===
```

### DIFF
--- a/doc/docker.md
+++ b/doc/docker.md
@@ -43,6 +43,22 @@ it automatically when present.
 If you plan to manage disk images or use `qemu-img` snapshots on the host (outside
 containers), install the `qemu-utils` package locally (which provides `qemu-img`).
 
+### Alternative: Using Nix directly without Docker
+
+You can also use Nix to enter a development shell or build Heads directly without Docker:
+
+```bash
+# Enter a development shell with all dependencies
+nix develop
+
+# Or run a single command in the environment
+nix develop --command make BOARD=x230-hotp-maximized
+```
+
+Note: `nix develop` provides QEMU, `swtpm`, and other required dependencies in the shell
+environment, so separate host installs are not needed for this workflow. The Docker
+workflow is still recommended for its canonical isolation and reproducibility benefits.
+
 ---
 
 ## Docker wrapper scripts
@@ -146,6 +162,8 @@ and mount your `${HOME}/.Xauthority` (if present); if the file is missing the he
 warn and will not attempt automatic cookie creation (GUI may fail).
 
 ### `./docker_local_dev.sh`
+
+**`HEADS_FORCE_DOCKER_REBUILD=1`** — force rebuild from flake.nix/flake.lock regardless of git status. Also attempts to delete the cached nix store result/link before rebuilding. Takes precedence over `HEADS_SKIP_DOCKER_REBUILD=1`.
 
 **`HEADS_SKIP_DOCKER_REBUILD=1`** — skip automatically rebuilding the local image when
 `flake.nix`/`flake.lock` are dirty.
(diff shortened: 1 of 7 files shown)


tlaurion:
The 100 layers default produces good enough result but should be improved. todo added

PR #2114, opened 2026-05-13
initrd/etc/gpg_functions.sh: check flash.sh exit status before reboot

### LINKED ISSUE #2113: Issue with adding gpg key and generating totp on maximized tpm1.2 board without factory reset
Moderation @tlaurion: this impacts only tpm 1.2, master (https://github.com/linuxboot/heads/commit/dc0cd70cd46e5b4c33e4f115e7942cceb69464c7)  tpm2 workflows not impacted from testing.

On TPM2, flashing without preserving settings will result in reboot and next prompt being a choice to factory-reset/inject public key and things go well there. Config -> wipe settings works as expected(+tpm reset), resulting in oem-factory-reset on next boot that can be cancelled, leading to gpg menu (add public key there). Reboot, and then reseal secrets, all UX guided as expected. 
----

**Describe the bug**
In order to use the maximized board with tpm1.2 and existing gpg key on nk3 mini heads fails to properly set everything up.

**To Reproduce**
1. From heads menu choose „Flash the firmware with a new ROM, erase settings“
2. Options > change configuration settings > clear GPG keys and reset all user settings > reset the tpm and reboot as suggested 
3. Generate new totp this will give multiple failures with integrity report without beeing able to sign files on / boot because the tpm passphrase does not match

**Expected behavior**
Beeing able to generate totp and sign boot files after adding public key into bios.

**Screenshots**
Please see the screenshot below

Image

**Additional context**
In the discussion with @tlaurion after reflashing the bios and directly reseting the tpm reboot step is not enforced.

tlaurion:
Fixes #2113

### initrd/etc/gpg_functions.sh

Check flash.sh exit status in gpg_flash_rom(). On success show a
confirmation dialog and reboot automatically; on failure show an
error dialog instead of rebooting unconditionally.

### initrd/bin/gui-init.sh

- Source /etc/gpg_functions.sh for the integrity report helper
- Add report_integrity_measurements() call before TPM reset paths
  so the user sees current integrity state before deciding to reset
- Restore /bin/reboot.sh at end of reset_tpm() so the reset flow
  completes with a clean reboot

### DIFF
--- a/initrd/bin/gui-init.sh
+++ b/initrd/bin/gui-init.sh
@@ -428,12 +428,16 @@ EOF
 				return 1
 				;;
 			# "Reset the TPM" from the TOTP failure whiptail menu.
-			# The gate runs first to verify /boot integrity.  If the gate
-			# fails *because* TPM reset is required (e.g. stale counters),
-			# the || tpm_reset_required bypass lets reset_tpm() proceed —
-			# it clears counters and creates a fresh one.
+			# Show the integrity report so the user can see the state,
+			# but do not force the investigation / signing path —
+			# that would attempt TPM counter operations requiring the
+			# current owner password, which is unknown (that is why
+			# we are resetting).  reset_tpm() handles everything:
+			# new password, counter create, /boot signing, TOTP/HOTP
+			# generation, DUK reseal, and reboot.
 			p)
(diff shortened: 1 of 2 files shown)


tlaurion:
will remove hotp fix and split it in https://github.com/linuxboot/heads/issues/2115

PR #2068, opened 2026-03-06
UX improvements: TPM reseal (HOTP/TOTP/DUK) adds integrity report; detects disk/tpm swap and guide user into action, add terminal colors and guidance! Reduced quiet noise

tlaurion:
## Summary

Improve TPM/TOTP/HOTP recovery and reseal behavior by adding integrity-first
gating, clearer failure handling, and stronger rollback preflight checks.

- Add integrity report + investigation flows in GUI, with explicit actions before reseal/reset paths
- Introduce TPM reset-required markers and rollback preflight validation to fail early on inconsistent TPM state
- Make unseal/seal paths safer and more recoverable (nonfatal unseal mode, clearer reset/reseal guidance, better TPM1/TPM2 handling)
- Improve kexec signing reliability with explicit signing key selection and actionable GPG error diagnostics
- Avoid hiding interactive password/PIN prompts by removing inappropriate debug wrappers around sensitive interactive commands
- Add `run_lvm` wrapper and switch runtime scripts to reduce harmless LVM noise
- Refresh TPM2 primary-handle hash in update/signing flows to keep trust metadata in sync
- Add new qemu fbwhiptail prod_quiet board configs for TPM1 and TPM2; fix board-name values for existing qemu hotp prod_quiet variants
- Document QEMU canokey state reuse and TPM2 pcap capture debugging
- Ignore exported public key artifacts (`*.asc`) in `.gitignore`
- Add coloring to console (STATUS, STATUS_OK, INPUT -- see `doc/logging.md`); all wrappers colorized with ANSI escape characters
- And many other fixes found along the way.

### Dongle branding: always from USB VID:PID at runtime

- Removed all reads/writes of `/boot/kexec_hotp_key` -- the stored branding file caused stale labels (Librem Key shown as Nitrokey and vice versa after dongle swaps)
- `detect_hotpkey_branding()` is now called after `enable_usb` in every code path (`gui-init`, `seal-hotpkey`, `oem-factory-reset`, `gui_functions`)
- VID:PID table: `20a0:42b2`->NK3, `20a0:4108`->NK Pro, `20a0:4109`->NK Storage, `316d:4c4b`->Librem Key
- Dongle firmware version color-coded: green (current), yellow (upgrade available via nitropy), red (external programmer required); shown once per session

### HEADS_TTY output routing

[...]

PR #2117, opened 2026-05-14
initrd: fix TPM1 counter auth regression (#2068)

tlaurion:
## Summary

Fixes a TPM1 counter auth regression (PR #2068) where `increment_tpm_counter` was changed from hardcoded `-pwdc ''` (empty counter auth) to `-pwdc "${tpm_passphrase:-}"` (owner passphrase), while counters continued to be created with `-pwdc ''`. This caused every increment to compute SHA1(owner_pass) against a counter created with SHA1(""), producing persistent `TPM_AUTHFAIL`.

Per TCG TPM Main Spec Part 3, `TPM_CreateCounter` uses owner auth (`-pwdo`) but `TPM_IncrementCounter` uses the counter's own authData, **not** the owner password. The correct design for Heads' rollback counter is empty auth.

The repeated auth failures (3 per boot) triggered TPM 1.2 dictionary-attack lockout (`TPM_DEFEND_LOCK_RUNNING`), which persisted through `forceclear` on some implementations.

### `initrd/bin/tpmr.sh`
- `tpm1_counter_increment()`: detect explicit `-pwdc ''` and call `tpm counter_increment` directly, bypassing `_tpm_auth_retry`. Non-empty or absent `-pwdc` falls through to owner-auth retry path for migration of counters created by pre-fix code.

### `initrd/etc/functions.sh`
- `check_tpm_counter()`: create counters with `-pwdc ''` instead of owner passphrase
- `increment_tpm_counter()`: increment with `-pwdc ''` instead of owner passphrase; counter_create fallback uses `-pwdc ''`

### `initrd/bin/oem-factory-reset.sh`
- `-pwdc "${TPM_PASS:-}"` → `-pwdc ''` for consistency

### `doc/tpm.md`
- Documented TPM1 counter auth requirements under new "TPM1 counter auth" subsection

### DIFF
--- a/doc/tpm.md
+++ b/doc/tpm.md
@@ -10,8 +10,35 @@ See also: [architecture.md](architecture.md), [boot-process.md](boot-process.md)
 ## tpmr — unified TPM abstraction
 
 `initrd/bin/tpmr.sh` is a shell script wrapper that presents a single interface
-over both TPM 1.2 (`tpm` / `trousers`) and TPM 2.0 (`tpm2-tools`). All Heads
-scripts call `tpmr.sh` rather than invoking `tpm` or `tpm2` directly.
+over both TPM 1.2 and TPM 2.0. All Heads scripts call `tpmr.sh` rather than
+invoking TPM tools directly.
+
+### Boot chain and TPM tool selection
+
+```text
+initrd/init  (PID 1)
+  └─ CONFIG_BOOTSCRIPT → /bin/gui-init.sh          [board config]
+       ├─ source /etc/functions.sh                   [shared TPM helpers]
+       ├─ source /etc/gui_functions.sh               [whiptail wrappers]
+       └─ calls initrd/bin/tpmr.sh                   [TPM abstraction]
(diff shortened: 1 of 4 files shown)


notgivenby:
The fix did not work…will copy logs.

tlaurion:
> The fix did not work…will copy logs.

Found the bug and where the regression comes from. Damn that one was not easy. Pushing fix and updating the other pr

tlaurion:
The rabbit hole of tpm1 da lockout went too far. This pr will just fix regression and on later time i will hopefully add diagnostics and helpers to test further issues.

on x230, i have a STM (not infineon) that doesn't support proper tpm da policy getcap, so i cannot ask what is the threshold and even counter; which required me to modify tpm increment handler where da lockout is the only place in code to check if tpm is in da defend. 


I was not aware (or forgot) that older devices like x230 came with a not compliant tpm 1.2 spec revision, which complicates testing by a big margin. Let's just do the right thing from tpm reset up to increment counter for now and fix #2068 regression.

This PR now includes only the regression fix.

PR #2119, opened 2026-05-14
circleci,modules/coreboot,doc: share crossgcc toolchain across Dasharo 24.12 forks; blobs: add lib.sh and unify hash pre-checks

tlaurion:
`dasharo_nv4x` is the toolchain provider.  `dasharo_v56`, `dasharo_msi_z690`, `dasharo_msi_z790` consume its crossgcc via the second `coreboot_module()` argument.  All Dasharo boards except `novacustom-nv4x_adl` are `build` jobs (no cache) -- they get crossgcc from workspace and clone fork source fresh.

Fix pre-existing hyphen/underscore variable name typo in `modules/coreboot` that made the toolchain-module nobuild override a silent no-op.  Fix pre-existing portability issue: `printf` instead of `echo -e`.

Rename CI jobs for clarity: `[cache keys]`, `[ME GBE IFD]`, `[cross compiler]`, `[seed:coreboot-VERSION]`.  Update `doc/circleci.md` to document the cache model, job types, and Dasharo shared toolchain design.

Add `blobs/lib.sh` with `check_outputs()` unifying hash verification across all 10 blob download scripts.  All scripts now check output file hashes before and after processing, reporting CHECKING/OK/MISSING/HASH MISMATCH per file.  Fix fallthrough bugs, variable name mismatches, and missing pre-checks.  (2 commits)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,7 +23,12 @@ commands:
           command: |
             echo "Sourcing /devenv.sh since docker entrypoint doesn't do it as expected"
             source /devenv.sh
+            # Remove stale logs, .rom, and .zip from previous cache/workspace so
+            # store_artifacts only picks up files from this run (git-describe differs
+            # per commit, so old artifacts persist alongside new ones otherwise).
             rm -rf build/<< parameters.arch >>/log/*
+            rm -f build/<< parameters.arch >>/<< parameters.target >>/*.rom \
+                  build/<< parameters.arch >>/<< parameters.target >>/*.zip
             # Fresh checkout timestamps can make restored module stamp files look stale.
             # Refresh them first so restored musl-cross-make and dependency trees are reused.
             if [ -d "build/<< parameters.arch >>" ]; then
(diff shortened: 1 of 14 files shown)


tlaurion:
might be wrong, was on 0.9.2, verifying

tlaurion:
Will see what circleci gives, but yeah, that's why we build 4 buildstacks in master for dasharo forks per their different FSP requirements/changes per bords family, see comment above replicated below in master:
```
# Dasharo forks cannot share toolchains with upstream coreboot releases because
# Dasharo uses custom FSP packages (3rdparty/dasharo-blobs/) with different/missing
# headers vs upstream (e.g. MemInfoHob.h, FirmwareVersionInfoHob.h).
```

Seems like it worked locally? Actually, we can share build system and FSP are in specialized repo for board builds!

tlaurion:
Went from 52 minutes from musl-cross-make cold cache to 32 minutes for all boards builds!

tlaurion:
ok enough. merging, its clear enough and already a big improvement.

PR #2124, opened 2026-05-20
(WIP) initrd: fix TPM1/TPM2 add DA lockout handling/gating and test tooling

### LINKED ISSUE #2123: TPM1/TPM2 DA policy exposure, tooling to trigger DA defend mode and proper gating on increment and unseal
PR https://github.com/linuxboot/heads/pull/2117 fixes the #2068 tpm1 counter increment regression.

But testing on x230, I realized that DA policy cannot be extracted from TPM (STM) while qemu swtpm policy is too permissive.

We need tooling from tpmr.sh (bad_auth simulation, da_state to get TPM DA policy for current tries before threhold and how many attempts and end of TPM Defend triggered) to guide the user properly when this happens.

As of now, a TPM Defend lock activated can be reset by a TPM reset on both TPM1 and TPM2. But that doesn't give us insights on what causes the DA Defend, now exposes the DA policy, not permit us to play with it.

tlaurion:
WiP to fix #2123 

  - tpm1_counter_increment: empty -pwdc '' bypasses _tpm_auth_retry and calls tpm directly per TCG spec. Non-empty passphrase or no -pwdc falls through to owner-auth retry path (migration for counters created by pre-fix code) (is standalone #2117)
  - tpm1_bad_auth: emit STATUS when lockout detected or counter ID missing, instead of silent DEBUG-only return. STATUS deduplicated with marker file for loop safety. On lockout, queries da_state and shows remaining duration when available.
  - tpm1_da_state: convert raw hex firmware version (0x0D.0x0C) to human-readable decimal (13.12) in all output and log messages.
  - tpm1_da_state: when state=1 but timer is 0 or empty, print "duration unknown on this TPM" instead of misleading "0s remaining".
  - tpm1_da_state: declare rev_major_dec/rev_minor_dec as local variables and normalize variable names to match tpm2_da_state conventions.
  - tpm2_da_state: declare interval_hex, recovery_hex, interval, recovery as local variables (previously leaked into global scope).
  - da_state output in bad_auth: extract only the => summary line for STATUS on console and the DA: machine line for DEBUG in log. Avoids duplicate re-emission of diagnostic lines already logged internally.
  - tpm1_unseal: detect DA lockout from unseal failure output ("defend lock"), emit WARN, query da_state for remaining time, set marker file so callers show lockout-specific guidance instead of generic reseal.
  - gui-init.sh update_totp: when marker present, show DA-lockout-specific whiptail with TCG exponential backoff guidance instead of generic OEM reset message. Shows exact remaining time when da_state provides it.
  - unseal-totp.sh: check DA lockout marker and show lockout-specific fail_unseal message.
  - recovery shell: show concise DA state summary at STATUS (not INFO) level so it reaches console in Quiet mode.


tlaurion:
Superseded by #2206 (same branch `tlaurion:tpm_da_lockout_rework`, rebased on current master). This PR was closed because force-pushing its head branch invalidated the merge-base; #2206 carries the full 17-commit series.

PR #2130, opened 2026-06-22
ISO boot support with sysfb display handoff: kernel detection, kexec patches, test harness

tlaurion:
## Summary

Adds USB ISO boot with kernel-level display driver detection and a
complete framebuffer handoff chain (coreboot → kexec → target kernel).

---

### kexec-tools patches

- **0001-build-fixes**: cross-compilation fixes for musl-gcc toolchain
- **0002-EBDA-segment-0**: EBDA segment 0 preservation for Xen multiboot
- **0003-EFI-VGA-set-VLFB-type-for-sysfb-display**: sets orig_video_isVGA=0x23
  (VLFB) for EFI VGA framebuffers so the target kernel applies the 16-bit
  lfb_size shift needed by sysfb display drivers.

### ISO boot

- **7-step boot flow**: signature verify → mount → loopback.cfg parse →
  probing/fast-path gate → initramfs compat → params → boot menu
- **_strip_grub_vars()**: removes unresolved GRUB variables from loopback.cfg;
  universal fallback ADD params provide absolute ISO paths for all 7 frameworks
- **_check_kernel_probe_driver()**: decompresses target kernel bzImage and
  searches for built-in display driver symbols in priority order:
  vesadrm_probe → vesafb_probe → simpledrm_probe + sysfb_parse_mode
- **boot_marker()**: three-state [OK]/[~]/[X] per-entry compatibility markers
  combining filesystem and display driver results
- **verify_global_hashes()** moved to gui_functions.sh for cross-script access

### Display handoff

- EFI VGA framebuffers are normalized to VLFB (0x23) so the target kernel's
  sysfb_create_simplefb() applies the required 16-bit lfb_size shift
- Actual stride is read from /dev/fb0 fix.line_length rather than assumed
  from width\xd7bpp, correct for all pixel formats and alignment padding
- color-format masks are left as /dev/fb0 reports them; sysfb_parse_mode()
  matches via SIMPLEFB_FORMATS comparison

### Documentation

- **kexec_handoff.md**: 3 display paths (simpledrm, vesadrm, vesafb), patch
  details, stride handling, debug log diagnosis, per-distro results
- **iso_boot.md**: 7 initramfs frameworks, universal ADD param reference,
  GRUB variable stripping, detection symbol priority
- **architecture.md**: hardware root of trust (S-CRTM), Purism boot modes
- **boot-process.md**: updated ISO boot flow with marker descriptions

### Test harness

[...]

PR #2153, opened 2026-07-08
initrd: remove TPM2 primary handle hash check and creation

tlaurion:
The kexec_primhdl_hash.txt feature stored a sha256sum of the TPM2 primary handle and verified it on every boot.  It was originally added as a canary to detect TPM state changes, then accidentally dropped during the PR #2130 refactoring and re-added as a regression fix in PR #2145.

However, this check is redundant with existing protections:

- The TPM rollback counter (kexec_rollback.txt) already detects TPM state changes — a different or reset TPM either lacks the counter entirely or holds a different value, causing a fatal mismatch
- The TPM Disk Unlock Key unseal fails if the primary handle changed, since the key is sealed to the original handle — booting into an OS where the DUK cannot unseal is futile
- TOTP/HOTP secrets sealed to the TPM similarly fail to unseal on a changed handle

In every realistic scenario the primary handle check catches, either the rollback counter or the TPM unseal (DUK, TOTP, or HOTP) catches it as well, with a clearer error message and the same outcome.

The check also created UX problems: it was fatal (DIE) on mismatch even after an intentional TPM reset, requiring a workaround in the TPM reset flow (which explicitly removed kexec_primhdl_hash.txt). And it added another file to manage in /boot with no non-redundant security value.

Removed from:
- kexec-select-boot.sh: the boot-time verification check
- kexec-save-default.sh: PRIMHASH_FILE variable and creation block
- gui-init.sh: the TPM reset workaround (TODO comment + rm -f)
- functions.sh: refresh_tpm2_primary_handle_hash() and its caller

Part of discussions for going forward https://github.com/linuxboot/heads/discussions/1655
> Either way, I tend to think the primary handle digest on /boot is ineffective, so we should remove it to eliminate the headaches it causes currently.

@JonathonHall-Purism : oppositions?

### DIFF
--- a/initrd/bin/gui-init.sh
+++ b/initrd/bin/gui-init.sh
@@ -798,13 +798,7 @@ reset_tpm() {
 			# now that the TPM is reset, remove invalid TPM counter files
 			mount_boot
 			mount -o rw,remount /boot
-			#TODO: this is really problematic, we should really remove the primary handle hash
-
-			STATUS "Removing rollback and primary handle hashes under /boot"
-
-			DEBUG "Removing /boot/kexec_rollback.txt and /boot/kexec_primhdl_hash.txt"
 			rm -f /boot/kexec_rollback.txt
-			rm -f /boot/kexec_primhdl_hash.txt
 
 			# create Heads TPM counter before any others
 			check_tpm_counter /boot/kexec_rollback.txt "" "$tpm_owner_passphrase" ||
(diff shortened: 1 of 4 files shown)


tlaurion:
@JonathonHall-Purism is reluctance to change based on alternative approach you would prefer implemented here as per https://github.com/linuxboot/heads/discussions/1655:
> An alternative I can think of is to extend the TOTP/HOTP secret with the primary handle digest, so a change in the primary handle (an MITM attack) would cause TOTP/HOTP verification to fail. (E.g. the TOTP/HOTP secret would become sha256(unsealed_secret + prim_handle_digest).) This does not require storing any trust anchor for the TPM on the device itself, essentially it's stored on the TOTP/HOTP authenticator. But I'm also not sure whether this gains any security guarantee against a device-opening attacker.

PR #2155, opened 2026-07-08
initrd/bin/kexec-save-default.sh: skip symlinks when finding crypttabfiles

### LINKED ISSUE #2154: Installing from nixos-graphical-26.05.590.ec942ba042da-x86_64-linux.iso can't have TPM DUK setuped
Heads tries to find crypttab in decompressed initramfs, hits a symlink pointing to rootfs and fails there.

Installing OSes will need revamping as well under https://osresearch.net/InstallingOS/#compatibility will need more details for legacybios, which will need more efforts, but from top of my head; I 
- Flashed the iso on dedicated thumb drive since latest heads correctly detects it won't support fileboot
- From installer, choosing custom partitioning option:
  - create dedicated 2000MiB /ext4 partition
    - set it to /boot mountpoint
    - set the flag to legacybios checkbox
    - set the grub flag checkbox
    - optionally set boot label
  - Create swap partition
    - say you want it encrypted
      - type a Disk Recovery Key passphrase to be reused by other encrypted rootfs later
    - set the swap flag checkbox
    - optionally set swap label
  - create rootfs partition
    - say you want it encrypted
      -  type a Disk Recovery Key passphrase used for the swap partition earlier
    -  set the / mountpoint
    - set the root flag checkbox
- dismiss any warning given on uefi partition needed, /boot warning whatever.

tlaurion:
- NixOS initrds contain /etc/crypttab as a symlink to a Nix store path (e.g. /nix/store/[hash]-initrd-crypttab). When unpacked into /tmp/initrd_extract, the symlink target is an absolute path that does not exist on the Heads recovery system, making cat fail.
- find without -type f matched the dangling symlink, causing cat to error out inside a pipefail pipeline, which killed the crypttab parsing loop and caused kexec-save-default.sh to fail.
- Add -type f to only match regular files, skipping symlinks.

As a result, a crypttab is generated pointing key unsealed and injected TPM DUK through passphrase in additional cpio passed alongside initramfs on kexec call, and that injected cpio successfully decrypt my 2 luks container setup described under #2154

It is to note that unencrypted /boot support option from installer was sadly completely removed in the aftermath of https://github.com/linuxboot/heads/issues/1348. I am unaware of when /boot got completely removed, which I find sad, requiring manual partitioning. At least, this PR permits TPM DUK on "brainless" partitionning scheme, where iniramfs can decrypt rootfs partitioning with swap, without needing rootfs to be already decrypted. If someone is from/within NixOS and wants to tackle the installer issues, be my guest :)

Fixes https://github.com/linuxboot/heads/issues/2154

----

### Notes on NixOS installer custom partitioning requirements
- 2048M ext4 unencrypted, specify /boot mountpoint and **grub-bios ticked**
- 4096M swap, tick encrypted, put passphrase that will be same then rootfs
- Rest of unpartition space ext4, specify / mountpoint,  tick encrypted, put passphrase that will be same then swap

### DIFF
--- a/initrd/bin/kexec-save-default.sh
+++ b/initrd/bin/kexec-save-default.sh
@@ -326,7 +326,7 @@ if [ "$save_key" = "y" ]; then
 
 	DEBUG "Extracting $current_default_initrd to find crypttab files"
 	unpack_initramfs.sh "$current_default_initrd" "$initrd_decompressed"
-	crypttab_files=$(find "$initrd_decompressed" | grep crypttab 2>/dev/null) || true
+	crypttab_files=$(find "$initrd_decompressed" -type f | grep crypttab 2>/dev/null) || true
 
 	if [ ! -z "$crypttab_files" ]; then
 		DEBUG "Found crypttab files in $current_default_initrd"


tlaurion:
- Debug trace prior of fix https://github.com/linuxboot/heads/issues/2154#issuecomment-4961347600, showing only requirement for NixOS installer for unencrypted /boot is to tick grub-bios for installation to work with encrypted rootfs+encrypted swap to work  (which needs heads-wiki updates)
  - initramfs cannot be parsed properly for crypttab and fails expecting a file but finding symlink. 
- With this PR applied, bug resolved with debug trace https://github.com/linuxboot/heads/issues/2154#issuecomment-4961427735

PR #2158, opened 2026-07-14
initrd/gpg: add reprovision smartcard from GPG key backup flow

tlaurion:
## Changes

This PR adds a "reprovision smartcard from GPG key backup" flow and
factors common GPG/smartcard operations into shared helper functions
that are reused by both the OEM factory-reset path and the new
reprovision path.

### Commit 1: initrd/etc/functions.sh: fix print_tree hang

- `print_tree()` piped `find -print0` directly into `sort -z`; BusyBox
  sort buffers all NUL-delimited input, so the pipeline hung on
  non-trivial trees. Output is now written to a temp file, then sorted
  from that file; the `mktemp`-failure fallback stays sorted too so it
  byte-matches the sorted regeneration at verification time. Adds a
  DEBUG CWD line for future diagnosability.

### Commit 2: initrd changes

**Shared functions (gpg_functions.sh):**
- `gpg_card_factory_reset()` -- factory-reset + forcesig + key-attr setting (RSA and ECC/P-256)
- `gpg_keytocard_subkeys()` -- move subkeys to card slots 1-3 (sign, encrypt, auth)
- `gpg_set_card_identity()` -- cardholder name and login fields
- `gpg_reset_nk3_secret_app()` -- Nitrokey 3 Secrets app reset
- `gpg_card_change_pin()` -- PIN change via `gpg --card-edit`
- `reprovision_smartcard_from_backup()` -- flow: mount LUKS backup, detect key type, factory-reset card, restore subkeys, set identity, set custom PINs, reset TPM + sign /boot atomically, export public key, offer ROM flash

**Menu integration:**
- `gpg-gui.sh`: **k** option (always visible) in GPG Management Menu, guarded with `|| true` so benign returns (user decline, wrong passphrase, signing failure) stay in the menu under `set -e`
- `gui-init.sh`: **K** option in clean boot wizard, integrity report, TPM State Inconsistent dialog; the K arm re-runs the rollback preflight after reprovision so a successful run exits the gate instead of looping the error dialog
- `gui_functions.sh`: `report_integrity_measurements` K arms guarded against `set -e` exit

**OEM consolidation:**
- `oem-factory-reset.sh`: `gpg_key_factory_reset` and `gpg_key_change_pin` replaced with thin wrappers calling shared functions

**Security hardening:**
- `mount-usb.sh`: `--pass-file` stores passphrase in 0600 temp file, shredded on exit
- `gpg_functions.sh`: PINs via `--passphrase-file`, temp files shredded; admin PIN bounds 8-25 (minimum 8 per gnupg PW3 requirement, capped at 25 for Nitrokey 3 Secrets-app compatibility), user PIN 6-25
- `gpg_add_key_reflash` returns 1 instead of `exit 1`
- TPM reset first + atomic signing: `tpmr.sh reset` re-owns the TPM, stale `kexec_rollback.txt` / `kexec_primhdl_hash.txt` are dropped, then /boot is hashed and signed atomically via `update_checksums()` -> `kexec-sign-config.sh -p /boot -u [-r]`, which creates a fresh rollback counter. Manifests move into /boot only after signing AND `check_config` verification succeed -- any failure leaves the previous valid signatures in place and returns 1 before the ROM-flash offer.
- Master secret keys deleted in `_luks_cleanup` on all exit paths

### Commit 3: initrd/etc/functions.sh: name offending /boot entries in assert_signable

- `assert_signable`'s recovery message now lists every /boot entry containing bytes outside printable ASCII or a backslash, escaped via `escape_zero()` for safe display, so field reports are actionable.

### Commit 4: documentation

- `doc/configuring-keys.md`: reprovision entry points, TPM-reset guidance
- `doc/qemu.md`: hardlink preservation workflow using `~/Qemu_img/`, reprovision test steps, USB flash drive workflow; daily development cycle uses `~/Qemu_img/` consistently
- `doc/recovery-shell.md`: authentication scope, reprovision and configuration sections
- `doc/modules.md`: fix `real.gitclean` target descriptions to match Makefile; re-add `remove_canary_files` helper target row
- `doc/gpg.md`, `doc/architecture.md`: minor cross-reference updates


tlaurion:
#2186 moved initrd/tests (which was under heads.cpio->initrd.cpio.xz->payload) to /tests, so free space got obtained

PR #2133, opened 2026-06-24
Add Google kano (MrChromebox 26.03 coreboot fork) support - First full Chromebook port!

cwiggs:
This pull request adds support for the Acer Chromebook Spin 714 ("kano") board based on the MrChromebox coreboot fork (26.06), including CI integration, configuration files, and documentation updates. It also introduces a patch for musl compatibility in the MrChromebox fork, adds a new Linux kernel version (6.6.30), and documents TPM2 CR50 command support. The most important changes are grouped below:

**Kano (Acer Chromebook Spin 714) board support:**
* Adds `boards/kano/kano.config` for Heads and Linux build configuration for the kano board.
* Adds `config/coreboot-kano.config_defconfig` as a defconfig backup for kano, documenting non-default coreboot options.
* Integrates kano into CircleCI: adds `coreboot-mrchromebox` as a seed, defines the kano build job, and documents its workflow. [[1]](diffhunk://#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47L265-R265) [[2]](diffhunk://#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R530-R539) [[3]](diffhunk://#diff-0524be7e9894b22d0bd8ba295a7b7cbcc18941dc551ec381ebc87f95dc335a0dR197)
* Documents kano's handling of the RTC century byte and why `USE_PC_CMOS_ALTCENTURY` must be unset for this board. [[1]](diffhunk://#diff-08021b9a66f4086dd71612980fd0b9655ba38dca9f4493525112e0cf1f05a786L68-R80) [[2]](diffhunk://#diff-08021b9a66f4086dd71612980fd0b9655ba38dca9f4493525112e0cf1f05a786L86-R94) [[3]](diffhunk://#diff-08021b9a66f4086dd71612980fd0b9655ba38dca9f4493525112e0cf1f05a786R104-R106)

**MrChromebox coreboot fork integration:**
* Adds the MrChromebox coreboot fork (26.03) as a supported module in `modules/coreboot`, with commit hash pinning and submodule handling notes. [[1]](diffhunk://#diff-18936189b28399cf48703d0c1ec1df33e57c559de2a12f4438be00e6813bdb68R186-R190) [[2]](diffhunk://#diff-18936189b28399cf48703d0c1ec1df33e57c559de2a12f4438be00e6813bdb68R207-R218)
[...]

Issue #2087, opened 2026-04-15
Make sure CONFIG_USE_PC_CMOS_ALTCENTURY is not defined in all boards


tlaurion:
wanted behavior is that when cmos battery is disconnected because externally flashing, next boot sets the clock to time of build which is 1969 < 2023 and heads will guide user to set clock once and cmos batter yshould keep clock time drift as minimal.

proper time is needed for displaying propr TOTPM in UTC timezone with really low discrepency (<5 seconds causes issues while master shows TOTP with time at each seconds when booting default boot entry prior of prompting for DUK passphrase)

PR #2162, opened 2026-07-21
Update key-init.sh year sanity check for 2026

PeonyThePeon:


tlaurion:
@PeonyThePeon, first,  thank you for your contribution.

The goal of the check for 2024 (year check was implemented) was to check if CMOS battery was disconnected, which resets RTC system clock to take time of build in combination to CONFIG_USE_PC_CMOS_ALTCENTURY bit (https://github.com/linuxboot/heads/issues/2087), so key-init.sh checks system clock < date and guide user into configuring date+time in UTC/GMT as guided.

Changing the date in code to 2026 has no impact on this; any date > 1970 (timeless build) will trip the condition and guide user with https://github.com/linuxboot/heads/blob/master/initrd/bin/change-time.sh

depending on platform, with CONFIG_USE_PC_CMOS_ALTCENTURY, date can be either 1900 or 2000 (the century bit).

Again, emphasizing here on the role of the check; time drift is common in all systems not syncing clock through ntp (which most OSes do silently through different mechanisms). This is also why Heads shows date+time averywhere (?) with timezone (UTC/GMT) so user can check the time; since a time drift of only a couple of seconds will not generate the same TOTP code on phone vs Heads. As doc says (?) if TOTP is wrong, do not type any secret, and then GUI proposes to change time too under Settings and TOTP/HOTP mismatch, also guiding user to actually launch `change-time.sh`, which guide the user into checking an external, and proper synchronized clock source, to set date+time in UTC/GMT.


TLDR: Maybe the code comments  could be clearer here, but changing 2024 to 2026 is not a needed fix.
@PeonyThePeon Can you close this PR and open another to clarify what was unclear in code comments if you find this is still needed?

tlaurion:
Recent validation https://github.com/linuxboot/heads/pull/2133#issuecomment-4926469576

tlaurion:
Closing. If you think it's an error @PeonyThePeon please comment.

PR #2174, opened 2026-08-02
Fix build reproducibility regression from PR #2136 (musl-cross-make binutils bump)

tlaurion:
Fixes the build reproducibility regression from PR #2136
(musl-cross-make bump to 227df8b), which upgraded binutils from
2.33.1 to 2.44.

Closes: #2168. Also fixes: #2085, #2086.

## Root causes

1. **busybox — binutils regression**. ld.bfd 2.44 --gc-sections uses
   ASLR-influenced hash tables (no upstream fix through 2.47). Fixed:
   SOURCE_DATE_EPOCH=0 + trylink patch disables --gc-sections.

2. **gpg-agent — latent TPM-probe race exposed by cache invalidation**.
   gpg2's configure probes for libtss2-esys via AC_SEARCH_LIBS; the
   result depends on parallel build order. The old CI cache consistently
   placed gpg2 first; the binutils bump invalidated caches and reshuffled
   the build schedule. Fixed: --disable-tpm2d eliminates the entire
   probe chain.

### modules/musl-cross-make
- BUILD = x86_64-pc-linux-gnu pins config.guess
- -Wa,--no-pad-sections, --with-debug-prefix-map, --enable-compressed-debug-sections=no
- SOURCE_DATE_EPOCH from pinned commit epoch
- SOURCES = $(packages) skips internal downloads
- .sources sentinel pre-seeds tarballs before configure runs

### modules/busybox + patches/
- SOURCE_DATE_EPOCH=0
- trylink-reproducible patch disables --gc-sections when SOURCE_DATE_EPOCH set
- install.sh bypass avoids make install FORCE re-link

### modules/gpg2
- --disable-tpm2d removes BUILD_WITH_TPM2D/HAVE_INTEL_TSS/HAVE_LIBTSS defines

### modules/linux
- FORCE on modules.cpio ensures hashes.txt complete on every rebuild

### Tarball downloads (fixes #2086)
- fetch_source_archive.sh: --timeout=30 --tries=3 -4, audit logging
- fetch_coreboot_crossgcc_archive.sh: ftpmirror.gnu.org -> mirrors.kernel.org
- fetch_musl_cross_make_archive.sh: pre-seeds component tarballs into packages/

### modules/qrencode (fixes #2085)
- Wayback Machine URL (fukuchi.org offline)

### Documentation
- doc/reproducible-builds.md: all mechanisms documented
- doc/index.md reorganized; cross-refs from architecture/build-freshness/
  docker/modules.md

PR #2179, opened 2026-08-05
build: complete warm-build hashes.txt coverage (bzImage, tools.cpio, initrd.cpio.xz FORCE; busybox symlinks fix; doc cleanup)

tlaurion:
Supersedes #2178. Completes the warm-build `hashes.txt` completeness work started in #2174. Fixes emergency bug introduced by #2174.

### Makefile clean targets
- Rewrite `real.gitclean` / `gitclean_keep_packages` / `gitclean_keep_packages_and_build` @echo messages to name the `git clean -fxd` command and describe exact preserve rules
- Tighten `real.remove_canary_files-extract_patch_rebuild_what_changed` @echo messages: make clear this target deletes `.canary` stamps and the re-extract/rebuild cascade happens on the next `make` invocation

### doc/reproducible-builds.md verification section
- Split verification into "Same commit: everything should match" vs "Different commits: ROMs always differ" (GIT_HASH in /etc/config causes expected cascade)
- Remove stale `real.gitclean_keep_packages` prerequisite -- after #2174 FORCE fixes, warm builds produce complete `hashes.txt`
- Remove incorrect warm-build partial-hashes.txt claim
- Add CI `hashes.txt` download reference with CircleCI artifact URL example
- Add clean working tree prerequisite (dirty tree sets `GIT_STATUS=dirty`, changing ROM output)
- Restore `bzImage` in same-commit step-down path
- Upgrade comprehensive diff from hash-only to hash+path comparison

### bzImage FORCE + UNCHANGED
- Add `FORCE` to bzImage copy/hash rule prerequisite -- matching the `modules.cpio` fix in #2174 -- so warm builds always re-run the recipe that records the kernel hash
- Replace direct `do-copy`-to-target with cmp-based UNCHANGED pattern: copy to `.tmp`, compare, `mv` only when changed, print `UNCHANGED` when identical, always append hash/size to `hashes.txt`
- When unchanged, preserve mtime to prevent cascading `.bundled` rebuild

### Emergency fix: busybox symlinks broken by #2174
- #2174 replaced `make install` with `applets/install.sh --symlinks` to avoid re-linking without `SOURCE_DATE_EPOCH`. However, `install.sh` reads symlink names from `busybox.links`, which was not regenerated and is missing after cache restore or `make clean`. Result: only the bare `busybox` binary was installed; all 137 applet symlinks (`[`, `ash`, `cat`, `cp`, `ls`, ...) were absent from `tools.cpio`, breaking the initrd.
- Fix: regenerate `busybox.links` via `applets/busybox.mkll` before calling `install.sh`, using `HOSTCC=gcc` (needed outside busybox's make context) and `include/autoconf.h` (busybox 1.36.1 uses this, not `config.h`), with a `test -s` guard so empty generation fails the build loudly.

## Verification (commit `c4c59764b64`)

Built locally with `./docker_repro.sh make BOARD=EOL_t480-hotp-maximized` (warm build, `clean` GIT_STATUS).

137 busybox symlinks now present in `tools.cpio`:
```
$ cpio -itv < tools.cpio | grep -c -- "-> busybox"
137
$ cpio -itv < tools.cpio | grep -- "-> busybox" | head -5
lrwx------   0 root    0 root    7 Dec 31  1969 bin/[ -> busybox
lrwx------   0 root    0 root    7 Dec 31  1969 bin/[[ -> busybox
lrwx------   0 root    0 root    7 Dec 31  1969 bin/ash -> busybox
lrwx------   0 root    0 root    7 Dec 31  1969 bin/cat -> busybox
lrwx------   0 root    0 root    7 Dec 31  1969 bin/cp -> busybox
```

Local ROM hash: [pending]
CI cross-check: [pending CircleCI pipeline for `c4c59764b64`]

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -863,7 +863,8 @@ endif
 $(build)/$(initrd_dir)/tools.cpio: \
 	$(initrd_bins) \
 	$(initrd_libs) \
-	$(initrd_tools_dir)/etc/config
+	$(initrd_tools_dir)/etc/config \
+	FORCE
 	$(call do-cpio,$@,$(initrd_tools_dir))
 	@$(RM) -rf "$(initrd_tools_dir)"
 
@@ -921,10 +922,10 @@ $(build)/$(initrd_dir)/heads.cpio: $(HEADS_INITRD_FILES) FORCE
 
 # --- FINAL INITRD PACKAGING ---
 
-$(build)/$(initrd_dir)/initrd.cpio.xz: $(initrd-y)
+$(build)/$(initrd_dir)/initrd.cpio.xz: $(initrd-y) FORCE
 	$(call do,CPIO-XZ  ,$@,\
 	$(pwd)/bin/cpio-clean.pl \
-		$^ \
+		$(filter-out FORCE,$^) \
 	| xz \
 		--check=crc32 \
 		--lzma2=dict=1MiB \
--- a/Makefile
+++ b/Makefile
@@ -1103,7 +1103,7 @@ define overwrite_canary_if_coreboot_git
 endef
 
 real.clean:
-	@echo "Cleaning build artifacts and install directories, leaving crossgcc intact."
+	@echo "Removing module build output under build/ and clearing install/; downloaded sources (packages/) and crossgcc are left intact."
 	for dir in \
 		$(module_dirs) \
 		$(kernel_headers) \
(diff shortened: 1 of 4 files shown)


tlaurion:
pushing as bugfix otherwise busybox missing symlinks causes brick

PR #2196, opened 2026-08-27
WIP: Add TCG Opal disk unlock support

Sean-StarLabs:
## Summary

- add a Linux-ioctl backend to scan, query, unlock, and rollback TCG Opal global locking ranges as Admin1
- run the unlock gate from init before normal, basic, or network boot scripts
- make production enablement and backend selection immutable at build time
- reject production board configurations that omit the required kernel OPAL support
- optionally use coreboot's upstream OPAL S3 APMC v1 service
- add host functional coverage and a deterministic QEMU control-flow fixture

The prompt completes before the backend starts. Heads opens and unlinks its mode-0600 password file before passing the anonymous file descriptor to the C helper, so credentials are not placed in command arguments, the environment, or a production Bash variable. Credential buffers and anonymous storage are cleared after use.

If a later disk, status check, or S3 handoff fails, Heads relocks disks already unlocked during that attempt. A rollback failure powers off instead of exposing an unlocked disk through recovery.

## Functional validation

The host fixture exercised no-device and already-unlocked paths, successful unlock, wrong-password retry and exhaustion, cancellation, prompt failure after producing output, empty and oversized input, backend and S3 pre/post-unlock failures, rollback failure, malformed scan output, multiple disks, partial multi-disk rollback, post-unlock verification, and argv/environment/path secret checks.

The exact committed QEMU ROM completed the overlaid mock scan, mocked unlock and S3 handoff, confirmed the unlocked state, and entered `detect_boot_device` in that order. The fixture password and linked-password-file failure marker were absent from serial output.

QEMU does not emulate a TCG Opal device or the coreboot SMM service. Production OPAL ioctls, NVMe Security Receive, physical-address lookup, SMI handling, and S3 resume remain hardware validation. This PR remains Draft for that reason.

## Board requirements

Production boards must use Linux 6.1 or later and enable `CONFIG_BLK_SED_OPAL=y`. Boards selecting `CONFIG_HEADS_OPAL_S3_APMC_V1=y` must also enable `CONFIG_PROC_PAGE_MONITOR=y` and run coreboot with the ABI merged through [coreboot change 91045](https://review.coreboot.org/c/coreboot/+/91045).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -777,6 +777,36 @@ bin_modules-$(CONFIG_ZSTD) += zstd
 bin_modules-$(CONFIG_E2FSPROGS) += e2fsprogs
 bin_modules-$(CONFIG_EXFATPROGS) += exfatprogs
 
+ifeq "$(CONFIG_HEADS_OPAL)" "y"
+heads_opal_bin := $(build)/heads-opal/heads-opal
+heads_opal_config_stamp := $(build)/heads-opal/.config-validated
+
+$(heads_opal_config_stamp): $(linux_kconfig) $(CONFIG) Makefile
+	@mkdir -p "$(dir $@)"
+	@if [ "$(CONFIG_HEADS_OPAL_TEST_FIXTURE)" != "y" ] && \
+	    ! grep -q '^CONFIG_BLK_SED_OPAL=y$$' "$<"; then \
+		echo "CONFIG_HEADS_OPAL requires CONFIG_BLK_SED_OPAL=y in $<" >&2; \
+		exit 1; \
+	fi
+	@if [ "$(CONFIG_HEADS_OPAL_TEST_FIXTURE)" != "y" ] && \
+	    [ "$(CONFIG_HEADS_OPAL_S3_APMC_V1)" = "y" ] && \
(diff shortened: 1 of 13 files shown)


Makefile
```diff
@@ -777,6 +777,36 @@ bin_modules-$(CONFIG_ZSTD) += zstd
...
+		echo "CONFIG_HEADS_OPAL_S3_APMC_V1 requires CONFIG_PROC_PAGE_MONITOR=y in $<" >&2; \
+		exit 1; \
+	fi
+	@touch "$@"
+
+$(heads_opal_bin): util/heads-opal.c $(heads_opal_config_stamp) \
+		$(INSTALL)/include/linux/limits.h \
+		$(build)/$(musl-cross-make_dir)/.build
+	@mkdir -p "$(dir $@)"
+	$(heads_cc) -Os -Wall -Wextra -Werror \
```
tlaurion, line 803: Can this be built with Oz?

util/heads-opal.c
tlaurion, line 1: Ideally, this would be into a modules/* where this project evolves on its own?

tlaurion (changes requested):
First quick review, will try to review thoroughly in the next days.

Will also need to update threat model, TCG had its bypasses before, so "trusting trust" kinda issues here

